PR #4, opened 2025-02-21
Data streaming

samsja:


src/zeroband/train.py
```diff
@@ -85,14 +86,35 @@ def apply_fsdp(model: torch.nn.Module, reshard_after_forward: bool):
...
+    step = 0  # for debugging
+
+    path = Path(data_path) / f"step_{step}"
+
+    ## todo might fail here if there is no data in the folder.
+    files = []
+    if world_info.rank == 0:
+        files = os.listdir(path)
+
+    dist.broadcast_object_list([files], src=0)
```
apaz-cli: What about `dist.scatter_object_list()` instead?

samsja: huh I did not know it existed :monkey: great idea

src/zeroband/training/data.py
```diff
@@ -26,37 +35,71 @@ def __init__(self, seq_len: int, vocab_size: int):
...
+    def update_files(self, files: list[str]):
+        # Update the shared files list
+        worker_info = torch.utils.data.get_worker_info()
+        self._logger.info(f"Updating files for worker {worker_info.id} with {len(files)} files")
+        self._shared_files[:] = files[worker_info.id :: worker_info.num_workers]  # Clear and update atomically
+
+    def __iter__(self):
+        while True:
+            for file in self._shared_files:
+                table = pq.ParquetFile(file).read()
```
apaz-cli: This holds the entire file in memory. Probably want to stream it from disk? Unless the assumption is we create a new dataloader every iteration or something, with the new files from S3.

src/zeroband/training/data.py
```diff
@@ -26,37 +35,71 @@ def __init__(self, seq_len: int, vocab_size: int):
...
+                    for ids, adv in zip(output_tokens, advantages):
+                        yield {"input_ids": ids, "advantages": adv}
 
-    def load_state_dict(self, state_dict):
-        self.step = state_dict["step"]
-        itera = iter(self)
-        for _ in range(self.step):
-            next(itera)
+            self._logger.info("Waiting for new files")
+            time.sleep(0.5)
```
apaz-cli, line 98: I assume the sleep is a placeholder for the actual condition to be added later.

Ex: A notification from another thread that all the files have been received.

src/zeroband/training/data.py
```diff
@@ -26,37 +35,71 @@ def __init__(self, seq_len: int, vocab_size: int):
...
             input_ids = torch.randint(3, self.vocab_size, (len_,))
             advantages = torch.randn(len_)
-            ref_logprobs = torch.randn(len_, self.vocab_size)
             self.step += 1
-            yield {"input_ids": input_ids, "advantages": advantages, "ref_logprobs": ref_logprobs}
-            # yield {"input_ids": input_ids}
+            yield {"input_ids": input_ids, "advantages": advantages}
+
+
+class ParquetDataset(IterableDataset):
```
apaz-cli, line 50: I presume this is also supposed to be an UpdateableDataset?

apaz-cli (commented):
Looks pretty good, agree on the API. Appending parquet files to the dataloader is a good idea.

src/zeroband/training/data.py
samsja: The file are supposed to be small should not be a problem

src/zeroband/training/data.py
samsja: with Protocol we don't need to inherit. It a bit like rust trait

src/zeroband/train.py
samsja: hmm I tried to use it but it require to know the size of the list before hand.
Which I dont

PR #5, opened 2025-02-25
Data stream 2

samsja:
Okay so this new version of the data loader

As discussed offline we are following this pattern:

image

seems that we don't have performance issue, ( done on 2xa40 with a 150m)

Screenshot from 2025-02-24 17-37-44

Maybe need to check performance a bit more once we have e2e run


Jackmin801 (commented):
LFGTM! :p

PR #12, opened 2025-02-27
add e2e testing

samsja:


src/zeroband/rl_launcher.py
```diff
@@ -26,6 +28,47 @@ class Config(BaseConfig):
...
 
+    total_steps: int | None = None
+
+    rollout_path: str  # rollout_path is define at the top and is inherited by train and inference via the model_validator above
+
+    rollout_data: str  # going to be use by inference to save file and training to load
+
+    batch_size: int | None = None  # going to be use by inference to save file and training to load
+
+    @model_validator(mode="after")
```
apaz-cli, line 39: Are the model validators for better cli error messages?

(I ask because I do not know and also because they are ass)

I like this though.

src/zeroband/inference.py
```diff
@@ -137,23 +126,38 @@ def main(config: Config):  # -> list[dict[str, Any]]:
...
+        if config.rollout_path is not None:
+            last_step = list(Path(config.rollout_path).glob("step_*"))
+            if len(last_step) > 0:
+                last_step = max(last_step, key=lambda x: int(x.stem.split("_")[-1]))
+                maybe_new_step = int(last_step.stem.split("_")[-1])
+
+                if step < maybe_new_step:
+                    stable_file = last_step / "stable"
+
+                    if stable_file.exists():
```
apaz-cli, line 153: Shouldn't we wait until it does? I presume we don't want to continue with an outdated model.

src/zeroband/inference.py
```diff
@@ -165,14 +169,28 @@ def main(config: Config):  # -> list[dict[str, Any]]:
...
 
         pq.write_table(table, f"{step_path}/{uuid.uuid4()}.parquet")
+        total_samples += len(prompts)
+        logger.info(f"Generated {total_samples} total samples")
+
+        if config.step_batch_size is not None and total_samples % config.step_batch_size == 0:
+            # logger.info(f"Reached step batch size {config.step_batch_size}. Writing stable file")
+            stable_file = step_path / "stable"
+            with open(stable_file, "w"):
+                pass
```
apaz-cli: You can call `stable_file.touch()`

src/zeroband/inference.py
apaz-cli: See how I did it in mine:
https://github.com/PrimeIntellect-ai/prime-rl/pull/11/files#diff-9d12d62785006784653b8270e71b40961cb019a774053c8184d65a6a3b9b6351R171-R177

src/zeroband/rl_launcher.py
samsja: they are more to catch error when creating the config. You can say for instance. Batch size should always be divisible by micro batch size at the config level. So you can stop early if the config is wrong. Instead of crashing later at when trying to load the batch.

keeping this code at the config level allow to check that all of our config are always fine (we have a test doing this)

src/zeroband/inference.py
samsja: oh nice !!

src/zeroband/inference.py
samsja: >  I presume we don't want to continue with an outdated model

hmm I actually thought that it makes more sense for the model to continue inferencing instead of just waiting. So that if we have any de-sync problem happening during training it does not create too many blocking situation

PR #13, opened 2025-02-27
grpo padding +  better tests

samsja:


src/zeroband/models.py
```diff
@@ -21,8 +21,8 @@
 }
 
 
-def get_model_and_tokenizer(model_name: ModelName) -> tuple[torch.nn.Module, AutoTokenizer]:
+def get_model_and_tokenizer(model_name: ModelName) -> tuple[torch.nn.Module, AutoTokenizer, int]:
     tokenizer = AutoTokenizer.from_pretrained(name_to_hf_tokenizer[model_name])
     config_model = LlamaConfig.from_pretrained(name_to_hf_model[model_name], attn_implementation="flex_attention")
     model = LlamaForCausalLM.from_pretrained(pretrained_model_name_or_path=name_to_hf_model[model_name], config=config_model)
-    return model, tokenizer
+    return model, tokenizer, tokenizer.eos_token_id
```
Jackmin801: What do u think about having the padding logic do tokenizer.eos_token_id instead of returning it here? Seems like it would be simpler

Jackmin801 (approved):
nice! lgtm

src/zeroband/models.py
samsja: sure

samsja: yeah lol I am stupid, the data_loader function take the tokenized as input already

samsja: oh wait I remember because I did not want to do this. For different model (example qwen vs llama) we would potentially want to have different padding token (eos or smth else). So my idea was to get it from the get_model. So my impl was less stupid that I first thought. That being said gonna use what is suggested here for now

PR #18, opened 2025-02-28
Add type hint model

samsja:


src/zeroband/models.py
```diff
@@ -9,6 +8,7 @@
 )
 
 ModelName: TypeAlias = Literal["debugmodel", "150M", "1B", "Qwen32B", "Qwen1.5B", "Qwen7B"]
+ModelType: TypeAlias = LlamaForCausalLM | Qwen2ForCausalLM
```
Jackmin801, line 11: Hrmm actually would `transformers.modeling_utils.PreTrainedModel` have worked? Could make it that we dont need to keep adding to this in the future

Jackmin801: Oh but theyre removing the GenerationMixin soon. Hrmm

samsja: > Hrmm actually would transformers.modeling_utils.PreTrainedModel have worked? Could make it that we dont need to keep adding to this in the future

hmm but my goal is to be able to control click into the llama and qwen code easily. Moreover I don't think that the PreTrainedModel is precise enough. For example in our apply_fsdp code we relay on the fact that model.model.layers exists. This is true for both LlamaForCausalLM and Qwen2ForCausalLM but probably not for other pretrained model

PR #25, opened 2025-03-04
feat: add grpo advantage estimation and math reward function

justusmattern27:


src/zeroband/inference.py
```diff
@@ -168,13 +208,11 @@ def main(config: Config):  # -> list[dict[str, Any]]:
     for i in range(0, min(len(dataset), max_samples), config.batch_size):
         if config.rollout_path is not None:
             last_step = list(Path(config.rollout_path).glob("step_*"))
-            if len(last_step) > 0:
+            if last_step:
```
samsja, line 200: oh weird bug good catch

src/zeroband/rewards/math.py
```diff
@@ -0,0 +1,24 @@
...
+        gold = parse(verification_info["ground_truth"])
+        correct = verify(gold, response)
+        
+        if correct:
+            return 1
+        else:
+            return -1
+    
+    except Exception as e:
+        print(f"error verifying math: {e}")
```
samsja: can u use logger here ?

samsja: logger = get_logger("INFERENCE")

src/zeroband/rewards/math.py
```diff
@@ -0,0 +1,24 @@
...
+        response = parse(split_response[1])
+        gold = parse(verification_info["ground_truth"])
+        correct = verify(gold, response)
+        
+        if correct:
+            return 1
+        else:
+            return -1
+    
+    except Exception as e:
```
samsja, line 24: Is there a more precise Excpetion we could use here ? no big deal if no

justusmattern27: honestly not sure what kinds of exceptions math-verify throws since it's a third party library so would not do that now - we will rewrite rewards anyway with genesys

PR #31, opened 2025-03-06
Ad micro step

samsja:


src/zeroband/inference.py
```diff
@@ -28,30 +27,27 @@
...
     n: int = 8
 
 
-dataset_key = defaultdict(lambda: "prompt", {"justus27/difficulty-calibrated-deepscaler": "problem"})
-
-
 class Config(BaseConfig):
     name_model: ModelName = "150M"
-    dataset: str = "justus27/deepscaler-math-genesys-format"
+    dataset: Literal["justus27/deepscaler-math-genesys-format"] = "justus27/deepscaler-math-genesys-format"
```
Jackmin801: the reason for specialising to literal here isnt that clear to me

samsja: will remove

src/zeroband/rl_launcher.py
```diff
@@ -56,17 +56,19 @@ def validate_rollout_data(self):
...
     def validate_batch_size(self):
         if self.batch_size is not None:
-            self.inference.step_batch_size = self.batch_size
-            self.train.optim.batch_size = self.batch_size
-            self.train.data.batch_size = self.batch_size
+            assert self.batch_size % self.train.optim.step_per_rollout == 0, "batch_size must be divisible by step_per_rollout"
+
+            self.inference.step_batch_size = self.batch_size // self.inference.sampling.n
+            self.train.optim.batch_size = self.batch_size // self.train.optim.step_per_rollout
+            self.train.data.batch_size = self.batch_size // self.train.optim.step_per_rollout
```
Jackmin801: Why is the inference one different?

Jackmin801: Why are there 3 definitions of batch size actually? Seems confusing, especially since theyre all the same and I believe the data one gets multiplied by step_per_rollout to get back the original batch size anyways @.@

Jackmin801 (approved):
lgtm maybe

PR #33, opened 2025-03-07
add loss mask and fix input_ids

samsja:


src/zeroband/training/data.py
```diff
@@ -183,10 +184,14 @@ def __iter__(self):
                             workers_id=worker_id,
                         ):
                             try:
-                                ids = torch.tensor(token.as_py())
+                                input_ids = torch.tensor(in_token.as_py())
+                                output_ids = torch.tensor(out_token.as_py())
+                                ids = torch.cat([input_ids, output_ids], dim=0)
```
justusmattern27, line 189: we have to make sure that depending on how we tokenize the inputs & outputs, the tokenizer doesn't automatically add eos tokens to the input; if relevant, that is part of the inference code though, so no changes needed

justusmattern27 (approved):

apaz-cli (approved):
LGTM, I really like getting rid of ignore_index.

src/zeroband/training/data.py
samsja: well actually, we want to inference to add the EOS and it not being part of the ingore mask otherwise the model would never learn the output EOS.

Jackmin801 (approved):
Makes sense. Seems like theres some unnecessary casting but should be quickly no-oped so doesnt matter I think.

samsja:
> Makes sense. Seems like theres some unnecessary casting but should be quickly no-oped so doesnt matter I think.

yeah I could just cast at the collate fn

PR #35, opened 2025-03-08
fix grpo and add gen logprobs

samsja:


src/zeroband/training/loss.py
```diff
@@ -25,52 +26,47 @@ def grpo_loss(
...
+    """
+    Computes the GRPO loss.
+    """
+    # Get log probs from current policy
+    log_probs = F.log_softmax(logits, dim=-1)
 
-    # Policy loss
-    policy_loss = -torch.min(ratio * advantages, clipped_ratio * advantages)
+    # Extract per-token log probs for the tokens that were generated
+    per_token_logps = torch.gather(log_probs, dim=2, index=input_ids.unsqueeze(-1)).squeeze(-1)
```
Jackmin801: Would we save memory if we did `sparse_grad=True`? https://pytorch.org/docs/main/generated/torch.gather.html

src/zeroband/training/loss.py
```diff
@@ -25,52 +26,47 @@ def grpo_loss(
...
+    per_token_loss = torch.min(unclipped_loss, clipped_loss)
 
-    # Final loss (mean across batch)
-    loss = (policy_loss + beta * kl_penalty).mean()
+    # Apply mask and average
+    masked_loss = (per_token_loss * loss_mask).sum(dim=1) / loss_mask.sum(dim=1).clamp(min=1e-5)
 
-    return loss
+    # Return mean loss
+    return -masked_loss.mean()  # Negative because we want to maximize reward
```
Jackmin801: Maybe move the the negative up to the line where masked_loss is set?

src/zeroband/training/loss.py
```diff
@@ -25,52 +26,47 @@ def grpo_loss(
...
+    """
+    Computes the GRPO loss.
+    """
+    # Get log probs from current policy
+    log_probs = F.log_softmax(logits, dim=-1)
 
-    # Policy loss
-    policy_loss = -torch.min(ratio * advantages, clipped_ratio * advantages)
+    # Extract per-token log probs for the tokens that were generated
+    per_token_logps = torch.gather(log_probs, dim=2, index=input_ids.unsqueeze(-1), sparse_grad=True).squeeze(-1)
```
Jackmin801: Did you try running the training? Seems like it doesnt work because log_softmax has no sparse bwd impl

image

Jackmin801: Also, we might want to use the trl one. I think its faster. They are doing selection before the softmax whis is less flops

samsja: done

Jackmin801 (approved):
lfgtm

PR #36, opened 2025-03-09
allow to control the  async level

samsja:


src/zeroband/inference.py
```diff
@@ -267,16 +276,16 @@ def main(config: Config):
...
-            logger.info(f"Generated {total_problems} problems for step {step}")
+            logger.info(f"Generated {total_problems} problems for step {real_step}")
 
         # Compute rewards asynchronously, grouped as a dictionary.
         grouped_rewards = asyncio.run(compute_rewards_async(generated_tokens, verification_infos))
         # Compute normalized advantages per prompt.
         grouped_advantages = compute_advantages_grpo(grouped_rewards)
 
-        table = get_parquet_table(generated_tokens, grouped_advantages, grouped_rewards, step)
+        table = get_parquet_table(generated_tokens, grouped_advantages, grouped_rewards, ckpt_step)
```
Jackmin801, line 286: Why is it ckpt_step here? Is this used by the trainer?

samsja: its never used for now. 

Its more that if we want to look at the data after the run we can now which model checkpoint generated the data

Jackmin801 (approved):
Ah nice. Seems like it should work

PR #39, opened 2025-03-09
[Feat] toploc proofs in inference script

Jackmin801:


src/zeroband/inference.py
```diff
@@ -209,13 +214,31 @@ def main(config: Config):
...
     dataset = load_dataset(config.dataset, split="train")
     max_samples = config.max_samples or len(dataset)
 
+    saved_activations = {}
+
+    def logits_processor_hook(module, input):
+        assert isinstance(input[1], torch.Tensor)
+        assert isinstance(input[2], SamplingMetadata)
+        # If the lengths dont match its not a decode step
+        if len(input[2].seq_groups) != input[1].shape[0]:
```
Jackmin801, line 247: There shouldnt be a possibility that a prefill has the correct shape because it would mean the prompts were all only one token.

src/zeroband/inference.py
```diff
@@ -121,7 +126,7 @@ def get_parquet_table(
             output_logprobs_list.append(get_own_logprobs(output.logprobs))
             advantages_list.append(adv)
             rewards_list.append(reward)
-            proofs_list.append("I am toploc proof, handcrafted by jack".encode())
```
samsja, line 142: ngl this proof was good as well

src/zeroband/inference.py
```diff
@@ -209,13 +214,31 @@ def main(config: Config):
...
 
+    saved_activations = {}
+
+    def logits_processor_hook(module, input):
+        assert isinstance(input[1], torch.Tensor)
+        assert isinstance(input[2], SamplingMetadata)
+        # If the lengths dont match its not a decode step
+        if len(input[2].seq_groups) != input[1].shape[0]:
+            return
+        index = [i.seq_ids[0] for i in input[2].seq_groups]
```
samsja, line 250: not sure what his seq_ids and seq_groups are

src/zeroband/inference.py
Jackmin801: 100%

src/zeroband/inference.py
Jackmin801: The sequence id is a unique id generated for each sequence added to the `LLMEngine` for the scheduler to pick up and execute. https://github.com/vllm-project/vllm/blob/v0.7.3/vllm/engine/llm_engine.py#L581

The seq_groups here is actually `SequenceGroupToSample` https://github.com/vllm-project/vllm/blob/v0.7.3/vllm/model_executor/sampling_metadata.py#L19. Its used to store metadata useful for sampling but its also passed to the logits_processor (the module) because its used to prune hidden states and apply logits_processors (the functions that modify logits based on previously generated tokens).

src/zeroband/inference.py
```diff
@@ -209,13 +214,31 @@ def main(config: Config):
...
+        for ti, i in enumerate(index):
+            if i not in saved_activations:
+                saved_activations[i] = []
+            saved_activations[i].append(input[1][ti])
```
Jackmin801: Do we need to detach().clone().cpu() here?

samsja: I would do it just to be careful

Jackmin801:
The timing should be good now. Probably can still squeeze a bit but 1% is already great.
```bash
21:23:07 [INFO] [INFERENCE] [Rank 0] Batch throughput: 911.19 tok/sec (33494 tokens in 36.76s, avg seq len: 523.3)
21:23:07 [INFO] [INFERENCE] [Rank 0] Time taken by logits processor hook: 0.30s
21:23:07 [INFO] [INFERENCE] [Rank 0] Time taken by proofs: 0.10s
21:23:07 [INFO] [INFERENCE] [Rank 0] Generated 16 total problems with 128 total samples
21:23:07 [INFO] [INFERENCE] [Rank 0] real_step: 0, ckpt_step: 0, real_step - ckpt_step: 0, config.max_async_level: 2
21:23:43 [INFO] [INFERENCE] [Rank 0] Batch throughput: 933.40 tok/sec (33353 tokens in 35.73s, avg seq len: 521.1)
21:23:43 [INFO] [INFERENCE] [Rank 0] Time taken by logits processor hook: 0.30s
21:23:43 [INFO] [INFERENCE] [Rank 0] Time taken by proofs: 0.10s
21:23:43 [INFO] [INFERENCE] [Rank 0] Generated 24 total problems with 192 total samples
21:23:43 [INFO] [INFERENCE] [Rank 0] real_step: 0, ckpt_step: 0, real_step - ckpt_step: 0, config.max_async_level: 2
21:24:19 [INFO] [INFERENCE] [Rank 0] Batch throughput: 933.29 tok/sec (33359 tokens in 35.74s, avg seq len: 521.2)
21:24:19 [INFO] [INFERENCE] [Rank 0] Time taken by logits processor hook: 0.30s
21:24:19 [INFO] [INFERENCE] [Rank 0] Time taken by proofs: 0.10s
21:24:19 [INFO] [INFERENCE] [Rank 0] Generated 32 total problems with 256 total samples
```

src/zeroband/inferencing/toploc.py
```diff
@@ -0,0 +1,107 @@
...
+        self._proof_futures: Dict[int, Future] = {}
+
+    def _init_cache(self, device: torch.device, dtype: torch.dtype):
+        """Initialize the cache tensor and related tracking structures.
+
+        Args:
+            device (torch.device): Device to store the cache tensor on
+            dtype (torch.dtype): Data type for the cache tensor
+        """
+        self._cache: torch.Tensor = torch.empty(self.max_seqs, self.max_len, self.hidden_size, device=device, dtype=dtype)
```
samsja:
```suggestion
        self._cache = torch.empty(self.max_seqs, self.max_len, self.hidden_size, device=device, dtype=dtype)
```

src/zeroband/inference.py
```diff
@@ -274,6 +295,7 @@ def inference(config: Config):
         start_time = time.time()
         generated_tokens = llm.generate(prompts, sampling_params, use_tqdm=False)
         end_time = time.time()
+        toploc_cache.maybe_generate_proofs_in_background(force_generate=True)
```
samsja, line 301: I am not sure to understand why we don't have this running periodocally in the background.

(not saying we should do it but what is the reason ?)

src/zeroband/inferencing/toploc.py
```diff
@@ -0,0 +1,107 @@
...
+from toploc import build_proofs_bytes
+from concurrent.futures import ThreadPoolExecutor, Future, wait
+
+
+class TopLocCache:
+    """A cache implementation for managing sequence data and generating proofs.
+
+    This class provides functionality to store sequence data in a tensor cache and
+    asynchronously generate proofs using a thread pool executor when the sequence reaches max_len.
+
```
samsja, line 12:
```suggestion
    """A cache implementation for managing sequence data and generating proofs.

    This class provides functionality to store sequence data in a tensor cache and
    asynchronously generate proofs using a thread pool executor when the sequence reaches max_len.

    It allows us to generate the proof as the same time as we are generating tokens

```

src/zeroband/inference.py
Jackmin801: This one is forced. it should only be executed at the end to generate proofs for the leftover hidden states.

The class consumes the hidden states every 32 hidden states to make a proof. but if the generation is 100 output tokens, we will leave the generate with 100 % 32 remaining tokens in the cache. This makes sure we generate a proof for these leftover tokens

src/zeroband/inferencing/toploc.py
Jackmin801: thred -> thread, apool -> pool

src/zeroband/inference.py
samsja: I see, makes sense. I think the name is a bit confusing. We could rename to 

`toploc_cache.force_generate_left_over_proofs()` which would just call `        toploc_cache.maybe_generate_proofs_in_background(force_generate=True)`

but I am cherry-picking no need to do it tbh

src/zeroband/inferencing/toploc.py
samsja: skill issue

samsja: wait, but I did not change this line myself. could it be that cursor learns to do typo when watching me code ????? Would be terrible for me

Jackmin801: hahaha that would be crazy

src/zeroband/inference.py
Jackmin801: hrmm yea I think an extra function is a bit more confusing. A comment could help though

PR #40, opened 2025-03-09
silence flex attn stuff

samsja:


src/zeroband/train.py
```diff
@@ -51,6 +54,8 @@ class TrainConfig(BaseConfig):
 
     attn_impl: AttnImpl = "flex_attention"
 
+    verbose_compile: bool = False
```
Jackmin801: would u consider using env var instead? its not set very often and isnt rlly a train config

samsja: sure

Jackmin801 (approved):

PR #42, opened 2025-03-10
fix reshard after forwards

samsja:
this pr should not change behavior but is cleaner, could catch bug later


Jackmin801 (approved):
sure lgtm

PR #43, opened 2025-03-11
Fix async

samsja:


src/zeroband/inference.py
```diff
@@ -340,8 +337,8 @@ def inference_run(config: Config) -> list[mp.Process]:
     if config.dp > 1:
         processes = []
 
-        if config.step_batch_size is None:
-            config.step_batch_size = config.batch_size // config.dp
+        if config.step_batch_size is not None:
```
Jackmin801: is this if still needed? we dont allow none anymore right?

Jackmin801 (approved):
nice! lfgtm

src/zeroband/inference.py
samsja: true

PR #44, opened 2025-03-11
fix dataloader

samsja:


src/zeroband/training/data.py
```diff
@@ -53,33 +53,46 @@ def __iter__(self) -> Generator[dict[str, Any], Any, None]:
...
+
+    while True:
+        files = list(step_path.glob("*.parquet"))
+
+        rows = 0
+        if len(files) > 0:
+            try:
+                dataset = ds.dataset(files, format="parquet")
+                rows = dataset.count_rows()
+            except Exception as e:
```
samsja, line 76: this excpetion allows to catch race condition on the parquet file not being fully written to disk

Jackmin801: yea looks ok

Jackmin801 (approved):
nice! lgtm

PR #50, opened 2025-03-12
add gcp storage

samsja:


src/zeroband/training/data_prefetch.py
```diff
@@ -0,0 +1,220 @@
...
+                # here we only take the last max_buffer_steps steps. the key are the step numbers and the values are the blobs
+
+            for step_number in step_to_download:
+                files = self.filter_to_download(steps_blobs[step_number])
+                for file in files:
+                    self.thread_pool_download.submit(step_number, self._download_files, file)
+                    # we want to download the file from the oldest step first
+                    self.files_downloaded.append(file.name)
+
+            # TODO bring back deleting logic
```
Jackmin801, line 159: We will delete them manually. ez. how fast can they grow?

src/zeroband/training/data_prefetch.py
```diff
@@ -0,0 +1,220 @@
...
+
+            step_to_download = sorted(list(steps_blobs.keys()))
+            if self.max_buffer_steps is not None:
+                step_to_download = step_to_download[-self.max_buffer_steps :]
+                # here we only take the last max_buffer_steps steps. the key are the step numbers and the values are the blobs
+
+            for step_number in step_to_download:
+                files = self.filter_to_download(steps_blobs[step_number])
+                for file in files:
+                    self.thread_pool_download.submit(step_number, self._download_files, file)
```
Jackmin801: is it ok to pass step_number as kwarg here? the expectation is that threadpool work submission has first arg as fn so I was confused here that we passed an int

Jackmin801 (approved):
naisu!

PR #53, opened 2025-03-13
fix: deepscaler math reward

justusmattern27:


samsja (changes requested):
I cannot run the code 

```
uv sync
error: Failed to parse `uv.lock`
  Caused by: Dependency `antlr4-python3-runtime` has missing `source` field but has more than one matching package
(prime-rl) 

```

samsja:
need to run the pre-commit hook as well

samsja (approved):

PR #55, opened 2025-03-14
[Feat] max tokens for whole seq and qwen test for protocol testing

Jackmin801:


src/zeroband/inference.py
```diff
@@ -227,7 +227,8 @@ def inference(config: Config):
     llm = LLM(
         model=name_to_hf_model[config.name_model],
         tensor_parallel_size=config.tp,
-        max_seq_len_to_capture=int(1.5 * config.sampling.max_tokens),  # 1.5 makes sure we capture the entire output even with prefill
+        max_seq_len_to_capture=config.sampling.max_tokens,
+        max_model_len=config.sampling.max_tokens,
```
samsja: should we remove max_tokens from sampling param ? Not sure if there is a conflict or nor if we have both max_model_len and max_sampling. Like which is has priority ?

Jackmin801: max_sampling?

Jackmin801: Eh we should have max_model_len as its own arg and set the default for max_tokens to None

samsja (approved):
lgtm

PR #57, opened 2025-03-14
Reordering Fixes

apaz-cli:
Reorder liger optimizations before activation checkpointing

Move the arguments to the loss function to cuda after the forward pass 

Memory profile up to the second full step (To show the materialized mixed precision model)


samsja (approved):
lgtm

PR #59, opened 2025-03-17
Feat add fa2 script

samsja:


Jackmin801 (approved):
Does the build work? If so lgtm

samsja:
> Does the build work? If so lgtm

only way to find out is to merge

or to test locally

gonna merge

PR #73, opened 2025-03-21
fix loss mask

samsja:


Jackmin801 (approved):
nice! lgtm

src/zeroband/train.py
```diff
@@ -216,8 +216,8 @@ def train(config: Config):
             # Load args
             batch = next(train_dataloader_iterator)
             input_ids = batch["input_ids"].to("cuda")
-            loss_mask = batch["loss_mask"].bool()
-            average_rewards += batch["rewards"][loss_mask].mean() / gradient_accumulation_steps
+            loss_mask = batch["loss_mask"]
+            average_rewards += (batch["rewards"] * loss_mask).mean() / gradient_accumulation_steps
```
Jackmin801: Oh. The original one is correct. The mean is different because `(a*b).numel() != a[b].numel()`

Jackmin801 (changes requested):

src/zeroband/train.py
samsja: yeah my bad

samsja:
Screenshot from 2025-03-21 15-17-42

PR #80, opened 2025-03-23
add suport for fp32 inference

samsja:


src/zeroband/inferencing/toploc.py
```diff
@@ -4,6 +4,9 @@
 from concurrent.futures import ThreadPoolExecutor, Future, wait
 
 
+DISABLE_MAX_LEN = 8192
```
Jackmin801: Is this used?

src/zeroband/inferencing/toploc.py
```diff
@@ -21,15 +24,27 @@ class TopLocCache:
...
         self.max_seqs = max_seqs
         self.max_len = max_len
         self.hidden_size = hidden_size
         self.device = device
         self._cache: Optional[torch.Tensor] = None
-        self.proofs: Dict[int, List[bytes]] = {}
-        self._executor = ThreadPoolExecutor(max_workers=8)
-        self._proof_futures: Dict[int, Future] = {}
+
+        self.proofs: Dict[int, List[bytes]]
```
Jackmin801: What do u think about instantiating empty dict here?

Jackmin801 (commented):
Eh do we not have to create an infinite iterator for the proof in the parquet table func?

src/zeroband/inferencing/toploc.py
samsja: I did not want to repeat the type hint  hence did it after

Jackmin801: oh but this is within scope always right so its ok?

samsja: okay fair

samsja: fixed

Jackmin801 (approved):
nice! lgtm

PR #85, opened 2025-03-24
[Feat] shardcast broadcast

Jackmin801:
Eh the uv.lock is a bit sus actually. 3k lines for 1 pure python package? O.o

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ dependencies = [
     "pylatexenc>=2.10",
     "liger_kernel",
     "google-cloud-storage",
+    "shardcast >= 0.1.2",
 ] 
 
 [build-system]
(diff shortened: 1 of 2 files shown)


Jackmin801:
Waiting for https://github.com/PrimeIntellect-ai/shardcast/pull/5 to merge and new wheel release. Should fix the giant uv lock change issue

src/zeroband/train.py
```diff
@@ -87,6 +88,7 @@ class Config(BaseConfig):
     entropy_loss_coeff: float = 0.001
 
     on_policy_log_prob: bool = False
+    max_async_level: int = 2  # the amount of rollout checkpoints to keep
```
samsja, line 91: should not this be the same value as in inference?. If not lets rename to avoid confusion ?

Jackmin801: oh its already the same name as the config in inference though

Jackmin801: eh will merge first, can have a small PR to change later if needed

PR #94, opened 2025-03-27
fix async level

samsja:


Jackmin801 (commented):
oof I think the >= is wrong ;-; 
because sync where real_step=0, ckpt_step=0, max_async_level=0 would try to load ckpt at the first step

Jackmin801 (approved):
lgtm

PR #95, opened 2025-03-27
[Feat] shardcast bg

Jackmin801:


src/zeroband/inferencing/envs.py
```diff
@@ -0,0 +1,23 @@
+from typing import TYPE_CHECKING, Any, List
+import os
+
+if TYPE_CHECKING:
+    SHARDCAST_SERVERS: list[str] | None = None
+    SHARDCAST_BACKLOG_VERSION: int = -1
+
```
samsja, line 7: noice :)

samsja: uh wait why putting default value for type checking ?

Jackmin801: well it doesnt matter I guess but its easier to read what the default is

PR #120, opened 2025-04-04
[fix] toploc proofs work with cuda graph enabled

Jackmin801:


src/zeroband/inference.py
```diff
@@ -260,14 +261,20 @@ def inference(config: Config):
...
     def logits_processor_hook(module, input):
-        assert isinstance(input[1], torch.Tensor)
-        assert isinstance(input[2], SamplingMetadata)
-        # If the lengths dont match its not a decode step
-        if len(input[2].seq_groups) != input[1].shape[0]:
+        hidden_states, sampling_metadata = input[1], input[2]
+        assert isinstance(hidden_states, torch.Tensor)
+        assert isinstance(sampling_metadata, SamplingMetadata)
+        # This check is true only for prefills
+        if max(sampling_metadata.selected_token_indices) > len(sampling_metadata.seq_groups):
```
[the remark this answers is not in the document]
Jackmin801: eh no

src/zeroband/inference.py
```diff
@@ -260,14 +261,20 @@ def inference(config: Config):
...
             return
-
-        index = [i.seq_ids[0] for i in input[2].seq_groups]
-        toploc_cache.add(index, input[1])
+        # This pruning is required when cuda graph padding is enabled.
+        hidden_states = _prune_hidden_states(hidden_states, sampling_metadata)
```
samsja, line 271: hmm might lock us forever to this version of vllm tho. but I don;t think we have a choice. Would be worth opening an issue for vllm 1.0 to support our usecase with toploc

samsja (approved):
lgftm

src/zeroband/inference.py
Jackmin801: Yea we should. The hook trick wont work in vllm 1.0 because the main proc no longer owns the module

PR #121, opened 2025-04-04
Fix fsdp 1

samsja:


src/zeroband/train.py
```diff
@@ -50,16 +53,23 @@ class OptimConfig(BaseConfig):
     step_per_rollout: int = 1
 
 
+class FSDPConfig(BaseConfig):
+    backward_prefetch: Literal["pre", "post"] = (
+        "pre"  # pre == faster but more memory, post == slower but less memory https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.BackwardPrefetch
```
apaz-cli, line 58: I presume we tried both?

src/zeroband/training/checkpoint.py
```diff
@@ -3,8 +3,10 @@
 from pathlib import Path
 import time
 import torch
-from torch.distributed.checkpoint.state_dict import get_model_state_dict, StateDictOptions
+from torch.distributed.checkpoint.state_dict import get_model_state_dict, StateDictOptions, get_state_dict, set_state_dict
+
 from safetensors.torch import save_file
+import torch.distributed.checkpoint as dcp
```
apaz-cli, line 9: Do you know what's different here/why you use dcp with fsdp1 and get_model_state_dict() suffices for 2?

src/zeroband/training/checkpoint.py
```diff
@@ -64,26 +71,32 @@ def load_checkpoint_fsdp_state(
     """
     Load the checkpoint state.
     """
+    import warnings
+
+    warnings.filterwarnings("ignore", message="Please use DTensor instead")
```
apaz-cli, line 76: 😭

src/zeroband/train.py
```diff
@@ -207,12 +237,12 @@ def train(config: Config):
     if world_info.rank == 0 and config.wandb:
         wandb.init(project=config.project, config=config.model_dump())
 
-    if config.train.torch_compile:
-        model = torch.compile(model) if not TYPE_CHECKING else model
```
apaz-cli, line 211: Makes sense. Oops.

src/zeroband/training/checkpoint.py
samsja: its the same

src/zeroband/train.py
```diff
@@ -207,12 +248,12 @@ def train(config: Config):
...
-
     if config.ckpt.resume:
         load_checkpoint_fsdp_state(model, [optimizer], training_progress, scheduler, config.ckpt.resume)
 
+    if config.train.torch_compile:
+        model = torch.compile(model) if not TYPE_CHECKING else model
```
JohannesHa, line 255: this should not move?

PR #122, opened 2025-04-05
Fix mfu

samsja:


src/zeroband/training/utils.py
```diff
@@ -48,6 +48,8 @@ def get_peak_flops(device_name: str) -> int:
             return 756e12
         else:  # for H100 SXM and other variants
             return 989e12
+    elif "H200" in device_name:
+        return 1979e12  # sxm  https://resources.nvidia.com/en-us-data-center-overview-mc/en-us-data-center-overview/hpc-datasheet-sc23-h200
     else:  # for other GPU types, assume A100
         return 312e12
```
Jackmin801, line 54: Maybe we should set a default value here that makes it obvious that the number is unreliable?

Jackmin801 (approved):
nice! lgtm

src/zeroband/training/utils.py
samsja: hmmmmmmmm I think I agree with you but now my brain has remenber the mfu that I should get on my 3090 with respect to A100 flops :laughing:

samsja: so I am just going to merge sorry

PR #123, opened 2025-04-05
[fix] disable async output proc

Jackmin801:


Jackmin801:
We should test the speed with and without first. If its significant. Might be worth changing the caching logic to accomodate it

PR #131, opened 2025-04-07
add timing for log prob

samsja:


src/zeroband/train.py
```diff
@@ -434,7 +445,11 @@ def train(config: Config):
         time_rollout_step = time.time() - time_start
         logger.info(f"Finished rollout {rollout_step} step {training_progress.step}")
         if world_info.rank == 0 and config.wandb:
-            wandb.log({"rollout_step": rollout_step, "step": training_progress.step, "time_rollout_step": time_rollout_step})
+            new_metrics = {"rollout_step": rollout_step, "step": training_progress.step, "time_rollout_step": time_rollout_step}
+            if time_logprob is not None:
```
Jackmin801: is it possible to be none? it seems to me based on the code changes that if config.on_policy_log_prob is false, it will error here on undefined local variable

samsja: good catch

Jackmin801 (approved):
nice! lgtm

PR #132, opened 2025-04-07
remove rl launcher

samsja:


Jackmin801 (approved):
RIP. May it return on a finer day to launch greater models

samsja:
> RIP. May it return on a finer day to launch greater models

will bring it back at some point for sure.

I am just pumping my commits and change of line by removing and adding it back later

PR #134, opened 2025-04-07
[feat] Random seed

Jackmin801:


src/zeroband/inference.py
```diff
@@ -317,7 +323,14 @@ def logits_processor_hook(module, input):
                 attempt_count += 1
 
         # Get batch
-        batch = dataset.select(range(i, min(i + config.batch_size, len(dataset))))
+        if node_address_int is not None:
+            # TODO: What if we have multiple sample per real step?
+            # Its impossible right now but we need to fix this if accept counter is used.
+            generator = np.random.default_rng(node_address_int + real_step)
+            indexes = generator.integers(0, len(dataset), config.batch_size)
+            batch = dataset.select(indexes)
```
samsja: can u add a comment here explaining that we reset the shuffling seed at each step so that we can deterministically get all the data ?

src/zeroband/inference.py
```diff
@@ -245,10 +245,17 @@ def inference(config: Config):
     logger = get_logger(f"INFERENCE {rank}")
     sampling_params = SamplingParams(**config.sampling.model_dump())
 
-    generator = np.random.default_rng(config.seed + rank) if config.seed is not None else np.random.default_rng()
-    # not sure what is the default seed for np.random.default_rng so doing this to make sure we use the default value
+    if os.environ.get("NODE_ADDRESS") is not None:
+        dataset = load_dataset(config.dataset, split="train")
+        assert rank == 0, "DP is not supported when NODE_ADDRESS is set"
+        node_address_int = int(os.environ.get("NODE_ADDRESS"), 16)
+        logger.info(f"Seeding with {node_address_int} ({os.environ.get('NODE_ADDRESS')})")
```
samsja: Can u add a comment here that we don't shuffle if we have the node address env var because we do the shuffling per batch ?

src/zeroband/inference.py
```diff
@@ -245,10 +245,17 @@ def inference(config: Config):
     logger = get_logger(f"INFERENCE {rank}")
     sampling_params = SamplingParams(**config.sampling.model_dump())
 
-    generator = np.random.default_rng(config.seed + rank) if config.seed is not None else np.random.default_rng()
-    # not sure what is the default seed for np.random.default_rng so doing this to make sure we use the default value
+    if os.environ.get("NODE_ADDRESS") is not None:
+        dataset = load_dataset(config.dataset, split="train")
+        assert rank == 0, "DP is not supported when NODE_ADDRESS is set"
```
samsja, line 307: we should probably assert that the seed is None here so that we don't try to use seed and NODE_ADRESS at the same time

Jackmin801: added

src/zeroband/inference.py
Jackmin801: yuop

src/zeroband/inference.py
Jackmin801: yeep

samsja (approved):

Jackmin801:
Tests didnt pass but we can just revert if it doesnt work

Jackmin801:
tested locally so shud be fine

PR #144, opened 2025-04-08
Seq packing

samsja:


apaz-cli (approved):
Good. Just adds tests that call the model with position_ids, doesn't actually change any functionality.

PR #147, opened 2025-04-08
[Feat] inference acceptcounter resume

Jackmin801:


src/zeroband/inference.py
```diff
@@ -370,15 +372,26 @@ def logits_processor_hook(module, input):
         ckpt_step = 0
         real_step = 0
 
+    current_step_batch_counter = (
+        1  # This is used by the seeding logic to make sure we dont generate the same samples twice if we do multiple batches for a step
+    )
```
samsja:
```suggestion
 # This is used by the seeding logic to make sure we dont generate the same samples twice if we do multiple batches for a step
    current_step_batch_counter = 1 
```

I hate when the formater do this weird stuff haha

src/zeroband/inference.py
```diff
@@ -370,15 +372,26 @@ def logits_processor_hook(module, input):
...
+                current_step_batch_counter = 1
+            else:
+                current_step_batch_counter += 1
+
         logger.info(
             f"real_step: {real_step}, ckpt_step: {ckpt_step}, real_step - ckpt_step: {real_step - ckpt_step}, config.async_level: {config.async_level}"
         )
         if config.rollout_path is not None and real_step - ckpt_step > config.async_level:
-            ckpt_step += 1
+            ckpt_step = real_step - config.async_level
```
samsja, line 394: not sure to understand what it does here

src/zeroband/inference.py
```diff
@@ -370,15 +372,26 @@ def logits_processor_hook(module, input):
...
         real_step = 0
 
+    current_step_batch_counter = (
+        1  # This is used by the seeding logic to make sure we dont generate the same samples twice if we do multiple batches for a step
+    )
     total_problems = 0
     total_tokens = 0
 
     for i in range(0, min(len(dataset), max_samples), config.batch_size):
+        if config.step_endpoint is not None:
```
samsja, line 381: can u add a comment of what is happening here ?

src/zeroband/inference.py
Jackmin801: this way of doing the math covers what it used to do a as well as the new resume logic. The if statement can be reordered to `ckpt_step < real_step - config.async_level`. Say the LHS is $N$, we will trigger when RHS is $N+1$. +=1 would thus bring LHS to $N + 1$.

PR #158, opened 2025-04-08
Sequence Packing Round 2

apaz-cli:
Let me know if I messed up the batch size stuff. Also definitely needs a validation run.

### DIFF
--- a/src/zeroband/train.py
+++ b/src/zeroband/train.py
@@ -253,8 +253,9 @@ def train(config: Config):
 
                 for rollout_step in range(config.optim.step_per_rollout):
                     batch_rollout: list[DatasetOutput] = next(train_dataloader_iterator)
-                    batch_packed, num_grad_acc_steps = packed_batch(batch_rollout, config.data.seq_length, tokenizer.pad_token_id, 1)
-
+                    batch_packed, num_grad_acc_steps = packed_batch(
+                        batch_rollout, config.data.seq_length, tokenizer.pad_token_id, config.train.micro_bs
+                    )
                     data_per_rollout = []
                     for grad_acc_step in range(num_grad_acc_steps):
                         time_data_loading = time.time()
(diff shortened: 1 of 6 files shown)


src/zeroband/train.py
```diff
@@ -250,20 +249,24 @@ def train(config: Config):
...
-                        batch = next(train_dataloader_iterator)
+                        batch = batch_packed[grad_acc_step]
 
                         time_data_loading = time.time() - time_data_loading
                         total_time_data_loading += time_data_loading
 
                         input_ids = batch["input_ids"].to("cuda")
 
-                        logits: Float[torch.Tensor, "batch seq vocab"] = model(input_ids=input_ids).logits.contiguous()
+                        logits: Float[torch.Tensor, "batch seq vocab"] = model(input_ids=input_ids, position_ids=batch["position_ids"]).logits.contiguous()
```
samsja: don't you need to move position-ids to GPU ?
or maybe hf is doing it for us

src/zeroband/training/data.py
```diff
@@ -296,125 +308,190 @@ def __iter__(self):
...
+        
 
-        inputs_ids = []
+        input_ids = []
         advantages = []
         rewards = []
         task_rewards = []
         length_penalties = []
-        target_lens = []
+        #target_lens = []
```
samsja: so we don't need this target len anymore ?

apaz-cli: Correct, train.py doesn't actually use it.

src/zeroband/train.py
apaz-cli: You don't. I think HF is doing it as well.

src/zeroband/train.py
```diff
@@ -250,20 +249,24 @@ def train(config: Config):
...
         with torch.no_grad():
             if config.on_policy_log_prob:
-                data = []
+                data: list[tuple[list[BatchOutput], int]] = []
 
                 for rollout_step in range(config.optim.step_per_rollout):
-                    for grad_acc_step in range(gradient_accumulation_steps):
+                    batch_rollout: list[DatasetOutput] = next(train_dataloader_iterator)
+                    batch_packed, num_grad_acc_steps = packed_batch(batch_rollout, config.data.seq_length, tokenizer.pad_token_id, 1)
+
```
samsja: it should not be 1 here

PR #160, opened 2025-04-09
put length prompt into system prompt

justusmattern27:


Jackmin801 (approved):
lgtm!

apaz-cli (approved):
LGTM but should we put the length part of the prompt at the end? I wonder if this allows for some kvcache reuse. It also may not matter, idk. But usually I think you'd want to put it at the end so not to break prefill caching when people go to actually use the model in production.

samsja (approved):

justusmattern27:
> LGTM but should we put the length part of the prompt at the end? I wonder if this allows for some kvcache reuse. It also may not matter, idk. But usually I think you'd want to put it at the end so not to break prefill caching when people go to actually use the model in production.

this is what is done in "instruction" variant

PR #167, opened 2025-04-10
add debug math

samsja:
https://wandb.ai/primeintellect/rl-Qwen1.5B/runs/6lsn2qd2?nw=nwusersamsja_pi
Screenshot from 2025-04-10 11-38-55

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,26 @@ inference
 uv run python src/zeroband/inference.py @ configs/inference/debug.toml
 ```
 
+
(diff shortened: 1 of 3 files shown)


justusmattern27 (approved):

kalomaze (approved):
ulimit is in the same place as updated readme (trainer, not the vllm tmux window), maximum steps shouldn't cause problems either (trainer will just idle when samples are exhausted), looks fine

PR #179, opened 2025-04-11
add schema validation for the dataloader

samsja:


Jackmin801 (approved):
Ah so basically check it before trying to load it in dataset right? Wont the dataset creation error though? Was the behavior that dataset didnt error because its lazy and then something errors later on outside of the try catch?

samsja:
> the dataset creation error though

yeah dataset does not error if we don't have the right schema, what will happened in that the scanner later will fail because of a missing column.

What will failed before is that if the type of the data is not good, lets say someone write a list of int instead of a list of list of int

PR #200, opened 2025-04-14
[fix] go back to start if we didnt load in 30s

Jackmin801:


samsja (approved):
congrats on doing the 200 pr of this repo

PR #213, opened 2025-04-16
fix all reduce

justusmattern27:


Jackmin801 (approved):
1 liners are the best PRs

samsja (approved):
good catch mister system engineer

PR #259, opened 2025-05-02
fix default model nake

samsja:


Jackmin801 (approved):
woah

mikasenghaas (approved):
thank you for your service!

PR #260, opened 2025-05-02
[PRI2-468] Improve logger and world info

mikasenghaas:
This PR improves the logger and world info classes:
- Moves both classes into `utils` submodule from the` training` submodule as they are used for both training and inference (+updates all imports)
- Adds unit tests for both in `tests/units/test_{world_info|logger}`
- Allows for multiple loggers with diferent names per process, i.e. remove singleton structure and handle it in the native logging module's registry
- Changes the `ZEROBAND_LOG_LEVEL` env var to `PRIME_LOG_LEVEL`
- Integrates the world info class into `inference.py`


src/zeroband/utils/world_info.py
```diff
@@ -0,0 +1,70 @@
...
+    global _WORLD_INFO
+    _WORLD_INFO = None
+
+
+if __name__ == "__main__":
+    import torch.distributed as dist
+
+    print(get_world_info())
+    if dist.is_initialized():
+        dist.destroy_process_group()
```
samsja: do we still need this ?

src/zeroband/utils/logger.py
```diff
@@ -0,0 +1,61 @@
...
+class PrimeFormatter(Formatter):
+    def __init__(self, world_info: WorldInfo):
+        super().__init__()
+        self.world_info = world_info
+
+    def format(self, record):
+        record.local_rank = self.world_info.local_rank
+        log_format = "{asctime} [{name}] [{levelname}] [Rank {local_rank}] {message}"
+        formatter = logging.Formatter(log_format, style="{", datefmt="%m-%d %H:%M:%S")
+        return formatter.format(record)
```
samsja, line 17: should we add color here like you did in the vllm repo ?

src/zeroband/utils/world_info.py
mikasenghaas: It's used in the `test_init_with_torchrun` test - but maybe there is a more elegant solution?

src/zeroband/utils/logger.py
mikasenghaas: I did not include color yet, because I'm waiting for response from Jannik to maybe align formatting with protocol (having one clean, colored logging format). Will be easy to add later.

Also, in `prime-vllm` I use [loguru](https://loguru.readthedocs.io/en/stable/) as dep. I really like it - but not sure if we want to add a dependency for logging here. Essentially, making the logging more "fancy" I left for future PR.

src/zeroband/utils/world_info.py
samsja: hoh okay I did not notice, feel free to add comments here to say that it is used be the test then !

src/zeroband/utils/logger.py
samsja: Oh I see, damn this loguru things look great, but yeah future PR

samsja (approved):
pre-approving, feel free to merge once you add the comments in the __main__

PR #262, opened 2025-05-03
[PRI2-471] Skip model tests if `flash_attn` not available

mikasenghaas:
Couldn't test full suite locally without having `flash_attn` installed


samsja (changes requested):
hmm actually kind of hesitant with this PR. This means that if we have a bug and somehow the flash attn is not install we won't catch the failing test.

I would rather return sdpa instead of skipping the test

mikasenghaas:
Both `sdpa` and `eager` fail test `test_model_with_sequence_packing` with

```
        if correct_position_ids:
            torch.testing.assert_close(output_packed_left, output_base)
>           torch.testing.assert_close(output_packed_right, output_base)
E           AssertionError: Tensor-likes are not close!
E           
E           Mismatched elements: 121428 / 128000 (94.9%)
E           Greatest absolute difference: 0.546875 at index (0, 0, 19292) (up to 1e-05 allowed)
E           Greatest relative difference: 28416.0 at index (0, 1, 20342) (up to 0.016 allowed)
```

which is why I was hesitant to run those as fallbacks

samsja (approved):
lfgtm !

PR #265, opened 2025-05-06
[PRI2-480] Toploc refactor

mikasenghaas:
Mainly for me to get used with how TOPLOC works in inference script. Moved some logic that cluttered the main script into a separate file. Main change of PR is that we have isolated unit tests for the TOPLOC tests and for generation with a dummy vLLM model using the hooks. This should also make it easier to catch issues when bumping vLLM.


src/zeroband/inference/toploc.py
```diff
@@ -14,20 +21,21 @@ class TopLocCache:
...
-        max_len (int): Maximum length of each sequence
         hidden_size (int): Size of the hidden dimension for each sequence element
+        max_len (int): Maximum length of each sequence, defaults to 32
         device (Optional[torch.device]): Device to store the cache tensor on.
             If None, the device of the first sequence will be used.
             Defaults to None.
     """
 
-    def __init__(self, max_seqs: int, max_len: int, hidden_size: int, device: Optional[torch.device] = None, disable: bool = False):
+    def __init__(self, max_seqs: int, hidden_size: int, max_len: int = 32, device: Optional[torch.device] = None, disable: bool = False):
```
samsja:
```suggestion
    def __init__(self, max_seqs: int, hidden_size: int, max_len: int = 32, device: torch.device |  None = None, disable: bool = False):
```

src/zeroband/inference/toploc.py
```diff
@@ -129,3 +140,49 @@ def wait_for_proofs(self):
             return
 
         wait(list(i for i in self._proof_futures.values() if i is not None))
+
+
+def toploc_hook(_, inputs: tuple, toploc: TOPLOC):
```
samsja:
```suggestion
def toploc_hook(_, inputs: tuple, toploc: TOPLOC):
```
_ this might break passing by kwargs at some point ?

Wondering if we should rather do *args,**kwargs, this would be more pythonic

mikasenghaas:
Ah damn, final test checking for exact match is failing on CI machine. Might be difficult to reproduce exactly..

src/zeroband/inference/toploc.py
```diff
@@ -129,3 +140,49 @@ def wait_for_proofs(self):
...
+        raise ValueError(f"Lengths dont match: {len(sampling_metadata.seq_groups)} {hidden_states.shape}")
+
+    # Get seq_ids from seq_groups
+    seq_ids = [seq_group.seq_ids[0] for seq_group in sampling_metadata.seq_groups]
+
+    # Add hidden states to TOPLOC cache with corresponding seq_ids
+    toploc.add(seq_ids, hidden_states)
+
+
+def setup_toploc(llm: LLM, disable: bool = False, **toploc_kwargs) -> Tuple[TOPLOC, Optional[RemovableHandle]]:
```
samsja:
```suggestion
def setup_toploc(llm: LLM, disable: bool = False, **toploc_kwargs) -> Tuple[TOPLOC, RemovableHandle | None]:
```

src/zeroband/inference/toploc.py
```diff
@@ -129,3 +140,49 @@ def wait_for_proofs(self):
...
+    """Initializes the TOPLOC cache and register a hook to dynamically populate the cache during inference"""
+    # Initialize the cache
+    toploc = TOPLOC(disable=disable, **toploc_kwargs)
+
+    # Register hook to add hidden states to TOPLOC cache
+    logits_processor: nn.Module = llm.llm_engine.model_executor.driver_worker.model_runner.model.logits_processor
+    handle: Optional[RemovableHandle] = None
```
samsja:
```suggestion
    handle: RemovableHandle | None = None
```

src/zeroband/inference/toploc.py
mikasenghaas: Not an optional fan?

src/zeroband/inference/toploc.py
```diff
@@ -129,3 +140,49 @@ def wait_for_proofs(self):
...
+    if not disable:
+        handle = logits_processor.register_forward_pre_hook(partial(toploc_hook, toploc=toploc))
```
samsja:
```suggestion
    if not disable:
        handle = logits_processor.register_forward_pre_hook(partial(toploc_hook, toploc=toploc))
   else:
        handle = None
     
```

samsja: this avoid the need for a type hint, cleaner imo

src/zeroband/inference/toploc.py
mikasenghaas: basically the first two arguments to a pre-hook are `module` and `inputs`. the first one we don't need right now, hence the underscore. the toploc arg is just there to access it as a local var but we need to partially apply the function to pass as a hook. i'm not sure we will run into the issue you point out. the only danger is if inputs changes with a vllm bump, but there is nothing really we can do except for paying close attention to how their interface changes (and have tests that catch this)

tests/units/test_toploc.py
```diff
@@ -0,0 +1,130 @@
...
+
+    # Generate proofs for all sequences
+    toploc.maybe_generate_proofs_in_background(force_generate=True)
+    toploc.wait_for_proofs()
+
+    # Check TOPLOC cached proofs
+    proof = list(toploc.proofs.values())[0]
+    assert proof == [
+        b'\xff\xd9\xe3\xed\xdf\xe3]\x86\xbc\xa6\r\x0e\x1e\xf5\xfaM\x0e\xc5\xd5\xd8\xb9%\xc7\xfc\x1e\x1a:\xb0\x9f\xe6\x8b\x0f\x95I\xb5\xbc\xc4\x8eG\xc9y\x19\x1an|\xdd+\x04\x90\xc6d\xb5<r\xbb\xb3\xefK\xc7\x94+N\x16{\xfe~\x93\xe7\xfbi\x00\x1b\x1f\xdc\xbfq6&^<\xbb|\xe8&a5\xa3Q~u\x85\x87\x07\xb3\xfd\x95[\xb0\x01\xc6lE=\x00\x11\x99\xb8J\x84\xcc\x16\x0e;F\xe4\xc5\xf4\x8a@\t7mdE_\x99|\x17\xd4+^\xc7\x96\x8b3\xbeaF1P\xd4x\xa2\x06\xc5\xd1\x88pdm\x02\x84\xb7\x10\x01\x18\xc5\xf2V\xe3\x7f\xf6zF\x03<\xc7\x81"\xeetJ\x11\x7fd\x9ctO\xbfD\xbbK\xaf\xfdU\xbdd\x0b\xbez\x1c\xe5\x05\x0e\x93\xba\xd4\x1d\xcbZE\t\xa0\xc4\x15v\xea-\x96o\xe3{\xdb)TIJ\x04Pk\xc3<\x81\xe0\xf3\x8d\tr\xcb:\xb6\x10\x82\x12\xd0=\x1a\xec\x1a\xe8is\x9a\x987hW\xbd\x95\xfaX\x81/CaNo\x9c'
+    ]
```
samsja: I would put this into a const var and add a comment that it was generated with commit HHH so that we remember how to generate it again

tests/units/test_toploc.py
```diff
@@ -0,0 +1,152 @@
...
+
+    # Check TOPLOC cached proofs
+    assert len(toploc.proofs) == max_seqs
+    expected_num_proofs = (num_output_tokens + max_len - 1) // max_len
+    assert all(map(lambda proof: len(proof) == expected_num_proofs, toploc.proofs.values()))
+
+    # Check concatenated byte string proofs
+    proofs = [b"".join(proofs) for proofs in sorted(toploc.proofs.values())]
+    assert len(proofs) == max_seqs
+    assert all(map(lambda proof: type(proof) is bytes, proofs))
```
samsja:
```suggestion
    assert all([type(proof) is bytes for proof in proofs])
```

samsja (approved):
lfgtm

samsja (approved):

src/zeroband/inference/toploc.py
```diff
@@ -1,10 +1,16 @@
...
 from concurrent.futures import ThreadPoolExecutor, Future, wait
+from vllm.model_executor import SamplingMetadata
+from vllm.model_executor.layers.logits_processor import _prune_hidden_states
+from vllm import LLM
 
+from functools import partial
 
-class TopLocCache:
+
+class TOPLOC:
```
Jackmin801: Dont quite get the name change here

src/zeroband/inference/toploc.py
```diff
@@ -14,34 +20,32 @@ class TopLocCache:
 
     Args:
         max_seqs (int): Maximum number of sequences that can be stored in the cache
-        max_len (int): Maximum length of each sequence
         hidden_size (int): Size of the hidden dimension for each sequence element
```
Jackmin801, line 23: The max_len argument is still relevant right?

src/zeroband/infer.py
```diff
@@ -388,38 +393,14 @@ def inference(config: Config):
...
-    if config.dtype == "fp32":
-        config.toploc = False
-
-    model = llm.llm_engine.model_executor.driver_worker.model_runner.model
-    toploc_cache = TopLocCache(
-        max_seqs=config.batch_size * config.sampling.n,
-        max_len=32,
-        hidden_size=model.config.hidden_size,
+    # Setup TOPLOC
+    toploc, _ = setup_toploc(
```
Jackmin801: name change is propagated here as well

Jackmin801 (approved):
nice! looks great. happy to have tests now hehe

src/zeroband/inference/toploc.py
mikasenghaas: reverted

src/zeroband/infer.py
mikasenghaas: yep, missed that. thanks!

PR #274, opened 2025-05-08
[PRI2-483] W&B QoL changes

kalomaze:
Segmented metrics into different panels:

- train
- rewards
- losses
- lengths
- completions logging implementation (every ~5 steps)

image


src/zeroband/train.py
```diff
@@ -176,13 +176,139 @@ def get_logprobs(model: ModelType, input_ids: torch.Tensor, position_ids: torch.
     return logprobs
 
 
+def log_to_wandb(metrics_dict, tokenizer=None, batch=None, step=None):
+    """Log metrics to wandb with proper grouping."""
```
samsja: lets move this to src/zeroband/training/utils.py

src/zeroband/train.py
```diff
@@ -176,13 +176,139 @@ def get_logprobs(model: ModelType, input_ids: torch.Tensor, position_ids: torch.
...
+            if k in metrics_dict and metrics_dict[k] is not None:
+                wandb_metrics[f"{prefix}{k}"] = metrics_dict[k]
+
+    # Log everything
+    wandb.log(wandb_metrics, step=metrics_dict.get("step", step))
+
+
+def log_prompt_response_samples(tokenizer, batch, step, sample_history=None):
+    """Log samples using wandb.Table with accumulated history.
+    Only logs every 5 steps to reduce overhead.
```
samsja: same lets move  this to src/zeroband/training/utils.py

src/zeroband/train.py
```diff
@@ -422,6 +561,10 @@ def train(config: Config):
                 loss = loss / num_grad_acc_steps
 
                 inputs_ids_shape = input_ids.shape
+
+                # REMOVED: WandB logging from here - it's now done once before the grad_acc_step loop
+
+                # Now we can delete the batch data
```
samsja: hmm I think its still better to keep at after al the grad step, I imagine that you had to move it up because of the deletete batch here right ? 

if yes lets just remove batch from the del

kalomaze: this comment might be a residual artifact of my revisions and not be applicable to the current version of the code

samsja: feel free to delete

src/zeroband/train.py
mikasenghaas: also, can we type these

src/zeroband/train.py
```diff
@@ -176,13 +176,139 @@ def get_logprobs(model: ModelType, input_ids: torch.Tensor, position_ids: torch.
...
+            if sample_history["rewards"] is not None:
+                table_data["reward"] = sample_history["rewards"]
+            if sample_history["task_rewards"] is not None:
+                table_data["task_reward"] = sample_history["task_rewards"]
+
+            # Create DataFrame
+            df = pd.DataFrame(table_data)
+
+            # Create and log the table with accumulated data
+            table = wandb.Table(dataframe=df)
```
mikasenghaas: we have to recreate the full table on every logging step?!

src/zeroband/train.py
```diff
@@ -176,13 +176,139 @@ def get_logprobs(model: ModelType, input_ids: torch.Tensor, position_ids: torch.
...
+
+            # Update last logged step
+            sample_history["last_logged_step"] = step
+
+            print(f"Logged table with {len(df)} rows at step {step}")
+
+        # Return the updated history for the next call
+        return sample_history
+
+    except Exception as e:
```
mikasenghaas: does this exception throw frequently? would this be associated to wandb internal errors?

src/zeroband/train.py
kalomaze: wandb limitation :(

src/zeroband/train.py
kalomaze: mainly just don't want to risk any server side jank happening with wandb causing the run itself to fail but that's probably preemptive(?)

src/zeroband/train.py
kalomaze: done

src/zeroband/train.py
kalomaze: also done

kalomaze:
Should be ready
also please note that by default these do not show up on wandb for whatever reason, but this is trivially reconfigurable:

image
image
image

src/zeroband/training/utils.py
```diff
@@ -227,3 +231,132 @@ def reshard_module(model: torch.nn.Module):
...
+            # Create and log the table with accumulated data
+            table = wandb.Table(dataframe=df)
+
+            # Log using step in wandb.log call
+            wandb.log({"completions": table}, step=step)
+
+            # Update last logged step
+            sample_history["last_logged_step"] = step
+
+            # print(f"Logged table with {len(df)} rows at step {step}")
```
justusmattern27: i assume this can be deleted

src/zeroband/train.py
```diff
@@ -331,6 +349,10 @@ def train(config: Config):
                 loss = loss / num_grad_acc_steps
 
                 inputs_ids_shape = input_ids.shape
+
+                # REMOVED: WandB logging from here - it's now done once before the grad_acc_step loop
```
justusmattern27: would also delete this comment

justusmattern27 (approved):

justusmattern27:
would delete a few comments but otherwise lgtm

mikasenghaas (approved):
lgtm too

src/zeroband/train.py
kalomaze: done, + pruned some other redundant ones

src/zeroband/training/utils.py
kalomaze: yup

mikasenghaas:
Renamed `total_problems` -> `total_samples` because we call a problem during inference as one prompt ID from the dataset for which we generate multiple rollouts, which make up the training samples. Also made it so the count starts at 0 by default.
Screenshot 2025-05-16 at 2 57 19 PM
Screenshot 2025-05-16 at 2 58 51 PM

kalomaze:
> Renamed `total_problems` -> `total_samples` because we call a problem during inference as one prompt ID from the dataset for which we generate multiple rollouts, which make up the training samples. Also made it so the count starts at 0 by default. Screenshot 2025-05-16 at 2 57 19 PM Screenshot 2025-05-16 at 2 58 51 PM

good stuff, i noted the total problems counter being wack as well but didnt think to clean that up in this one

PR #272, opened 2025-05-08
[PRI2-493] Switch to Qwen3 family for debugging/ testing + clean up configs

mikasenghaas:
This PR switches to using the Qwen3 family for debugging and testing purposes. This replaces our [random weight small Llama models](https://huggingface.co/PrimeIntellect). Also cleans up the training and inference configs and updates the README.

PR #277, opened 2025-05-09
Fix wrong dtype 

mikasenghaas:
Revert change from (#272) which casts model to bf16 on load. We cannot do this for FSDP training.


mikasenghaas:
Last Commit (3c38caf, Green Line)
Second Last Commit (8579a48, Pink Line)
This PR (Orange Line)

Screenshot 2025-05-09 at 4 33 20 PM

PR #284, opened 2025-05-13
Protocol testing

mikasenghaas:


Dockerfile
```diff
@@ -77,5 +77,5 @@ ENV PATH="/root/prime-rl/.venv/bin:$PATH"
 COPY --from=builder /root/prime-rl/src ./src
 COPY ./configs/ ./configs/
 
-ENTRYPOINT ["python", "src/zeroband/inference.py"]
-CMD ["@", "configs/inference/Qwen1.5B/multistep.toml", "--output_path", "/share"]
+ENTRYPOINT ["python", "src/zeroband/infer.py"]
+CMD ["--help"]
```
samsja: what is this --help ?

src/zeroband/inference/pipeline.py
```diff
@@ -55,6 +55,7 @@ def setup_comm(world_size: int, iroh_seed: int | None, iroh_peer_id: str | None)
 
     # Setup node (with or without seed)
     if iroh_seed is not None:
+        logger.debug(f"Using IROH seed: {iroh_seed}")
```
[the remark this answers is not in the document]
samsja: This copilot is annoying, can we disable it @Jackmin801

Dockerfile
mikasenghaas: ooh, yeah i forgot we don't use `argparse` so this flag actually does nothing lol. the idea was that if the docker image is run without any args then it should print a list of all available arguments + descriptions, but this is not supported with pydantic based configs ig. will revert

samsja (approved):
lgtm

PR #287, opened 2025-05-16
Sort imports

mikasenghaas:
Enables import sorting via `ruff` by setting the `"I"` flag in `pyproject.toml`. In this way running 

```bash
uv run ruff check --fix .
```

will lint, check format and sort imports across the entirely project. The ruff pre-commit hook automatically sorts imports.


mikasenghaas:
Note: Need to be careful with W&B logs - they are put in `wandb` directory which makes ruff treat it as a local package which leads to wrong sorting (detected by CI)

samsja (approved):

PR #288, opened 2025-05-19
Add sentence unscrambling and ASCII tree formatting tasks

kalomaze:
### 1. Sentence Unscrambling

Requires the model to rearrange a series of text blocks with a scrambled order such that it matches the original order of the ground truth text. A 2024 randomly split subset of FineWeb-Edu was chosen for this first data iteration.

Minimum blocks/sentences included in the current dataset: 4
Maximum: 10

**Example Input:**
````
Rearrange these scrambled blocks of text into a coherent order:

```
*1*: Apple revolutionized smartphones with the iPhone, introducing touchscreens to the mass market.
*2*: Before smartphones, most people used basic mobile phones primarily for calls and text messages.
*3*: Today, smartphones are essential tools that billions of people worldwide use daily.
*4*: Following the iPhone, companies like Samsung and Google developed their own smartphone platforms.
```

Keep the block numbers in your answer, and wrap it in <unscrambled_text> tags. Keep the order of the numbers the same in the output, but change the order of the content so that it makes the most chronological sense.
````

**Expected Output:**
```
<think>
*...reasoning...*
</think>

<unscrambled_text>
*1*: Before smartphones, most people used basic mobile phones primarily for calls and text messages.
*2*: Apple revolutionized smartphones with the iPhone, introducing touchscreens to the mass market.
*3*: Following the iPhone, companies like Samsung and Google developed their own smartphone platforms.
*4*: Today, smartphones are essential tools that billions of people worldwide use daily.
</unscrambled_text>
```

image
*The reward function for this task focuses on the longest correct consecutive ordering of sentences, encouraging the model to construct coherent sequences rather than isolated correct placements.*

### 2. ASCII Tree Formatting

This task requires the model to convert a textual description of a directory structure into an ASCII tree format. The descriptions were generated programatically rather than synthetically; the latter option would ultimately be better for future dataset variation once verifying fuzzy accuracy (i.e. through classifiers/LLM as judge esque approaches) starts to mature in our pipelines.

**Example Input:**
````
Your task is to convert an English description into an ASCII tree format using this style:

```
fonts
+--Inter
|  '--static
+--Space_Grotesk
|  '--static
'--Space_Mono
```

Convert this description:

precompiledScriptPlugins-externalPlugins is a base directory that holds 3 subdirectories: groovy, kotlin and tests.
In the root directory, 'groovy' contains 1 subdirectory elements: buildSrc.
At level 2, in directory path 'groovy → buildSrc', 'buildSrc' retains 1 directory entries: src.
At level 3, in directory path 'groovy → buildSrc → src', Within 'src', you'll find 1 contained folders for code structure.
In nested path 'groovy → buildSrc → src → main', 'main' stores 1 contained folders: groovy.
At level 5, in directory path 'groovy → buildSrc → src → main → groovy', 'groovy' awaits content.
At the main level, 'kotlin' keeps 1 folders: buildSrc.
In nested path 'kotlin / buildSrc', 'buildSrc' maintains 1 subfolders: src.
In nested path 'kotlin → buildSrc → src', 'src' encompasses 1 subdirectories that organize the source code.
At level 4, in directory path 'kotlin / buildSrc / src / main', 'main' incorporates 1 directories: kotlin.
At level 5, in directory path 'kotlin → buildSrc → src → main → kotlin', The 'kotlin' data structure has nothing inside.
At the top level, The 'tests' folder holds no subdirectories at the moment.
````

**Expected Output:**
````
<think>
*...reasoning...*
</think>

<ascii_formatted>
precompiledScriptPlugins-externalPlugins
+--groovy
| `--buildSrc
|   `--src
|     `--main
|       `--groovy
+--kotlin
| `--buildSrc
|   `--src
|     `--main
|       `--kotlin
`--tests
</ascii_formatted>
````

image

*The reward function uses difflib to compare the model's output with the ground truth, weighting both overall similarity (30%) and the longest continuous matching block (70%). Additional penalties are applied if the tree structure is invalid or missing distinctive elements seen in the example, like the "--" and "|" characters.*

### CODE REVIEW

File: configs/inference/simple_ascii_tree.toml [hunk 1]
```diff
@@ -0,0 +1,11 @@
...
+dataset = "kalomaze/ascii-tree-mix-it1"
+batch_size = 16
+dp = 4
+rollout_path = "outputs"
+output_path = "data_rollout"
+max_model_len = 2048
+
+[sampling]
+temperature = 1.0
+n = 16
```
samsja, line 11: any reason for having different sampling param that the default for this task ? Or was it just for debugging

configs/training/simple_ascii_tree.toml
```diff
@@ -0,0 +1,23 @@
...
+[train]
+micro_bs = 2 # change to 8 for H200
+reshard_after_forward = true
+
+[optim]
+batch_size = 64
+warmup_steps = 20
+total_steps = 100000000000000
+step_per_rollout = 4
+grad_norm_clip = 0.0001
```
samsja, line 13: interesting, woud love to see some ablation later on on the grad norm clip value :)

src/zeroband/inference/genesys/ascii_tree_formatting.py
```diff
@@ -0,0 +1,48 @@
+from typing import Dict
+import re
+import difflib
+
+
+def compute_reward(completion: str, verification_info: Dict):
+    # Extract answer
```
samsja, line 7: can u add a short docstring explaning how the rewards is computed ? No need to be super specific tho

src/zeroband/inference/genesys/unscramble_sentence.py
```diff
@@ -0,0 +1,62 @@
+from typing import Dict
+import re
+
+
+def compute_reward(completion: str, verification_info: Dict):
```
samsja, line 5: same please add short docstring

configs/training/simple_ascii_tree.toml
willccbb: have been playing with diff grad norms as well on my own tasks as per @kalomaze 's suggestion and am also finding that heavy clip (at least 1e-3) coupled with higher LR is more stable and reaches higher max values vs. more typical settings (on a text reversal task which i will PR here soon :)). kalo has done enough experimentation here that i think this is totally fine for now, we can do more sweeps before an intellect-2.5/3 run.

seems like maybe a sensible alternative to KL penalty without the memory overhead.

justusmattern27:
agree with sami but looks good to me beyond that; let's please rebase so that we don't have the wandb changes in this PR

willccbb (approved):
Tasks both look sensible + in line with the existing format. LGTM

mikasenghaas (commented):
agree with comments made before - looks good!

PR #294, opened 2025-05-21
update wandb logs to avoid import conflict

samsja:


src/zeroband/train.py
```diff
@@ -147,7 +147,7 @@ def train(config: Config):
     training_progress = TrainingProgress(total_tokens=0, step=config.start_step, total_samples=total_samples)
 
     if world_info.rank == 0 and config.wandb:
-        wandb.init(project=config.project, config=config.model_dump())
+        wandb.init(project=config.project, config=config.model_dump(), dir="wandb_logs")
```
mikasenghaas, line 150: make an arg with default value?

samsja: we will never change this args tho

PR #295, opened 2025-05-21
Git clone via HTTPS in install script

mikasenghaas:


mikasenghaas:
Keep SSH setup but pass `ssh -A ...` flag to forward all keys from the local SSH agent to the remote machine, which will automatically authenticate.

PR #296, opened 2025-05-21
reasoning gym, code understanding, bump to python 3.11

justusmattern27:
dataset: https://huggingface.co/datasets/justus27/reasoning-gym-genesys/


src/zeroband/inference/genesys/reasoning_gym.py
```diff
@@ -0,0 +1,129 @@
+import json
+import re
+from reasoning_gym.factory import get_score_answer_fn
```
samsja, line 3: Probably need to add this to pyporject.toml

pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "zeroband"
 version = "0.1.0"
 description = "ZeroBand is a production ready codebase for decentralized training of LLM"
 readme = "README.md"
-requires-python = ">=3.10, <3.11"
+requires-python = "==3.11.*"
```
samsja, line 6:
```suggestion
requires-python = ">=3.10"
```

justusmattern27: we're intentionally bumping to 3.11 as discussed

samsja: sg yeah

src/zeroband/inference/genesys/reasoning_gym.py
```diff
@@ -0,0 +1,134 @@
...
+    if len(split_response) == 1:
+        return None
+    tail = split_response[1].strip()
+    json_str = _find_last_json_block(tail)
+    if json_str is None:
+        return None
+    try:
+        return json.loads(json_str)
+    except json.JSONDecodeError:
+        return None
```
Jackmin801: these function are duplicated in `code_output_prediction.py`. what do u think about putting them in one place like code_utils?

src/zeroband/inference/genesys/code_utils.py
```diff
@@ -709,10 +723,10 @@ def check_correctness(in_outs: dict | None, generation, timeout=10, debug=True):
     The global timeout is to catch some extreme/rare cases not handled by the timeouts
     inside `run_test`"""
 
-    manager = multiprocessing.Manager()
+    manager = _fork_ctx.Manager()
     result = manager.list()
     metadata_list = manager.list()
-    p = multiprocessing.Process(target=_temp_run, args=(in_outs, generation, debug, result, metadata_list, timeout))
+    p = _fork_ctx.Process(target=_temp_run, args=(in_outs, generation, debug, result, metadata_list, timeout))
```
Jackmin801: do u happen to know why this is required? its kind of strange to me that it only works with fork and not spawn considering its a sandbox.

Jackmin801 (approved):
nice! lgtm. minor comments :)

src/zeroband/inference/utils.py
```diff
@@ -65,3 +67,22 @@ def generate_target_length_prompts(config: LenRewardsConfig, batch_size: int):
...
+    data = data.map(
+        _add_token_lengths_batched,
+        batched=True,
+        batch_size=tokenize_batch_size,
+        desc=f"Calculating prompt lengths to filter out lengths > {max_length}",
+    )
+
+    data = data.filter(lambda x: x["token_length"] <= max_length)
+
+    return data
```
apaz-cli, line 88: Should the `token_length` column stick around?

pyproject.toml
```diff
@@ -27,9 +27,11 @@ dependencies = [
     "shardcast >= 0.3.1",
     "asyncio>=3.4.3",
     "aiohttp>=3.10.5",
-    "pyext>=0.7",
+    "pyext @ git+https://github.com/justusmattern27/PyExt.git",
     "prime-iroh>=0.2.1",
-    "setuptools"
+    "setuptools",
+    "reasoning-gym @ git+https://github.com/open-thought/reasoning-gym.git",
```
mikasenghaas, line 33: why install from github here?

src/zeroband/infer.py
```diff
@@ -87,13 +87,19 @@ def inference(config: Config):
...
+        logger.info(f"✨ Removed long prompts - {len(dataset)} samples remaining")
+
     # Optionally filter dataset
     if config.difficulty_filtering:
         dataset = dataset.filter(
             lambda x: x[config.difficulty_filtering.solve_rate_field] >= config.difficulty_filtering.min_solve_rate
             and x[config.difficulty_filtering.solve_rate_field] <= config.difficulty_filtering.max_solve_rate
         )
 
+    os.environ["TOKENIZERS_PARALLELISM"] = "false"  # to avoid HF warning
```
mikasenghaas: what change introduced makes this necessary? can we move this to the `src/zeroband/utils/envs.py` (we pre-set a bunch of env variables here across the project). if it should only be set for inference could also live in `src/zeroband/inference/envs.py`

src/zeroband/inference/genesys/code_utils.py
```diff
@@ -18,6 +18,19 @@
 import numpy as np
 from pyext import RuntimeModule
 
+# private fork context just for the sandbox
+_fork_ctx = _mp.get_context("fork")
+
+
+# helper to make sure the child never trips world_info assertions
```
mikasenghaas: ahh i dont like this. world info should not be triggered by any inference code. can you trace back which import is triggering this?

src/zeroband/inference/genesys/reasoning_gym.py
```diff
@@ -66,12 +67,16 @@ def _extract_json_field(completion: str, field: str) -> str | None:
...
-    verification_info["reasoning_gym_entry"]["metadata"]["output"] = tuple([tuple(b) for b in verification_info["reasoning_gym_entry"]["metadata"]["output"]])
+    verification_info["reasoning_gym_entry"]["metadata"]["output"] = tuple(
+        [tuple(b) for b in verification_info["reasoning_gym_entry"]["metadata"]["output"]]
+    )
     return _extract_post_string(completion), verification_info
 
 
 def extract_answer_rearc(completion: str, verification_info: dict):
-    verification_info["reasoning_gym_entry"]["metadata"]["output"] = tuple([tuple(b) for b in verification_info["reasoning_gym_entry"]["metadata"]["output"]])
+    verification_info["reasoning_gym_entry"]["metadata"]["output"] = tuple(
```
mikasenghaas, line 36: imo good usecase for map, way easier to read that this is just a list of lists to tuple of tuple conversion. also, why is this needed?

src/zeroband/infer.py
mikasenghaas: oh this is the multiprocessed data filtering by prompt length right? yeah, i remember having a similar issue so probably fine to keep this env variable, but let's still move it to new file

pyproject.toml
samsja: no pip install faik

src/zeroband/inference/genesys/reasoning_gym.py
justusmattern27: it's a workaround for a bug in reasoning-gym

src/zeroband/infer.py
justusmattern27: ok sure

pyproject.toml
justusmattern27: yes there is but the most recent release does not yet have what we need

src/zeroband/inference/genesys/code_utils.py
justusmattern27: ah it turns out that this was solved with the tp fix as well, so removing it

src/zeroband/inference/genesys/reasoning_gym.py
justusmattern27: done

src/zeroband/inference/genesys/code_utils.py
justusmattern27: removed

pyproject.toml
JohannesHa: should we pin the commit hash tho?

PR #297, opened 2025-05-22
Add cleanup debug

samsja:


justusmattern27 (approved):

justusmattern27:
lgtm

Jackmin801 (approved):
lgtm

mikasenghaas (commented):
What would be the the use-case for not cleaning up? Point being, can we remove the CLI arg?

configs/inference/simple_math.toml
```diff
@@ -4,7 +4,9 @@ batch_size = 14
 dp = 6
 rollout_path = "outputs"
 output_path = "data_rollout"
+clean_output_path_start = true
```
mikasenghaas: Maybe just `clean_output_path`?

samsja: fair

samsja:
> What would be the the use-case for not cleaning up? Point being, can we remove the CLI arg?

two use case

* restart from ckpt 
* production where we never want to delete folder by mistakes (could lead to ckpt loss)

mikasenghaas (approved):
nice! merge

PR #298, opened 2025-05-22
add lcs reward

samsja:


src/zeroband/infer.py
```diff
@@ -71,7 +71,17 @@ def inference(config: Config):
 
     # Load  dataset
     logger.info(f"Loading dataset {config.dataset}")
-    dataset = load_dataset(config.dataset, split="train")
+
+    if config.dataset == "agentlans/wikipedia-paragraphs":
```
mikasenghaas: can we not preprocess the data, upload to hf and then add to config? cleaner imo

justusmattern27: I concur

justusmattern27 (changes requested):
lgtm but i agree that we should bring the dataset in the genesys format instead of adding an if statement for processing

src/zeroband/infer.py
samsja: hmmm I guess we can, but isn't it losing in hackability? 

I was thinking of having some kind of get_dataset function that do processing base on the dataset.

samsja:
> lgtm but i agree that we should bring the dataset in the genesys format instead of adding an if statement for processing

sure my bad. Also I am not sure how to do this genesys format

configs/training/simple_reverse_two_gpu.toml
```diff
@@ -0,0 +1,25 @@
...
+project = "reverse-debug"
+
+[train]
+micro_bs = 32
+reshard_after_forward = true
+
+[optim]
+batch_size = 64
+warmup_steps = 1
+total_steps = 100000000000000
```
mikasenghaas, line 12: does the trainer stop automatically? would be nice if successful ci runs in wandb show as successful and not as error, or have to be killed

src/zeroband/utils/models.py
```diff
@@ -43,6 +43,8 @@
     "mikasenghaas/Qwen3-14B-1.2",
     "mikasenghaas/Qwen3-30B-A3B-0.2",
     "mikasenghaas/Qwen3-30B-A3B-1.2",
+    # Reverse SFT
+    "willcb/Qwen2.5-0.5B-Reverse-SFT",
```
mikasenghaas: agree that we should probably move away from model literal, getting kinda annoying. for another pr tho

tests/integration/test_short_rl_training.py
```diff
@@ -0,0 +1,47 @@
...
+import subprocess
+
+import pytest
+
+pytestmark = [pytest.mark.gpu, pytest.mark.slow]
+
+
+@pytest.fixture(scope="session")
+def username():
+    return os.environ.get("USERNAME_CI", os.getlogin())
```
mikasenghaas: how does the auth to wandb work in ci?

tests/integration/test_short_rl_training.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@pytest.fixture(scope="session")
+def branch_name():
+    branch = subprocess.check_output(["git", "rev-parse", "--abbrev-ref", "HEAD"]).decode("utf-8").strip()
+    return branch
+
+
+@pytest.fixture(scope="session")
```
mikasenghaas: do these need to be fixtures? can they not just live in the test

configs/training/simple_reverse_two_gpu.toml
samsja: added a stop criteria in the tests

tests/integration/test_short_rl_training.py
samsja: I added a token in the ci as env var (wandb token)

tests/integration/test_short_rl_training.py
samsja: they could but not sure what wouldbe the benefit

justusmattern27 (approved):

script/preprocess_dataset_example.py
mikasenghaas, line 1: do we want this file to live in prime-rl? maybe keep internal for now, and at some point this will live in the proper place, e.g. genesys/ verifier or similar

tests/e2e/test_short_rl_training.py
mikasenghaas, line 1: oh, do we want to move to using the same fixtures for running parallel processes as the integration tests use? less code duplication and more features, like timeouts, using process outputs across multiple tests etc.

tests/unit/training/test_model.py
```diff
@@ -4,16 +4,19 @@
 from zeroband.utils.models import get_model_and_tokenizer
 
 BS = 1
-SEQ_LEN = 16
+SEQ_LEN = 8
+
+pytestmark = [pytest.mark.gpu, pytest.mark.skip]
```
mikasenghaas: why are we skipping here?

samsja: oups

tests/e2e/test_short_rl_training.py
samsja: ah yeah true

script/preprocess_dataset_example.py
samsja: I can remove it yeah

mikasenghaas:
Screenshot 2025-05-26 at 9 42 25 PM

PR #301, opened 2025-05-23
add wandb_logs folder to .gitingore

samsja:


mikasenghaas (approved):
sure, still not a big fan of the output dir structure haha. i think all runtime outputs should be in one folder and we set the inference and training configs accordingly , but we can delay that for now

samsja:
> sure, still not a big fan of the output dir structure haha. i think all runtime outputs should be in one folder and we set the inference and training configs accordingly , but we can delay that for now

hmm hmm hmm lets see later I think its okay for now

PR #304, opened 2025-05-24
fix 3.11 update

samsja:


.github/workflows/tests.yaml
```diff
@@ -41,6 +41,6 @@ jobs:
           cache-dependency-glob: "uv.lock"
 
       - name: Install dependencies
-        run: uv sync && uv sync --extra fa
+        run: uv venv --python 3.11 && uv sync && uv sync --extra fa
```
mikasenghaas: this should not be necssary?

install.sh
```diff
@@ -46,6 +46,7 @@ main() {
     fi
     
     log_info "Installing dependencies in virtual environment..."
+    uv venv --python 3.11
```
mikasenghaas: also here. doesn't it respect the pyproject which pins to 3.11?

.github/workflows/tests.yaml
samsja: hmmm based on what ?

mikasenghaas (approved):

Jackmin801 (approved):
nice! why did the uv lock change though?

samsja:
> nice! why did the uv lock change though?

I think that the previous pr did not update the uv lock accordingly

PR #305, opened 2025-05-24
Pipeline fix and remove model literal

mikasenghaas:
Adds more model shards (used in benchmark) and fixes a critical bug for pipelines with intermediate nodes (>2 group size) where one hook was missing a required argument. Didn't catch yet because was only ever testing with PP=2 and we don't have CI for it yet. Probably worth adding one pipeline integration test with PP=4 @samsja?


samsja:
> Probably worth adding one pipeline integration test with PP=4 @samsja?
yes also would probably makes sense to use 4 GPU in CI

src/zeroband/utils/models.py
```diff
@@ -41,8 +41,20 @@
...
+    "mikasenghaas/Qwen3-32B-1.4",
+    "mikasenghaas/Qwen3-32B-2.4",
+    "mikasenghaas/Qwen3-32B-3.4",
     "mikasenghaas/Qwen3-30B-A3B-0.2",
     "mikasenghaas/Qwen3-30B-A3B-1.2",
+    "mikasenghaas/Qwen3-30B-A3B-0.4",
+    "mikasenghaas/Qwen3-30B-A3B-1.4",
+    "mikasenghaas/Qwen3-30B-A3B-2.4",
+    "mikasenghaas/Qwen3-30B-A3B-3.4",
 ]
```
samsja: as discussed offline lest remove this Literal its getting out of hand imo

samsja (approved):

src/zeroband/utils/models.py
mikasenghaas: yep

samsja (approved):
lgtm

PR #306, opened 2025-05-24
Add PP integration test

mikasenghaas:
PR to add multi-processed integration test for pipeline parallel. Still to do:
- Uses pre-sharded `Qwen3-0.6B` with PP=2
- Uses the same fixtures as other integration tests for running subprocesses in parallel (@samsja we should use the `run_processes` fixture for the RL training integration test)

For future:
Once we have a CI runner with 4 workers, we should test PP=4 to cover intermediate nodes behavior


tests/integration/conftest.py
```diff
@@ -1,32 +1,63 @@
...
+    except Exception as e:
+        raise e
 
-        return process
 
-    yield start_process
+def run_subprocesses_in_parallel(commands: list[Command], envs: list[Environment]) -> list[ProcessResult]:
+    """Start multiple processes in parallel using ProcessPoolExecutor and wait for completion."""
+    assert len(commands) == len(envs), "Should have an environment for each command"
+    with concurrent.futures.ProcessPoolExecutor(max_workers=len(commands)) as executor:
```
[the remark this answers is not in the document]
mikasenghaas: can't do because iroh endpoint has to run in process

tests/integration/inference/test_debug_pp.py
```diff
@@ -0,0 +1,99 @@
+from pathlib import Path
+from typing import Callable
+
+import pyarrow.parquet as pq
+import pytest
+
+from tests.integration import Command, Environment, ProcessResult
```
samsja, line 7: Ideally this should be imported from conftest.

Importing frm tests. will break at some point (ex running the test from a ide with debugger)

mikasenghaas: mmh, but these are type hints not fixtures. i think the only alternative is to not type the fixtures which im not a fan of. can you maybe try running in ide and see if breaks?

samsja: hmmm good point, lets keep it like this and fix it later if we have issue

samsja (approved):

PR #311, opened 2025-05-26
add wandb run param

samsja:


mikasenghaas (approved):
long term i think we should refactor to have a separate wandb subconfig as well, but good for now

samsja:
> long term i think we should refactor to have a separate wandb subconfig as well, but good for now

agree

PR #312, opened 2025-05-26
run math

samsja:


configs/training/math_only_5b.toml
```diff
@@ -0,0 +1,23 @@
...
+project = "intellect-2"
+
+collate_mode = "packing"
+
+[train]
+micro_bs = 4 # change to 8 for H200
+reshard_after_forward = true
+
+[data]
+seq_length = 16000
```
justusmattern27: seq_len should be 8k

configs/training/math_only_5b.toml
```diff
@@ -0,0 +1,23 @@
+model_name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B"
+project = "intellect-2"
+
+collate_mode = "packing"
+
+[train]
+micro_bs = 4 # change to 8 for H200
```
justusmattern27: not sure if micro_bs 4 will fit with 8k length

justusmattern27 (approved):

configs/training/math_only_5b.toml
samsja: I will check

PR #313, opened 2025-05-26
Improve inference logging and environment variable handling

mikasenghaas:


src/zeroband/infer.py
```diff
@@ -6,6 +6,10 @@
 import uuid
 from pathlib import Path
 
+# Import environment before any other imports
+# ruff: noqa: I001
+from zeroband.inference import envs
```
samsja, line 11:
```suggestion
from zeroband.inference import envs # ruff: noqa: I001

```

I assume if we put it at the top of the file it would disable for all import here ?

If I am wrong just click resolve

src/zeroband/infer.py
```diff
@@ -34,17 +37,23 @@
 def inference(config: Config):
     # Initialize the logger
     logger.info("Starting inference")
-    logger.info(f"TP={config.tp}, DP={config.dp}, PP={config.pp.world_size}")
+
+    # Log relevant configuration
+    logger.info(f"Model: {config.model_name}")
+    logger.info(f"Dataset: {config.dataset}")
+    logger.info(f"Parallelism: TP={config.tp}, DP={config.dp}, PP={config.pp.world_size}")
```
samsja, line 44: could be worth adding a pretty-print of the config here when using debug mode ?

samsja (approved):
lgtm

src/zeroband/infer.py
mikasenghaas: nope, its just the line below

src/zeroband/infer.py
mikasenghaas: yeah, im saving this for when we restructure the config. if we have proper subconfigs we can show a nice rich table/ console for each subconfig in an automated way but our config is too messy atm to do that.

PR #315, opened 2025-05-27
Refactor monitoring utils and implement SYN-2 metrics

mikasenghaas:
This PR refactors existing monitoring logic like `PrimeMetrics` (Unix socket logging) and `HttpMonitor` (API based logging) into a unified interface that is shared across the train and inference script and adds tests for it. The main idea is that the `Monitor` class has multiple (CLI-configurable) output handlers that it logs metrics to. In this way we can log to different sources (e.g. protocol worker via Unix socket, internal dashboard via API, debugging runs via file, etc.) in one call to `.log()`

```python
monitor = Monitor(monitor_config)
monitor.log({"step": 0, "loss": ...})
```

So far, I implemented three output types that can be toggled on and off via CLI arguments:
- `FileOutput`: Logs to file
- `SocketOutput`: Logs to Unix socket
- `APIOutput`: Logs to an API via HTTP

*For the future, I think it would be cool to move `wandb` logs into this format too. Another TODO is to allow for multiple instances of the same output class, e.g. logging the same metrics to different API endpoints.*

For example, if one wanted to log to a file, socket and an API from the inference script, one would run:

```bash
uv run python src/zeroband/infer.py @ configs/inference/debug.toml \
	--monitor.file.enable \
	--monitor.file.path file_monitor.jsonl \
	--monitor.socket.enable \
	--monitor.socket.path metrics.sock \
	--monitor.api.enable \
	--monitor.api.url http://0.0.0.0:8080/metrics \
	--monitor.api.auth_token tes
```

### Important Changes

There are some breaking changes, that might require attention from the protocol/ platform-side:

1. We always log a single metrics dictionary (`metrics: dict[str, Any]`) which is serialized using `json.dumps` before being written to file/ socket or posted to API. You can see the new format below. Previously, the `PrimeMetrics` class would write each key-value pair on a separate line as `{"key": key, "value": value}`, and the `HttpMonitor` would write a list of dictionaries of size 1 (unless the flush interval was set larger).

```txt
{"progress/batch_problems": 8, "progress/batch_samples": 64, "progress/batch_tokens": 1831}
{"performance/batch_tokens_per_second": 3323.6451193107387, "performance/batch_samples_per_minute": 6970.3971917822155, "performance/batch_avg_seq_length": 28.609375}
{"output/save_path": "outputs/step_0/0cc265c0-d5cc-4966-97fe-dcadf13f232a.parquet", "output/sha256": "c44d8831b9bc0090c0b455347da282e4113960666146406685b441bf63eae0c3"}
{"progress/batch_problems": 8, "progress/batch_samples": 64, "progress/batch_tokens": 2239}
{"performance/batch_tokens_per_second": 4499.759060206199, "performance/batch_samples_per_minute": 7717.317905847166, "performance/batch_avg_seq_length": 34.984375}
{"output/save_path": "outputs/step_1/096bc233-88c9-4be8-9349-591bae6e3c42.parquet", "output/sha256": "0fd552a790623d62a809a66f33a5499eca7b071db512dc886e4b69ae7de6f398"}
{"progress/batch_problems": 8, "progress/batch_samples": 64, "progress/batch_tokens": 1815}
{"performance/batch_tokens_per_second": 3800.7191182219453, "performance/batch_samples_per_minute": 8041.19086169271, "performance/batch_avg_seq_length": 28.359375}
{"output/save_path": "outputs/step_2/2f3d5cda-a96d-44ba-b67b-58e78491d66e.parquet", "output/sha256": "2dd42f1710c07112138ce80f611445dfeecb9afbf8eef4f7e2be5bb4b5deeace"}
{"progress/batch_problems": 8, "progress/batch_samples": 64, "progress/batch_tokens": 2849}
{"performance/batch_tokens_per_second": 5021.725348434069, "performance/batch_samples_per_minute": 6768.489062122438, "performance/batch_avg_seq_length": 44.515625}
{"output/save_path": "outputs/step_3/6e2f81da-0f16-4107-bda0-b58c622dc548.parquet", "output/sha256": "6696da9cf6deefb600a69fb54529bf3b19b486d1d4b9914357cc4a86f2bf8866"}
```

2. Renamed environment variables (and typed them in `src/zeroband/utils/envs.py`) to be more descriptive: 
- `PRIME_TASK_BRIDGE` -> `PRIME_SOCKET_PATH`
- `PRIME_API_BASE_URL` -> `PRIME_API_URL` (i.e. it takes the full API endpoint, not just the base for more flexibility)
- `PRIME_DASHBOARD_AUTH_TOKEN` -> `PRIME_API_AUTH_TOKEN`
- Removed `PRIME_DASHBOARD_METRIC_INTERVAL`

The full list of available environment variables is listed in `src/zeroband/utils/envs.py` - all variables that can/ should be set by the Prime worker are prefixed with `PRIME_*`

3. Right now, the `PRIME_TASK_ID` environment variable is ignored. If needed, we can log from the application (`infer.py` or `train.py`). Or, if all outputs need it, we can add to base class

4. No batching in API outputs anymore to align interface, I don't think we will run into the issue of flooding our server (especially since the batching was count based not time based)

5. I removed the deduplication method `_remove_duplicates` from the `HttpMonitor` class. I am not sure why it was there because we only logged metrics *once* per step (i.e. no duplicates) from `train.py`. If we do need this, this should be handled in the file not the monitoring classes imo.

### Inference Metrics

From the inference script we are monitoring the following values:

- System Metrics: `system/cpu_percent`, `system/memory_percent`, `system/memory_usage`, `system/memory_total`, `system/gpu_{i}_memory_used`, `system/gpu_{i}_memory_total`, `system/gpu_{i}_utilization`
- Progress Metrics: `progress/batch_problems`, `progress/batch_samples`, `progress/batch_tokens`
- Performance Metrics: `performance/batch_tokens_per_second`, `performance/batch_samples_per_minute` `performance/batch_avg_seq_length`
- Output Metrics: `output/save_path`, `output/sha256`

### Examples

**Example 1: Run with no metrics logged (default)**

```bash
uv run python src/zeroband/infer.py @ configs/inference/debug.toml
```

**Example 2: Log to file**

```bash
uv run python src/zeroband/infer.py @ configs/inference/debug.toml --monitor.file.enable --monitor.file.path outputs/file_monitor.jsonl
```

**Example 2: Log to unix socket**

Run fake socket in one terminal using `uv run fake_socket.py ` ([Gist](https://gist.github.com/mikasenghaas/03b65ab27947e9735a732d644d070219)). Then, run

```bash
uv run python src/zeroband/infer.py @ configs/inference/debug.toml --monitor.socket.enable --monitor.socket.path metrics.sock
```

**Example 3: Log to API**

Run fake API in one terminal using `uv run fake_api.py` ([Gist](https://gist.github.com/mikasenghaas/25204c771c183687116350593af4cbde)). Then, run

```bash
uv run python src/zeroband/infer.py @ configs/inference/debug.toml --monitor.api.enable --monitor.api.url http://0.0.0.0:8080/metrics --monitor.api.auth_token test
```

Ticket: PRI2-517


src/zeroband/infer.py
```diff
@@ -242,25 +242,39 @@ def inference(config: Config):
...
-        avg_sequence_length = batch_tokens / num_batch_samples
+        batch_avg_seq_length = batch_tokens / num_batch_samples
         logger.info(
-            f"Batch throughput: {batch_throughput:.2f} tok/sec ({batch_tokens} tokens in {end_time - start_time:.2f}s, avg seq len: {avg_sequence_length:.1f})"
+            f"Batch throughput: {batch_throughput:.2f} tok/sec ({batch_tokens} tokens in {end_time - start_time:.2f}s, avg seq len: {batch_avg_seq_length:.1f})"
         )
 
+        # Log performance metrics
+        perf_metrics = {
+            "performance/batch_throughput": batch_throughput,
```
manveerxyz: nit: might be worth adding the unit here just for clarity

src/zeroband/inference/metrics.py
```diff
@@ -0,0 +1,185 @@
...
+from zeroband.utils.logger import get_logger
+
+# Module logger
+logger = get_logger("INFER")
+
+
+class MetricsConfig(BaseConfig):
+    # Type of metrics to log
+    type: Literal["file", "socket", "null"] = "null"
+
```
samsja: would it be a way to log both to socket and to stdout ?

src/zeroband/infer.py
mikasenghaas: in the key string or just as a comment?

src/zeroband/inference/metrics.py
mikasenghaas: oh yea, so the whole point was that we can add multiple "metric handlers" that log identical metrics to different sinks which is configurable by just passing a list of metric types via cli. i imagine the research team having a server that listens to metrics from all workers and aggregates them for internal use

mikasenghaas:
> I am not sure that we ever need to log to file.
> 
> I would rather suggest to :
> 
> Only have one instance of the metric class which is the PrimeMetric ( like it is currently) and just log with the logger to stoud.
> 
> I am not a fan of adding abstract class before we actually need them

I agree that we probably don't want to log to file. See it more as a POC for what could be a REST API server hosted by the research team to get live statistics on how the run is going. In production we would want to log the metrics to the protocol worker via Unix socket and to the research team via REST API and configure both of these via CLI without any code changes.

We already log to stdout from all subclasses.

samsja:
> it more as a POC for what could be a REST API server hosted by the research team to get live statistics on how the run is going

I would suggest reviving this abstraction when we move away from a PoC phase.  I am very much against early abstraction in training codebase. Basically each time I was inheritance and abstract class I ask myself if I really need this right now

mikasenghaas:
Yeah, I get your point. I guess my point is that if we want to monitor SYNTHETIC-2 (which I think we do) this abstraction becomes useful because we can call one method `metrics.log({...})` in the inference code which logs to all configured endpoints. I think this is much cleaner and more maintainable than adding a custom metrics class for each type of logging we do (see http monitor, prime metrics, etc.) with custom code from the application.

samsja:
> Yeah, I get your point. I guess my point is that if we want to monitor SYNTHETIC-2 (which I think we do) this abstraction becomes useful because we can call one method `metrics.log({...})` in the inference code which logs to all configured endpoints. I think this is much cleaner and more maintainable than adding a custom metrics class for each type of logging we do (see http monitor, prime metrics, etc.) with custom code from the application.

I get your point, lets merge it then.

I guess that I did not saw the point of having the file metrics and hence we could just live with just the socket. But also I realized that the file metrics is probably useful for debugging / developing new metrics

src/zeroband/infer.py
mikasenghaas: Explictly sets `batch_tokens_per_second` and `batch_samples_per_minute`

mikasenghaas:
Screenshot 2025-05-27 at 4 51 59 PM

src/zeroband/utils/monitor.py
```diff
@@ -0,0 +1,259 @@
...
+        self.config = config
+        self.lock = threading.Lock()
+        logger.debug(f"Initializing {self.__class__.__name__} ({str(self.config).replace(' ', ', ')})")
+
+    def _serialize_metrics(self, metrics: dict[str, Any]) -> str:
+        return json.dumps(metrics)
+
+    @abstractmethod
+    def log(self, metrics: dict[str, Any]) -> None:
+        pass
```
samsja: I am not sure about the term output here. Took me couple of minutes to understand what it does

samsja:
```suggestion
        ...
```

samsja: I am not convinced anymore that we need this base class. I think that the code is minimal enough to duplicate everywhere. 

It's essentially just a json.dumps + lock, this does not justify a class IMO.

If we want to define a behavior I would suggest rather to create a Protcol


(you can have tell but I really don't like inheritance when its not absolutely needed).

mikasenghaas: fair let me think about it

samsja (approved):
lgtm

PR #317, opened 2025-05-27
Fix race condition in PrimeMetric initialization causing AttributeError

JannikSt:
Fixes a race condition in PrimeMetric.__init__() where the background metrics thread was starting before socket_path was initialized, causing "AttributeError: 'PrimeMetric' object has no attribute 'socket_path'".


samsja (approved):
lgtm

mikasenghaas (commented):
I actually also found this bug and will be fixed as part of #315 with a general refactor or our monitoring tools. Keeping open for now as reference.

PR #320, opened 2025-05-27
Sanskrit Literature Retrieval 

Shashi456:
I just saw that #319 exists; I'm happy to close this in case there's a conflict. 

https://huggingface.co/datasets/paws/sanskrit-verses-gretil - Here's the dataset. Mostly XML extractions, but I think there can be more extractions, and even these have a few deformed ones. The HTML extractions will need per-book basis scripts, which I am working on. 

Tried training a Gemma-2B-IT with rollouts from Qwen3-8B, but the results weren't great. This was just 200 steps, so I couldn't exactly see any learning happening here. But qwen3 did guess a bunch of the genre's right at least when it came to building the rollout dataset [reward > 0.3] and pass@8.

image


Abhi12122000:
@Shashi456 I'm currently extracting the HTML files in GRETIL too, though in a slightly different format. Have mostly done the Vedas for now, but am expanding. Here's the dataset:- https://huggingface.co/datasets/the-ak-2000/vedic-texts

Would you want to merge the datasets? My source extraction was quite granular, so I can modify my dataset a little to match whatever's standard.

ghost:
No conflict at all! When we parsed the books, we actually tried to go for simplicity due to the sheer amount of data in the data bank. Do you think it would be feasible to train on enough data for a model to make such fine grained predictions about the position of a quote? Interested in your opinion as well @Abhi12122000

Abhi12122000:
Since I parsed the Vedas dataset, I can speak for that. 
1.) I think we could manage to train genre/subgenre, book, chapter, verse (optional). Anything finer than that might be difficult to pick up (but I don't have any data to back that up). 
2.) Also, along with just the numbers for book, chapter, verse (for eg. 11, 3, 14), maybe we could describe (and score) the source of origin more descriptively?
For example, for a sample text, we could score based on similarity to the following source:- 
`"WorkTitle": "rigveda padapatha", "Genre": "padapatha", "Book": "Mandala 8", "Chapter": "Sukta 24", "Verse": "Richa 17"`, instead of just `padapatha, 8, 24, 17`.
For another piece of text belonging to the Brahmana, we could have, for example:-
`"WorkTitle": "satapatha brahmana (madhyandina)", "Genre": "brahmana", "Book": "Kanda 14", "Chapter": "Adhyāya 2", "Verse": "Brāhmaṇa 1, Khaṇḍikā 17"`, instead of just `brahmana, 14, 2, 1, 17(optionally)`.
This diversity (even in Sanskrit terms) might ensure that the model learns the sources correctly, instead of just spitting random numbers.
This definitely makes the reward calculation much more difficult, but can be tackled. 
Just thoughts, what do you guys think @Shashi456 @apokryphosx?

PR #323, opened 2025-05-28
[PRI2-562] Use safe serialization in PP

mikasenghaas:
Implements safe serialization for communication between pipeline stages:
- `serialize_tensors`/ `deserialize_tensors` uses `safetensors`
- `serialize_sampler_output`/ `deserialize_tensors` uses `msgspec` (this is the vLLM default anyway, so supported out of the box)

### DIFF
--- a/src/zeroband/inference/pipeline.py
+++ b/src/zeroband/inference/pipeline.py
@@ -1,12 +1,13 @@
-import pickle
 import time
 from functools import partial
 from typing import Tuple
 
+import msgspec
 import torch
 import torch.nn as nn
 from prime_iroh import Node
 from pydantic_config import BaseConfig
+from safetensors.torch import load, save
 from vllm import LLM
 from vllm.model_executor.layers.sampler import SamplerOutput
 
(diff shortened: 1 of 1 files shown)


src/zeroband/inference/pipeline.py
```diff
@@ -26,6 +27,29 @@ class PipelineConfig(BaseConfig):
     iroh_peer_id: str | None = None
 
 
+def serialize_tensors(tensor_dict: dict[str, torch.Tensor]) -> bytes:
+    """Safely serializes a tensor to bytes."""
+    return save(tensor_dict)
```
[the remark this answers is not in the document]
mikasenghaas: I think `safetensors` handles this for us - at least in integration/ e2e testing there were no problems and the vLLM tensors we send are definitely GPU tensors.

src/zeroband/inference/pipeline.py
```diff
@@ -26,6 +27,29 @@ class PipelineConfig(BaseConfig):
...
+    """Safely deserializes a dictionary of tensors from bytes."""
+    tensor_dict = load(data)
+    if device is not None:
+        return {key: tensor.to(device) for key, tensor in tensor_dict.items()}
+    return tensor_dict
+
+
+def serialize_sampler_output(output: SamplerOutput) -> bytes:
+    """Safely serializes a vLLM SamplerOutput object"""
+    return msgspec.json.encode(output)
```
samsja, line 45: what is the reason to use msgspec ?

mikasenghaas: what would be the alternative? it's the way vLLM does serialization, so I assumed it's a good default

samsja: oh okay then good enough :) I was just wondering

PR #329, opened 2025-05-28
Add RL tmuxinator

mikasenghaas:
This PR proposes to use `tmuxinator` as a general entrypoint for local-only RL experiments that require starting a training and inference process. I think this will be super useful and clear for new users to get started with using prime-rl and streamline our internal experimentation workflow.

The tmuxinator script should be viewed as an entrypoint that consumes a configuration for training and inference as well as a GPU allocation. It then automatically starts a tmux session with three windows: 
- `RL`: Training and inerence process in two panes
- `Monitor`: File monitor logs from training and inference in two pane
- `System`: Runs `htop` and `nvtop`

## Default

By default, it will run our CI RL run which is very minimal, completes fast but has includes all the functionality you would want from a local RL run.

```bash
tmuxinator
```

Screenshot 2025-06-17 at 10 00 45 PM

Screenshot 2025-06-17 at 10 01 12 PM
 
## Running an experiment

Most experiments, should be uniquely identifiyed by three things:
1. A training script (ideally this is as little as a single config file, say `@configs/training/exp1.toml`) and the number of GPUs to run on
2. An inference script (again, a single (matching) config file, say @configs/inference/exp1.toml`) and the number of GPUs to run on
3. Optionally, a name for the experiment

All of these can be set in the tmuxinator script via environment variables when starting the tmuxinator script. For example, to run the 2 GPU simple math run from the README:

```
TRAIN_ARGS=@configs/training/simple_math.toml INFER_ARGS="@configs/inference/simple_math.toml --parallel.dp 1 --max-batch-size 512" WANDB_GROUP=simple-math-test tmuxinator
```

### DIFF
--- a/.tmuxinator.yaml
+++ b/.tmuxinator.yaml
@@ -4,12 +4,14 @@ windows:
   - RL:
       layout: even-horizontal
       panes:
-        - Training:
-            - export CUDA_VISIBLE_DEVICES=0
-            - uv run torchrun src/zeroband/train.py @ configs/training/simple_reverse_two_gpu.toml --monitor.file.enable --monitor.file.path outputs/training_monitor.jsonl
+        - Training: 
+          - export CUDA_VISIBLE_DEVICES=0 
+          - clear
+          - uv run torchrun src/zeroband/train.py @ configs/training/simple_reverse_two_gpu.toml --monitor.file.enable --monitor.file.path outputs/training_monitor.jsonl
         - Inference:
-            - export CUDA_VISIBLE_DEVICES=1
-            - uv run src/zeroband/infer.py @ configs/inference/simple_reverse_two_gpus.toml --monitor.file.enable --monitor.file.path outputs/inference_monitor.jsonl
+          - export CUDA_VISIBLE_DEVICES=1 
+          - clear
+          - uv run src/zeroband/infer.py @ configs/inference/simple_reverse_two_gpus.toml --monitor.file.enable --monitor.file.path outputs/inference_monitor.jsonl
   - Monitor:
       layout: even-horizontal
       panes:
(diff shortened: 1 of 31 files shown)


.tmuxinator.yaml
```diff
@@ -0,0 +1,22 @@
...
+        - Inference:
+          - export CUDA_VISIBLE_DEVICES=${INFER_GPUS:-0}
+          - uv run src/zeroband/infer.py ${INFER_ARGS:-@configs/inference/simple_reverse_two_gpus.toml} --monitor.file.path infer-monitor.jsonl --monitor.wandb.group ${WANDB_GROUP:-$(date +%y%m%d-%H%M%S)} --monitor.wandb.prefix infer
+        - Training: 
+          - export CUDA_VISIBLE_DEVICES=${TRAIN_GPUS:-1}
+          - uv run torchrun src/zeroband/train.py ${TRAIN_ARGS:-@configs/training/simple_reverse_two_gpu.toml} --monitor.file.path train-monitor.jsonl --monitor.wandb.group ${WANDB_GROUP:-$(date +%y%m%d-%H%M%S)}
+  - Monitor:
+      layout: even-horizontal
+      panes:
+        - Training Logs: watch -n 1 "cat train-monitor.jsonl | tail -n 10"
```
samsja:
```suggestion
        - Training Logs: watch -n 1 "cat train-monitor.jsonl | tail -n 10"
```
What we should use here is tail -f instead of watch

.tmuxinator.yaml
```diff
@@ -0,0 +1,22 @@
...
+        - Inference Logs: watch -n 1 "cat infer-monitor.jsonl | tail -n 10"
+  - System:
+      layout: even-horizontal
+      panes:
+        - CPU Monitor: htop
+        - GPU Monitor: nvtop
```
samsja: I would not enable this by default

.tmuxinator.yaml
mikasenghaas: nice, didn't know

.tmuxinator.yaml
mikasenghaas: disabled system and monitor logs

mikasenghaas:
Screenshot 2025-06-17 at 10 57 19 PM

PR #332, opened 2025-05-28
Fix unifying monitor format

mikasenghaas:
Reverts change to old logging format introduced in #328 for `SocketMonitor`

*Old Format*

```txt
{"key": "progress/batch_problems", "value", 8, "task_id": "dskfjs"}
{"key": "progress/batch_samples", "value": 64, "task_id": "dskfjs"}
{"key": "progress/batch_tokens", "value": 1831, "task_id": "dskfjs"}
```

*New Format*

```txt
{"progress/batch_problems": 8, "progress/batch_samples": 64, "progress/batch_tokens": 1831, "task_id": "dskfjs"}
```

**This PR should only be merged once the protocol worker (opens the socket port which handles aggregating and relaying the logs) supports the new format.**

Also moves adding prime metadata (e.g. `PRIME_TASK_ID` or `PRIME_RUN_ID` into the base class's `_serialize_metrics` method. This means that, if available, all public run metadata will be logged to all sources (file, socket, api, ...) on every call to `.log`. While this might be redundant (and some sources might not need all the logs), this is keeps the clean abstraction that all sources mirror the same logs.

### DIFF
--- a/src/zeroband/utils/monitor.py
+++ b/src/zeroband/utils/monitor.py
@@ -65,9 +65,18 @@ class Monitor(ABC):
     def __init__(self, config: MonitorConfig):
         self.config = config
         self.lock = threading.Lock()
+        self.metadata = {
+            "run_id": envs.PRIME_RUN_ID,
+            "task_id": envs.PRIME_TASK_ID,
+        }
+        self.has_metadata = any(self.metadata.values())
+        if not self.has_metadata:
+            logger.warning("No run metadata found. This is fine for local runs, but unexpected when contributing to a public run.")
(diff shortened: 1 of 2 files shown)


src/zeroband/utils/monitor.py
```diff
@@ -65,9 +64,18 @@ class Monitor(ABC):
     def __init__(self, config: MonitorConfig):
         self.config = config
         self.lock = threading.Lock()
+        self.metadata = {
+            "run_id": envs.PRIME_RUN_ID,
```
JannikSt: I think we talked about removing this one, right?

mikasenghaas: yep, will remove

PR #339, opened 2025-05-29
Use new pyext version and add ipykernel as dev dependency

justusmattern27:


samsja (changes requested):
need to update uv lock as well

samsja (approved):
lgtm

PR #341, opened 2025-05-30
Add task-specific reward tracking and logging

kalomaze:
image

Individually logs rewards by task type under `task_rewards` on wandb, and changes the parquet data structure to make this information easily accessible within the trainer.

### DIFF
--- a/src/zeroband/inference/rewards.py
+++ b/src/zeroband/inference/rewards.py
@@ -1,4 +1,3 @@
-
 import json
 import os
 from concurrent.futures import ThreadPoolExecutor
@@ -65,7 +64,6 @@ class RewardRequest(BaseModel):
             yield request_output, verification_info, task_type
 
 
-
 def unwrap_request_output(request_output: RequestOutput) -> ModelOutput:
     outputs = [ModelCompletion(index=o.index, text=o.text, token_ids=o.token_ids) for o in request_output.outputs]
     return ModelOutput(request_id=request_output.request_id, outputs=outputs)
@@ -97,6 +95,8 @@ class CompletionReward(BaseModel):
 class RequestRewards(BaseModel):
     request_id: str  # type(RequestOutput.request_id)
     rewards: list[CompletionReward]
+    task_type: str
+
 
 class RewardsResponse(BaseModel):
     rewards: list[RequestRewards]
@@ -136,11 +136,11 @@ def _compute_completion_reward(
             output_length = len(completion_output.token_ids)
             # Penalizes absolute deviation from target length
             if length_config.reward_type == "exact":
-                length_penalty = (abs(target_length - output_length) * length_config.reward_coef)
+                length_penalty = abs(target_length - output_length) * length_config.reward_coef
                 reward -= length_penalty
             # Rewards for being close to target length with a maximum reward
             elif length_config.reward_type == "max":
-                raw_value = (length_config.reward_coef * (target_length - output_length) + length_config.max_reward_delta)
+                raw_value = length_config.reward_coef * (target_length - output_length) + length_config.max_reward_delta
                 length_penalty = max(0, min(1, raw_value))
                 reward *= length_penalty
             # Zero reward if output exceeds target length
@@ -192,9 +192,7 @@ def _compute_request_rewards(
     for completion_reward, advantage in zip(completion_rewards, advantage_array):
         completion_reward.advantage = float(advantage)
 
-    return RequestRewards(
-        request_id=request_output.request_id, rewards=completion_rewards
-    )
+    return RequestRewards(request_id=request_output.request_id, rewards=completion_rewards, task_type=str(task_type))
 
 
 def compute_rewards(
--- a/src/zeroband/train.py
+++ b/src/zeroband/train.py
@@ -2,6 +2,7 @@ import logging
 import os
 import shutil
 import time
+from collections import defaultdict
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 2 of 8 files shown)


src/zeroband/train.py
```diff
@@ -300,6 +301,14 @@ def train(config: Config):
                 for target_lengths in batch["target_lengths"]:
                     metric_averager.update("target_lengths", target_lengths)
 
+                task_rewards_by_type = defaultdict(list)
+                for reward, task_type in zip(batch["task_rewards"], batch["task_types"]):
+                    task_rewards_by_type[task_type].append(reward.item())
+
+                for task_type, rewards in task_rewards_by_type.items():
+                    avg_reward = sum(rewards) / len(rewards)
+                    metric_averager.update(f"{task_type}", torch.tensor(avg_reward))
```
samsja:
```suggestion
                for reward, task_type in zip(batch["task_rewards"], batch["task_types"]):
                    metric_averager.update(f"{task_type}", torch.tensor(reward))
```

kalomaze: simplified to this

src/zeroband/inference/rewards.py
```diff
@@ -136,11 +136,11 @@ def _compute_completion_reward(
             output_length = len(completion_output.token_ids)
             # Penalizes absolute deviation from target length
             if length_config.reward_type == "exact":
-                length_penalty = (abs(target_length - output_length) * length_config.reward_coef)
+                length_penalty = abs(target_length - output_length) * length_config.reward_coef
```
mikasenghaas, line 139: is this caused by different ruff formatter version/ rules?

src/zeroband/inference/rewards.py
```diff
@@ -136,11 +136,11 @@ def _compute_completion_reward(
...
                 reward -= length_penalty
             # Rewards for being close to target length with a maximum reward
             elif length_config.reward_type == "max":
-                raw_value = (length_config.reward_coef * (target_length - output_length) + length_config.max_reward_delta)
+                raw_value = length_config.reward_coef * (target_length - output_length) + length_config.max_reward_delta
```
mikasenghaas, line 143: same here

src/zeroband/train.py
```diff
@@ -288,18 +288,38 @@ def train(config: Config):
...
                     metric_averager.update("length_penalties", length_penalties)
                 for target_lengths in batch["target_lengths"]:
                     metric_averager.update("target_lengths", target_lengths)
 
+                # Task-specific metrics with proper grouping
+                if "task_types" in batch:
+                    # Group rewards by task type
+                    task_type_rewards = {}
+                    for i, task_type in enumerate(batch["task_types"]):
+                        task_key = f"individual_task_{task_type}"
```
mikasenghaas, line 307: what's the reason why prefixing with `individual_task_` here? can't we just use `task_type`?

src/zeroband/train.py
```diff
@@ -288,18 +288,38 @@ def train(config: Config):
...
+                        if task_key not in task_type_rewards:
+                            task_type_rewards[task_key] = []
+                        task_type_rewards[task_key].append(batch["task_rewards"][i].item())
```
mikasenghaas: can probably write this a more concisely. defaultdict(list) to prevent check if key in dict and/ or dict comprehension

src/zeroband/train.py
```diff
@@ -288,18 +288,38 @@ def train(config: Config):
...
+                    for task_key, rewards in task_type_rewards.items():
+                        if rewards:
+                            avg_reward = sum(rewards) / len(rewards)
+                            metric_averager.update(task_key, torch.tensor(avg_reward))
+
+                    # Add aggregate task_reward metric
+                    all_task_rewards = [reward.item() for reward in batch["task_rewards"]]
+                    if all_task_rewards:
+                        avg_task_reward = sum(all_task_rewards) / len(all_task_rewards)
+                        metric_averager.update("task_reward", torch.tensor(avg_task_reward))
```
mikasenghaas, line 320: what is the difference between this code block and the one a couple lines up that you deleted?

```python
for task_rewards in batch["task_rewards"]:
     metric_averager.update("task_reward", task_rewards)
```

src/zeroband/training/utils.py
```diff
@@ -274,6 +275,11 @@ def log_to_wandb(
             if k in metrics_dict and metrics_dict[k] is not None:
                 wandb_metrics[f"{prefix}{k}"] = metrics_dict[k]
 
+    # Handle task-specific metrics based on the prefix
+    for key, value in metrics_dict.items():
+        if key.startswith("individual_task_"):
+            wandb_metrics[f"task_rewards/{key.replace('individual_task_', '')}"] = value
```
mikasenghaas, line 281: oh and here you delete the prefix again lol?

kalomaze: yes, the annoying thing was task_reward conflicting with the task_ prefixing i had earlier wrt matching the tasks to be added to the task_rewards panel, and i figured it'd be wise to internally make it clear that those are separate things.
(i also ensured that `task_reward` only gets logged once).

src/zeroband/train.py
kalomaze: this one seems to properly handle grad accumulation

src/zeroband/train.py
kalomaze: that seems like an easy way to get people confuddled by making the name of the string vs. the name of the ephemeral prefix that is used to properly separate it from `task_` (as its used in the aggregate metric) from -INDIVIDUAL- tasks too similar to each other. i don't want them to start the same way.

src/zeroband/inference/rewards.py
kalomaze: i believe so

mikasenghaas:
Screenshot 2025-05-30 at 11 18 04 AM

PR #338, opened 2025-05-29
Auto-detect maximum batch size

mikasenghaas:
Implements util function to compute maximum batch size (defined as the maximum number of sequences for which we can keep KV cache assuming worst-case scenario, i.e. all sequences go to `max_model_len` and have distinct GPU cache blocks). For heterogeneous pipelines, all-reduces the batch size using `min` reduction to ensure no cache eviction on any pipeline stage.

PR #343, opened 2025-05-30
Add sentence unscrambling + ASCII tree formatting tasks [remade PR]

kalomaze:
### 1. Sentence Unscrambling

Requires the model to rearrange a series of text blocks with a scrambled order such that it matches the original order of the ground truth text. A 2024 randomly split subset of FineWeb-Edu was chosen for this first data iteration.

Minimum blocks/sentences included in the current dataset: 4
Maximum: 10

**Example Input:**
````
Rearrange these scrambled blocks of text into a coherent order:

```
*1*: Apple revolutionized smartphones with the iPhone, introducing touchscreens to the mass market.
*2*: Before smartphones, most people used basic mobile phones primarily for calls and text messages.
*3*: Today, smartphones are essential tools that billions of people worldwide use daily.
*4*: Following the iPhone, companies like Samsung and Google developed their own smartphone platforms.
```

Keep the block numbers in your answer, and wrap it in <unscrambled_text> tags. Keep the order of the numbers the same in the output, but change the order of the content so that it makes the most chronological sense.
````

**Expected Output:**
```
<think>
*...reasoning...*
</think>

<unscrambled_text>
*1*: Before smartphones, most people used basic mobile phones primarily for calls and text messages.
*2*: Apple revolutionized smartphones with the iPhone, introducing touchscreens to the mass market.
*3*: Following the iPhone, companies like Samsung and Google developed their own smartphone platforms.
*4*: Today, smartphones are essential tools that billions of people worldwide use daily.
</unscrambled_text>
```

image
*The reward function for this task focuses on the longest correct consecutive ordering of sentences, encouraging the model to construct coherent sequences rather than isolated correct placements.*

### 2. ASCII Tree Formatting

This task requires the model to convert a textual description of a directory structure into an ASCII tree format. The descriptions were generated programatically rather than synthetically; the latter option would ultimately be better for future dataset variation once verifying fuzzy accuracy (i.e. through classifiers/LLM as judge esque approaches) starts to mature in our pipelines.

**Example Input:**
````
Your task is to convert an English description into an ASCII tree format using this style:

```
fonts
+--Inter
|  '--static
+--Space_Grotesk
|  '--static
'--Space_Mono
```

Convert this description:

precompiledScriptPlugins-externalPlugins is a base directory that holds 3 subdirectories: groovy, kotlin and tests.
In the root directory, 'groovy' contains 1 subdirectory elements: buildSrc.
At level 2, in directory path 'groovy → buildSrc', 'buildSrc' retains 1 directory entries: src.
At level 3, in directory path 'groovy → buildSrc → src', Within 'src', you'll find 1 contained folders for code structure.
In nested path 'groovy → buildSrc → src → main', 'main' stores 1 contained folders: groovy.
At level 5, in directory path 'groovy → buildSrc → src → main → groovy', 'groovy' awaits content.
At the main level, 'kotlin' keeps 1 folders: buildSrc.
In nested path 'kotlin / buildSrc', 'buildSrc' maintains 1 subfolders: src.
In nested path 'kotlin → buildSrc → src', 'src' encompasses 1 subdirectories that organize the source code.
At level 4, in directory path 'kotlin / buildSrc / src / main', 'main' incorporates 1 directories: kotlin.
At level 5, in directory path 'kotlin → buildSrc → src → main → kotlin', The 'kotlin' data structure has nothing inside.
At the top level, The 'tests' folder holds no subdirectories at the moment.
````

**Expected Output:**
````
<think>
*...reasoning...*
</think>

<ascii_formatted>
precompiledScriptPlugins-externalPlugins
+--groovy
| `--buildSrc
|   `--src
|     `--main
|       `--groovy
+--kotlin
| `--buildSrc
|   `--src
|     `--main
|       `--kotlin
`--tests
</ascii_formatted>
````

image

*The reward function uses difflib to compare the model's output with the ground truth, weighting both overall similarity (30%) and the longest continuous matching block (70%). Additional penalties are applied if the tree structure is invalid or missing distinctive elements seen in the example, like the "--" and "|" characters.*

---
Original PR: https://github.com/PrimeIntellect-ai/prime-rl/pull/288

### DIFF
--- a/configs/inference/simple_ascii_tree.toml
+++ b/configs/inference/simple_ascii_tree.toml
@@ -1,0 +1,11 @@
+model_name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
+dataset = "kalomaze/ascii-tree-mix-it1"
+batch_size = 16
+dp = 4
+rollout_path = "outputs"
+output_path = "data_rollout"
+max_model_len = 2048
+
+[sampling]
+temperature = 1.0
+n = 16
--- a/configs/inference/simple_ascii_tree.toml
+++ b/configs/inference/simple_ascii_tree.toml
@@ -1,6 +1,6 @@
 model_name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
 dataset = "kalomaze/ascii-tree-mix-it1"
-batch_size = 16
+batch_size = 256
 dp = 4
 rollout_path = "outputs"
 output_path = "data_rollout"
--- a/configs/inference/simple_unscramble.toml
+++ b/configs/inference/simple_unscramble.toml
@@ -1,0 +1,7 @@
+model_name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
+dataset = "kalomaze/fwedu-unscramble-mix-it1"
+batch_size = 32
+dp = 4
+rollout_path = "outputs"
+output_path = "data_rollout"
+max_model_len = 2048
--- a/configs/inference/simple_unscramble.toml
+++ b/configs/inference/simple_unscramble.toml
@@ -1,6 +1,6 @@
 model_name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
 dataset = "kalomaze/fwedu-unscramble-mix-it1"
-batch_size = 32
+batch_size = 256
 dp = 4
 rollout_path = "outputs"
 output_path = "data_rollout"
--- a/configs/training/simple_ascii_tree.toml
+++ b/configs/training/simple_ascii_tree.toml
@@ -1,0 +1,23 @@
+model_name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
+project = "ascii-tree-debug"
+
+[train]
+micro_bs = 2 # change to 8 for H200
+reshard_after_forward = true
+
+[optim]
+batch_size = 64
+warmup_steps = 20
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: configs/inference/simple_ascii_tree.toml [hunk 1]
```diff
@@ -0,0 +1,11 @@
+model_name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
+dataset = "kalomaze/ascii-tree-mix-it1"
+batch_size = 16
```
mikasenghaas: the definition of `batch_size` changed in #338.

before `batch_size` was the number of problems we'd sample from the dataset (and the total number of sequences per infer step is `batch_size * sampling.n`

now, `batch_size` is the total number of sequences per infer step (and can be auto-inferred to evict kv cache). 

you want to adjust the config to `batch-size = old_batch_size * sampling.n = 16 * 16 = 256` to align definitions

configs/inference/simple_unscramble.toml
```diff
@@ -0,0 +1,7 @@
+model_name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
+dataset = "kalomaze/fwedu-unscramble-mix-it1"
+batch_size = 32
```
mikasenghaas: same here

mikasenghaas (approved):
two minor comments, else looks good

mikasenghaas:
Screenshot 2025-05-30 at 11 11 34 AM

PR #344, opened 2025-05-30
Install git in install script

apaz-cli:
Small quality of life thing, but has been bothering me a while

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -20,9 +20,14 @@ main() {
     log_info "Updating apt..."
     sudo apt update
 
-    log_info "Installing cmake python3-dev..."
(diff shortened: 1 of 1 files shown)


mikasenghaas:
I had the same thought but then @samsja enlightened me: Better to keep cloning from SSH and put this in your `~/.ssh/config`

```txt
Host *
  AddKeysToAgent yes
  UseKeychain yes
  ForwardAgent yes
```

As long as your local ssh agent has access to your GitHub key, every SSH session will have access it to and so you can pull/ push however much you like

samsja:
> Small quality of life thing, but has been bothering me a while

SSH forward is probably the way otherwise HTTPS cloning we cannot do dev.

Also I like installing git yeah

PR #345, opened 2025-05-30
Shard model dynamically

mikasenghaas:
We don't require pre-sharding the model and pointing to a specific model shard repo on HF when starting pipeliend inference, but monkey-patch the `make_layers` function in vLLM. In this way we trigger all the vLLM native behavior sharding behavior but without setting up a PP communciation group via `torch.dist`. Main advantages:
- We don't have to spend time on pre-sharding
- We save disk space, because we only ever download the full model and then pick the relevant shards
- We have make `toml` configs for each model-task combination independent of the underlying topology (this will enable much cleaner tasks to the orchestrator API)

### DIFF
--- a/src/zeroband/inference/pipeline.py
+++ b/src/zeroband/inference/pipeline.py
@@ -59,8 +59,8 @@ def deserialize_sampler_output(data: bytes) -> SamplerOutput:
 
 def setup_comm(config: PipelineConfig) -> Node | None:
     """
-    Setup communication via PRIME-IROH. Forms a ring topology between the model shards
-    with unidirectional communication flow.
+    Setup P2P communication via using `prime-iroh` nodes. Forms a ring topology
+    between the model shards with unidirectional communication flow.
 
     Args:
         config: The pipeline configuration
@@ -73,7 +73,7 @@ def setup_comm(config: PipelineConfig) -> Node | None:
 
     # Setup node (with or without seed)
     if config.iroh_seed is not None:
-        logger.debug(f"Using IROH seed: {config.iroh_seed}")
+        logger.debug(f"Using seed: {config.iroh_seed}")
         # If seed is provided, create a new node with the seed
         node = Node.with_seed(num_streams=1, seed=config.iroh_seed)
     else:
--- a/src/zeroband/inference/pipeline.py
+++ b/src/zeroband/inference/pipeline.py
@@ -57,34 +57,19 @@ def deserialize_sampler_output(data: bytes) -> SamplerOutput:
     return msgspec.json.decode(data, type=SamplerOutput)
 
 
-def setup_pipeline(config: PipelineConfig, llm: LLM) -> Node:
-    """
-    Setup PRIME-IROH communication and hooks for pipeline parallel inference.
-
-    Args:
-        llm: The LLM model shard instance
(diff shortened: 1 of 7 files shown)


mikasenghaas:
Screenshot 2025-05-30 at 2 35 27 PM

samsja:
> * because we only ever download the full model and then pick the relevant shards

can u clarify if we 
* do download the full model
* only download the shard relevant

src/zeroband/inference/pipeline.py
```diff
@@ -114,25 +100,62 @@ def setup_comm(config: PipelineConfig) -> Node:
...
+    def _patched_make_layers(num_hidden_layers: int, layer_fn: LayerFn, prefix: str) -> Tuple[int, int, torch.nn.ModuleList]:
+        from vllm.distributed.utils import get_pp_indices
+
+        start_layer, end_layer = get_pp_indices(num_hidden_layers, config.rank, config.world_size)
+        modules = torch.nn.ModuleList(
+            [PPMissingLayer() for _ in range(start_layer)]
+            + [maybe_offload_to_cpu(layer_fn(prefix=f"{prefix}.{idx}")) for idx in range(start_layer, end_layer)]
+            + [PPMissingLayer() for _ in range(end_layer, num_hidden_layers)]
+        )
+        return start_layer, end_layer, modules
```
samsja: put a bit more comment here to explain

src/zeroband/inference/utils.py
```diff
@@ -115,3 +117,30 @@ def compute_max_batch_size(llm: LLM) -> int:
...
     return max_batch_size
+
+
+def rgetattr(obj: Any, attr_path: str) -> Any:
+    """
+    Tries to get a (nested) attribute from an object. For example, if
+    `model.layers` is an attribute of `model`, then `rgetattr(model,
+    "model.layers")` will return the attribute `layers`. If `layers` is not an
+    attribute of `model`, then `rgetattr(model, "layers")` will raise an
+    `AttributeError`.
```
samsja: It's not super obvious to me wha this func do at first read. Maybe an example would make it clearer

mikasenghaas:
> > * because we only ever download the full model and then pick the relevant shards
> 
> can u clarify if we
> 
> * do download the full model
> * only download the shard relevant

we download the full model to disk and load the shard to gpu memory. could probably even patch the download, but this is more difficult. it's also more annoying for local development. with the current setup it always has the full model and you can do whatever pp sharding you want.

samsja:
> > > * because we only ever download the full model and then pick the relevant shards
> > 
> > 
> > can u clarify if we
> > 
> > * do download the full model
> > * only download the shard relevant
> 
> we download the full model to disk and load the shard to gpu memory. could probably even patch the download, but this is more difficult. it's also more annoying for local development. with the current setup it always has the full model and you can do whatever pp sharding you want.

yeah makes sense, maybe we will need to do it at some point but its fine for now.

(I was not sure about the meaning initially hence I ask for clarification)

mikasenghaas:
i actually had to double check myself haha

mikasenghaas:
Screenshot 2025-05-30 at 3 26 32 PM

mikasenghaas:
Screenshot 2025-06-01 at 9 14 55 AM

PR #346, opened 2025-05-30
Fix and refactor applying chat template

mikasenghaas:


mikasenghaas:
Very interesting! Because this PR removes adding an "artificial" think token for non-thinking models like Qwen2.5-0.5B (SFT) parsing the final answer fails (currently it parses `solution = completion.split("</think")[1]`), leading to zero rewards during the entire training. There is no way for the model to know about this formatting quirk.

Screenshot 2025-06-02 at 2 09 06 PM

mikasenghaas:
I created a new analogous dataset asking the model to respond with the reversed text within `<answer>...</answer>` tags (I do not have permissions to push to `PrimeIntellect` HF, so had to use personal).

```python
import json
from datasets import load_dataset

dataset = load_dataset("PrimeIntellect/reverse_text_dataset_debug_50_seq_len")
def reverse_text(ex):
    text = ex["text"][:50]
    prompt = f'Reverse the text in quotation marks: "{text}"\n\nPut your final answer in <answer>...</answer> tags.'
    verification_info = json.dumps({"ground_truth": text[::-1]})
    return {"prompt": prompt, "verification_info": verification_info, "task_type": "reverse_text"}

dataset = dataset.map(reverse_text, remove_columns=["title", "text"])
dataset.push_to_hub("mikasenghaas/reverse_text_dataset_debug_50_seq_len")
```

And then, to compute the reward, I parse from the answer tags using regex similar to @kalomaze reward functions for unscrambling text and ASCII tree formatting. This fixes the CI training run:

Screenshot 2025-06-02 at 2 53 22 PM

Some interesting additional findings:
- We start from a higher base solve rate (~0.2 instead of ~0.05) which makes sense to me given that we start from a model already SFT'ed for this task. It seems like giving more precise instructions on formatting leads to the base model solving more samples
- We learn quicker and get a slightly higher reward after the final 100 steps at around 0.8
- We still don't overfit to reward 1.0 which I find a bit suspicious but ig more training/ slightly different hparams would solve

mikasenghaas:
Some more interesting ablations that I did on the way:

1. **Train from thinking model (`Qwen/Qwen3-0.6B`)**: I wanted to try training from a thinking model to test the difference in the chat template for thinking/non-thinking mode to make sure that training still works (ie. the model produces <think>...</think> and then <answer>...</answer>). TLDR: I couldn't really check for this because the Qwen3 model was not SFTed. Here, the model is barely learning within 100 steps. Looking at the outputs, we see that the `128` token limit was not enough to even produce the answer tag (the model was still thinking): 

Screenshot 2025-06-02 at 3 09 29 PM

I increased the `max_model_len` to 512 to allow the model to finish thinking to get some positive training signal.

Screenshot 2025-06-02 at 3 09 54 PM

That helped a little but the model still thinks for too long. Now, I reset the `max_model_len` to 128 and disable thinking.

Screenshot 2025-06-02 at 3 15 17 PM

It seems like the model is not even trying to reverse the text but is still getting ~0.2 reward (LCS ratio might not be the best to use here). I find it a bit suspicious that we cannot learn anything on a non-SFTed model here. Even though it's a small model, I would expect some learning to happen within 100 steps.

2. **Use custom "task tag"**: I change the prompt to ask to put the answer in a custom `<reversed_text>...</reversed_text>` tag. Training works but starts from a lower reward in the beginning. I suspect that Will's SFT used used `<answer>` tags.

Screenshot 2025-06-02 at 3 00 10 PM

samsja (approved):

mikasenghaas:
Screenshot 2025-06-04 at 11 09 17 AM

PR #348, opened 2025-05-31
add barrier before saving ckpt

samsja:


mikasenghaas:
did this fix the weight loading race condition?

samsja:
> did this fix the weight loading race condition?

still working on it to replicate

Jackmin801 (approved):
lgtm! I think even if we cant repro maybe its worth to merge this safer line and task whoever wants to optimize this later with having to prove the race condition doesnt happen

samsja:
> lgtm! I think even if we cant repro maybe its worth to merge this safer line and task whoever wants to optimize this later with having to prove the race condition doesnt happen

yeah this is also my thinking. Tho defo need to look at the whole ckpt performance before next run

PR #349, opened 2025-06-01
pydantic schema adherance task

justusmattern27:


src/zeroband/inference/genesys/pydantic_json_adherance.py
mikasenghaas, line 1: high recommend setting up ruff locally:)) install hook `uv run pre-commit install` to catch format/lint errors at commit time. or, even more convenient, install ruff vscode extension and format on file save

justusmattern27: yeah i usually do - this is pushed from a vm where i didnt do it

justusmattern27:
dataset: https://huggingface.co/datasets/justus27/pydantic-adherance-test

src/zeroband/inference/genesys/pydantic_json_adherance.py
```diff
@@ -0,0 +1,48 @@
...
+    if payload is None:
+        return 0
+
+    try:
+        Model = _load_model_from_code(
+            verification_info["pydantic_config"],
+            verification_info["model_name"],
+        )
+    except Exception:
+        return 0
```
mikasenghaas: zero reward if there is a dataset error?

src/zeroband/inference/genesys/pydantic_json_adherance.py
```diff
@@ -0,0 +1,48 @@
...
+    except Exception as e:
+        raise RuntimeError(f"config code failed to execute: {e!r}") from e
+
+    cls = getattr(module, model_name, None)
+    if cls is None or not issubclass(cls, BaseModel):
+        raise RuntimeError(f"{model_name} not found or not a Pydantic BaseModel")
+
+    # cheap structural self-check (never instantiates)
+    cls.model_json_schema()
+    return cls
```
mikasenghaas, line 27: running `exec` on arbitrary code string seems a bit sketchy

samsja: @apaz-cli what should we do here ?

justusmattern27: it's LLM generated code that we have validated and looked at before - later this should be part of the sandbox but while we're still doing local code execution imo it's fine. It's very similar to what we're doing with unit tests rn

src/zeroband/inference/genesys/pydantic_json_adherance.py
justusmattern27: ah yeah, this should just be filtered out beforehand, will remove

samsja (approved):
lgtm

PR #350, opened 2025-06-01
[feat] Inference work counting

Jackmin801:


src/zeroband/inference/work_counting.py
```diff
@@ -0,0 +1,86 @@
+from transformers.models.deepseek_v3.configuration_deepseek_v3 import DeepseekV3Config
+from transformers.models.qwen3.configuration_qwen3 import Qwen3Config
+
+# Note: Only matmuls are counted
+
+
+def get_inference_input_output_flops_qwen3(config: Qwen3Config, input_tokens: int, output_tokens: int) -> tuple[int, int]:
```
mikasenghaas: nit: call these `num_input_tokens` and `num_output_tokens`? we are not yet super consistent with differentiating this (`input_tokens: List[int]`, `num_input_tokens: int`) in the main script but i want to get there eventually:)

src/zeroband/inference/work_counting.py
```diff
@@ -0,0 +1,86 @@
...
+    ## LM Head
+    lm_head_flops = 2 * vocab_size * hidden_size
+    ## Total
+    input_linear_flops = (q_flops + k_flops + v_flops + o_flops + mlp_flops + lm_head_flops) * input_tokens
+    output_linear_flops = (q_flops + k_flops + v_flops + o_flops + mlp_flops + lm_head_flops) * output_tokens
+
+    # SDPA
+    ## 4lhqt from mm
+    input_sdpa = 4 * num_hidden_layers * head_dim * num_attention_heads * ((input_tokens + 1) * input_tokens // 2)
+    output_sdpa = 4 * num_hidden_layers * head_dim * num_attention_heads * ((output_tokens + input_tokens + 1) * output_tokens // 2)
```
mikasenghaas: maybe add a little comment for `((input_tokens + 1) * input_tokens // 2)` and `((output_tokens + input_tokens + 1) * output_tokens // 2)`? took me a second to figure out this term

mikasenghaas (commented):
very nice! just wondering - right now it seems like we need a util function for each model with an architectural change. two ideas to not do this:
1. can we rely on any vllm internal computation in this direction that we can piggyback on?
2. can we get away with a "less precise" but directionally correct computation, ie. have one function for dense and one for moe models and then be fine with qwen3 moe and deepseek moe having the same function although not the exact same count?

src/zeroband/inference/work_counting.py
```diff
@@ -0,0 +1,86 @@
...
+    sparse_mlp_flops = num_sparse_layers * (
+        2 * shared_experts * 3 * moe_intermediate_size * hidden_size  # Shared experts
+        + 2 * experts_per_tok * 3 * moe_intermediate_size * hidden_size  # Routed experts
+        + 2 * routed_experts * hidden_size  # Router
+    )
+    ## LM Head
+    lm_head_flops = 2 * vocab_size * hidden_size
+    ## Total
+    input_linear_flops = (q_flops + kv_flops + o_flops + dense_mlp_flops + sparse_mlp_flops + lm_head_flops) * input_tokens
+    output_linear_flops = (q_flops + kv_flops + o_flops + dense_mlp_flops + sparse_mlp_flops + lm_head_flops) * output_tokens
```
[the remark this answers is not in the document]
Jackmin801: no

Jackmin801:
> 1. can we rely on any vllm internal computation in this direction that we can piggyback on?

I dont think so. I guess you could count the number of params or memory used by the VLLM engine but then you'd have issues when it is PP / TP and you also would not be able to account for the sdpa. It also needs to be consistent from prime-rl to toploc-validator in different deployments, of which taking the config as input seems to me more stable.

> 2. can we get away with a "less precise" but directionally correct computation, ie. have one function for dense and one for moe models and then be fine with qwen3 moe and deepseek moe having the same function although not the exact same count?

For the particular case of qwen3 and deepseekv3 I dont think so. The MLA used in deepseekv3 is quite a different calculation for the GQA used by qwen3. We could opt for initializing a fake model just to count the parameters though e.g.
```python
from transformers import AutoConfig
from transformers.models.deepseek_v3.configuration_deepseek_v3 import DeepseekV3Config
from transformers.models.deepseek_v3.modeling_deepseek_v3 import DeepseekV3ForCausalLM
import torch

torch.set_default_dtype(torch.bfloat16)
config: DeepseekV3Config = AutoConfig.from_pretrained("deepseek-ai/DeepSeek-R1-0528")
old_device = torch.get_default_device()
torch.set_default_device("meta")
model = DeepseekV3ForCausalLM._from_config(config)
torch.set_default_device(old_device)

print(f"Total params: {sum(p.numel() for p in model.parameters())}")
```
But this function takes 25s to complete for deepseekv3 while the one that calculates it from config takes a few micro seconds. And extra logic would still be required for calculating the sdpa term which would probably access the config.

src/zeroband/inference/work_counting.py
Jackmin801: Eh sure. Changed to
```python
    # SDPA
    ## 4lhqt from mm
    ## Each subsequent token sees 1 more ctx so the total is the sum of an arithmetic series
    input_ctx_sum = (input_tokens + 1) * input_tokens // 2
    output_ctx_sum = (output_tokens + input_tokens + 1) * output_tokens // 2
    input_sdpa = 4 * num_hidden_layers * head_dim * num_attention_heads * input_ctx_sum
    output_sdpa = 4 * num_hidden_layers * head_dim * num_attention_heads * output_ctx_sum
```

src/zeroband/inference/work_counting.py
Jackmin801: sure!

src/zeroband/inference/work_counting.py
```diff
@@ -0,0 +1,101 @@
...
+from transformers.models.qwen3.configuration_qwen3 import Qwen3Config
+from transformers.models.qwen3_moe.configuration_qwen3_moe import Qwen3MoeConfig
+
+# Note: Only matmuls are counted
+
+
+def get_inference_input_output_flops_qwen3(
+    config: Qwen3Config | Qwen3MoeConfig, num_input_tokens: int, num_output_tokens: int
+) -> tuple[int, int]:
+    """Get input and output flops for Qwen3 inference"""
```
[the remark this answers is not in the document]
Jackmin801: docstring added

src/zeroband/inference/work_counting.py
```diff
@@ -0,0 +1,101 @@
...
+    lm_head_flops = 2 * vocab_size * hidden_size
+    ## Total
+    input_linear_flops = (q_flops + k_flops + v_flops + o_flops + mlp_flops + lm_head_flops) * num_input_tokens
+    output_linear_flops = (q_flops + k_flops + v_flops + o_flops + mlp_flops + lm_head_flops) * num_output_tokens
+
+    # SDPA
+    ## 4lhqt from mm
+    ## Each subsequent token sees 1 more ctx so the total is the sum of an arithmetic series
+    input_ctx_sum = (num_input_tokens + 1) * num_input_tokens // 2
+    output_ctx_sum = (num_output_tokens + num_input_tokens + 1) * num_output_tokens // 2
```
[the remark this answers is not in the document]
Jackmin801: I feel like this formula is simple and well known enough to not be its own function

src/zeroband/inference/utils.py
```diff
@@ -6,6 +6,7 @@
 from vllm.model_executor.model_loader.loader import _process_weights_after_loading
 
 from zeroband.inference.rewards import LenRewardsConfig
+from zeroband.inference.work_counting import get_inference_input_output_flops  # noqa: F401
```
mikasenghaas, line 9: why do we need noqa here?

mikasenghaas (approved):

tests/unit/inference/test_work_counting.py
```diff
@@ -0,0 +1,39 @@
...
+from zeroband.inference.work_counting import get_inference_input_output_flops_deepseek_v3, get_inference_input_output_flops_qwen3
+
+
+@pytest.mark.parametrize(
+    "model_name_or_path, active_params",
+    [
+        ("deepseek-ai/DeepSeek-R1-0528", 37e9),  # 37B Active Params from https://arxiv.org/pdf/2412.19437
+    ],
+)
+def test_get_inference_input_output_flops_deepseek_v3(model_name_or_path: str, active_params: int):
```
samsja, line 13: Is it a test than need to download v3 ? If yes need to mark it "very slow" and probably skip it from ci

tests/unit/inference/test_work_counting.py
```diff
@@ -0,0 +1,39 @@
...
+        ("Qwen/Qwen3-1.7B", 1.7e9),
+        ("Qwen/Qwen3-4B", 4e9),
+        ("Qwen/Qwen3-8B", 7.6e9),  # This is only 8B because it has untied embs somehow
+        ("Qwen/Qwen3-14B", 14e9),
+        ("Qwen/Qwen3-32B", 32e9),
+        ("Qwen/Qwen3-30B-A3B", 3e9),
+        ("Qwen/Qwen3-235B-A22B", 22e9),
+    ],
+)
+def test_get_inference_input_output_flops_qwen3(model_name_or_path: str, active_params: int):
```
samsja, line 35: Should only run the small model in ci

samsja (approved):
Lgtm

src/zeroband/inference/utils.py
Jackmin801: because the import isnt used so ruff will remove it.
I felt it was clunky to have to import from many places in the main script so this line is so u can import it from utils.
I didnt want to move it to utils because its logic is quite isolatable and needs to be copied over to toploc-validator

PR #351, opened 2025-06-01
Add SYNTHETIC-2 inference configs

mikasenghaas:
This PR defines the config files for all SYNTHETIC-2 models.

### Debug Run

For debugging, we run the small `Qwen/Qwen3-0.6B` for five inference steps (at maximum auto-detected batch size) on a single GPU at 8k context with TOPLOC enabled and monitor logging to file `file_monitor.jsonl`

```bash
uv run python src/zeroband/infer.py @ configs/inference/synthetic-2/debug.toml
```

### SYNTHETIC-2 Models

We have a config for each of the SYNTHETIC-2 models. To test them locally (including logging to socket), first run a fake echo Unix socket in a separate terminal

```bash
curl -s https://gist.githubusercontent.com/mikasenghaas/03b65ab27947e9735a732d644d070219/raw | uv python -
```

Further, for local testing you probably want to set `--sampling.max-tokens 1000` (to not run full context generation) and `--output-path outputs` (because outside of Docker you might not have permission to write to `/data`). Append these flags to each of the inference start commands below.

1. `Qwen3-4B`

```bash
export PRIME_SOCKET_PATH=metrics.sock 
uv run python src/zeroband/infer.py @ configs/inference/synthetic-2/qwen3-4b.toml
```

2. `DeepSeek-R1-0528-Qwen3-8B`

```bash
export PRIME_SOCKET_PATH=metrics.sock 
uv run python src/zeroband/infer.py @ configs/inference/synthetic-2/deepseek-r1-0528-qwen3-8b.toml
```

3. `Qwen3-32B`

```bash
export PRIME_SOCKET_PATH=metrics.sock 
uv run python src/zeroband/infer.py @ configs/inference/synthetic-2/qwen3-32b.toml
```

4. `DeepSeek-R1-0528` (can only test on 8xH200)

```bash
export PRIME_SOCKET_PATH=metrics.sock 
uv run python src/zeroband/infer.py @ configs/inference/synthetic-2/deepseek-r1-0528.toml
```

### DIFF
--- a/configs/inference/synthetic-2/debug.toml
+++ b/configs/inference/synthetic-2/debug.toml
@@ -6,8 +6,8 @@ toploc = true
 
 [monitor.file]
 enable = true
-path = "file_monitor.jsonl"
+path = "outputs/file_monitor.jsonl"
 
 [sampling]
-max_tokens = 100
+max_tokens = 1000
 n = 1
--- a/configs/inference/synthetic-2/debug.toml
+++ b/configs/inference/synthetic-2/debug.toml
@@ -1,0 +1,10 @@
+model_name = "Qwen/Qwen3-0.6B"
+dataset = "PrimeIntellect/INTELLECT-2-RL-Dataset"
+
+max_model_len = 16384
+max_samples = 100
+toploc = true
+
+[sampling]
+max_tokens = 100
+n = 1
--- a/configs/inference/synthetic-2/debug.toml
+++ b/configs/inference/synthetic-2/debug.toml
@@ -1,9 +1,12 @@
 model_name = "Qwen/Qwen3-0.6B"
-dataset = "PrimeIntellect/INTELLECT-2-RL-Dataset"
 
-max_model_len = 16384
-max_samples = 100
+max_model_len = 8192
+total_step = 5
 toploc = true
+
+[monitor.file]
+enable = true
+path = "file_monitor.jsonl"
 
 [sampling]
 max_tokens = 100
--- a/configs/inference/synthetic-2/deepseek-r1-0528-qwen3-8b.toml
+++ b/configs/inference/synthetic-2/deepseek-r1-0528-qwen3-8b.toml
@@ -1,2 +1,15 @@
 model_name = "deepseek-ai/DeepSeek-R1-0528-Qwen3-8B"
+
 max_model_len = 16384
+tp = "auto"
+output_path = "/data"
+toploc = true
+
+[monitor.socket]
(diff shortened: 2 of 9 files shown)


justusmattern27 (commented):
i think we should have a higher temperature for sampling, probably 0.7 or 0.8. Also curious what @kalomaze things

mikasenghaas:
Added the default sampling parameters used in the official evals for Qwen-3 and DeepSeek models, as per their HF model cards:
- `Qwen3-4B`: https://huggingface.co/Qwen/Qwen3-4B
- `Qwen3-32B`: https://huggingface.co/Qwen/Qwen3-32B
- `DeepSeek-R1-0825-Qwen3-8B`: https://huggingface.co/deepseek-ai/DeepSeek-R1-0528
- `DeepSeek-R1-0825`: https://huggingface.co/deepseek-ai/DeepSeek-R1-0528

@kalomaze and @justusmattern27 should intervene if we should use different defaults.

mikasenghaas:
NB:
1. The Qwen3 models set `min_p=0.0` which we don't yet support in our sampling config. Should probably add this?
2. They recommend different configs for thinking and non-thinking mode. Did we decide whether we want to run some omdels in non-thinking mode @justusmattern27? If yes, we should have a differet TOML config for that.
3. They propose to set the `presence_penalty` flag (see below) - should we set this?

```txt
For supported frameworks, you can adjust the presence_penalty parameter between 0 and 2 to reduce endless repetitions. However, using a higher value may occasionally result in language mixing and a slight decrease in model performance.

```

mikasenghaas:
@samsja there are some overlapping configs (ie. independent of model) such as `--output-path` which is always set to `/data`. A clean solution in terms of loading toml configs would be to:
1. Extract all shared configs in a file `configs/inference/synthetic-2/default.toml`
2. Have model-specific configs for all models, e.g. `configs/inference/synthetic-2/qwen3-4b.toml`

And then start the run by chaining the default with the model-specific config.

```bash
uv run python src/zeroband/infer.py @configs/inference/synthetic-2/default.toml @configs/inference/synthetic-2/qwen3-4b.toml
```

Here, the logic would be to load configs in the order in which they are put in CLI, which means that later config files take precedence (the model specific config can overwrite a default SYN-2 config, if necessary). 

Do you think this makes sense implementing in `pydantic-config` @samsja, afaik this is not yet supported.

mikasenghaas:
Screenshot 2025-06-05 at 1 57 37 PM

PR #352, opened 2025-06-02
[fix] dont norm intermediate toploc proofs

Jackmin801:
replace norm with identity in earlier pp stages so that the proofs are generated with the unnormed hidden activations which can be easily saved by the toploc-validator validation engine in the fwd prefill

### DIFF
--- a/src/zeroband/inference/toploc.py
+++ b/src/zeroband/inference/toploc.py
@@ -197,7 +197,7 @@ def setup_toploc_cache(
     if not disable:
         handle = logits_processor.register_forward_pre_hook(partial(toploc_cache_hook, toploc_cache=toploc_cache))
 
-    if pipeline_config is not None and pipeline_config.world_size > 1 and pipeline_config.rank < pipeline_config.world_size - 1:
+    if pipeline_config is not None and pipeline_config.pipeline_enabled and pipeline_config.rank < pipeline_config.world_size - 1:
         llm.llm_engine.model_executor.driver_worker.model_runner.model.model.norm = ArgsIdentity()
 
     return toploc_cache, handle
--- a/src/zeroband/inference/toploc.py
+++ b/src/zeroband/inference/toploc.py
@@ -9,6 +9,16 @@ from vllm import LLM
 from vllm.model_executor import SamplingMetadata
 from vllm.model_executor.layers.logits_processor import _prune_hidden_states
 
+from zeroband.inference.pipeline import PipelineConfig
+
+
+class ArgsIdentity(nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, *args):
+        return args
+
 
 class TopLocCache:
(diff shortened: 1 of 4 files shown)


src/zeroband/inference/toploc.py
```diff
@@ -174,7 +184,9 @@ def toploc_cache_hook(_, inputs: tuple, toploc_cache: TopLocCache):
     toploc_cache.add(seq_ids, hidden_states)
 
 
-def setup_toploc_cache(llm: LLM, disable: bool = False, **toploc_kwargs) -> tuple[TopLocCache, RemovableHandle | None]:
+def setup_toploc_cache(
```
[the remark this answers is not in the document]
Jackmin801: It didnt have a long docstring with params and stuff to begin with so the docstring didnt become wrong. Not sure if its worth adding

src/zeroband/inference/toploc.py
```diff
@@ -185,4 +197,7 @@ def setup_toploc_cache(llm: LLM, disable: bool = False, **toploc_kwargs) -> tupl
     if not disable:
         handle = logits_processor.register_forward_pre_hook(partial(toploc_cache_hook, toploc_cache=toploc_cache))
 
+    if pipeline_config is not None and pipeline_config.world_size > 1 and pipeline_config.rank < pipeline_config.world_size - 1:
```
mikasenghaas: pipeline config cannot be `None` so we can skip the first check. this should be caught by pydantic. also, the last check we should add as a model validation on the pydantic object not here. and then we have a single source of truth throughout the script for whether pipelining is happening or not (world_size > 1 seems like an intuitive one)

mikasenghaas (approved):

src/zeroband/inference/toploc.py
Jackmin801: Ah the pipeline config having none default is used by the toploc unit tests. We could create a pipeline config in the toploc test I guess

oh so as a bool? say `pipeline_config.pipeline_enabled`?

Jackmin801:
```python
class PipelineConfig(BaseConfig):
    # The rank of the current node in the pipeline
    rank: int = 0

    # The total number of nodes in the pipeline (e.g. the number of PP model shards)
    world_size: int = 1

    # The seed used to create the public node address (optional, will lead to deterministic connection strings)
    iroh_seed: int | None = None

    # The peer ID to connect to (optional, if not provided, the user will be prompted to enter it)
    iroh_peer_id: str | None = None

    # How many times to retry connection to peer (each retry takes ~30s)
    connection_num_retries: int = 10  # Each retry takes ~30s, so 10 retries is ~300s (5min)

    @property
    def pipeline_enabled(self) -> bool:
        """Returns True if pipeline parallelism is enabled (world_size > 1)."""
        return self.world_size > 1

```

Jackmin801: address both. thanks!

PR #353, opened 2025-06-02
Add prompts and completions to output parquet

mikasenghaas:
Adds fields `prompt` (`str`) and `completion` (`str`) to the output parquet written by the output worker. Uses the `prompt` before formatting (e.g. adding "thinking budget" information, applying chat template, etc.). Convenience feature to display generated samples on the dashboard without requiring a tokenizer.

*Note: This require `toploc-validator` to check for a match between `input_tokens` and `prompt` as well as `output_tokens` and `completion`.*

Based on PR #346, which should be merged before.

### DIFF
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -220,12 +220,12 @@ def inference(config: Config):
         for target_length, verification_info in zip(target_lengths, verification_infos):
             verification_info["target_length"] = target_length
 
-        prompts = format_prompts(
+        formatted_prompts = format_prompts(
             prompts, target_lengths, config.rewards.len_reward, tokenizer=tokenizer, enable_thinking=config.enable_thinking
         )
 
         start_time = time.time()
-        request_outputs = llm.generate(prompts, sampling_params, use_tqdm=False)
+        request_outputs = llm.generate(formatted_prompts, sampling_params, use_tqdm=False)
         end_time = time.time()
 
         # Dropping like this isn't ideal. But in practice, we shouldn't have any prompts that are too long.
@@ -299,6 +299,7 @@ def inference(config: Config):
         table = get_parquet_table(
             request_outputs,
             request_rewards,
+            prompts,
             proofs,
             ckpt_step,
             target_lengths,
--- a/src/zeroband/utils/parquet.py
+++ b/src/zeroband/utils/parquet.py
@@ -4,6 +4,8 @@ pa_schema = pa.schema(
     [
         ("input_tokens", pa.list_(pa.int32())),
         ("output_tokens", pa.list_(pa.int32())),
+        ("prompt", pa.string()),
+        ("completion", pa.string()),
         ("advantages", pa.float32()),
         ("rewards", pa.float32()),
         ("task_rewards", pa.float32()),
--- a/src/zeroband/inference/parquet.py
+++ b/src/zeroband/inference/parquet.py
@@ -8,6 +8,7 @@ from zeroband.utils.parquet import pa_schema
 def get_parquet_table(
     request_outputs: list[RequestOutput],
     request_rewards: list[RequestRewards],
+    prompts: list[str],
     proofs: list[bytes],
     step: int,
     target_lengths: list[int],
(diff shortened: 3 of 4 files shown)


mikasenghaas:
Checking the parquet output:

```python
import pyarrow.parquet as pq

table = pq.read_table("outputs/step_0/f5d2671a-3ae2-46be-8315-a61b60efbbc7.parquet")
df = table.to_pandas()

sample = df.iloc[0]
print(f"Prompt:\n{sample.prompt}\n\nCompletion:\n{sample.completion}")
```

Example from `Qwen3-0.6B` with `enable_thinking` set:

```bash
Prompt:
The angle between the line $x=-2$ and the line $\sqrt{3}x-y+1=0$ is ______.

Completion:
<think>
Okay, so I need to find the angle between the line x =
```

Example from `Qwen3-0.6B` without `enable_thinking` set:

```bash
Prompt:
The angle between the line $x=-2$ and the line $\sqrt{3}x-y+1=0$ is ______.

Completion:
<think>
Okay, so I need to find the angle between the line x =
```

mikasenghaas:
Screenshot 2025-06-02 at 3 38 06 PM

src/zeroband/utils/parquet.py
```diff
@@ -4,6 +4,8 @@
     [
         ("input_tokens", pa.list_(pa.int32())),
         ("output_tokens", pa.list_(pa.int32())),
+        ("prompt", pa.string()),
+        ("completion", pa.string()),
```
samsja, line 8: probably need to add comments in the parquet definition to explain what is the diff between promot and input_tokens .

Should we rename input_tokens to prompt_tokens ?

mikasenghaas: mmh, i think it's pretty clear from the name and the types. i do see an argument for aligning the naming in general to have `input_xyz` and `output_xyz`. could do  `input_text` and `output_text` instead of `prompt` and `completion`.

also, a convention i generally like is to use `token_ids` whenver there is tokenized inputs. this seems pretty standard across hf, vllm etc. but i didnt' change it yet bc it seemed clear enough for now. but if you think it's not im down to change

mikasenghaas: another alternative would be to go full vllm:
- `prompt`
- `prompt_token_ids`
- `completion`
- `completion_token_ids`

.gitignore
```diff
@@ -7,6 +7,7 @@ output/*
 outputs/*
 data/*
 data_rollout/*
+test.ipynb
```
Jackmin801: Im guessing this is for dev? rmb to remove. might confuse someone in the future

src/zeroband/infer.py
```diff
@@ -317,6 +293,7 @@ def inference(config: Config):
         table = get_parquet_table(
             request_outputs,
             request_rewards,
+            prompts,
```
Jackmin801, line 302: should we be saving the formatted prompts instead?

.gitignore
samsja: Or rather do 
```
*.ipynb
```

mikasenghaas: i would defo like to opt sami's suggestion. im using jupyters for debugging quite a bit and it's annoying to delete everytime/ be careful on not committing on

src/zeroband/infer.py
mikasenghaas: mmh the reason i save the raw prompts is that i thought it would the chat template would look weird in the dashboard which is the primary usecase for now

src/zeroband/utils/parquet.py
mikasenghaas: i will adopt this naming scheme then i take it?

.gitignore
samsja: I have the same issue

samsja (approved):
pre-approving

src/zeroband/utils/parquet.py
mikasenghaas: I actually decided against it. I think `{prompt|completion}_token_ids` is too verbose. When we say "tokens" this should imply that this this is a list of integers. I propose:
- `input_tokens` and `output_tokens` for the tokenized prompt and completion (`list[int]`)
- `num_{input|output}_tokens` to refer to the length of the above lists (`int`)
- `prompt` and `completion` for the string version of the above (`str`)
- `batch_input_tokens`, `batch_output_tokens` for a batch of tokenized prompts/ completions (`list[list[int]]`) and analogously `batch_prompts` and `batch_completions` (`list[str]`)

mikasenghaas:
Screenshot 2025-06-04 at 11 29 55 AM

PR #355, opened 2025-06-02
add reward to the log

samsja:


src/zeroband/infer.py
```diff
@@ -306,6 +306,13 @@ def inference(config: Config):
         request_rewards = compute_vllm_rewards(request_outputs, verification_infos, task_types, config.rewards)
         logger.info(f"Computed rewards and advantages in {time.time() - start:.2f}s")
 
+        total_rewards = sum(sum(r.reward for r in req.rewards) for req in request_rewards)
+        total_completions = sum(len(req.rewards) for req in request_rewards)
```
mikasenghaas: we already compute `total_completions` as `batch_samples` a bit further up, no?

src/zeroband/infer.py
```diff
@@ -306,6 +306,13 @@ def inference(config: Config):
         request_rewards = compute_vllm_rewards(request_outputs, verification_infos, task_types, config.rewards)
         logger.info(f"Computed rewards and advantages in {time.time() - start:.2f}s")
 
+        total_rewards = sum(sum(r.reward for r in req.rewards) for req in request_rewards)
```
mikasenghaas: and can we call this `batch_rewards` for consistency reasons

src/zeroband/infer.py
```diff
@@ -306,6 +306,13 @@ def inference(config: Config):
...
+        total_completions = sum(len(req.rewards) for req in request_rewards)
+        avg_reward_to_log = total_rewards / total_completions
+
+        monitor.log({"avg_reward": avg_reward_to_log})
```
mikasenghaas: if we want to log to monitor can we find a grouping for it? existing ones are `progress`, `performance` and `output` but im not sure it fits in any of those categories

also, is this the analogous thing of how we compute batch rewards during training? average of all, and not average over average. just making sure we don't have a different value in infer and training logs bc of this

samsja: we can put it into a "reward" category like during training ?

mikasenghaas: yea, i like that.

mikasenghaas (approved):

PR #357, opened 2025-06-02
Non root docker

samsja:
Validated that it is working by running 

```bash
python src/zeroband/infer.py @ configs/inference/debug.toml
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,24 +57,37 @@ RUN uv sync && uv sync --extra fa
 
 # Runtime stage
 FROM python:3.11-slim
-WORKDIR /root/prime-rl
+# 1. Create the non-root account (UID 1000 / GID 1000)  ────────────────
+ARG USER_ID=1000
+ARG GROUP_ID=1000
+RUN groupadd --gid $GROUP_ID appuser  && \
+    useradd  --uid $USER_ID --gid appuser --create-home --shell /bin/bash appuser && \
+    mkdir -p /opt/prime-rl/.venv/bin && \
+    chown -R appuser:appuser /opt/prime-rl
+   
+# 2. Put your code under /opt not /root, then hand ownership to appuser ─
+WORKDIR /opt/prime-rl
 
-RUN apt-get update && apt-get install -y --no-install-recommends --force-yes build-essential wget clang
+# still need a compiler for wheels that ship only as source
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        build-essential wget clang && \
+    rm -rf /var/lib/apt/lists/*
 
-# Copy virtual environment
-COPY --from=builder /root/prime-rl/.venv /root/prime-rl/.venv
-RUN rm /root/prime-rl/.venv/bin/python
-RUN ln -s /usr/local/bin/python /root/prime-rl/.venv/bin/python
-ENV PATH="/root/prime-rl/.venv/bin:$PATH"
+# Copy the virtual-env and fix permissions
+COPY --from=builder /root/prime-rl/.venv /opt/prime-rl/.venv
+RUN chmod -R 755 /opt/prime-rl/.venv && \
+    ln -sf /usr/local/bin/python /opt/prime-rl/.venv/bin/python
 
-# Note(Jack): Nothing should need to compile so we don't need these
-# COPY --from=builder /usr/local/cuda-12.4 /usr/local/cuda
-# ENV LD_LIBRARY_PATH=/usr/local/cuda-12.4/lib64:$LD_LIBRARY_PATH
-# ENV CUDA_HOME=/usr/local/cuda-12.4
-# ENV PATH="/usr/local/cuda-12.4/bin:$PATH"
-
-# Copy application files
+# Copy sources and configs
 COPY --from=builder /root/prime-rl/src ./src
-COPY ./configs/ ./configs/
+COPY ./configs ./configs
+
+# 3. Adjust permissions so the new user can read/write everything ───────
+RUN chown -R appuser:appuser /opt/prime-rl
+
+# 4. Activate the venv + switch user before ENTRYPOINT ──────────────────
+ENV PATH="/opt/prime-rl/.venv/bin:${PATH}"
+
+USER appuser         
 
 ENTRYPOINT ["python", "src/zeroband/infer.py"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -57,24 +64,34 @@ RUN uv sync && uv sync --extra fa
...
-RUN rm /root/prime-rl/.venv/bin/python
-RUN ln -s /usr/local/bin/python /root/prime-rl/.venv/bin/python
-ENV PATH="/root/prime-rl/.venv/bin:$PATH"
 
-# Note(Jack): Nothing should need to compile so we don't need these
-# COPY --from=builder /usr/local/cuda-12.4 /usr/local/cuda
-# ENV LD_LIBRARY_PATH=/usr/local/cuda-12.4/lib64:$LD_LIBRARY_PATH
-# ENV CUDA_HOME=/usr/local/cuda-12.4
-# ENV PATH="/usr/local/cuda-12.4/bin:$PATH"
+# 4. Activate the venv + switch user before ENTRYPOINT ──────────────────
```
Jackmin801: 1,2,4? also `-----------`

samsja: fixed

Dockerfile
```diff
@@ -35,6 +35,13 @@ RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.
 # ENV PATH="/root/.cargo/bin:${PATH}"
 # RUN echo "export PATH=\"/opt/conda/bin:/root/.cargo/bin:\$PATH\"" >> /root/.bashrc
 
+# 1. Create the non-root account (UID 1000 / GID 1000)  ────────────────
+ARG USER_ID=1000
+ARG GROUP_ID=1000
+RUN groupadd --gid $GROUP_ID appuser  && \
+    useradd  --uid $USER_ID --gid appuser --create-home --shell /bin/bash appuser
+
```
Jackmin801: I think its better to keep the builder root. I dont think we ever switch in the builder anyways so I think this doesnt rlly do anything useful

Dockerfile
```diff
@@ -35,6 +35,13 @@ RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.
 # ENV PATH="/root/.cargo/bin:${PATH}"
 # RUN echo "export PATH=\"/opt/conda/bin:/root/.cargo/bin:\$PATH\"" >> /root/.bashrc
 
+# Create the non-root account (UID 1000 / GID 1000)  ────────────────
+ARG USER_ID=1000
+ARG GROUP_ID=1000
+RUN groupadd --gid $GROUP_ID appuser  && \
+    useradd  --uid $USER_ID --gid appuser --create-home --shell /bin/bash appuser
+
```
Jackmin801: eh for some reason ur edit made this outdated but yea, maybe dont non-root the builder: https://chatgpt.com/share/683dfcd4-fd54-8011-8458-2e202b56495e

Dockerfile
```diff
@@ -57,24 +64,33 @@ RUN uv sync && uv sync --extra fa
 
 # Runtime stage
 FROM python:3.11-slim
-WORKDIR /root/prime-rl
 
-RUN apt-get update && apt-get install -y --no-install-recommends --force-yes build-essential wget clang
+# 2. Put your code under /appuser not /opt, then hand ownership to appuser ─
```
Jackmin801: comment from claude here is a little weird. was never in /opt?

Dockerfile
```diff
@@ -57,24 +64,33 @@ RUN uv sync && uv sync --extra fa
...
+# Create non-root user in runtime stage
+ARG USER_ID=1000
+ARG GROUP_ID=1000
+RUN groupadd --gid $GROUP_ID appuser && \
+    useradd --uid $USER_ID --gid appuser --create-home --shell /bin/bash appuser
+
+# Copy the virtual-env and fix permissions
+COPY --from=builder /appuser/prime-rl/.venv /appuser/prime-rl/.venv
+RUN ln -sf /usr/local/bin/python /appuser/prime-rl/.venv/bin/python && \
+    chown -R appuser:appuser /appuser
```
Jackmin801: id prefer this in its own line instead of as && as its kind of unrelated to venv

Dockerfile
```diff
@@ -57,24 +64,33 @@ RUN uv sync && uv sync --extra fa
...
+
+# Copy sources and configs
+COPY --from=builder /appuser/prime-rl/src ./src
+COPY ./configs ./configs
```
Jackmin801: is this covered by the chown?

Dockerfile
```diff
@@ -57,24 +64,33 @@ RUN uv sync && uv sync --extra fa
...
-RUN rm /root/prime-rl/.venv/bin/python
-RUN ln -s /usr/local/bin/python /root/prime-rl/.venv/bin/python
-ENV PATH="/root/prime-rl/.venv/bin:$PATH"
 
-# Note(Jack): Nothing should need to compile so we don't need these
-# COPY --from=builder /usr/local/cuda-12.4 /usr/local/cuda
-# ENV LD_LIBRARY_PATH=/usr/local/cuda-12.4/lib64:$LD_LIBRARY_PATH
-# ENV CUDA_HOME=/usr/local/cuda-12.4
-# ENV PATH="/usr/local/cuda-12.4/bin:$PATH"
+ENV PATH="/appuser/prime-rl/.venv/bin:${PATH}"
```
Jackmin801: Move up as this is related to venv

Dockerfile
samsja: I had issue tho when doing it that way, the user could not use the virtual env of the root, I guess its probably fixable but require more work

Jackmin801: oh it should be fixed by the chown later I think

Dockerfile
```diff
@@ -14,67 +16,71 @@ ENV PATH=$PATH:/usr/local/cuda/bin
...
+    --no-install-recommends \
+    --force-yes \
+    build-essential \
+    wget \
+    clang \
+    tmux \
+    iperf \
+    openssh-server \
+    git-lfs \
+    gpg \
```
Jackmin801, line 64: do we need these dependencies?

Jackmin801 (approved):
lfgtm! looks rlly clean now

Dockerfile
samsja: could be usefull for development at some point

PR #359, opened 2025-06-03
complex json output prediction

justusmattern27:
task: the model is given a few very simple questions - the main challenge is around relatively complex instructions for the json output format - this is very relevant for real world applications where users need structured outputs. Interestingly, even in easy variations of this task, relatively smart models like r1 qwen 7B distill and even R1-0528-Qwen3-8B fail very often, despite spending very little time on thinking (<1k tokens) - they could clearly have better performance by just double checking the formatting instructions, so im hopeful that this is a good RL task

Screenshot 2025-06-03 at 00 10 19


justusmattern27:
example dataset (might still extend): https://huggingface.co/datasets/justus27/complex-json-outputs

mikasenghaas (approved):

samsja (approved):
lgtm

PR #360, opened 2025-06-03
[Feat] toploc2

Jackmin801:


Jackmin801:
image
graphs look ok

src/zeroband/inference/config.py
```diff
@@ -18,6 +18,13 @@ class SamplingParamConfig(BaseConfig):
     seed: int | None = None
     logprobs: int | None = 0  # put to None to disable logprobs calculation
 
+    @model_validator(mode="after")
+    def convert_negative_logprobs_to_none(self):
+        """Convert negative logprobs values to None to disable logprobs calculation."""
+        if self.logprobs is not None and self.logprobs < 0:
+            self.logprobs = None
+        return self
```
mikasenghaas, line 26: Feels more intuitive to err when passing negative values?

src/zeroband/infer.py
```diff
@@ -81,7 +82,10 @@ def inference(config: Config):
         disable_async_output_proc=True,  # We have an off by 1 error in toploc without this flag when cuda graph padding is enabled.
         download_dir=config.download_dir,
         dtype="bfloat16" if config.dtype == "bf16" else torch.float32,
+        enable_chunked_prefill=False,  # This is required for toploc2 because chunked prefill seems to allow len(seq_groups) != len(selected_token_indices) which is unexpected
     )
+    if config.sampling.logprobs is None and envs.PRIME_GROUP_ID is not None:
+        llm.llm_engine.model_executor.driver_worker.model_runner.sampler = Toploc2Sampler()
```
mikasenghaas, line 88: would be nice to lift this constraint. ofc don't need for syn-2 but this seems avoidable and for any rl run we defo want toploc-2 + logprobs. what would be needed to make this happen?

src/zeroband/inference/toploc2.py
mikasenghaas, line 1: nit: imo all toploc stuff should live in a single `toploc.py` file and then we have settings and setup for both v1 and v2. but can do later

src/zeroband/inference/toploc2.py
```diff
@@ -0,0 +1,143 @@
...
+    _apply_min_p,
+    _apply_min_tokens_penalty,
+    _apply_top_k_top_p,
+    _build_sampler_output,
+)
+from vllm.model_executor.layers.utils import apply_penalties
+from vllm.model_executor.sampling_metadata import SamplingMetadata, SamplingTensors
+from vllm.sequence import Logprob
+
+GUMBEL_BATCH_SIZE = 2**16
```
mikasenghaas, line 22: maybe add a comment why this is necessary?

tests/integration/inference/test_debug_pp.py
```diff
@@ -16,7 +16,7 @@
     "src/zeroband/infer.py",
     "@configs/inference/debug.toml",
     "--model-name",
-    "mikasenghaas/Qwen3-0.6B-0.2",
+    "Qwen/Qwen3-0.6B",
```
mikasenghaas: nice that you found this, but can we actually move this out of this pr. it's also fixed in the config refactor pr - one less conflict for me to resolve:))

tests/integration/inference/test_debug_pp.py
```diff
@@ -35,7 +35,7 @@
     "src/zeroband/infer.py",
     "@configs/inference/debug.toml",
     "--model-name",
-    "mikasenghaas/Qwen3-0.6B-1.2",
+    "Qwen/Qwen3-0.6B",
```
mikasenghaas: same here

src/zeroband/infer.py
```diff
@@ -81,7 +82,10 @@ def inference(config: Config):
         disable_async_output_proc=True,  # We have an off by 1 error in toploc without this flag when cuda graph padding is enabled.
         download_dir=config.download_dir,
         dtype="bfloat16" if config.dtype == "bf16" else torch.float32,
+        enable_chunked_prefill=False,  # This is required for toploc2 because chunked prefill seems to allow len(seq_groups) != len(selected_token_indices) which is unexpected
     )
+    if config.sampling.logprobs is None and envs.PRIME_GROUP_ID is not None:
```
mikasenghaas: oh also, can we make toploc configurable, ie. simplest solution would be to add a `--toploc2` flag to configs. later down the line it would be nice to have a separate `ToplocConfig` with all toploc settings (at least two boolean flags to enable/disable v1/2)

mikasenghaas (approved):
hell yea! nice job. have we tested that the produced output tokens are identical for the sampler and toploc sampler, ie. are we *absolutely* sure we are not altering model behavior? could be nice to have a simple test for this?

also, i think adding toploc 2 into the configs is also important. let's get this merged soon, so that i can rebase onto the config refactor:)

samsja (approved):
lfgtm v2

src/zeroband/inference/config.py
Jackmin801: this is necessary to disable logprobs. I couldnt find a way to pass none and since the default is 0, it didnt seem like there was a way to make it None other than this

src/zeroband/infer.py
Jackmin801: yeep

Jackmin801: added logprobs into the toploc2 sampler

src/zeroband/inference/toploc2.py
Jackmin801: hrmm toploc1 and toploc2 are quite different. I think separate files make sense

Jackmin801: massive files are hard for the LLMs

src/zeroband/infer.py
Jackmin801: added flag

Jackmin801:
Should be ok. lets merge!
image

PR #364, opened 2025-06-04
add infinite dataset

samsja:


Jackmin801 (approved):
nice! lgtm

src/zeroband/infer.py
```diff
@@ -336,6 +338,8 @@ def inference(config: Config):
             logger.info(f"Reached total step {config.total_step}, stopping inference")
             break
 
+        i += batch_size
```
mikasenghaas: can we maybe find a more descriptive name than i here?

mikasenghaas (approved):
lgtm! at some point (probably when we refactor configs) we should think about which stopping criteria we want to support. could do step but that's quite abstract. obvious ones are max problems, max samples, maybe both and stop whichever is saturated first?

src/zeroband/infer.py
samsja: sure

samsja: done

PR #365, opened 2025-06-05
amd support

mikex86:


src/zeroband/inference/utils.py
```diff
@@ -5,7 +5,7 @@
 from safetensors import safe_open
 from transformers import AutoTokenizer
 from vllm import LLM
-from vllm.model_executor.model_loader.loader import _process_weights_after_loading
+from vllm.model_executor.model_loader.utils import process_weights_after_loading
```
mikasenghaas: this import is not defined?

mikex86: We were stuck on vllm 0.8.5 because we accessed an internal method. This method has been made public. We have to for the sake of amd support update to a newer version and now we can.

.github/workflows/release.yaml
```diff
@@ -62,3 +63,60 @@ jobs:
...
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile.rocm
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+
+      - name: Test import
+        run: docker run --entrypoint python primeintellect/prime-rl-rocm:latest -c "import zeroband.inference"
+
```
Jackmin801, line 120: should we test before push? also fine if we want to keep the build-push action and not split it

Jackmin801 (approved):
666

.github/workflows/release.yaml
mikasenghaas: yea probably a good idea in general, but we have this setup in this way for time now and i don't wanna spend time on puling apart the build and push now

mikasenghaas:
Screenshot 2025-06-19 at 8 11 36 PM

samsja (approved):
lgtm

PR #366, opened 2025-06-06
add problem_id field to parquet output

justusmattern27:
We need problem ids to group together solutions for the same problems for synthetic-2. This was vibecoded with codex btw

### DIFF
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -307,7 +307,7 @@ def inference(config: Config):
             proofs,
             ckpt_step,
             target_lengths,
-            [p["problem_id"] for p in problems],
+            problems,
         )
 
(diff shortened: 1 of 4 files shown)


src/zeroband/infer.py
```diff
@@ -307,6 +307,7 @@ def inference(config: Config):
             proofs,
             ckpt_step,
             target_lengths,
+            [p["problem_id"] for p in problems],
```
mikasenghaas: can we pass problems as func arg and do a zipped iteration in there

samsja (approved):
lgtm

src/zeroband/infer.py
justusmattern27: done

PR #367, opened 2025-06-06
[Fix] double bos

Jackmin801:


mikasenghaas (commented):
mmh we can push it as a hot fix for now, but i'm not a huge fan to have long term. seems super hacky and immediately breaks with any other special tokens than bos. have you checked that none of the below is possible?
1. set a flag to not set special tokens on apply chat template
2. disable applying special tokens in vllm
3. set a custom chat template passed to vllm (where we can define what to do with special tokens)

Jackmin801:
> mmh we can push it as a hot fix for now, but i'm not a huge fan to have long term. seems super hacky and immediately breaks with any other special tokens than bos. have you checked that none of the below is possible?
> 
> 1. set a flag to not set special tokens on apply chat template
> 2. disable applying special tokens in vllm
> 3. set a custom chat template passed to vllm (where we can define what to do with special tokens)

1. yea doesnt seem so. 
2. we'd have to make a custom tokenizer class that wraps this functionality which seems kind of overkill
3. we can change the chat template but I think thats not very maintainable

The 2nd test that checks tokenize=True == tokenize=False actually covers other special tokens. The main issue is actually that `apply_chat_template` doesnt actually support `tokenize=False` well. The ground truth of what its expected to do should be with `tokenize=True` which is being compared in that test

mikasenghaas:
sg, let's merge (and keep an eye out for any vllm/ hf upstream changes that might make our life easier)

PR #368, opened 2025-06-06
Model infer clipping

samsja:


configs/inference/stability_ablation/math_only_7b.toml
justusmattern27, line 1: why is the batch size so huge? Here we have 352*4*16 = 22k, why?

justusmattern27 (approved):

configs/inference/stability_ablation/math_only_7b.toml
samsja: the argument change recenclty. Now the batch size means problem * sampling.n. We divide later by n to get the num problem

justusmattern27: i see makes sense

justusmattern27 (approved):

src/zeroband/infer.py
```diff
@@ -76,7 +76,12 @@ def inference(config: Config):
         dtype="bfloat16" if config.dtype == "bf16" else torch.float32,
     )
     tokenizer = llm.get_tokenizer()
-    sampling_params = SamplingParams(**config.sampling.model_dump())
+
+    # Adjust sampling params based on config
+    sampling_config = config.sampling.model_dump()
+    if not config.enable_logprobs:
+        sampling_config["logprobs"] = None
+    sampling_params = SamplingParams(**sampling_config)
```
Jackmin801: what do you think about making the default None and enable_logprobs setting it to 1?

src/zeroband/inference/parquet.py
```diff
@@ -22,10 +40,18 @@ def get_parquet_table(
         assert request_output.request_id == request_rewards.request_id
         for output, reward in zip(request_output.outputs, request_rewards.rewards):
             assert output.index == reward.completion_id
+
+            # Extract logprobs if enabled and available
+            output_logprobs = extract_logprobs(output.logprobs) if enable_logprobs else None
+            # For input logprobs, we typically don't have them from vllm, so set to None or zeros
```
Jackmin801: I believe there is a way to get them. However, we shouldnt need them so maybe it could be better to set that as the comment

src/zeroband/train.py
```diff
@@ -160,14 +166,25 @@ def train(config: Config):
...
 
-        if config.kl_coef is not None:
+        if config.grpo.kl_coef is not None:
             model_reference = torch.compile(model_reference) if not TYPE_CHECKING else model_reference
 
-    if config.kl_coef is not None:
+        if config.logprob_mode == "infer_model":
+            model_for_logprob_only = torch.compile(model_for_logprob_only) if not TYPE_CHECKING else model_for_logprob_only
+
+    tensor_offloaded_repository: dict[int, OffloadedTensor] = {}
```
Jackmin801, line 175: why do we need a repository?

Jackmin801 (approved):
nice! lgtm.

configs/training/stability_ablation/math_only_1b.toml
```diff
@@ -13,9 +13,9 @@ path = "data_rollout"
 
 [optim]
 warmup_steps = 1
-batch_size = 512
+batch_size = 2048
 total_steps = 10000000000000000000
-step_per_rollout = 4
+step_per_rollout = 1
```
mikasenghaas, line 18: yayy

src/zeroband/infer.py
```diff
@@ -76,7 +76,12 @@ def inference(config: Config):
         dtype="bfloat16" if config.dtype == "bf16" else torch.float32,
     )
     tokenizer = llm.get_tokenizer()
-    sampling_params = SamplingParams(**config.sampling.model_dump())
+
+    # Adjust sampling params based on config
+    sampling_config = config.sampling.model_dump()
+
+    sampling_config["logprobs"] = 0
```
mikasenghaas: why are we setting this unconditionally at runtime?

src/zeroband/inference/config.py
```diff
@@ -50,6 +49,9 @@ class Config(BaseConfig):
     # Whether to enable thinking for the model. Used by the `format_prompts` function to prepend a thinking prompt
     enable_thinking: bool = True
 
+    # Whether to calculate and save logprobs in the output
+    enable_logprobs: bool = False
```
mikasenghaas: ah are we just diverging in terms of args definition from vllm here? what's the reason for that?

src/zeroband/inference/parquet.py
```diff
@@ -1,17 +1,35 @@
 import pyarrow as pa
 from vllm import RequestOutput
+from vllm.sequence import SampleLogprobs
 
 from zeroband.inference.rewards import RequestRewards
 from zeroband.utils.parquet import pa_schema
 
 
+def extract_logprobs(sample_logprobs: SampleLogprobs | None) -> list[float] | None:
+    """Extract logprobs from vllm output. Returns None if logprobs are not available."""
```
mikasenghaas: can we add some more docs to this? it's not entirely clear without reading the context in which the function is called whether it operates at token or sequence level

mikasenghaas: upon closer reading, it seems like we get the logprobs (list[list[float]]) for the entire sequence but only save the top logprob per token and return as list[float]? this should defo be in the docs of this function imo

src/zeroband/training/config.py
```diff
@@ -54,6 +54,33 @@ def check_path_and_interval(self):
...
+    epsilon_high: float = 0.2
+    clip_ratio: float = 4.0
+
+
+class RatioConfig(BaseConfig):
+    type: Literal["ratio"] = "ratio"
+    clip_ratio: float = 8.0
+
+
+LossTypeConfig: TypeAlias = ClippingConfig | KlCovConfig | RatioConfig
```
mikasenghaas: clipping and ratio do not seem to be loss types? i'm missing a bit of context here, but something like `GRPOVariantsConfig` seems more correct?

src/zeroband/training/config.py
```diff
@@ -92,6 +112,10 @@ class Config(BaseConfig):
 
     normalize_batch_to_token_count: bool = False
 
+    logprob_mode: Literal["infer_model", "vllm"] = "infer_model"
```
mikasenghaas: i don't get these two keys - they seem to be the same? what is the naming idea here?

src/zeroband/training/config.py
samsja: good point

src/zeroband/inference/config.py
samsja: sami inconsistency is a mark of prime software

src/zeroband/infer.py
samsja: this paramters should not be set via sampling config. Its not really set at runtime tho, just different parametes passing

src/zeroband/training/config.py
samsja: updated to "recompute_infer_logprobs" instead of "infer_model"

src/zeroband/inference/config.py
samsja: roll back to just using the logprobs param

src/zeroband/infer.py
mikasenghaas: why not set via sampling config?

src/zeroband/training/config.py
mikasenghaas: what about `compute_logsprobs: Literal["training", "inference"]` or `recompute_logprobs: bool`

PR #373, opened 2025-06-09
Fix connection lost during all reduce

mikasenghaas:
[WIP] Fixes an issue where a connection lost exception is raised when all-reducing the local max batch size.


mikasenghaas:
This is the error message reported by @JannikSt.

```txt
06-09 00:01:18 [INFER] [DEBUG] Initial tensor: 5
[rank0]: Traceback (most recent call last):
[rank0]:   File "/app/src/zeroband/infer.py", line 404, in <module>
[rank0]:     main(config)
[rank0]:   File "/app/src/zeroband/infer.py", line 366, in main
[rank0]:     inference(config)
[rank0]:   File "/app/src/zeroband/infer.py", line 92, in inference
[rank0]:     batch_size = all_reduce(node, torch.tensor(local_batch_size), config=config.pp, op=torch.min).item()
[rank0]:                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/app/src/zeroband/inference/pipeline.py", line 491, in all_reduce
[rank0]:     send_future.wait()
[rank0]: RuntimeError: connection lost
```

This error messages indicates that a `ConnectionError` ([Docs](https://docs.rs/iroh/latest/iroh/endpoint/enum.ConnectionError.html)) is thrown during the async send operation in `prime-iroh`. There are a number of potential reasons for a connection error to occur, with the most important being `ConnectionClosed` and `TimedOut`. In the former case, a node crashed and we cannot establish connection. In this case we want to propagate the error and crash the prime-rl worker. The latter case should not occur and would imply a bad default. We can likely adjust the default QUIC timeout within `prime-rl`.

mikasenghaas:
Let's try to reproduce the issue. We create a pipeline with four workers (as here we might have some nodes setting up all connections to their adjacent peers while non-peers are still not set up).

First, set environment variables

```bash
export PRIME_LOG_LEVEL=DEBUG
export RUST_LOG=prime_iroh=debug
export PRIME_GROUP_ID=0x550143DCA43E8a59C27E7c12F91Dd76c7609Dde8
export CUDA_VISIBLE_DEVICES=
```

*Set the `CUDA_VISIBLE_DEVICES` to the same GPU rank as PP rank*

Then, in a TMUX session in separate panes, start the workers in the group

```bash
# Node 1
uv run src/zeroband/infer.py @configs/inference/debug.toml \
	--pp.rank 0 \
	--pp.world-size 4 \
	--pp.iroh-seed 9 \
	--pp.iroh-peer-id ff572e291402ae6a3952e54459c349acd635908e2dd34a7c02f04c88d8a616a6 \
	--pp.connection_num_retries 100 \
	--batch-size auto
```

```bash
# Node 2
uv run src/zeroband/infer.py @configs/inference/debug.toml \
	--pp.rank 1 \
	--pp.world-size 4 \
	--pp.iroh-seed 10 \
	--pp.iroh-peer-id f69f4d12b2283bc43a6dd8f0e83df69ffa91cc9e76cca77c0f85b3fa9854f55a \
	--pp.connection_num_retries 100 \
	--batch-size auto
```

```bash
# Node 3
uv run src/zeroband/infer.py @configs/inference/debug.toml \
	--pp.rank 2 \
	--pp.world-size 4 \
	--pp.iroh-seed 11 \
	--pp.iroh-peer-id c15efa1d4b0a2f4473c694703df14a70c1da9bca8772db974fd4631c87b90463 \
	--pp.connection_num_retries 100 \
	--batch-size auto
```

```bash
# Node 4
uv run src/zeroband/infer.py @configs/inference/debug.toml \
	--pp.rank 3 \
	--pp.world-size 4 \
	--pp.iroh-seed 13 \
	--pp.iroh-peer-id 00d21610e478bc59b0c1e70505874e191bf94ab73cb1f9246f963f9bc0a1b253 \
	--pp.connection_num_retries 100 \
	--batch-size auto
```

mikasenghaas:
Experiments:
✅ Correct inference: Starting all workers correctly sets up communication and leads to correct generation
✅ Connection retries reached: Starting all but the PP rank 3 with `10` (default) connection retries raises an exception on PP rank 2 because it cannot (correctly) connect to PP rank 3. PP rank 1 has setup its communication and starts the all-reduce but does not time out. 
✅ Crash worker: Starting all ranks correct, but crash PP rank 3 after succcessfully seting up peer connection raises and exception on PP rank 2 with `ConnectionError` (`ConnectionClosed`)
✅ Increasing the connection retries to 100 (as in production) and starting all but PP rank 3...

mikasenghaas:
I cannot reproduce triggering a connection lost exception caused by a delay in the peer setup. The only way I could trigger an exception in a send  was to actually crash one the peer processes, which is expected behavior.

I will close this issue for now. Can reopen if we see the issue again in production.

PR #375, opened 2025-06-09
[PRI2-619] Fix node address not found

mikasenghaas:
Fixes an issue where the prime-iroh connection retry mechanism can spin in a tight loop, retrying 500 times almost instantly instead of over the intended duration which would manifest in the following error message.

RuntimeError: Failed to connect 636e74e8a8->33021e04dd after 500 tries (left: 0)
No addressing information for NodeId(33021e04dd), unable to connect

Based on the retry interval (~30s), 500 retries should take roughly 4 hours — which is impossible given the actual observed time (a few milliseconds). The suspected reason for this is that we pre-compute the peer addresses on the protocol side which means that if a node tries to connect to a peer before the peer has started his endpoint, the address is not reachable and we immediately throw an exception (without 30s timeout which happens if the address was found but the node is not yet reachable).


mikasenghaas:
Screenshot 2025-06-09 at 5 11 09 PM

PR #377, opened 2025-06-09
Hotfix E2E test

mikasenghaas:
Uses the vLLM internal request ID as a fallback if the data sample does not have a problem ID. This is a hotfix to have the e2e tests running again.


mikasenghaas:
Screenshot 2025-06-09 at 4 31 38 PM

PR #379, opened 2025-06-09
[fix] use token prompt

Jackmin801:
The apply_chat_template function is just kinda weird with tokenize=False. Lets just do tokenize=True and pass token prompt. This is actually what llm.chat does anyways but theres some other edge case stuff in llm.chat that im not sure we need and they scare me a bit.


samsja (approved):
lgtm would wait for a screenshot of the ci wandbrun before merging tho

Jackmin801:
image
image

Jackmin801:
hrmm thats not the commit tho

Jackmin801:
image

PR #382, opened 2025-06-10
DRAFT: verifiers prime-rl integration

willccbb:
Main changes so far:
- initial integration of `verifiers` environments to prime-rl
- "async" OAI-compatible wrapper around LLM inference (waits for requests, then submits single batch)
- removed length penalties
- "reverse text" task runs, but possibly some bugs to fix -- reward curves look weird

TODOs:
- decide on env loading config 
- default env class for
- ensure TOPLOC works (batch ordering may change)
- tests vs. main for perf + reward curves


pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "reasoning-gym @ git+https://github.com/open-thought/reasoning-gym.git",
     "tomli>=2.2.1",
     "pydantic[email]>=2.11.5",
+    "verifiers[all]",
```
samsja, line 39: screenshot-20250610-142716
do we really need all ? seems a bit many depth

samsja (commented):
I like how it looks, probably will need more testing / iterating before merging:

Couple of comments already:

* lets use list,dict as type hint instead of List, Dict
* Wondering how to best integrate the fake async vllm engine with toploc and pipeline p (@mikasenghaas @Jackmin801 )
* how does reloading the weight work in this setup ?

justusmattern27:
We should probably move to a setting where users define environments outside of the infer.py file instead of hardcoding them in there - is this what is meant by "decide on env loading config"?

src/zeroband/infer.py
```diff
@@ -286,12 +422,39 @@ def inference(config: Config):
...
+        for req in request_outputs:
+            req_reward = SimpleNamespace()
+            req_reward.request_id = getattr(req, "request_id", str(reward_idx))
+            req_reward.task_type = "default"  # Could extract from env results if available
+            req_reward.rewards = []
+            for i, output in enumerate(req.outputs):
+                reward_obj = SimpleNamespace()
+                reward_obj.completion_id = i
+                reward_obj.reward = rewards[reward_idx] if reward_idx < len(rewards) else 0.0
+                reward_obj.advantage = 0.0  # Will be computed later in pipeline
```
justusmattern27, line 452: is the advantage computation already implemented?

justusmattern27: will be on the trainer side i assume

samsja: hmm defo should be on inference side

PR #384, opened 2025-06-10
high entropy minority token training

justusmattern27:


src/zeroband/training/config.py
```diff
@@ -54,25 +54,29 @@ def check_path_and_interval(self):
...
 
 
-class RatioConfig(BaseConfig):
+class RatioConfig(BaseGRPOVariantConfig):
     type: Literal["ratio"] = "ratio"
     clip_ratio: float = 8.0
 
 
-GRPOVariantsConfig: TypeAlias = ClippingConfig | KlCovConfig | RatioConfig
+GRPOVariantsConfig: TypeAlias = Annotated[Union[ClippingConfig, KlCovConfig, RatioConfig], Field(discriminator="type")]
```
samsja, line 79: hmm why do we need this ?

justusmattern27: we're getting a validation error if we use the old definition but use the highest_entropy_ratio_loss from BaseGRPOVariantConfig

samsja: I see makes sense

PR #385, opened 2025-06-11
Pre-download model weights

mikasenghaas:
We download model weights from the main process to not rely on vLLM's internal model weights download, which in production has hung a couple of times when downloading large models while using tensor paralelism. Also makes the HF environment variables `HF_HUB_CACHE` and `HF_HUB_DISABLE_PROGRESS_BARS` more visible by adding them to the inference envs, and disables progress bars by default.

*At some point we should move this before splitting into the DP subprocesses, but that should be part of a general DP refactor so will keep like this for now.*


samsja (approved):
lgtm :flight_arrival:

PR #390, opened 2025-06-11
fix grpo normalization

justusmattern27:


mikasenghaas:
ah a classic

mikasenghaas:
Screenshot 2025-06-11 at 12 52 50 PM

we are so back

PR #392, opened 2025-06-12
add back torch compile

samsja:


mikasenghaas:
bold moves

samsja:
> bold moves

trying my best to reduce memory usage

PR #393, opened 2025-06-12
Implement elastic reasoning

mikasenghaas:
This PR implements configuring a "thinking" and "solution" budget as defined in the paper Scalable Chain of Thoughts via Elastic Reasoning ([arXiv](https://arxiv.org/pdf/2505.05315)). The main idea is that we want to force a model to produce a solution even if the CoT is unfinished, with the hope being that an unfinished thinking trace still has value.

In the original paper the authors propose appending a `</think>` token if the thinking budget is reached. In their official code ([link](https://github.com/SalesforceAIResearch/Elastic-Reasoning/blob/8dacccd1ff60e82caa9033844cb0f07b97884259/verl/verl/workers/rollout/vllm_rollout/vllm_rollout.py#L303)) they do two separate calls to the inference engine, one with the think budget and one with the solution budget. If the last sampled token in the first stage is not `</think` they append it manually. In our implementation we use a post-hook on the sampler which replaces the token sampled when reaching the thinking budget by `</think`. This is more elegant because we do not require any other code changes to our inference script - we still do a single call to `.generate()` during an inference step.

One can configure the elastic reasoning by setting the `--sampling.max-solution-tokens` configuration which will dyncamilly compute the think budget as `sampling.max-tokens - sampling.max-solution-tokens`. We chose this API because we can set a constant default value for the solution budget and scale the thinking budget with the context length of the model

## Examples

1. No thinking budget set (default)

```bash
uv run src/zeroband/infer.py @ configs/inference/debug.toml --sampling.max-tokens 10 --log-level debug
```

This produces the following output

```txt
<|im_start|>assistant
<think>
Okay, let me try to solve this
```

2. Thinking budget is set

We dynamically compute the thinking budget as the difference of the maximum number of output tokens (this is `min(context length - num input tokens, max output tokens)`). In this example, the context length is 40k by default and we set the max number of output tokens to 10, so here we will almost always hit the thinking budget after 5 output tokens

```bash
uv run src/zeroband/infer.py @ configs/inference/debug.toml --sampling.max-tokens 10 --sampling.max-solution-tokens 5 --log-level debug
```

This produces the following output

```txt
<|im_start|>assistant
<think>
Okay,</think>

We are given a
```

In the second example, we do not set `--sampling.max-tokens`and so the thinking budget is dynamically computed as the difference between the context length (2k)  and the number of input tokens for each sequence, leading to a dynamic thinking budget.

```bash
uv run src/zeroband/infer.py @ configs/inference/debug.toml --model.max-model-len 2048 --sampling.max-tokens None  --sampling.max-solution-tokens 5 --log-level debug
```

### DIFF
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -137,7 +137,7 @@ def inference(config: InferenceConfig):
     # Setup elastic reasoning
     # Note: This hook has to be set up before the PP communication hooks to ensure that possibly modified sampler outputs are propagated to group peers
     if config.sampling.elastic_reasoning_enabled:
-        setup_elastic_reasoning(config.sampling, llm=llm)
+        setup_elastic_reasoning(config.sampling, llm)
 
     # Setup pipeline parallel communication and hook
     node = setup_comm(config.parallel.pp)
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -136,7 +136,8 @@ def inference(config: InferenceConfig):
 
     # Setup elastic reasoning
     # Note: This hook has to be set up before the PP communication hooks to ensure that possibly modified sampler outputs are propagated to group peers
-    setup_elastic_reasoning(config.elastic_reasoning, llm=llm)
+    if config.sampling.elastic_reasoning_enabled:
+        setup_elastic_reasoning(config.sampling, llm=llm)
 
     # Setup pipeline parallel communication and hook
     node = setup_comm(config.parallel.pp)
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -26,6 +26,7 @@ from zeroband.inference.pipeline import all_reduce, patch_model_load, setup_comm
 from zeroband.inference.rewards import compute_vllm_rewards
 from zeroband.inference.toploc import setup_toploc_cache
 from zeroband.inference.toploc2 import Toploc2Sampler
+from zeroband.inference.elastic_reasoning import setup_elastic_reasoning
 from zeroband.utils.monitor import setup_monitor
 from zeroband.inference.utils import (
     filter_data_by_prompt_length,
@@ -132,6 +133,10 @@ def inference(config: InferenceConfig):
     # Initialize sampling parameters
     logger.info(f"Initializing sampling parameters ({config.sampling})")
     sampling_params = SamplingParams(**config.sampling.model_dump())
+
+    # Setup elastic reasoning
+    # Note: This hook has to be set up before the PP communication hooks to ensure that possibly modified sampler outputs are propagated to group peers
+    setup_elastic_reasoning(config.elastic_reasoning, llm=llm)
 
     # Setup pipeline parallel communication and hook
     node = setup_comm(config.parallel.pp)
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -24,6 +24,7 @@ from zeroband.inference.config import Config as InferenceConfig
 from zeroband.inference.parquet import get_parquet_table
 from zeroband.inference.pipeline import all_reduce, patch_model_load, setup_comm, setup_hooks
 from zeroband.inference.rewards import compute_vllm_rewards
+from zeroband.inference.mask import get_mask_cache
 from zeroband.inference.toploc import setup_toploc_cache
 from zeroband.inference.toploc2 import Toploc2Sampler
 from zeroband.inference.elastic_reasoning import setup_elastic_reasoning
@@ -368,10 +369,17 @@ def inference(config: InferenceConfig):
         else:
             sampling_seeds = [None] * batch_samples
 
+        # Get the masked out tokens for this batch
+        output_masks = get_mask_cache().construct_masks(request_outputs)
+        num_masked_tokens = sum(len(mask) - sum(mask) for mask in output_masks)
+        if num_masked_tokens > 0:
+            logger.info(f"Masked out {num_masked_tokens} tokens")
+
         # Get parquet table
         table = get_parquet_table(
             request_outputs,
             request_rewards,
+            output_masks,
             prompts,
             proofs,
             ckpt_step,
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -312,6 +312,11 @@ def inference(config: InferenceConfig):
         total_samples += batch_samples
         logger.success(f"Generated {batch_samples} samples for {batch_problems} problems in {end_time - start_time:.2f}s")
 
+        # Give a warning about sequences that did not terminate
+        batch_unterminated_samples = sum(sum(1 for output in req.outputs if output.finish_reason != "stop") for req in request_outputs)
+        if batch_unterminated_samples > 0:
+            logger.warning(f"{batch_unterminated_samples}/{batch_samples} samples did not terminate.")
+
         # Print example
         first_prompt = tokenizer.decode(request_outputs[0].prompt_token_ids)
         first_completion = tokenizer.decode(request_outputs[0].outputs[0].token_ids)
--- a/src/zeroband/inference/config.py
+++ b/src/zeroband/inference/config.py
@@ -341,6 +341,7 @@ class RLConfig(BaseConfig):
         int,
         Field(
             default=2,
+            ge=0,
             description="Maximum number of steps that inference can be ahead of training.",
         ),
     ]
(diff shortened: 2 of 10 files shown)


src/zeroband/inference/config.py
```diff
@@ -280,6 +281,9 @@ class Config(BaseSettings):
     # The RL configuration. If None, inference will run in a non-RL setting.
     rl: Annotated[RLConfig | None, Field(default=None)]
 
+    # The elastic reasoning configuration
+    elastic_reasoning: Annotated[ElasticReasoningConfig, Field(default=ElasticReasoningConfig())]
```
samsja, line 285:
```suggestion
    elastic_reasoning: ElasticReasoningConfig,  =  ElasticReasoningConfig() 
 
```

isn't better to keep like this ?

src/zeroband/inference/elastic_reasoning.py
```diff
@@ -0,0 +1,73 @@
...
+from vllm import LLM
+from vllm.model_executor.layers.sampler import SamplerOutput, SamplingMetadata
+
+from zeroband.utils.config import BaseConfig
+from zeroband.utils.logger import get_logger
+
+
+class ElasticReasoningConfig(BaseConfig):
+    """Configures elastic reasoning from `Scalable Chain of Thoughts via Elastic Reasoning` (https://arxiv.org/abs/2505.05315)."""
+
```
samsja: lets keep config on the config.py of inference

src/zeroband/inference/elastic_reasoning.py
```diff
@@ -0,0 +1,73 @@
...
+    enable: Annotated[bool, Field(default=False, description="Whether to enable elastic reasoning.")]
+    think_budget: Annotated[
+        int,
+        Field(
+            default=1024,
```
samsja: I like the default being pass by the = rather that defualt

src/zeroband/inference/elastic_reasoning.py
```diff
@@ -0,0 +1,73 @@
...
+    solution_budget: Annotated[int, Field(default=1024, description="Number of tokens to allow for solution.")]
+    stop_think_token_id: Annotated[
+        int,
+        Field(
+            default=151668, description="The token ID of the `</think>` token which is swapped in to force the model to produce a solution."
+        ),
+    ]  # </think
+
+
+def swap_think_token(_, __, kwargs, outputs, config: ElasticReasoningConfig) -> SamplerOutput | None:
```
samsja:
```suggestion
def swap_think_token(_, __, kwargs, outputs, config: ElasticReasoningConfig) -> SamplerOutput | None:
```

what is +_ and __ doing ?

src/zeroband/inference/elastic_reasoning.py
```diff
@@ -0,0 +1,73 @@
...
+        for seq_group, seq_outputs in zip(sampling_metadata.seq_groups, sampling_output.outputs):
+            for seq_id, seq_data, seq_output in zip(seq_group.seq_data.keys(), seq_group.seq_data.values(), seq_outputs.samples):
+                num_output_tokens = len(seq_data.output_token_ids) + 1  # Increment by 1 because the current token is not yet in the output
+                if num_output_tokens == config.think_budget:
+                    get_logger("INFER").debug(
+                        f"Thinking budget reached for sequence {seq_id}. Replacing sampled token {seq_output.output_token} with {config.stop_think_token_id}"
+                    )
+                    seq_output.logprobs = {config.stop_think_token_id: seq_output.logprobs[seq_output.output_token]}
+                    seq_output.output_token = config.stop_think_token_id
+
```
samsja: we need to have way to flag that this token was generated manually. 

We need to mask it for the loss

src/zeroband/infer.py
```diff
@@ -130,6 +131,10 @@ def inference(config: InferenceConfig):
     logger.info(f"Initializing sampling parameters ({config.sampling} seed={config.seed})")
     sampling_params = SamplingParams(**config.sampling.model_dump(), seed=config.seed)
 
+    # Setup elastic reasoning
+    # Note: This hook has to be set up before the PP communication hooks to ensure that possibly modified sampler outputs are propagated to group peers
+    setup_elastic_reasoning(config.elastic_reasoning, llm=llm)
```
samsja:
```suggestion
    if config.elastic_reasoning is not None:
            setup_elastic_reasoning(config.elastic_reasoning, llm=llm)
```

src/zeroband/inference/elastic_reasoning.py
```diff
@@ -0,0 +1,73 @@
...
+    return sampling_output
+
+
+def setup_elastic_reasoning(config: ElasticReasoningConfig, llm: LLM):
+    # Skip if disabled
+    if not config.enable:
+        return
```
samsja:
```suggestion
```

src/zeroband/inference/elastic_reasoning.py
```diff
@@ -0,0 +1,73 @@
...
+from vllm.model_executor.layers.sampler import SamplerOutput, SamplingMetadata
+
+from zeroband.utils.config import BaseConfig
+from zeroband.utils.logger import get_logger
+
+
+class ElasticReasoningConfig(BaseConfig):
+    """Configures elastic reasoning from `Scalable Chain of Thoughts via Elastic Reasoning` (https://arxiv.org/abs/2505.05315)."""
+
+    enable: Annotated[bool, Field(default=False, description="Whether to enable elastic reasoning.")]
```
samsja:
```suggestion
```

src/zeroband/inference/config.py
samsja:
```suggestion
    elastic_reasoning: ElasticReasoningConfig | None = None
```

samsja (commented):
code looks good, I added some stuff with respect to config. 

lets rather use the X | None pattern instead of having enable flag

src/zeroband/inference/elastic_reasoning.py
mikasenghaas: oh, okay. i was actually step by step moving subconfigs into submodules whenver possible (see rewards and pipeline). if we decide to do this, then i would move back all the configs to be consistent. executive decision given?

src/zeroband/inference/elastic_reasoning.py
mikasenghaas: we pass default for all configs now so this would be a changed across all config values. 

the main reason i decided for this is because i think it's more readable (less line breaks) and because if we ever want to dynamically set a default param we can do so by setting `default_factory: Callable`, e.g. `id: Field(default_factory: lambda: uuid.hex)` which is closer in syntax to `default` than the `=` way

src/zeroband/inference/elastic_reasoning.py
mikasenghaas: it's capturing the nn module that is being called (`_`) and the args (`__`) which are both not used but are required for the hook to be installed correctly

src/zeroband/inference/elastic_reasoning.py
mikasenghaas: mmh, we should only need to do this if it causes any issues. my guess would be that the effect on the overall traininig dynamic to be quite minimal, but will check with some more tests

not super clear to me how to propagate that "masking" information out of the hook

src/zeroband/infer.py
mikasenghaas: yep will change to that

src/zeroband/inference/elastic_reasoning.py
mikasenghaas: moving the subconfigs for this pr into `src/zeroband/inference/config.py` and will move the rest in separate pr

src/zeroband/inference/elastic_reasoning.py
samsja: i see makes sense

src/zeroband/inference/elastic_reasoning.py
samsja: defo hard requirement here imo. Its quite clear that this would create issue during rl training because the logprob would be super wrong. Also, this is standard with things like tool call

justusmattern27:
re discussion: </think> tokens should definitely be masked imo - if we train the model on positive advantage samples with the token unmasked, we specifically encourage the model to interrupt its thinking which will degrade performance; we want to encourage the model to think for longer, and having a bunch of training data in there of short sequence lengths will most likely cause the model to interrupt thinking after short amounts of time

PR #398, opened 2025-06-14
Fix TOPLOC crash when trying to generate one token

mikasenghaas:
Quick fix for an issue with the TOPLOC cache being not initialized when we only generate a single token

```bash
uv run src/zeroband/infer.py @ configs/inference/debug.toml --sampling.max-tokens 1
```


samsja (approved):

mikasenghaas:
Screenshot 2025-06-13 at 5 36 15 PM

PR #399, opened 2025-06-14
add null reward for syn gen

samsja:


justusmattern27 (approved):
wow, lgtm 🚀 😍 🔥

configs/inference/synthetic-2/deepseek-r1-0528-qwen3-8b.toml
```diff
@@ -5,4 +5,7 @@ max_model_len = 16384
 [sampling]
 n = 8
 temperature = 0.6
-top_p = 0.95
\ No newline at end of file
+top_p = 0.95
+
+[rewards]
+compute_reward = false
```
mikasenghaas: we don't need to repeat this here. the config is shared across all model configs (config composition ftw:))

src/zeroband/infer.py
```diff
@@ -91,6 +91,11 @@ def inference(config: InferenceConfig):
     logger.info(f"Initializing dataset (name={config.data.name}, split={config.data.split})")
     start_time = time.time()
     dataset = load_dataset(config.data.name, split=config.data.split)
+
+    if not config.rewards.compute_reward:
+        logger.info("Reward computation is disabled, setting task_type to null_reward")
+        dataset = dataset.map(lambda x: {"task_type": "null_reward"})
```
mikasenghaas, line 97: this worked?

samsja: apparently yes

mikasenghaas: can you run with `--log-level debug` and check that there is a valid generation

mikasenghaas: im scared

samsja: yep

configs/inference/synthetic-2/deepseek-r1-0528-qwen3-8b.toml
samsja: lfg

mikasenghaas:
Screenshot 2025-06-13 at 6 57 19 PM

mikasenghaas:
To disable reward computation have to set `--no-rewards.compute-reward`, e.g.

```bash
uv run src/zeroband/infer.py @ configs/inference/debug.toml --log-level debug --no-rewards.compute-reward
```

mikasenghaas (approved):
it's slightly awkward that we map the whole dataset which takes adds a couple of seconds to the script startup, but probs fine for now

PR #402, opened 2025-06-14
update configs for better stability during async training

samsja:
This pr do:

* Use one optimizer step per rollout. Warn if any other value is set
* use temp 1.0 and n=16 as default sampling
* normalize per batch as true for default
* Remove obsolete configs
* use dr_grpo as default


faresobeid (approved):

src/zeroband/inference/config.py
```diff
@@ -45,7 +45,7 @@ class SamplingConfig(BaseConfig):
     temperature: Annotated[
         float,
         Field(
-            default=0.6,
+            default=1.0,
```
mikasenghaas, line 35: mmh im not sure i like changing the default in the inference defaults as this is a very rl-specific default. for anything non-rl, we should usually default to the recommended default from the model card which is typically 0.6-0.7. one option is to move this value for each config file that is associated with a rl run. or we make a new rl base config file that is always composed with an rl experiment

mikasenghaas (approved):
one comment on the temperature default, else looks great

src/zeroband/inference/config.py
samsja: hmmm but prime-rl is RL by default. It does makes sense to me to use the best temp for RL

mikasenghaas: ok sure, we just have to be careful then to set this for our syn data runs

samsja:
screenshot-20250616-091848

PR #403, opened 2025-06-14
Remove prompts exceeding model context by default

mikasenghaas:


mikasenghaas:
We close for now to avoid duplicating this change on the TOPLOC validator. Ideally, we will move to the API structure, in which case we can reopen and merge this PR without requiring a change on the TOPLOC validator.

PR #404, opened 2025-06-14
Serialize current inference step

mikasenghaas:
This PR adds the `--step-path` config (`Path | None`, defaults to None) which writes the current inference step (the one the inference worker is working on right now if the program is running/ the last inference step before termination) to the file path specified. If the path is not set, we do not write create or write to the file.

Should be set by protocol to a file in a different volume to ensure persistent group restarts.

For example, the below command will create a file in `step` in the `rollouts` directly with a *single line and a single integer* indicating the current inference step, e.g. at step 0 the file would read `0` and there is no work submissions yet. At step 1 the file would read `1` and there is a subdirectory `rollouts/step_0` with the outputs of the finished step 0, and so on..

```bash
uv run src/zeroband/infer.py @ configs/inference/debug.toml --step-path rollouts/step
```

### DIFF
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -187,6 +187,16 @@ def inference(config: InferenceConfig):
         llm = reload_model_weights(llm, path_file)
         real_step = ckpt_step
 
+    # Check if we should resume from step_path file
+    if config.step_path is not None and config.step_path.exists():
+        try:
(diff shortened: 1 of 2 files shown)


Jackmin801 (commented):
There also needs to be logic to read it right?

samsja (approved):

PR #405, opened 2025-06-14
fix: pin `vllm` version to `0.8.x`

rasdani:
without pinning version `0.9.0.1` gets installed.

in this version `vllm.model_executor.model_loader.loader` moved to `vllm.model_executor.model_loader.utils`
https://github.com/vllm-project/vllm/blob/v0.8.5/vllm/model_executor/model_loader/loader.py
https://github.com/vllm-project/vllm/blob/v0.9.0/vllm/model_executor/model_loader/utils.py

which breaks this import here
https://github.com/PrimeIntellect-ai/prime-rl/blob/cdb3e25d87bb5feb7bcca179b581fad7e9c67034/src/zeroband/inference/utils.py#L9

either pin version with this PR or bump `vllm` + and update import statement.

@samsja I updated my local `uv`. hopefully my lock files are ok now.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -37,7 +37,7 @@ wheels = [
 
 [[package]]
 name = "aiohttp"
-version = "3.12.2"
+version = "3.12.13"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "aiohappyeyeballs" },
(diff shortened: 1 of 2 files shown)


samsja:
> without pinning version `0.9.0.1` gets installed.
> 
> in this version `vllm.model_executor.model_loader.loader` moved to `vllm.model_executor.model_loader.utils` https://github.com/vllm-project/vllm/blob/v0.8.5/vllm/model_executor/model_loader/loader.py https://github.com/vllm-project/vllm/blob/v0.9.0/vllm/model_executor/model_loader/utils.py
> 
> which breaks this import here
> 
> https://github.com/PrimeIntellect-ai/prime-rl/blob/cdb3e25d87bb5feb7bcca179b581fad7e9c67034/src/zeroband/inference/utils.py#L9
> 
> either pin version with this PR or bump `vllm` + and update import statement.
> 
> @samsja I updated my local `uv`. hopefully my lock files are ok now.

hmmm I don't think that 0.9 should be instlal uv you do uv sync. The pin version in the uv.lock is 0.8.5.post1

samsja:
> > without pinning version `0.9.0.1` gets installed.
> > in this version `vllm.model_executor.model_loader.loader` moved to `vllm.model_executor.model_loader.utils` https://github.com/vllm-project/vllm/blob/v0.8.5/vllm/model_executor/model_loader/loader.py https://github.com/vllm-project/vllm/blob/v0.9.0/vllm/model_executor/model_loader/utils.py
> > which breaks this import here
> > https://github.com/PrimeIntellect-ai/prime-rl/blob/cdb3e25d87bb5feb7bcca179b581fad7e9c67034/src/zeroband/inference/utils.py#L9
> > 
> > either pin version with this PR or bump `vllm` + and update import statement.
> > @samsja I updated my local `uv`. hopefully my lock files are ok now.
> 
> hmmm I don't think that 0.9 should be instlal uv you do uv sync. The pin version in the uv.lock is 0.8.5.post1

or is the issue that vllm got updated in uv.lock when doing an update (via uv sync with a new dependencies ? )

Also I agree that it makes sense to pin the pyproject.toml as well since we only support this specific version

uv.lock
```diff
@@ -964,7 +952,7 @@ wheels = [
 
 [[package]]
 name = "huggingface-hub"
-version = "0.32.1"
```
samsja, line 967: curious why this one got change

mikasenghaas (approved):

samsja (changes requested):
lets wait I still need to do the uv.lock

samsja:
so it seems that smth is wrong with the uv lock. Here is my version https://github.com/PrimeIntellect-ai/prime-rl/pull/408 

I suggest we merge my PR instead which is base on yours so both of us will be accredited (like the one before)

samsja:
closign as PR was merged via https://github.com/PrimeIntellect-ai/prime-rl/pull/408

PR #407, opened 2025-06-15
refactor config management and add toml inheritance

samsja:
# refactor

this pr refactor the config management. Specifically it abstracts all the custom PydanticSetting we used into their own class. User just had to call utis.pydantic_config import BaseSettings, BaseConfig.

## Inheritance  API proposal

allow to define inheritance in toml

Firt file: debug.toml
```toml
toml_files = ["base.toml"]

[model]
name = "Qwen/Qwen3-0.6B"

```

the first file would tell to inherit the second

second file base.toml
```toml
smth_else = true
```

both debug.toml and base.toml would leave under the same root dir (configs).

Unde the hood this resolve to the same as doing 

```bash
uv run test.py @debug.toml @base.toml
```


src/zeroband/utils/config.py
```diff
@@ -71,20 +71,48 @@ def __str__(self) -> str:
         return f"file={file_str}, socket={socket_str}, api={api_str}, system_log_frequency={self.system_log_frequency}"
 
 
+def look_for_inheritance_in_toml(path: str) -> list[str]:
+    """
+    Look for inheritance in a toml file. Return a list of extra toml files to load.
```
mikasenghaas: definitely add here that this is recursive and add to pr description as well. at the beginning i thought we only support depth 1, but this is actually more general and nicer. let's have one example of this in action

configs/training/base.toml
mikasenghaas, line 1: nice for demonstration purposes but we shouldn't commit a base training config imo - everything in there should just be set as a default in the `config.py` file. inheritance/ composition makes a lot of sense for stuff like ablations

mikasenghaas (commented):
already super nice. a couple of questions/ thoughts:
- personallly, the `# @` syntax is a bit awkward. i would never expect a comment to have an influence on my program so i think (even with documentation) this can be very confusing to users. im guessing you did this to no fuck up the toml syntax, but can we maybe we find allowed toml syntax that allows us to specify
- it seems like the line in which you search for a toml inheritance does not have an influece on the precedence it gets. imagine file1 sets `a=1` and file 2 `a=2` and this is file 3

```toml
# @file1.toml

a = 3

# @file2.toml
```

then for this config (iiuc) we would load the files in order 1,2,3 and so the value would be `a=3`. however, from reading the file i would expect that file 2 is loaded last and has precedence. this would require a bit more changes bc we cannot just return a list of paths and let pydantic-settings handle everything but i think this particular usecase is quite confusing as well. a simpler way to cirumvent this is that we "assert" to only load "lower-priority" configs by loading only parsing potential filepaths before the first non-comment line is reached (potentially even showing a warning if we detect a file inheritance after configs were set)

configs/training/base.toml
samsja: yes yes ofc just as demo purpose

samsja:
> already super nice. a couple of questions/ thoughts:
> 
> * personallly, the `# @` syntax is a bit awkward. i would never expect a comment to have an influence on my program so i think (even with documentation) this can be very confusing to users. im guessing you did this to no fuck up the toml syntax, but can we maybe we find allowed toml syntax that allows us to specify
> * it seems like the line in which you search for a toml inheritance does not have an influece on the precedence it gets. imagine file1 sets `a=1` and file 2 `a=2` and this is file 3
> 
> ```toml
> # @file1.toml
> 
> a = 3
> 
> # @file2.toml
> ```
> 
> then for this config (iiuc) we would load the files in order 1,2,3 and so the value would be `a=3`. however, from reading the file i would expect that file 2 is loaded last and has precedence. this would require a bit more changes bc we cannot just return a list of paths and let pydantic-settings handle everything but i think this particular usecase is quite confusing as well. a simpler way to cirumvent this is that we "assert" to only load "lower-priority" configs by loading only parsing potential filepaths before the first non-comment line is reached (potentially even showing a warning if we detect a file inheritance after configs were set)

right now we only support doing it at the first line ( so you can only inherit from one file).

Yeah I also file that the # @ is akwads. 

I tried using _base_ = "helo.toml" as syntax, but unfo  this was not compatible with pydantic. 


I guess we could have some other pydantic field just for this inheritance. What could we use tho ?

mikasenghaas:
i do think that we should support inheriting multiple files so whatever solution we find should be compatible with that

mikasenghaas:
could put an additional field in `TrainingConfig` and `InferenceConfig` called `toml_files: list[Path] | None`. we would never use this field at runtime but still rely on parsing it out so that the toml file plugin can actually load the configs. added benefit of this solution is that we see which files were inherited when logging to e.g. wandb (essentially knowing all the sources that went into constructing the config dict)

samsja:
> could put an additional field in `TrainingConfig` and `InferenceConfig` called `toml_files: list[Path] | None`. we would never use this field at runtime but still rely on parsing it out so that the toml file plugin can actually load the configs. added benefit of this solution is that we see which files were inherited when logging to e.g. wandb (essentially knowing all the sources that went into constructing the config dict)

I guess this is the easiest yeah

src/zeroband/training/config.py
```diff
@@ -132,6 +132,14 @@ class DataConfig(BaseConfig):
 class Config(BaseSettings):
     """Configures training"""
 
+    toml_files: Annotated[
+        list[str] | None,
+        Field(
+            default=None,
+            description="List of extra TOML files to load. If provided, will override all other config files. This field will be read during cli parsing before this config is instantiated.",
```
mikasenghaas: should add that it's only read from within files right? setting `--toml-files` from cli has no effect, right?

src/zeroband/utils/config.py
```diff
@@ -71,20 +72,52 @@ def __str__(self) -> str:
...
+    return_data = [str(path)]
+    try:
+        with open(path, "rb") as f:
+            data = tomli.load(f)
+
+        if "toml_files" in data:
+            maybe_new_files = [path.parent / file for file in data["toml_files"]]
+
+            files = [file for file in maybe_new_files if str(file).endswith(".toml") and file.exists()]
+            # todo which should probably look for infinite inheritance loops here
```
mikasenghaas: yea, good point. let's do this

src/zeroband/training/config.py
samsja: very good point

configs/training/debug.toml
```diff
@@ -1,3 +1,5 @@
+toml_files = ["base.toml"]
```
samsja: also will be deleted before merging

src/zeroband/utils/pydantic_config.py
```diff
@@ -0,0 +1,129 @@
...
+    model_config = ConfigDict(extra="forbid")
+
+
+class BaseSettings(PydanticBaseSettings):
+    """
+    Base settings class for all configs.
+    """
+
+    # These are two somewhat hacky workarounds inspired by https://github.com/pydantic/pydantic-settings/issues/259 to ensure backwards compatibility with our old CLI system `pydantic_config`
+    _GLOBAL_TOML_FILES: ClassVar[list[str]] = []
```
mikasenghaas: rename this to `_TOML_FILES` bc it's not global anymore?

src/zeroband/utils/pydantic_config.py
```diff
@@ -0,0 +1,129 @@
...
+    toml_files: Annotated[
+        list[str] | None,
+        Field(
+            default=None,
+            description="List of extra TOML files to load. If provided, will override all other config files. Note: This field is only read from within configuration files - setting --toml-files from CLI has no effect.",
+        ),
+    ]
+
+    @classmethod
+    def set_global_toml_files(cls, toml_files: list[str]) -> None:
```
mikasenghaas: and then this to `set_toml_files`

mikasenghaas (commented):
nice! i like the new base settings class and how we set toml files as class attribute.

not sure im the biggest fan of having 3 diff config files all the same level and `config.py` only acting as a "import" hop. either we have all the code inside `config.py` as before or we make `config` a submodule with 3 files like this 

- `config/__init__.py` (prev `config.py`)
- `config/shared.py` (prev `shared_config.py`)
- `config/base.py` (prev `pydantic_config.py`)

ig third option would be to move pydantic_config into its own pypi project and in which case all code left is `shared_config.py` which we couuld rename to `config.py`

samsja:
> nice! i like the new base settings class and how we set toml files as class attribute.
> 
> not sure im the biggest fan of having 3 diff config files all the same level and `config.py` only acting as a "import" hop. either we have all the code inside `config.py` as before or we make `config` a submodule with 3 files like this
> 
> * `config/__init__.py` (prev `config.py`)
> * `config/shared.py` (prev `shared_config.py`)
> * `config/base.py` (prev `pydantic_config.py`)
> 
> ig third option would be to move pydantic_config into its own pypi project and in which case all code left is `shared_config.py` which we couuld rename to `config.py`

I remove shared.

since pydantic_config is going to be its own package soon I would keep it like this. I don't want it to make it its own package yet tho as we might still add stuff in the next days

src/zeroband/utils/pydantic_config.py
```diff
@@ -0,0 +1,164 @@
...
+
+def check_path_and_handle_inheritance(path: str, seen_files: list[str]):
+    """
+    Recursively look for inheritance in a toml file. Return a list of all toml files to load.
+
+    Example:
+        If config.toml has `toml_files = ["base.toml"]` and base.toml has
+        `toml_files = ["common.toml"]`, this returns ["config.toml", "base.toml", "common.toml"]
+    """
+    print(f"Checking {path}")
```
mikasenghaas: delete this?

mikasenghaas (approved):

PR #410, opened 2025-06-16
sami/fix rl config

samsja:


src/zeroband/training/config.py
```diff
@@ -96,7 +96,7 @@ class RatioConfig(BaseGRPOVariantConfig):
     clip_ratio: Annotated[float, Field(default=8.0)]
 
 
-GRPOVariantsConfig: TypeAlias = Annotated[Union[ClippingConfig, KlCovConfig, RatioConfig], Field(discriminator="type")]
```
mikasenghaas, line 99: this is unrelated to the pr purpose or am i missing smth?

mikasenghaas (approved):

src/zeroband/training/config.py
samsja: for some reason having this break the PR. But I think that we don't need this part anymore so just removing it fix it

PR #411, opened 2025-06-16
Refactor logger

mikasenghaas:
This PR introduces a more beautiful and informative logger based on `loguru`.

## Features

1. **Dynamic logging format**: By default, we log minimal information (time, log level and message). However, when specificying the log level as debug, we automatically switch to a debug logging format which also shows information about the process ID, thread ID and log location (file and line number)
2. **Parallelism-aware**: By default, we only log from the main local rank (this is DP during inference) and the rank from world info during training. One can optionally enable logging from all ranks by setting (`--log.all-ranks`). The parallelism rank (DP/ PP for inference, and Rank for training) are logged even if not using the debug format.
3. **Timezone-aware**: By default, we log the time in the current timezone (can be set via `TZ` env variable). By specifying (`--log.utc`) we always log time in UTC timezone.

### Regular Log Format (All Levels)

Screenshot 2025-06-16 at 8 48 03 PM
Screenshot 2025-06-16 at 8 47 55 PM

### Debug Log Format (All Levels)

Screenshot 2025-06-16 at 8 48 28 PM
Screenshot 2025-06-16 at 8 48 37 PM

### Inference

Screenshot 2025-06-16 at 8 51 47 PM

### Training

Screenshot 2025-06-16 at 8 54 11 PM

## Implementation

To enable custom training/ inference loggers (e.g. inference should handle DP/ PP while training should handle parallelism based on world info) but still allow shared modules, like `monitor.py` to use the logger, we extend the singleton schema in the following way: The global logger instance lives in a shared module `src/zeroband/utils/config.py` with light-weight setters/getters. Each submodule (`training` and `inference`) defines its logger by exposing a `setup_logger` function which is called at the start of the entrypoint of that submodule (e.g. `train.py` and `infer.py`

TODOS:
- [x] Use non-bold message text
- [x] Use current time zone
- [x] Define logger getter in shared module
- [x] Write and fix tests
- [x] Enable/ disable logging from all parallel ranks
- [x] Implement training logger
- [x] Update log message in inference


mikasenghaas:
Screenshot 2025-06-16 at 9 31 24 PM

samsja:
screenshot-20250617-094852


I would like two things:


* reduce the space between the time abd the logs. I get that its for aligning but most of the time SUCESS is the biggest word so we just wast terminal space imo.

* remove UTC from the time (I need more space when doing tmux lol)

* potentiall adding a | after the time and after the INFO so that there is a nice border / columns

mikasenghaas:
> screenshot-20250617-094852
> 
> I would like two things:
> 
> * reduce the space between the time abd the logs. I get that its for aligning but most of the time SUCESS is the biggest word so we just wast terminal space imo.
> * remove UTC from the time (I need more space when doing tmux lol)
> * potentiall adding a | after the time and after the INFO so that there is a nice border / columns

re:
1. so pad with 7 instead of 8 characters? (assumes that longest word is SUCCESS, althouh CRITICAL has 8 chars). i would reallly like to avoid taking the padding away completetely as this looks much worse imo
2. yep can do. do you want to disable this both for local time and utc (might be nice to keep the utc, so that then basically if you don't see a timezone you know it's taking your local one)
3. i will check if i like to look, but i was actually liking the little clutter look

samsja:
> > screenshot-20250617-094852
> > I would like two things:
> > 
> > * reduce the space between the time abd the logs. I get that its for aligning but most of the time SUCESS is the biggest word so we just wast terminal space imo.
> > * remove UTC from the time (I need more space when doing tmux lol)
> > * potentiall adding a | after the time and after the INFO so that there is a nice border / columns
> 
> re:
> 
> 1. so pad with 7 instead of 8 characters? (assumes that longest word is SUCCESS, althouh CRITICAL has 8 chars). i would reallly like to avoid taking the padding away completetely as this looks much worse imo
> 2. yep can do. do you want to disable this both for local time and utc (might be nice to keep the utc, so that then basically if you don't see a timezone you know it's taking your local one)
> 3. i will check if i like to look, but i was actually liking the little clutter look

1. 7 is fine I guess. But even when SUCCESS is printed there is large gap rn (seems more that one caracter)
2. lets keep for utc then, but lets make local the default then
3. feel free to add screenshot of both but I do like column especially if we have padding

mikasenghaas:
Screenshot 2025-06-17 at 11 35 09 AM

PR #418, opened 2025-06-17
[feat] enable toploc2 by default

Jackmin801:


Jackmin801:
The plots look ok so should be fine. Should we run more tests to make sure this sampler is ok?
image

Jackmin801:
hrmmm what actually happens for the path when we arent setting the sampling_params.seed in the node address path? Is the same sampling_params.seed from the config being used for every generation?

samsja:
> The plots look ok so should be fine. Should we run more tests to make sure this sampler is ok? image

Probably math sense to run the math_only_1b run, should take 24 hours to converge

samsja (approved):
lgtm

mikasenghaas:
> hrmmm what actually happens for the path when we arent setting the sampling_params.seed in the node address path? Is the same sampling_params.seed from the config being used for every generation?

I am not sure I get your question? We only have one seed parameter in the top-level config that, if set, we also pass into the sampling params to get fully reproducible outputs.

Jackmin801:
Should be ok and doesnt seemto significantly impact perf.
image

PR #422, opened 2025-06-18
Implement W&B multi-logging

mikasenghaas:
This PR enables logging to W&B via any number of submodules (e.g. for a RL training run from the training, inference and evals workers) by implementing the `WandbMonitor` class. By default, logging to W&B (like all other monitors) is disabled. 

### Single Run

To enable logging to W&B on a single run, simply pass which project to log to `--monitor.wandb.project`:

```bash
uv run src/zeroband/infer.py @ configs/inference/debug.toml --monitor.wandb.project example-project
```

For convenience, one can set the run name via `--monitor.wandb.name`.

### Multi-Run

To enable logging from multiple processes onto a single W&B dashboard, we use the grouping feature. For example, to log an RL run into `mika-testing` project as group `test` run the following command.

```bash
# Inference
CUDA_VISIBLE_DEVICES=0 uv run src/zeroband/infer.py @ configs/inference/simple_reverse_two_gpus.toml --monitor.wandb.project mika-testing --monitor.wandb.group test --monitor.wandb.name inference
```

```bash
# Training
CUDA_VISIBLE_DEVICES=1 uv run torchrun src/zeroband/train.py @ configs/training/simple_reverse_two_gpu.toml --monitor.wandb.project mika-testing --monitor.wandb.group test --monitor.wandb.name training
```

### DIFF
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -36,11 +36,11 @@ from zeroband.inference.utils import (
     format_prompts,
 )
 from zeroband.training.mp import EnvWrapper
-from zeroband.utils.utils import ensure_process_group_cleanup
+from zeroband.utils.utils import clean_exit
 from zeroband.inference.logger import setup_logger
 
 
-@ensure_process_group_cleanup
+@clean_exit
 def inference(config: InferenceConfig):
     # Initialize the logger
     dp_rank = int(os.environ.get("DP_RANK", 0))
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -59,7 +59,7 @@ def inference(config: InferenceConfig):
     logger.success(f"Downloaded model weights in {time.time() - start_time:.2f}s")
 
     # Initialize metrics
-    monitor = setup_monitor(config.monitor, config.task_id)
+    monitor = setup_monitor(config.monitor, config.task_id, config)
 
     # Patch vLLM's model loading to load model shard
     patch_model_load(config=config.parallel.pp)
@@ -314,6 +314,7 @@ def inference(config: InferenceConfig):
             "progress/batch_problems": batch_problems,
             "progress/batch_samples": batch_samples,
             "progress/batch_tokens": batch_tokens,
+            "step": real_step,
         }
         monitor.log(progress_metrics)
 
@@ -330,6 +331,7 @@ def inference(config: InferenceConfig):
             "performance/batch_tokens_per_second": batch_tokens_per_second,
             "performance/batch_samples_per_minute": batch_samples_per_minute,
             "performance/batch_avg_seq_length": batch_avg_seq_length,
+            "step": real_step,
         }
         monitor.log(perf_metrics)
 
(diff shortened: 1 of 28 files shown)


src/zeroband/training/utils.py
```diff
@@ -257,56 +257,6 @@ def reshard_module(model: torch.nn.Module):
             module.reshard()
 
 
-def log_to_wandb(
-    metrics_dict: dict[str, Any],
-    tokenizer: PreTrainedTokenizer | None = None,
-    batch: dict[str, torch.Tensor] | None = None,
-    step: int | None = None,
-) -> None:
-    """Log our metrics to wandb, grouped by type."""
```
samsja, line 266: where did this moved ?

mikasenghaas: this was extremely bloated. it's much cleaner to already structure (using the `train/`, `/lengths`, etc. prefixes) our logs when we aggregate in the metrics averager and keep the simple abstraction that the wandb monitor logs any metrics dict as is to wandb. it has exactly the same behavior as before (metrics are logged with the same key) but doesn't require additional code

mikasenghaas:
(This PR also changes the steps per rollout parameter to our new default vaue of 1 for the CI run as introduced in #402 - this leads to the training only running for 50 steps, not 100)

Screenshot 2025-06-17 at 10 41 45 PM

src/zeroband/infer.py
```diff
@@ -59,7 +59,7 @@ def inference(config: InferenceConfig):
     logger.success(f"Downloaded model weights in {time.time() - start_time:.2f}s")
 
     # Initialize metrics
-    monitor = setup_monitor(config.monitor, config.task_id)
+    monitor = setup_monitor(config.monitor, config.task_id, config)
```
Jackmin801, line 62: could we just pass config?

src/zeroband/train.py
```diff
@@ -150,11 +148,8 @@ def train(config: TrainingConfig):
     total_samples = config.start_total_samples if config.start_total_samples is not None else 0
     training_progress = TrainingProgress(total_tokens=0, step=config.start_step, total_samples=total_samples)
 
-    if world_info.rank == 0 and config.wandb:
-        wandb.init(project=config.project, config=config.model_dump(), dir="wandb_logs", name=config.wandb_run_name)
-
     # Setup the monitor
-    monitor = setup_monitor(config.monitor)
+    monitor = setup_monitor(config.monitor, run_config=config)
```
Jackmin801, line 152: here too

Jackmin801 (approved):
nice! lfgtm!

src/zeroband/infer.py
mikasenghaas: oh yea true, ig my idea was that a submodule should take its subconfig as first unnamed arg but it is a bit silly on second look you're right

mikasenghaas: oh, acc im not sure the typing works out here. if the multimonitor takes the entire config then i would have to pass config as type infer or train config in order for that config to have the attributes that i need like `config.monitor.file`. monitor is a shared module tho so i would not like to make assumptions about who consumes it ideally. whoever does, must own a the correct config for it. and if they want, they can also pass the general config to log to wandb, but this can be any pydantic model

does that make sense?

src/zeroband/training/utils.py
samsja: fair

samsja (approved):
lgtm

mikasenghaas:
Screenshot 2025-06-18 at 5 01 39 PM

PR #424, opened 2025-06-18
Bump vLLM to 0.9.1

mikasenghaas:
This PR bumps vLLM to 0.9.1 to enable merging #365 which requires the newest vLLM version. The bump required the following two changes to `prime-rl`:
1. Changing the import from `from vllm.model_executor.model_loader.loader import _process_weights_after_loading` to `from vllm.model_executor.model_loader.utils import process_weights_after_loading`
2. Removing the `max_seq_len_to_capture` argument from the engine arguments (we just use the default from now on which I wanted to do for a longer time anyway)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ dependencies = [
     "zstandard",
     "pyarrow",
     "wandb",
-    "vllm~=0.8.5",
+    "vllm~=0.9.1",
     "jaxtyping",
     "beartype",
     "toploc>=0.1.6",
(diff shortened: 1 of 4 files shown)


mikasenghaas:
CI run

Screenshot 2025-06-18 at 9 39 01 PM

samsja:
> CI run
> 
> Screenshot 2025-06-18 at 9 39 01 PM

looks good, I would still double check with toploc before merging cc @Jackmin801

mikasenghaas:
This is from the debug math run (`simple_math` config files), should all be good.

Screenshot 2025-06-19 at 2 34 05 PM

PR #428, opened 2025-06-18
save ckpt every n steps

justusmattern27:


src/zeroband/training/config.py
```diff
@@ -55,6 +55,7 @@ class CkptConfig(BaseConfig):
     path: Annotated[str | None, Field(default=None)]
     interval: Annotated[int | None, Field(default=None)]
     resume: Annotated[str | None, Field(default=None)]
+    save_every: Annotated[int | None, Field(default=None)]
```
samsja: lets rather call it rollout_interval to stick with "interval" as above

mikasenghaas: mmh, when we do refactor the training config we should have a better way to differentiate the full checkpoint from the weight checkponint. super not clear rn what is saved when but let's do at another time

mikasenghaas (approved):

src/zeroband/training/config.py
samsja: agree but also lets do it later

samsja (approved):
lgtm

src/zeroband/train.py
```diff
@@ -491,7 +491,9 @@ def train(config: TrainingConfig):
 
                 if len(previous_ckpt_rollout) > config.async_level:
                     path_to_delete = previous_ckpt_rollout.pop(0)
-                    if path_to_delete.exists():
+                    ckpt_step = int(str(path_to_delete).split("_")[-1])
+
+                    if path_to_delete.exists() and ckpt_step % config.ckpt.interval_rollout != 0:
```
samsja:
```suggestion
                    if path_to_delete.exists() and (config.ckpt is not None and ckpt_step % config.ckpt.interval_rollout != 0):
```

samsja (approved):
lgtm

PR #430, opened 2025-06-19
Fix work submissions

mikasenghaas:
This PR reverts me being greedy and trying to log the input and output FLOPs to W&B, not knowing that protocol expects the work submission logs to come in a single socket call. Oh well..


Jackmin801 (approved):
makes sense! lgtm

PR #432, opened 2025-06-19
Do one generation per prompt for synthetic data gen

mikasenghaas:
This PR ensures that we will never run into the issue where `sampling.n` > `max_batch_size` which is auto-computed. Previously, some model-topology groups could not fit a large enough batch size to sample 8 sequnces at the same time. This will also lead to better throughput because we will never truncate the true batch size in the case where `max_batch_size` is not a multiple of `sampling.n`


justusmattern27 (approved):

mikasenghaas:
Screenshot 2025-06-19 at 2 30 17 PM

PR #435, opened 2025-06-20
Sami/refactor/perf logging

samsja:


src/zeroband/training/data.py
```diff
@@ -61,7 +61,7 @@ def __iter__(self) -> Generator[DatasetOutput, Any, None]:
             if self.generate_logprobs and len_ > 1:
                 # Generate random negative values for log probabilities
                 # Exclude first token (BOS) to match expected format
-                logprobs = -torch.abs(torch.randn(len_ - 1))  # Negative values for log probs
+                logprobs = -torch.abs(torch.randn(len_))  # Negative values for log probs
```
mikasenghaas: what is this?

mikasenghaas (approved):
lgtm!

src/zeroband/train.py
```diff
@@ -192,7 +192,7 @@ def train(config: TrainingConfig):
         batch_size=config.optim.batch_size * config.optim.step_per_rollout,
         data_config=config.data,
         step_count_init=step_count_init,
-        use_vllm_logprobs=config.recompute_logprobs,
+        use_vllm_logprobs=True,
```
Jackmin801: eh so always use vllm logprobs now?

samsja: this is just for dataset, so we always load the logprob for logprob, but we then later decide if we override them or not

samsja: I should probably just remove this parameters and always return the logprob

src/zeroband/training/data.py
samsja: for some reason this value was never tested and was wrong

mikasenghaas (approved):

PR #436, opened 2025-06-20
Implement continuous evals

mikasenghaas:
This PR implements offline and online, with two entrypoints. We support both running evals from a standalone script (`src/zeroband/eval.py`) and as part of the inference script to evaluate the base model.

The evals share a lot of code and configs with inference, e.g. the `ModelConfig`, `SamplingConfig`, `ParallelConfig` and `MultiMonitorConfig` are all reused between inference and eval. This allows us to more or less for free set up sampling parameters, parallelism, and log everything to a joint W&B dashboard. We support all benchmarks for which we have converted the dataset to `genesys` format and uploaded on HF in the `Benchmarks` collections, and have implemented the corresponding reward function in the `genesys` submodule in `prime-rl`. Right now we support the following
- Benchmarks: MATH-500 (`math500`), AIME-24 (`aime24`), AIME-25 (`aime25`) which all point to datasets in our HF org in the [benchmark collection](https://huggingface.co/collections/PrimeIntellect/benchmarks-685b37ed5828796ca3412b80)
- Metrics: We compute different scores. For any benchmark, we compute the `score` as the average reward obtained over all samples (num problems * sampling.n). For benchmarks with binary rewars, we also compute `pass@k`, i.e.  pass@1, pass@2, pass@4, ..., pass@{sampling.n} by partioning the per-problem rewards into k-sized chunks, then for each rewarding with 1 if any got a 1 reward, and then take the average over the sublists

### Eval Script

1. **Debug Mode** Get a first feel for the script by running with debug config. Will run a small model (`Qwen/Qwen3-0.6B`) on single benchmark (`MATH-500`) and only sample 16 tokens per prompt (will get 0 score).

```bash
uv run src/zeroband/eval.py @ configs/eval/debug.toml
```

2. **Eval parameters** Evaluation specific configurations are stored in `EvalConfig`. The most important parameter is the list of benchmarks `--eval.benchmarks` which allows run multiple benchmarks from the supported registry. The `--eval.online` subconfiguration allows to run evaluation on model checkpoints generated during a training run by specifying the checkpoint path and interval to evaluate checkponits for.

```bash
uv run src/zeroband/eval.py --eval.benchmarks math500,aime24 --eval.online.ckpt_path checkpoints --eval.online.interval 100
```
*NB: By default, we search `checkpoints` at interval `100`, i.e. `checkpoints/step_100`, `checkpoints/step_200`, .... If those defaults are fine it is enough to specify `--eval.online` to eval checkpoints.*

3. **Model/ sampling/ parallel settings** The evaluation config reuses the same `ModelConfig`, `SamplingConfig` and `ParallelConfig` syntax.

```bash
uv run src/zeroband/eval.py --model.name deepseek-ai/DeepSeek-R1-0528-Qwen3-1.5B --model.max-model-len 8192 --sampling.n 16 --parallel.tp 4
```

4. **Wandb settings**: You can log the benchmark results to W&B in the same way as for training and inference by using the `monitor.wandb` subconfig. It will always log the benchmark results with prefix `eval/{benchmark}`, e.g. `eval/math-500/score` or `eval/math-500/pass@k`, as well as timings under `eval/math-500/benchmark-time`

```bash
uv run src/zeroband/eval.py @ configs/eval/debug.toml --monitor.wandb.project mika-testing
```

### Inference Script

The `EvalConfig` is also part of the `infer.py` script and can be configured in the exact same way as the standalone script. By default no evals are run, to activate the default evaluation (evaluate base model and every 100 steps by searching `checkpoints`)

1. **RL w/o evals**: By default, we don't enable running evals. For example, this is our CI RL run without any evals

```bash
# Inference
export CUDA_VISIBLE_DEVICES=0
uv run src/zeroband/infer.py @ configs/inference/simple_reverse_two_gpus.toml --monitor.wandb.offline
```

```bash
# Training
export CUDA_VISIBLE_DEVICES=1
ulimit -n 65336
uv run torchrun src/zeroband/train.py @ configs/training/simple_reverse_two_gpu.toml --monitor.wandb.offline
``` 

2. **RL w/ evals during inference**: We can run evaluations directly from the inference worker. As of now, it will only run evals from DP rank 0, which might block the other DP workers if the batch size is not set up in a way to oversample

```bash
# Inference
export CUDA_VISIBLE_DEVICES=0
uv run src/zeroband/infer.py @ configs/inference/simple_reverse_two_gpus.toml --eval.online --eval.online.interval 10 --monitor.wandb.offline  --model.max-model-len 4096 --sampling.max-tokens 256 
```

```bash
# Training
export CUDA_VISIBLE_DEVICES=1
ulimit -n 65336
uv run torchrun src/zeroband/train.py @ configs/training/simple_reverse_two_gpu.toml --ckpt.interval_rollout 10 --monitor.wandb.offline
```

This is an example of how it would look like on [W&B](https://wandb.ai/primeintellect/mika-testing/workspace?nw=nwusermikasenghaas_)

3. **RL w/ evals from standalone process**: One can also set up online evals as part of a RL run from a standalone process. This option is the most flexible because it allows to set up eval-specific parameters, e.g. using different `sampling.n` for evals than during training

```bash
# Inference
export CUDA_VISIBLE_DEVICES=0
uv run src/zeroband/infer.py @ configs/inference/simple_reverse_two_gpus.toml --monitor.wandb.offline
```

```bash
# Training
export CUDA_VISIBLE_DEVICES=1
ulimit -n 65336
uv run torchrun src/zeroband/train.py @ configs/training/simple_reverse_two_gpu.toml --ckpt.interval_rollout 10 --monitor.wandb.offline
```

```bash
# Evals
export CUDA_VISIBLE_DEVICES=2
uv run src/zeroband/eval.py --model.name willcb/Qwen2.5-0.5B-Reverse-SFT --eval.online --eval.online.interval 10 --monitor.wandb.offline --sampling.max-tokens 256 --sampling.n 1
```

## Benchmarks

Evaluating `deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B` on MATH-500, AIME-24 and AIME-25 at 8k context

```bash
uv run src/zeroband/eval.py @ configs/eval/model/r1.toml --model.name deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B --model.max-model-len 8192 --eval.benchmarks math500,aime24,aime25 --sampling.n 8 --parallel.tp 4 --use-tqdm
```

| Benchmark | Time (s) | Score | Pass@1 | Pass@2 | Pass@4 | Pass@8 |
|-----------|----------|-------|--------|--------|--------|--------|
| MATH-500  | 3226.71  | 0.75  | 0.75   | 0.83   | 0.88   | 0.90   |
| AIME-24   | 440.41   | 0.19  | 0.19   | 0.24   | 0.30   | 0.40   |
| AIME-25   | 433.51   | 0.23  | 0.23   | 0.28   | 0.32   | 0.37   |

## TODO

- [x] Support running multiple evals (e.g. AIME24, MATH-500)
- [x] Implement estimating pass@1, ..., pass@k
- [x] Follow up: Only enable for binary rewards
- [x] Support a fixed set of benchmarks via `data.name : Literal` with a registry for benchmark datasets in genesys format on PrimeIntellect HF org 
- [x] Implement continuous evals
- [x] Share logic so that eval can run from inference worker
- [x] Check prefixes when logging to wandb
- [ ] Implement DP parallelism for inference


mikasenghaas:
Screenshot 2025-06-24 at 5 33 44 PM

justusmattern27:
@mikasenghaas is the recommended way to run online evals to run the eval script, and thus (e.g.) use gpus 0,1 for training, gpus 2-6 for inference and gpu 7 for evals?

justusmattern27 (approved):
lgtm

src/zeroband/infer.py
```diff
@@ -205,64 +197,71 @@ def inference(config: InferenceConfig):
...
+            and dp_rank == 0
+            and config.eval
+            and config.eval.online
+            and ckpt_step % config.eval.online.interval == 0
+            and ckpt_step > last_eval_step
+        ):
+            last_eval_step = ckpt_step
+            logger.info(f"Running evals for checkpoint step {ckpt_step}")
+            for benchmark in config.eval.benchmarks:
+                run_benchmark(llm, benchmark, config.model, config.sampling, ckpt_step, seed=config.seed, use_tqdm=config.use_tqdm)
```
samsja, line 241: its a big if statement lol

mikasenghaas:
> @mikasenghaas is the recommended way to run online evals to run the eval script, and thus (e.g.) use gpus 0,1 for training, gpus 2-6 for inference and gpu 7 for evals?

yes, rn this is probably most efficient, epecially when the inference is doing DP and evals are taking quite long, because the evals in the inference script are only run by dp 0 which might block the entire training if the the batch size is not set to oversample. so yea, either you reserve one/ two gpus for eval if you need online evals. or you only split with train/ infer during training, keep checkpoint and then run evals on all gpus afterwards. you can still log everything to the same wandb group in the latter case so it's really mostly about seeing/ not seeing the evals while the  training is running. 

with the refactor we will be able to get better parallelized online evals almost for free, so no real point in overengineering this feature at this point

src/zeroband/infer.py
mikasenghaas: indeed hahah

samsja (commented):
lgtm

PR #440, opened 2025-06-20
Minor fixes

mikasenghaas:
Fixes two wrong logs that have been annoying me for a while. Also makes the tqdm progress bar on generation optional, disables by default, by enables for SYN-2.


mikasenghaas:
Screenshot 2025-06-20 at 1 49 24 PM

PR #443, opened 2025-06-21
Run at manual batch size for single nodes

mikasenghaas:
Run at large max batch size by default, but disable for groups during synthetic-2


Jackmin801 (approved):
lgtm!

PR #445, opened 2025-06-21
[fix] dont reset generator for premption prefill

Jackmin801:
These are the changes to the function
image

The issue is described in https://github.com/vllm-project/vllm/pull/19940

### DIFF
--- a/src/zeroband/inference/monkeypatch_sampling_metadata.py
+++ b/src/zeroband/inference/monkeypatch_sampling_metadata.py
@@ -1,0 +1,175 @@
+from typing import Optional
+
+import torch
+from vllm.model_executor.sampling_metadata import SamplingMetadataCache, SequenceGroupToSample
+from vllm.sampling_params import SamplingType
+from vllm.sequence import SequenceGroupMetadata
+
+
+def fixed_prepare_seq_groups(
+    seq_group_metadata_list: list[SequenceGroupMetadata],
+    seq_lens: list[int],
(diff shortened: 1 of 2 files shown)


src/zeroband/inference/monkeypatch_sampling_metadata.py
```diff
@@ -0,0 +1,175 @@
...
+from vllm.sequence import SequenceGroupMetadata
+
+
+def fixed_prepare_seq_groups(
+    seq_group_metadata_list: list[SequenceGroupMetadata],
+    seq_lens: list[int],
+    query_lens: list[int],
+    device: str,
+    generators: Optional[dict[str, torch.Generator]] = None,
+    cache: Optional[SamplingMetadataCache] = None,
```
samsja, line 15:
```suggestion
    generators: dict[str, torch.Generator] | None = None,
    cache: SamplingMetadataCache] | None = None,
```

samsja (approved):
pre-approving, still added a comment tho

src/zeroband/inference/monkeypatch_sampling_metadata.py
Jackmin801: hrmm I think keeping it as similar as possible to the way it is defined in vllm is probably better. this code wont live too long so should be fine

mikasenghaas (approved):
nice, can you maybe summarize in the pr description what the patch is doing, e.g. how it differs from the original function

PR #446, opened 2025-06-21
refactor: rename async var

samsja:


Jackmin801 (approved):
catyes

mikasenghaas (commented):
im not sure i understand the name change. so on infer you have `async_level` which triggers a reload of the model checkpoints if the inference is `async_level` steps ahead of the last loaded checkpoint and on the trainer we have `max_async_level` delete stale checkpoints. i dont see a reason for why these two should be named differently. if we name then one thing, i think `max_async` (or more verbose `max_async_level` is the correct name)

also, another thing that came to mind, we don't have a mechanism for actually loading the next ckpt step if it is available from trainer before hitting async level which is a bit silly.

they way i would expect async rl to work: in the best case (trainer and inference take exactly the same time) max async>2 is not hit in which case the inference is always ~on-policy (maybe off by one bc trainer has to wait for first rollouts)

samsja:
> im not sure i understand the name change. so on infer you have `async_level` which triggers a reload of the model checkpoints if the inference is `async_level` steps ahead of the last loaded checkpoint and on the trainer we have `max_async_level` delete stale checkpoints. i dont see a reason for why these two should be named differently. if we name then one thing, i think `max_async` (or more verbose `max_async_level` is the correct name)
> 
> also, another thing that came to mind, we don't have a mechanism for actually loading the next ckpt step if it is available from trainer before hitting async level which is a bit silly.
> 
> they way i would expect async rl to work: in the best case (trainer and inference take exactly the same time) max async>2 is not hit in which case the inference is always ~on-policy (maybe off by one bc trainer has to wait for first rollouts)

on the infer it should be name async_level and not max_async. because It is always exactly doing rl.async_level async steps.

On the trainer tho its the other way around. It can do up to max_async_level async, so its really is a max

samsja:
>  if we name then one thing, i think max_async (or more verbose max_async_level is the correct name)

max suggest it could be less, but this is never possible for infer. it is for train tho.

Happy to name both async_level as well tho

PR #449, opened 2025-06-22
Implement chunked generation

mikasenghaas:
This PR implements a (hacky) way to achieve higher throughput by implementing chunked generation. The idea is to start generating with a high batch size (higher than we could go conservatively without triggering cache eviction) with a reduced context length and then incrementally (in powers of 2) increase the context length until we hit the max model length. To enable this configuration, set the `--contexts` flag which defines the list of context lengths to chunk.

Note: This implementation assumes that we only generate one sample per problem, which is fine because this is our SYNTHETIC-2 config anyways.

### DIFF
--- a/src/zeroband/infer.py
+++ b/src/zeroband/infer.py
@@ -297,15 +297,13 @@ def inference(config: InferenceConfig):
             unordered_proofs: dict[int, bytes] = defaultdict(bytes)
 
             max_model_len = llm.llm_engine.model_config.max_model_len
-            max_input_tokens = max(len(token_prompt["prompt_token_ids"]) for token_prompt in unordered_token_prompts.values())
-            contexts_max_tokens = [context - max_input_tokens for context in sorted(config.contexts)]
-            assert min(contexts_max_tokens) > 0, "Contexts must be larger than the max input tokens"
             assert max(config.contexts) <= max_model_len, "Contexts must be smaller than the max model length"
-            for i, (context, max_tokens) in enumerate(zip(config.contexts, contexts_max_tokens)):
+            for i, context in enumerate(config.contexts):
+                max_input_tokens = max(len(token_prompt["prompt_token_ids"]) for token_prompt in unordered_token_prompts.values())
+                max_tokens = context - max_input_tokens
+                assert max_tokens > 0, "Context must be larger than the max input tokens"
                 sampling_params.max_tokens = max_tokens
-                logger.info(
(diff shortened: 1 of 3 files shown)


src/zeroband/infer.py
```diff
@@ -281,22 +282,114 @@ def inference(config: InferenceConfig):
...
+            for i, context in enumerate(config.contexts):
+                # Auto-compute the max batch size for the current context
+                logger.info(f"Running at context {context}")
+                local_max_batch_size = compute_max_batch_size(llm, max_model_len=context)
+                if node:
+                    max_batch_size = int(
+                        all_reduce(node, torch.tensor(local_max_batch_size), config=config.parallel.pp, op=torch.min).item()
+                    )
+                else:
+                    max_batch_size = local_max_batch_size
```
Jackmin801, line 312: this case actually isnt needed right? we would ever run with contexts without pp?

src/zeroband/infer.py
```diff
@@ -281,22 +282,114 @@ def inference(config: InferenceConfig):
...
+                logger.info(f"Auto-computed maximum batch size for context {context}: {max_batch_size}")
+
+                # Micro-batch the sequences to not trigger cache eviction
+                prompt_ids = list(unordered_token_prompts.keys())
+                micro_batch_prompt_ids = [prompt_ids[i : i + max_batch_size] for i in range(0, len(prompt_ids), max_batch_size)]
+                micro_batches = [
+                    {prompt_id: unordered_token_prompts[prompt_id] for prompt_id in micro_batch_prompt_ids}
+                    for micro_batch_prompt_ids in micro_batch_prompt_ids
+                ]
```
Jackmin801, line 321:
```suggestion
                batched_micro_batch_prompt_ids = [prompt_ids[i : i + max_batch_size] for i in range(0, len(prompt_ids), max_batch_size)]
                micro_batches = [
                    {prompt_id: unordered_token_prompts[prompt_id] for prompt_id in micro_batch_prompt_ids}
                    for micro_batch_prompt_ids in batched_micro_batch_prompt_ids
                ]
```

Jackmin801 (commented):
remember to fix the request_outputs such that the `completion` text is the unchunked one, the `output_token_ids` are the unchunked one and the `input_tokens_ids` are the unchunked one

`completion` will be used by dashboard
the `token_ids` are used by flop counting

Jackmin801:
There might be an edge case where most of the seqs cannot reach max_model_len because they will be blocked by the longest sequence which is used to calculate the conservative max_tokens.

The clipping ratio here is a bit weird but lets merge for now. Not sure what it is about.
image

PR #453, opened 2025-06-23
fix verbose logging

samsja:


mikasenghaas (approved):
bit hacky but better than now:))

PR #458, opened 2025-06-24
Misc developer experience improvements

mikasenghaas:
Streamlines the development experience:
- Adds recommended extensions into `.vscode/extensions.json` which will trigger a message within VS Code/ Cursor upon opening the project to install all recommended extensions. **Does not have any effect on non-VSC editors and can easily be ignored if you wish to**
- Adds `ipywidgets` to dev settings so that progress bars are shown in Jupyter notebooks


.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
+    "editor.formatOnSave": true,
+    "editor.codeActionsOnSave": {
+        "source.fixAll.ruff": "always",
+        "source.organizeImports.ruff": "always"
+    },
```
samsja:
```suggestion
    "editor.formatOnSave": true,
    "editor.codeActionsOnSave": {
        "source.fixAll.ruff": "always",
        "source.organizeImports.ruff": "always"
    },
```

this is way too specific imo

pyproject.toml
```diff
@@ -64,6 +64,7 @@ dev-dependencies = [
     "pytest>=7.0.0",
     "faker",
     "ipykernel>=6.29.5",
+    "ipywidgets>=8.1.7",
```
samsja: isn't dependency?

.vscode/settings.json
samsja: I also personnaly always disable format on save. It does not makes sense to me to have such default

pyproject.toml
mikasenghaas: Screenshot 2025-06-24 at 1 28 49 PM

without it any tqdm progress bar in jupyter shows an error that prompts you to install this dep

mikasenghaas:
;(

PR #461, opened 2025-06-25
Fix monitor step

mikasenghaas:
Backwards compatibitliy change that logs the train/infert/eval step via argument to `wandb.log` (will show as `Step` in wandb) and as "metric" (will show as small `step`)


samsja (approved):
lgtm

PR #467, opened 2025-06-26
Support any torch backend

mikasenghaas:
This PR changes the install instructions to use `uv pip install vllm --torch-backend auto` to automatically resolve the correct torch build given the driver. This is to make B200s work out of the box.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,7 @@ name: Release
 
 on:
   push:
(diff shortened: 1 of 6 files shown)


mikasenghaas:
Image built locally with the most recent `Dockerfile.cuda` from this PR and pushed to `mikasenghaas/prime-rl:cu128`. Will not merge this because it most likely breaks backwards compatibility.

PR #468, opened 2025-06-27
gradio script

willccbb:
Launch any model endpoint with `vllm serve ...` (at `0.0.0.0:8000` by default)

From same machine, run `uv run scripts/chat_ui.py`

Creates public link for chat UI, support for thinking tokens, configurable temp/top_p/max_tokens, setting model name + endpoint

Example (temporary): https://9b43932e3e245c5813.gradio.live

### DIFF
--- a/scripts/chat_ui.py
+++ b/scripts/chat_ui.py
@@ -1,0 +1,220 @@
+# /// script
+# requires-python = ">=3.12"
+# dependencies = [
+#     "gradio",
+#     "openai",
+# ]
+# ///
+import argparse
+import re
+
+import gradio as gr
+from openai import OpenAI
+
+
+def chat_function(message, history, endpoint, model_name, key, temperature, top_p, max_tokens):
+    """
+    Chat function that communicates with OpenAI-compatible endpoint.
+    """
+    client = OpenAI(api_key=key, base_url=f"http://{endpoint}/v1" if not endpoint.startswith("http") else endpoint)
+
+    messages = []
+
+    for msg in history:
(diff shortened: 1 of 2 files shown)


scripts/chat_ui.py
```diff
@@ -0,0 +1,220 @@
...
+
+import gradio as gr
+from openai import OpenAI
+
+
+def chat_function(message, history, endpoint, model_name, key, temperature, top_p, max_tokens):
+    """
+    Chat function that communicates with OpenAI-compatible endpoint.
+    """
+    client = OpenAI(api_key=key, base_url=f"http://{endpoint}/v1" if not endpoint.startswith("http") else endpoint)
```
samsja, line 19: does it always recreate the open ai worker when the function is called by gradio ? could makes sense to initiate the client once IMO if possible

willccbb: the server is done once in advance manually

recreating the client is pretty cheap (~0.1s), it's just specifying the endpoint wrapper, no background tasks. this is fairly common from what ive seen

benefit of this way is that you could change models/endpoints more easily mid-chat. but am happy to make it static + passed as an arg if you'd like

samsja: oh I see good point, lets keep it like this then

samsja (approved):
lgtm

mikasenghaas (approved):
incredible, i love it. will probably be moved somewhere after refactor but lets merge for now

sankalp1999:
waow

PR #469, opened 2025-06-27
Custom inference entrypoint

mikasenghaas:
Implement custom inference endpoint which mirrors the vLLM inference API but is configurable and startable like we own the full thing. Enables config-passthrough, i.e. one can set configs that are not defined in our config subset which are routed to vLLM's internal configs.


samsja (approved):
beautiful

PR #472, opened 2025-06-28
refactor: use torch save instead of safe tensor

samsja:
moving to torch.save as safetensor save is not compatible with async thread ckpt

### DIFF
--- a/src/zeroband/inference/utils.py
+++ b/src/zeroband/inference/utils.py
@@ -4,7 +4,6 @@ from typing import Any
 
 import torch
 from datasets import Dataset
-from safetensors import safe_open
 from vllm import LLM
 from vllm.model_executor.model_loader.utils import process_weights_after_loading
 from vllm.transformers_utils.tokenizer import AnyTokenizer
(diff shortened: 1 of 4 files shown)


src/zeroband/inference/utils.py
```diff
@@ -57,17 +56,17 @@ def reload_model_weights(llm: LLM, ckpt_path: Path):
...
-                # Skip empty keys
-                if not key:
-                    continue
-                yield key, f.get_tensor(key)
-
-        # Load weights
-        model.load_weights(weights_iterator())
+    state_dict = torch.load(ckpt_path, map_location="cpu")
+
+    # Create a better weight iterator that filters out empty keys and handles prefixes
```
mikasenghaas, line 61: why do we need this acc?

mikasenghaas (approved):

src/zeroband/inference/utils.py
samsja: hmmmm let me check

samsja: I tried wo but it ain't working

PR #475, opened 2025-06-28
fix: think splitting in `git_diff` reward

rasdani:
If `</think>` is not present in the completion the last fenced diff block got rewarded no matter what.

During training this led to intermediate patches in the unfinished reasoning trace being rewarded.
The fix now ensures proper splitting of the final answer.

The change drops support for models not using `</think>`, in favor of the fix.


samsja (approved):
lgtm !

PR #481, opened 2025-06-30
Cleanup trainer (+configs)

mikasenghaas:


src/zeroband/inference/config.py
```diff
@@ -40,17 +41,9 @@ def __str__(self) -> str:
         return f"tp={self.tp} dp={self.dp}"
 
 
-class ModelConfig(BaseConfig):
+class ModelConfig(BaseModelConfig):
```
samsja: this feels a bit too much to me, I would just repeat the filed, its just one string

src/zeroband/training/ac.py
```diff
@@ -0,0 +1,22 @@
...
+    layers_ckpt = 0
+    start_time = time.time()
+    for layer_id, transformer_block in model.model.layers.named_children():
+        if layer_id % ac_config.interval == 0:
+            transformer_block = checkpoint_wrapper(transformer_block, preserve_rng_state=False)
+            model.model.layers.register_module(layer_id, transformer_block)
+            layers_ckpt += 1
+    logger.info(
+        f"Applied activation checkpointing to {layers_ckpt}/{len(model.model.layers)} layers in {time.time() - start_time:.2f} seconds"
+    )
```
samsja: I would put this into a shared file with apply_fsdp

src/zeroband/training/config.py
```diff
@@ -3,67 +3,92 @@
...
+            default=False,
+            description="Whether to compile the model using `torch.compile`. Currently discouraged because it was found to destabilize training.",
+        ),
+    ]
+
+
+class FSDPConfig(BaseConfig):
+    """Configures FSDP."""
+
+    reshard: Annotated[bool, Field(default=True, description="Whether to reshard the model after each forward pass.")]
```
samsja:
```suggestion
    reshard_after_forward: Annotated[bool, Field(default=True, description="Whether to reshard the model after each forward pass.")]
```

this is confusing imo

src/zeroband/training/config.py
```diff
@@ -3,67 +3,92 @@
...
-    interval: Annotated[int | None, Field(default=None)]
-    interval_rollout: Annotated[int | None, Field(default=None)]
-    resume: Annotated[str | None, Field(default=None)]
+    buffer_size: Annotated[
+        int,
+        Field(
+            default=1,
+            description="Number of weight checkpoint steps to keep in memory. This is useful to keep some weight-only checkpoints for online evals.",
+        ),
+    ]
```
samsja: this is not clear why we need this and not the max_async_level

src/zeroband/training/config.py
```diff
@@ -93,43 +118,43 @@ class RatioConfig(BaseGRPOVariantConfig):
...
-    fake: Annotated[bool, Field(default=False)]
 
+    fake: Annotated[FakeDataLoaderConfig | None, Field(default=None)]
 
-class PathConfig(BaseConfig):
-    """Configures a path used for input/ output operations"""
 
-    path: Annotated[Path, Field(description="Path to write to.")]
+class ActivationCheckpointConfig(BaseConfig):
+    """Configures activation checkpointing."""
```
samsja: lets avoid config for just one field, this is too versbose imo.

Lets fuse reshard_after_forward and this one into a hardwareconfig or smth

src/zeroband/training/config.py
```diff
@@ -166,47 +191,67 @@ class Config(BaseSettings):
...
+    optim: Annotated[OptimizerConfig, Field(default=OptimizerConfig())]
 
     # The checkpoint configuration
-    ckpt: CkptConfig = CkptConfig()
+    ckpt: Annotated[CheckpointConfig, Field(default=CheckpointConfig(path="checkpoints", clean=True))]
 
-    # The data configuration
-    data: DataConfig = DataConfig()
+    # The weight checkpoint configuration
+    weights: Annotated[WeightCheckpointConfig, Field(default=WeightCheckpointConfig())]
```
samsja, line 213:
```suggestion
    weights_ckpt: Annotated[WeightCheckpointConfig, Field(default=WeightCheckpointConfig())]
```

src/zeroband/training/config.py
```diff
@@ -166,47 +191,67 @@ class Config(BaseSettings):
...
+    # The loss configuration
+    loss: Annotated[GRPOLossConfig, Field(default=GRPOLossConfig())]
+
+    # The activation checkpoint configuration
+    ac: Annotated[ActivationCheckpointConfig | None, Field(default=None)]
 
-    # The GRPO loss configuration
-    grpo: GRPOLossConfig = GRPOLossConfig()
+    # The FSDP config
+    fsdp: Annotated[FSDPConfig, Field(default=FSDPConfig())]
```
samsja: let use a shared hardware config here

src/zeroband/training/config.py
```diff
@@ -166,47 +191,67 @@ class Config(BaseSettings):
...
+        Field(
+            default=True,
+            description="Whether to recompute the logprobs. If True, will always recompute logprobs and overwrite those potentially found in the training batch.",
+        ),
+    ]
 
     normalize_batch_to_token_count: Annotated[bool, Field(default=True)]
 
-    recompute_logprobs: Annotated[bool, Field(default=True)]
+    seed: Annotated[int | None, Field(default=None, description="Random seed for the training.")]
```
samsja: where do we use seed in the trianing ?

src/zeroband/training/data.py
```diff
@@ -1,67 +1,62 @@
...
 
 
-class BatchOutput(TypedDict):
-    # token level
-    input_ids: Int[torch.Tensor, "micro_bs seq"]
-    advantages: Float[torch.Tensor, "micro_bs seq"]
-    loss_mask: Int[torch.Tensor, "micro_bs seq"]
+class MicroBatch(TypedDict):
+    # Token level
+    token_ids: Int[torch.Tensor, "micro_bs seq"]
```
samsja: lets stick to input_ids as its the hugging face key

src/zeroband/training/data.py
```diff
@@ -1,67 +1,62 @@
...
-    def __init__(self, max_seq_len: int, pad_token_id: int, micro_bs: int, batch_size: int):
-        self.max_seq_len = max_seq_len
-        self.pad_token_id = pad_token_id
-        self.micro_bs = micro_bs
-        self.batch_size = batch_size
+    def __init__(self, config: FakeDataLoaderConfig):
+        self.batch_size = config.batch_size
+        self.micro_batch_size = config.micro_batch_size
+        self.num_micro_batches = self.batch_size // self.micro_batch_size
+        self.seq_len = config.seq_len
```
samsja, line 29:
```suggestion
        self.config = config
```

src/zeroband/training/data.py
```diff
@@ -1,67 +1,62 @@
...
-    """
-    Simply load the data from the data path.
-    """
+    """Loads serialized data from a data path written by the orchestrator."""
 
     def __init__(self, data_path: Path, start_step: int):
         self.data_path = data_path
         self.current_step = start_step
-        self.world_info = get_world_info()
+        self.world = get_world()
```
samsja, line 51: world_info seems more of a correct name to me

src/zeroband/training/utils.py
```diff
@@ -1,128 +1,26 @@
...
-    def get_mfu(self) -> float | None:
-        tokens_per_second = self.get_tokens_per_second()
-        if tokens_per_second is None:
-            return None
-        return 100 * self.num_flop_per_token * tokens_per_second / self.gpu_peak_flops / self._world_info.world_size
+def seed_everything(seed: int):
+    np.random.seed(seed)
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
```
samsja: I don't think that we should seed for 3 reasons:

* there is no such thing as reproducible training, even with seed
* using seed can lead to worse performance because cuda backend will try to use more deterministic kernel
* there is no clear benefit in having a seed

lets embrace random

src/zeroband/training/utils.py
```diff
@@ -155,62 +53,33 @@ def __len__(self):
...
+) -> torch.Tensor:
+    logits: Float[torch.Tensor, "batch seq vocab"] = model(
+        input_ids=input_ids, position_ids=position_ids
+    ).logits.contiguous()
+
+    input_ids_shifted = input_ids[:, 1:]
+    logits_shifted = logits[:, :-1, :] / temperature
+    logprobs = selective_log_softmax(logits_shifted, input_ids_shifted)
+    del logits, logits_shifted
+    return logprobs
```
samsja: lets move this to loss.py instead

src/zeroband/training/utils.py
```diff
@@ -155,62 +53,33 @@ def __len__(self):
...
+
+
 OffloadedTensor: TypeAlias = list[tuple[torch.Tensor, int]]
 
 
-def offload_model_to_cpu(model: ModelType) -> OffloadedTensor:
+def offload_model_to_cpu(model: Model) -> OffloadedTensor:
```
samsja, line 35:
```suggestion
def offload_model_to_cpu(model: Model) -> OffloadedTensor:
```

this should be modelType as its a type

src/zeroband/training/world.py
```diff
@@ -5,7 +5,7 @@
 import zeroband.training.envs as envs
 
 
-class WorldInfo:
+class World:
```
samsja, line 8:
```suggestion
class WorldInfo:
```

not sure about just World

src/zeroband/utils/config.py
```diff
@@ -6,6 +6,32 @@
...
+    path: Annotated[Path, Field(description="Path to write to.")]
+
+    clean: Annotated[
+        bool,
+        Field(
+            default=False,
+            description="Whether to clean the path at the beginning of the run. If True, will delete the entire directory.",
+        ),
+    ]
+
```
samsja: I don't liek this PathConfig, this is too much, lets just duplicate the path and clean on most path

src/zeroband/training/config.py
mikasenghaas: we can support both with alias, the motivation for this was brevity

src/zeroband/training/config.py
mikasenghaas: larger buffer = safer. in the previous setup async level 0 was not working because the newly written weight ckpt would be immediately deleted. the buffer size essentially disentangles using async level for defining the infer step and how to delete old weight ckpts. it does use the max async level as a floor tho using model validation

src/zeroband/inference/config.py
mikasenghaas: this is to ensure consitency in the way we name things across subcomponents. orchestrator, eval and trainer all use the `model.name`. i basically see no reason why this should not be shared. the only downside ig is that you might have to hop files to check configs/ generally be aware of inheritance which is fine imo. a solution to this could also be inspecting configs with `--help` from CLI?

src/zeroband/training/ac.py
mikasenghaas: what would be the name? what ties fsdp and ac together? maybe `model.py`?

src/zeroband/training/config.py
mikasenghaas: i see that argument. the reason was that i was treating them as submodules, but maybe they should just be utils

src/zeroband/training/config.py
mikasenghaas: mmh in this case i would make it part of ckpt config, i.e. `ckpt.weights`...

src/zeroband/training/config.py
mikasenghaas: could acc be reasonably part of model config? hardware is too generic imo

src/zeroband/training/config.py
mikasenghaas: previously we hardcoded torch seed to 42. i jsut made this configurable

src/zeroband/training/data.py
mikasenghaas: ok

src/zeroband/training/data.py
mikasenghaas: i like brevity whenever possible

src/zeroband/training/utils.py
mikasenghaas: ok, should we als reomve the cuda manual seed then? this was hardcoded before

src/zeroband/training/utils.py
mikasenghaas: mmh it's a union of llama and qwen models tho

src/zeroband/training/world.py
mikasenghaas: brevity is nice, `world.rank` etc. make sense imo?

src/zeroband/utils/config.py
mikasenghaas: ;((

src/zeroband/training/ac.py
mikasenghaas: actually this fit quite perfeclty into model.py and simplified a lot of the redundant setup code for the train and logprob model

src/zeroband/training/config.py
mikasenghaas: yep, fused into model configs

src/zeroband/training/config.py
mikasenghaas: fits nicely into model

src/zeroband/training/config.py
mikasenghaas: removed any seeding on trainer

src/zeroband/training/data.py
mikasenghaas: the reason was that it could be confused with only the input (=prompt) token ids, but ig hf standard is convincing

src/zeroband/training/config.py
```diff
@@ -3,67 +3,100 @@
...
-    path: Annotated[str | None, Field(default=None)]
-    interval: Annotated[int | None, Field(default=None)]
-    interval_rollout: Annotated[int | None, Field(default=None)]
-    resume: Annotated[str | None, Field(default=None)]
+    buffer_size: Annotated[
+        int,
+        Field(
+            default=1,
+            ge=1,
+            description="Number of weight checkpoint steps to keep in memory. This is useful to keep some weight-only checkpoints for online evals.",
```
samsja: I am not sure when we need this weight only for eval in the trainer

mikasenghaas: im not sure i get this?

mikasenghaas: because the trainer is in charge of removing the weight ckpt atm?

PR #494, opened 2025-07-01
Debug math run

samsja:


src/zeroband/training/orchestrator/orchestrator.py
```diff
@@ -205,6 +221,8 @@ async def orchestrate(config: OrchestratorConfig, setup_queue: Queue | None = No
             "progress/infer/total_tokens": total_tokens,
             "progress/infer/total_samples": total_samples,
             "progress/train/step": ckpt_step,  # Shared W&B axis
+            "progress/train/epoch": epoch,
+            "progress/train/epoch_step": epoch_step + 1,
```
mikasenghaas: let's not log epoch step, not very informative

src/zeroband/training/orchestrator/orchestrator.py
```diff
@@ -97,18 +97,34 @@ async def orchestrate(config: OrchestratorConfig, setup_queue: Queue | None = No
...
+        if epoch_step == 0:
+            epoch += 1
+            logger.info(f"Starting epoch {epoch}")
+            # Reshuffle dataset at the beginning of each epoch
+            dataset = dataset.shuffle(seed=(config.seed or 0) + epoch - 1)
+
+        step += 1
+        logger.info(
+            f"Orchestrator step {step} (epoch: {epoch}, epoch_step: {epoch_step + 1}/{steps_per_epoch}, checkpoint step: {ckpt_step})"
+        )
```
mikasenghaas: hmm im not sure i like the step = 0 ... step += 1 at the beg of loop. i think a for loop is much more explicit about the definition of a step and the only point we'd need this here is to check the epoch step with modulo?

src/zeroband/training/orchestrator/orchestrator.py
```diff
@@ -97,18 +97,34 @@ async def orchestrate(config: OrchestratorConfig, setup_queue: Queue | None = No
     dataset = dataset.shuffle(seed=config.seed)
 
     # Iterate over dataset in batches
-    total_steps = min(config.max_steps or math.inf, len(dataset) // config.batch_size)
-    logger.info(f"Starting training loop ({total_steps=}, {config.batch_size=})")
+    max_steps = config.max_steps or math.inf
+    steps_per_epoch = len(dataset) // (config.batch_size // config.sampling.n)
```
mikasenghaas, line 101: oop nice catch, this was acc a bug i introduced before not dividing the batch szie by sampling n to compute max steps. nice

configs/inference/simple_math.toml
```diff
@@ -1,19 +1,6 @@
...
+max_model_len = 2048 # test 
 
 [parallel]
-dp = 6
-
-[rl]
-ckpt_path = "checkpoints"
-clean_ckpt_path = true
\ No newline at end of file
+dp = 6
```
mikasenghaas, line 6: btw, we might want to update our priors on how many dp/ tp inference servers we need with async servers, but fine for now

mikasenghaas: for maximal perf i mean

src/zeroband/training/orchestrator/orchestrator.py
mikasenghaas: but epoch yes!

samsja: fair

configs/inference/simple_math.toml
samsja: agree

PR #516, opened 2025-07-04
Sami/refactor/infer side car 2

samsja:


src/zeroband/training/utils.py
```diff
@@ -66,3 +73,90 @@ def wake_up_model_from_cpu(model: Model, tensors: OffloadedTensor):
...
+        try:
+            os.dup2(devnull_fd, 1)  # Redirect stdout
+            os.dup2(devnull_fd, 2)  # Redirect stderr
+            self.fn(self.config, vllm_args=[])
+        finally:
+            os.dup2(old_stdout_fd, 1)  # Restore stdout
+            os.dup2(old_stderr_fd, 2)  # Restore stderr
+            os.close(devnull_fd)
+            os.close(old_stdout_fd)
+            os.close(old_stderr_fd)
```
mikasenghaas, line 124: can you explain why we need this?

mikasenghaas (commented):
code looks good, but the more im seeing this sidecar stuff the less i can help the following feeling: it seems like we are not embracing the modularity of our components. i get that it is a convenient to define configs for one experiment in one file, but does that mean our trainer *needs* to spawn all the depedency processes? it seems like this should not be the concern of the trainer btu the person doing the run. also, in this case our configs don't work the second we want to decouple the processes, e.g. host inference somewhere else, run the orchestrator separate from the trainer. ig my point is, can we think of a alternative solution that maintains the convenience of not having redundant configs (e.g. model name on trainer, orchestrator and inference) but keeps the modularity (i.e. actually start the components separate)

PR #517, opened 2025-07-05
[fix] grad acc scaling

Jackmin801:
the previous behavior scaled loss down too much. lets just scale them all in one code chunk


samsja (approved):
lgtm

PR #521, opened 2025-07-05
Ratio/sami way

samsja:


src/zeroband/training/train.py
```diff
@@ -233,15 +233,19 @@ def train(config: TrainingConfig):
...
             loss.backward()
 
             loss_metrics["loss/loss"] += loss.detach().clone()
             loss_metrics["loss/entropy"] += entropy.detach().clone()
-            loss_metrics["loss/clip_ratio"] += clip_ratio.detach().clone()
+
+            if config.loss.variant == "clip":
+                loss_metrics["loss/clip_ratio"] += ratio_info.detach().clone()
+            elif config.loss.variant == "ratio":
+                loss_metrics["loss/mean_ratio"] += ratio_info.detach().clone()
```
mikasenghaas: i really don't think this is scalable. this ~works for two variants with one differing arg, the second we want to export more to subcomponents of logs this becomes extremtly ugly and seems like it's very easily solved by having the return type be `loss: Tensor, loss_componets: dict[str, Tensor]` and then you simply iterate over the loss comps for logging

mikasenghaas: another solution would be to `get_monitor` in the loss function and log directly and have the return type be jsut `loss: Tensor`

mikasenghaas: ah this might have issues tho when we want to apply normalizing etc, my vote is still with proposal 1

samsja: we are not in the variant game, we will pick whatever works best

mikasenghaas:
i would rather keep it the way we currently do it, than doing this

samsja:
> i would rather keep it the way we currently do it, than doing this

sg to me I can close

PR #524, opened 2025-07-05
only do one forward when doing recompute logprob

samsja:


Jackmin801 (approved):
this would make it higher vram usage than the later loop? also do we know how much faster this makes it?

samsja:
> this would make it higher vram usage than the later loop? also do we know how much faster this makes it?

I am running benchmark rn. 

Yes this would use higher vram, BUT this will still be way lower that the max vram that is achieved during training because of the activation

samsja:
lets keep open I will do this at some point

mikasenghaas:
but for any proper run we are compute bound at bs=1 anyways, so no real point?

samsja:
still will matter for prod run where we can fine grain inference and training

mikasenghaas:
fine grain?

PR #525, opened 2025-07-05
add 7b simple math config 

samsja:


configs/training/simple_math_7b.toml
```diff
@@ -0,0 +1,28 @@
+[monitor.wandb]
+project = "math-debug"
```
mikasenghaas, line 2: general q: shoul we keep experiments with different model sizes in the same wandb project

mikasenghaas (approved):

configs/training/simple_math_7b.toml
samsja: I don't think so, also its too late to chate this for math-debug

PR #529, opened 2025-07-06
sami/refactor/rename clip ratio

samsja:


mikasenghaas (commented):
sure, but then let's name it to what it actually is. mean_ratio could be the mean ratio of anything. im not super sure what the best name should be because i dont understand the variable enough. i think @Jackmin801 had an idea

PR #534, opened 2025-07-07
[fix] start loop with break condition

Jackmin801:
this should be the correct way to do it because then max_steps=0 correctly does not execute the loop. putting the break at the end would have the loop execute once before it realizes we actually asked for 0


mikasenghaas (approved):
🚀

PR #539, opened 2025-07-07
refactor eval

samsja:


src/zeroband/training/orchestrator/config.py
```diff
@@ -185,7 +172,20 @@ class EvalConfig(BaseConfig):
...
+            description="Interval at which to evaluate the model.",
+        ),
+    ] = 100
+
+    eval_base_model: Annotated[
+        bool,
+        Field(
+            description="Whether to evaluate the base model we are training on.",
+        ),
+    ] = False
```
mikasenghaas, line 188: i think this should not be configurable. ie. if eval is on, you eval the base always. otherwise the eval plot in wandb only starts at `interval` so we don't know what happened in the first `interval` steps

src/zeroband/training/orchestrator/orchestrator.py
```diff
@@ -183,20 +177,28 @@ async def orchestrate(config: OrchestratorConfig, setup_queue: Queue | None = No
...
-                await run_benchmark(
-                    client,
-                    benchmark,
-                    config.model,
-                    config.sampling,
-                    ckpt_step,
-                )
+            # Only evaluate on step 0 if eval_base_model is set
+            if ckpt_step == 0 and not config.eval.eval_base_model:
+                pass
```
mikasenghaas: move this in the main condition to avoid nested if {if else}

mikasenghaas (approved):
nice

PR #542, opened 2025-07-08
Implement inference benchmark

mikasenghaas:
This PR adds the boolean `--bench` flag to the orchestrator. It will automatically set the maximum number of steps to run to 6 (1 warmup + 5 real ones), the async level to infinity (to prevent waiting for the non-existent trainer) and disables W&B. 

Screenshot 2025-07-08 at 12 37 20 PM

For example, to benchmark from debug mode first run the inference server using

```bash
uv run infer @ configs/inference/debug.toml 
```

And then benchmark against it using

```bash
uv run orchestrator @ configs/training/orchestrator/debug.toml --bench
```

Some things that changed along the way:
- Implemented generic functions `format_num` and `format_time` to convert numeric values/ times into human-readable format


samsja:
> This PR adds the boolean `--bench` flag to the orchestrator. It will automatically set the maximum number of steps to run to 6 (1 warmup + 5 real ones), the async level to infinity (to prevent waiting for the non-existent trainer) and disables W&B.
> 
> Screenshot 2025-07-08 at 12 37 20 PM
> For example, to benchmark from debug mode first run the inference server using
> 
> ```shell
> uv run infer @ configs/inference/debug.toml 
> ```
> 
> And then benchmark against it using
> 
> ```shell
> uv run orchestrator @ configs/training/orchestrator/debug.toml --bench
> ```
> 
> Some things that changed along the way:
> 
> * Implemented generic functions `format_num` and `format_time` to convert numeric values/ times into human-readable format
hmm I think we should enable wandb tho, still useful for comparing two run

src/zeroband/training/orchestrator/config.py
```diff
@@ -290,3 +297,12 @@ def validate_batch_size(self):
         if self.batch_size < self.micro_batch_size:
             raise ValueError("Batch size must be greater than or equal to micro batch size")
         return self
+
+    @model_validator(mode="after")
+    def validate_bench(self):
+        if self.bench:
+            self.max_steps = 6  # Run for 1 warmup step + 5 evaluation steps
+            self.async_level = 1e9  # Never wait for RL weight checkpoints
+            self.monitor.wandb = None  # Disable W&B
```
samsja: I would remove this

src/zeroband/utils/utils.py
```diff
@@ -125,3 +126,53 @@ def wait_for_path(path: Path, interval: int = 1, log_interval: int = 10) -> None
             logger.debug(f"Waiting for path `{path}` for {wait_time} seconds")
         time.sleep(interval)
         wait_time += interval
+
+
+def to_dict(list_of_dicts: list[dict[str, Any]]) -> dict[str, list[Any]]:
```
samsja: the name of the function is not explicit enough

src/zeroband/training/orchestrator/config.py
mikasenghaas: that's the whole point of the flag? i removed the wandb part

samsja: sorry I wanted to point only to wandb lol

PR #543, opened 2025-07-08
Refactor weight checkpointing

mikasenghaas:


src/zeroband/training/orchestrator/client.py
```diff
@@ -8,6 +8,7 @@
 from vllm.entrypoints.openai.api_server import TokenizeResponse
 
 from zeroband.training.orchestrator.config import ClientConfig, ModelConfig, SamplingConfig
+from zeroband.training.weights import WeightCheckpointManager
```
samsja: lets not create coupling and recreate the function on this side (like we used to do between inference and training)

src/zeroband/training/orchestrator/client.py
```diff
@@ -61,7 +62,7 @@ async def reload_weights(client: AsyncOpenAI, path: Path, step: int) -> None:
     """Make a HTTP post request to the vLLM server to reload the weights."""
     logger = get_logger()
     url = str(client.base_url)[:-4] + "/reload_weights"
-    model_path = path / f"step_{step}" / "pytorch_model.bin"
+    model_path = WeightCheckpointManager.get_model_path(path, step)
```
samsja: I don't like this pattern, too much coupling for a simple string manipulation

src/zeroband/training/orchestrator/utils.py
```diff
@@ -51,7 +52,7 @@ def parse_completions(chat_completions: list[ChatCompletion]) -> list[str]:
 
 
 def wait_for_weight_checkpoint(path: Path, step: int, interval: int = 1, log_interval: int = 10) -> None:
-    model_path = Path(path) / f"step_{step}" / "pytorch_model.bin"
+    model_path = WeightCheckpointManager.get_model_path(path, step)
```
samsja: same lets remove

src/zeroband/training/orchestrator/client.py
samsja: otherwise we can create a very simple helper func that does this outisde of the classs

src/zeroband/training/train.py
```diff
@@ -274,10 +272,8 @@ def train(config: TrainingConfig):
         optimizer.zero_grad()
 
         # Save the weight checkpoint
-        step_path = Path(config.weights.path) / f"step_{progress.step + 1}"
         save_weights_start_time = time.time()
-        model_path = save_weight_checkpoint(model, tokenizer, step_path, async_save=config.weights.save_async)
-        active_weight_checkpoint_paths.append(step_path)
+        model_path = weight_ckpt_manager.save(model, tokenizer, step=progress.step + 1)
```
samsja: why + 1 here ? Lets add comment to explain it

src/zeroband/training/train.py
```diff
@@ -286,16 +282,8 @@ def train(config: TrainingConfig):
...
-        candidate_weight_step_to_delete = int(candidate_path_to_delete.name.split("_")[-1])
-        keep_for_eval = config.weights.interval and candidate_weight_step_to_delete % config.weights.interval == 0
-        # For checkpointing step x, we need all weight checkpoints in [x-async_level, x] (for logprob model)
-        # To get [n-k, n] with interval n and buffer k over all natural numbers x, we use the condition (n - (x % n)) % n <= k
-        keep_for_ckpt = config.ckpt and (config.ckpt.interval - (candidate_weight_step_to_delete % config.ckpt.interval)) % config.ckpt.interval <= config.async_level 
-        if not (keep_for_eval or keep_for_ckpt):
-            logger.debug(f"Removing past weight checkpoint {candidate_path_to_delete} ({keep_for_eval=}, {keep_for_ckpt=})")
-            shutil.rmtree(candidate_path_to_delete, ignore_errors=True)
+        candidate_step = max(progress.step - (config.async_level + 1), 0)
+        weight_ckpt_manager.clean(candidate_step)
```
samsja: this feel off, I would hide this logic inside the clean function. The clean() should take progress.step as input

src/zeroband/training/orchestrator/client.py
mikasenghaas: hm, yeah, the idea obviously was that this is in fact shared logic and so if we define it in one place there's no chance of a mismatch between creator and consumer

mikasenghaas: but yea the import from train is not nice

src/zeroband/training/weights.py
```diff
@@ -1,82 +1,146 @@
...
+        1. The step is an evaluation step (e.g. step % weights.interval == 0)
+        2. The step is a checkpoint step or at most async_level steps earlier
+        """
+        if self.config.save_async:
+            thread = threading.Thread(
+                target=self._clean,
+                args=(step,),
+                name=f"weight-checkpoint-clean-{step}",
+            )
+            thread.start()
```
samsja: Now that we have a class I would add the thread to a list of thread we could decide later to handle the termination of this threads. for now its fine

src/zeroband/training/train.py
mikasenghaas: its bc of the step definition. at any train step t, we are working we update from t to t+1. at the time of writing the ckpt we have alredy updated the model but the step counter. if we update the step counter at the time of model update (which feels natural) all logs after will have a diff step, and so our "step definition" changes in mid loop. have discussed this with jack and length and he wants it like this haha

src/zeroband/training/train.py
mikasenghaas: ah yea maybe correct. the reason for this is that if felt the cleaner interface is to guarantee that if you call clean on a step then you actually want to clean that step (and not the async off step). let me know if i should change, i dont mind too much

src/zeroband/training/weights.py
mikasenghaas: ah yea, probably makes sense to review thread safety in general for this kinda stuff. i also feel like we might have race conditions in the current logic but for now i just copied the logic we had

src/zeroband/training/orchestrator/client.py
samsja: I don't like the class, I would be fine with a function that import from train

samsja (approved):

PR #544, opened 2025-07-08
backport two tasks to registry

kalomaze:


src/zeroband/training/environments/registry.py
```diff
@@ -85,13 +89,163 @@ def lcs_ratio(x: str, y: str) -> float:
     return vf_env
 
 
+def load_unscramble_environment(env_args: dict = {}) -> Environment:
+    # Load the unscramble dataset
+    dataset = load_dataset("kalomaze/unscramble-mix-it2", split="train")
```
mikasenghaas, line 93: can we start having these under the prime hf org? would be nice to have a collection `Environments` which has datasets for our custom environments

src/zeroband/training/environments/registry.py
```diff
@@ -85,13 +89,163 @@ def lcs_ratio(x: str, y: str) -> float:
...
+
+    def process_dataset(example):
+        verification_info = json.loads(example["verification_info"])
```
mikasenghaas, line 96: we might wanna move away from the `verification_info` field in environment datasets which was essentially a cheap way to write rewards function in a standardized way before by just json dumping everything. im not super familar with the way task samples are linked to their ground truth in  reward funcs in `verifiers` but i'd imagine we can make the "ground truth" field be more task-dependent and then handle the parsing with the reward func, e.g. for a math question it would be a latex string, for a code question it would be a list of unit tests as strings etc. 

@willccbb would know for sure

src/zeroband/training/environments/registry.py
```diff
@@ -85,13 +89,163 @@ def lcs_ratio(x: str, y: str) -> float:
...
+        example["answer"] = verification_info["ground_truth"]
```
mikasenghaas, line 97: re myself (lol): yea so seems like it requires an answer key that can be mapped which is niced. in this case i would move the format of the underlying dataset to just have the answer key directly whenver we own the dataset

mikasenghaas (commented):
nice, nice, nice. starting to look good with our environments

src/zeroband/training/environments/registry.py
```diff
@@ -1,3 +1,7 @@
+import re
```
willccbb: can you move any imports inside the functions that need them? idea is that each function is a self-contained "factory" for 

eventually the environments hub (pihub) will take the place of the registry, and we'll want to copy/paste sections

kalomaze:
https://github.com/PrimeIntellect-ai/prime-rl/pull/582
^ Will be instead merged here

PR #545, opened 2025-07-08
Implement single RL entrypoint

mikasenghaas:
This PR implements a convenience feature for starting single-node RL experiments.

TODOs:
- [ ] Tie shared configs in `rl.py`
- [ ] Enable running a subset of the processes from `rl.py`
- [ ] Implement single-node device placement
- [ ] Improve logs from the RL entrypoint itself (all, some, no logs?)
- [ ] Enable restarting an experiment from within tmux session with file streamed logs


mikasenghaas:
> * I have the feeling that the tmuxninator right now overfit on one workflow we should make it a bit more generic and just split the logs into 3 window, maybe a nvtop and call it a day. If someone want a more complex workflow it can add it outside of the repo

think of it as a recommendation. anybody can configure whatever they want. its very customziable by design.

at some point, i had the following setup:
1. Window: Only entrypoint, use this to optionall start inference server as bg process and then rl on top
2. Window: Trainer+Orchestraor file logs as before
3. Window: NVTOP + HTOP

so essentially just removing the extra inference window. the reason i added is that i thought it might sometimes be nice to see the inference logs

PR #551, opened 2025-07-11
Update tmuxinator

samsja:
screenshot-20250711-154128


mikasenghaas:
as discussed, i think its better to keep one window for the RL entrypoint and then do `even-vertical` on the 3 logs. be mindful that the inference logs are from `.stdout` while the rest is from `.log`

PR #552, opened 2025-07-11
Log uncaught exceptions to file logs

mikasenghaas:
If a subprocess (trainer, orchestrator) fails, this is reflected in the file logs to immediately debug the issue


samsja (approved):
lgtm

PR #570, opened 2025-07-12
Fix for difficulty filtering double-negative, min + max each optional

willccbb:
also removing unused ThinkParser

### DIFF
--- a/src/prime_rl/environments/registry.py
+++ b/src/prime_rl/environments/registry.py
@@ -44,26 +44,23 @@ Provide the final numerical answer inside \\boxed{{...}}."""
 def load_intellect_math_environment(env_args: dict = {}) -> Environment:
     import json
 
-    from verifiers.utils.data_utils import extract_boxed_answer
-
     from prime_rl.orchestrator.genesys.math import compute_math_reward
 
     train_dataset = load_dataset("PrimeIntellect/INTELLECT-2-only-math", split="train").map(
         lambda x: {"question": x["prompt"], "info": json.loads(x["verification_info"]), "task": "simple-math"}
     )
     solve_rate_field = env_args.get("solve_rate_field", None)
-    if not solve_rate_field:
-        min_solve_rate = env_args["min_solve_rate"]
-        max_solve_rate = env_args["max_solve_rate"]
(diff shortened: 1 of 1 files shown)


src/prime_rl/environments/registry.py
```diff
@@ -44,24 +44,21 @@ def correct_answer_reward_func(completion, answer, **kwargs) -> float:
...
         lambda x: {"question": x["prompt"], "info": json.loads(x["verification_info"]), "task": "simple-math"}
     )
     solve_rate_field = env_args.get("solve_rate_field", None)
-    if solve_rate_field is not None:
-        min_solve_rate = env_args["min_solve_rate"]
-        max_solve_rate = env_args["max_solve_rate"]
-        train_dataset = train_dataset.filter(
-            lambda x: x[solve_rate_field] >= min_solve_rate and x[solve_rate_field] <= max_solve_rate
-        )
+    if solve_rate_field:
```
samsja:
```suggestion
    if solve_rate_field is not None:
```

src/prime_rl/environments/registry.py
```diff
@@ -44,24 +44,21 @@ def correct_answer_reward_func(completion, answer, **kwargs) -> float:
...
+        min_solve_rate = env_args.get("min_solve_rate", None)
+        max_solve_rate = env_args.get("max_solve_rate", None)
+        if min_solve_rate is not None:
+            train_dataset = train_dataset.filter(lambda x: x[solve_rate_field] >= min_solve_rate)
+        if max_solve_rate is not None:
+            train_dataset = train_dataset.filter(lambda x: x[solve_rate_field] <= max_solve_rate)
     train_dataset = train_dataset.remove_columns(["prompt", "verification_info"])
 
-    parser = vf.ThinkParser(extract_fn=extract_boxed_answer)
```
samsja, line 63: so we don't have think parser anymore ? How is it working rn for the reward and thinking ?

willccbb: ThinkParser wasn't actually being used before haha, Parser is only used in multi-turn or if you explicitly reference it in reward functions

mikasenghaas (approved):
looks good

PR #574, opened 2025-07-13
Sami/refactor/async save

samsja:


mikasenghaas (approved):
we only save weights async? not full?

samsja:
> we only save weights async? not full?

right now the feature to save full async is not implemented

PR #577, opened 2025-07-13
Datapool (dataloader for dynamic datasets, online difficulty filtering, buffer, etc.)

justusmattern27:


justusmattern27:
PR implements a datapool, which is essentially a new dataloader that lets us work with more dynamic datasets and implement many of the data loading methods from the literature (e.g. online difficulty filtering).

The datapool lets you add training data on the fly, maintain a buffer to re-use generated training samples, and implement priority pools (e.g. to have 10% of your training batch from a low-priority and 90% of your batch from a high-priority batch).

On top of this, we can implement much of what we discussed and have seen in papers. Two data loading strategies are alredy implemented:

- online difficulty filtering: ignore samples below and above a certain avg reward threshold, e.g. to ignore zero advantages). The implementation uses a buffer so that we don't keep samples from oversampling

- difficulty adjusted sample prioritization: keep a low prio and high prio data pool, and for every batch, sample 90% from the low prio and 10% from high prio pool. Put every prompt with zero advantages in the low priority pool after every batch


I've run experiments on hendrycks math to make sure everything works:
Baseline: https://wandb.ai/primeintellect/prime-rl/runs/j5a20h0b/workspace?nw=nwuserjustus27
Baseline with extra hard questions added for 2/3 of the dataset (with 0 solve rate for all): https://wandb.ai/primeintellect/prime-rl/runs/0yh09m18?nw=nwuserjustus27
Difficulty Prioritization (with 2/3 extra hard questions added to make the effect noticable): https://wandb.ai/primeintellect/prime-rl/runs/znycsz3j?nw=nwuserjustus27
Hendrycks Math with online difficulty filtering (= ignore zero advantages): https://wandb.ai/primeintellect/prime-rl/runs/bs6udv9g?nw=nwuserjustus27

configs/orchestrator/hendrycks_math/1b_no_difficulty_prioritization.toml
mikasenghaas, line 1: we need to find a good way to store/log variants of an experiment. i think it gets too verbose if we have a toml for each minor variant, wdyt would be good solution here @samsja

src/prime_rl/environments/registry.py
```diff
@@ -102,6 +102,71 @@ def correct_answer_reward_func(completion, info, **kwargs) -> float:
     return vf_env
 
 
+def load_hendrycks_math_environment_with_additional_difficult_questions(env_args: dict = {}) -> Environment:
```
mikasenghaas, line 105: i think the same goes for environments, we want to support iterations on data mixes in environments (e.g. variants of an environment) but having separate ones in the registry seems too verbose. ig this might be solved by pi-hub? @willccbb

src/prime_rl/environments/registry.py
```diff
@@ -102,6 +102,71 @@ def correct_answer_reward_func(completion, info, **kwargs) -> float:
...
+    from verifiers.utils.data_utils import extract_boxed_answer
+
+    from prime_rl.orchestrator.genesys.math import compute_math_reward
+
+    train_dataset = load_dataset("justus27/math-hendrycks-genesys-format", split="train").map(
+        lambda x: {
+            "question": x["prompt"],
+            "info": json.loads(x["verification_info"]),
+            "task": "simple-math",
+            "priority": "high",
```
mikasenghaas, line 117: does any dataset that we do priority sampling on need this column?

src/prime_rl/orchestrator/config.py
```diff
@@ -154,6 +154,79 @@ class CheckpointConfig(BaseConfig):
     ] = None
 
 
+class DifficultyPrioritizationStrategy(BaseConfig):
+    enabled: Annotated[
+        bool,
+        Field(
+            description="Whether to enable difficulty-based prioritization for training samples.",
+        ),
+    ] = False
```
mikasenghaas, line 163: we converged to the config design where `subconfig: Subconfig | None = None` to enable/ disable subconfigs, let's do it here too

src/prime_rl/orchestrator/config.py
```diff
@@ -154,6 +154,79 @@ class CheckpointConfig(BaseConfig):
...
+    ] = 0.1
+
+
+class OnlineDifficultyFilteringStrategy(BaseConfig):
+    enabled: Annotated[
+        bool,
+        Field(
+            description="Whether to enable online filtering of samples based on difficulty.",
+        ),
+    ] = False
```
mikasenghaas, line 188: same here

src/prime_rl/orchestrator/config.py
```diff
@@ -154,6 +154,79 @@ class CheckpointConfig(BaseConfig):
...
 
+class DifficultyPrioritizationStrategy(BaseConfig):
+    enabled: Annotated[
+        bool,
+        Field(
+            description="Whether to enable difficulty-based prioritization for training samples.",
+        ),
+    ] = False
+
+    priority_data_field: Annotated[
```
mikasenghaas, line 165: wdyt of just `priority_field`?

src/prime_rl/orchestrator/config.py
```diff
@@ -154,6 +154,79 @@ class CheckpointConfig(BaseConfig):
...
+        str | None,
+        Field(
+            description="Field name in the dataset that contains priority information. If None, all samples are treated as high priority by default.",
+        ),
+    ] = None
+
+    low_priority_batch_fraction: Annotated[
```
mikasenghaas, line 172: `low_priority_fraction`? trying to delete unnecessary characters in configs as much as possible

src/prime_rl/orchestrator/config.py
```diff
@@ -154,6 +154,79 @@ class CheckpointConfig(BaseConfig):
...
+    max_sample_tries: Annotated[
+        int,
+        Field(
+            gt=0,
+            description="Maximum number of attempts to sample within the reward thresholds.",
+        ),
+    ] = 3
+
+
+class DataLoadingConfig(BaseConfig):
```
mikasenghaas, line 225: call this `DataPoolConfig` to tie config name to class?

src/prime_rl/orchestrator/config.py
```diff
@@ -154,6 +154,79 @@ class CheckpointConfig(BaseConfig):
...
+    online_difficulty_filtering_strategy: OnlineDifficultyFilteringStrategy = OnlineDifficultyFilteringStrategy()
+    difficulty_prioritization_strategy: DifficultyPrioritizationStrategy = DifficultyPrioritizationStrategy()
```
mikasenghaas, line 227: i get that these are not mutually exclusive? is this generally the case? ie. can we stack any sampling technique on top of each other? if not, this design sort of design might get awkward to handle in the client-code

src/prime_rl/orchestrator/data.py
```diff
@@ -1,14 +1,30 @@
...
 
+@dataclass
+class GeneratedSample:
+    prompt_tokens: List[int]
+    completion_tokens: List[int]
+    completion_logprobs: List[float]
+    prompt_masks: List[int]
+    completion_masks: List[int]
+    reward: List[float]
+    advantages: List[float]
```
mikasenghaas, line 25: are the types wrong? one sample only has one reward/ advantage? also, why do we need this dataclass at all?

src/prime_rl/orchestrator/orchestrator.py
mikasenghaas, line 1: the changes to the orchestrator seem very intrusive. bunch of nested conditionals and loops, does not seem very scalable and should not live on orchestrator imo. do you think we can get a simpler outward facing api for the datapool, maybe just something like `datapool.sample(n)` with all of the complexity handled internally?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -23,13 +23,16 @@
...
+from prime_rl.orchestrator.data import DataPool, prepare_batch, GeneratedSample
 from prime_rl.orchestrator.logger import setup_logger
 from prime_rl.orchestrator.utils import (
     process_env_results,
     compute_advantages,
     wait_for_weight_checkpoint,
     print_benchmark,
+    flatten_keep,
+    create_generated_samples,
+    unpack_generated_samples
```
mikasenghaas, line 35: what's the reason for the nesting then flattening logic?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -173,49 +169,119 @@ async def orchestrate(config: OrchestratorConfig):
...
+        
+        all_generated_samples = datapool.get_buffered_samples()
+        
+        if len(all_generated_samples) >= config.batch_size:
+            all_generated_samples, samples_to_buffer = all_generated_samples[:config.batch_size], all_generated_samples[config.batch_size:]
+            datapool.add_to_buffer(samples_to_buffer)
+        
+        else:
+            for sampling_iteration in range(config.data_loading.online_difficulty_filtering_strategy.max_sample_tries):
+
```
mikasenghaas, line 183: what happens if we do not have at least `batch_size` samples in the difficulty range after `max_sample_tries`?

src/prime_rl/environments/registry.py
willccbb: been working on implementing a lightweight version of the registry setup within verifiers which we can probably start using soon for local registration without needing full pihub setup. basic idea is that environments should take kwargs for anything we want to be easily configurable within the same env loading module, e.g. data mix, flags for prompt/format choices.

maybe a rule of thumb could be that any non-adaptive prompt-level preprocessing is handled by the environment (as environments are only initialized once, then the dataset is handed off to the orchestrator), and the orchestrator is responsible for adaptive/online filtering, as well as distribution balance if we want this to be adjusted during training (e.g blending in more code/reasoning data gradually).

configs/orchestrator/hendrycks_math/1b_no_difficulty_prioritization.toml
willccbb: would be nice if there's a way we can have environment-specific toml files that support arbitrary args.

example:
```toml
[environment]
id = "reverse-text"

[environment.args]
use_thinking = false
max_num_sentences = 5
reward_type = "LCS"
```

which then gets passed as:
```python
# orchestator
vf_env = load_environment(env_id=config.environment.id, env_args=config.environment.args)

# registry -- load_environment
def load_environment(env_id, env_args)
    return REGISTRY[env_id](**env_args)
```
where the fields in `args` aren't pre-defined in advance in the config, and can be different for each environment.

thoughts on how to do this within our config setup? would also be nice to support this in verifiers directly as the default way of configuring/loading an env

samsja: afaik the only difference here is that we have a different env ? If yes I suggest that we just override by cli or use toml inheritance on hendrycks math

src/prime_rl/orchestrator/config.py
samsja: agree

samsja (commented):
mini comment can we use List -> list and Dict -> dict

PR #579, opened 2025-07-13
Deprecate genesys

mikasenghaas:
This PR removes all currently unused task rewards functions from `prime_rl.orchestrator.genesys` (only keeps `math_reward` for backwards compatibility). 

Also cleans the project dependencies which should make the ROCM image build again.


justusmattern27:
lgtm

PR #580, opened 2025-07-13
use real mean for importance ratio

samsja:


src/prime_rl/trainer/loss.py
```diff
@@ -141,7 +141,7 @@ def grpo_loss_ratio(
     loss = _apply_mask(per_token_loss, loss_mask, max_tokens)
 
     with torch.no_grad():
```
mikasenghaas: no need for the no_grad anymore?

src/prime_rl/trainer/loss.py
```diff
@@ -141,7 +141,7 @@ def grpo_loss_ratio(
...
-        ratio_avg = _apply_mask(ratio, loss_mask, max_tokens)
+        ratio_avg = ratio.mean()
```
mikasenghaas: why does `_apply_mask` not do the same?

src/prime_rl/trainer/loss.py
```diff
@@ -141,7 +141,7 @@ def grpo_loss_ratio(
     loss = _apply_mask(per_token_loss, loss_mask, max_tokens)
 
     with torch.no_grad():
-        ratio_avg = _apply_mask(ratio, loss_mask, max_tokens)
+        ratio_avg = (loss_mask * ratio).sum() / loss_mask.sum()
```
mikasenghaas: i still dont get why we would only do this for computing `ratio_avg` and not everywhere `_apply_mask` is used? won't we have the same issue with the loss?

PR #581, opened 2025-07-13
Deprecate W&B group

mikasenghaas:
This PR removes the ability to set a W&B group via `--monitor.wandb.group` which was previously used to group together runs. We found the UI and handling to be overcomplicated, so now we group experiments runs by using a shared prefix in the run name (which is also editable)

### DIFF
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -219,7 +219,7 @@ class TrainerConfig(BaseSettings):
     @model_validator(mode="after")
     def auto_setup_bench(self):
         if self.bench:
-            self.max_steps = 6  # 1 Warmup + 5 Benchmark
+            self.max_steps = 4  # 1 Warmup + 3 Benchmark
             if not self.data.fake:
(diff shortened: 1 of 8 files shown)


src/prime_rl/trainer/config.py
```diff
@@ -219,7 +219,10 @@ class TrainerConfig(BaseSettings):
     @model_validator(mode="after")
     def auto_setup_bench(self):
         if self.bench:
-            self.max_steps = 6  # 1 Warmup + 5 Benchmark
+            self.max_steps = 4  # 1 Warmup + 3 Benchmark
             if not self.data.fake:
                 self.data.fake = FakeDataLoaderConfig()
+
+            if self.monitor.wandb:
+                self.monitor.wandb.project = f"{self.monitor.wandb.project}-bench"
```
samsja: not sure about this one seems to intrusive

either we use a fix bench project and the user is in charge to pick to add -bench

samsja (approved):

PR #583, opened 2025-07-13
Fix normalizing by batch tokens

mikasenghaas:
This PR fixes an issue where we wrongly normalize the micro-batch by the number of tokens per micro-batch, instead of per-batch. We want to normalize with respect to the entire batch

Let $x=(x_1+\dots+x_n)/n$ be values over $n$ micro batches with normalization constant $c_1+\dots+c_n=c$. 

Now, we want to compute the normalized batch average

$$
\frac{x}{c} = \frac{(x_1+\dots+x_n)/n}{c} =  \frac{x_1+\dots+x_n}{nc} = \frac{x_1}{nc} + \dots + \frac{x_n}{nc}
$$

However, currently we compute

$$
 \frac{x_1}{nc_1} + \dots + \frac{x_n}{nc+n}
$$

which is not mathematically equivalent.

### DIFF
--- a/src/prime_rl/trainer/loss.py
+++ b/src/prime_rl/trainer/loss.py
@@ -16,7 +16,7 @@ def grpo_loss(
     original_logprobs: Float[Tensor, "batch seq_minus_1"],
     loss_mask: Int[Tensor, "batch seq"],
     temperature: float,
-    max_tokens: int,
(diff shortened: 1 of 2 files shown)


samsja (approved):
lgtm

PR #586, opened 2025-07-13
Feature/port pydantic adherence env

JohannesHa:
ported the pydantic adherence env to verifiers format. made the environment completely self contained in the `load_pydantic_adherence_environment` function incl the parser and helper functions for now before we switch to pi hub.


samsja (approved):
lgtm

PR #587, opened 2025-07-14
Add B200 support

mikasenghaas:
We install `torch>2.7.0` from test `cu128` index (should become stable soon) to fix a bug where the FSDP all-gather hung. With this PR we have B200 support for the trainer and inference out of the box.


pyproject.toml
```diff
@@ -43,19 +43,19 @@ requires = ["hatchling"]
 build-backend = "hatchling.build"
 
 [project.optional-dependencies]
-# Entrypoint-specific dependencies
 fa = ["flash-attn>=2.8.0"]
 
 [tool.uv]
 no-build-isolation-package = ["flash-attn"]
+prerelease = "allow"
```
samsja, line 50: do we need prerelease if we end up using 2.8.1 ? Or is this version a pre-release version ?

mikasenghaas: i think without the prerelease flag we cannot use either the test or nightly releases

PR #591, opened 2025-07-14
Support running multiple experiments per node

mikasenghaas:
This PR adds first-class support for parallelizing experiments within a node, i.e. using subsets of GPUs to run multiple experiments in parallel. For this to work the following things have to be uniquely set by each experiment:

1. Inference server port <> orchestrator client port
2. Orchestrator rollout path <> trainer data path
3. Trainer weighs path <> orchestrator weight path
4. The log file names for streaming logs
5. The rendezvous endpoint for torchrun

For convenience, 2), 3) and 4) are handled automatically with minimal change to our logic by introducing the `--exp-id` flag to the RL run which, if set, creates subdirectories in the base `logs`, `rollouts` and `weights` directories. 5. is automatically handled by choosing a random port. 1) is left to the end user.

## Explanation

> *Taken from README*

For small models/ quick ablations, it can be more efficient to parallelize experiments within a node (e.g. split your GPUs to run two experiments in parallel). Because the trainer communicates with the orchestrator via a shared file system, and the orchestrator communicates with the inference engine via an OAI-compatible API, the connection points have to be uniquely set. For example, if you have access to 4 GPUs you can run two 2 GPU training runs in parallel as follows:

Start the first experiment as normal (*will use the first 2 GPUs*)

```bash
./tmux.sh exp-1
```

```bash
# Start the first experiment
uv run rl \
  --trainer @ configs/trainer/reverse_text.toml \
  --orchestrator @ configs/orchestrator/reverse_text.toml \
  --inference @ configs/inference/reverse_text.toml \
  --exp-id exp-1
```

For the second experiment, configure the match a new server port on the orchestrator and inference
```bash
./tmux.sh exp-2
```

```bash
# Start the second experiment
CUDA_VISIBLE_DEVICES=3,4 uv run rl \
  --trainer @ configs/trainer/reverse_text.toml \
  --orchestrator @ configs/orchestrator/reverse_text.toml \
  --inference @ configs/inference/reverse_text.toml \
  --inference.server.port 8001 \
  --orchestrator.client.port 8001 \
  --exp-id exp-2
```

## Screenshot

First experiment runs as `exp-1` session

Screenshot 2025-07-14 at 4 29 51 PM

Second experiment runs as `exp-2` session

Screenshot 2025-07-14 at 4 30 00 PM

## Example W&B run

Logs from 2 `reverse-text` runs in parallel from the same node

Screenshot 2025-07-14 at 4 58 50 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,9 +92,7 @@ uv run rl \
 </details>
 
 
-## Entrypoints
-
-### Setup
+## Additional Setup
 
 1. If you want to log your runs to W&B (`wandb`), log in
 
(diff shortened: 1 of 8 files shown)


samsja (approved):
lgtm

PR #593, opened 2025-07-14
DRAFT: wordle envs in registry, test script, multi-turn logprobs

willccbb:
included:
- wordle envs (think, nothink)
- scripts/test_env.py for offline inference
- SFT'd model for nothink version of wordle (traces finish in 1024 toks, n=128 inference batches take ~3min on 6 GPUs, may want to optimize env implementation + drop TA/nltk dependencies)

in progress:
- multi-turn vLLM logprobs/tokens in verifiers' process_env_results

TODO:
- rl.py configs with willcb/Qwen3-1.7B-Wordle
- test wordle nothink with rl.py

reward curve (verifiers trainer)
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ no-build-isolation-package = ["flash-attn"]
 
 [tool.uv.sources]
 torch = [{ index = "pytorch-cu128" }]
-verifiers = { git = "https://github.com/willccbb/verifiers", rev = "8363357" }
+verifiers = { git = "https://github.com/willccbb/verifiers", rev = "9136da4" }
 
 [[tool.uv.index]]
 name = "pytorch-cu128"
--- a/scripts/test_env.py
+++ b/scripts/test_env.py
@@ -1,0 +1,52 @@
+import argparse
+import os
+import numpy as np
+from openai import OpenAI
+from verifiers.utils.logging_utils import print_prompt_completions_sample
+
+from prime_rl.environments.registry import load_environment
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--env", "-e", type=str, default="pydantic-adherence")
+    parser.add_argument("--model", "-m", type=str, default="gpt-4.1-mini")
+    parser.add_argument("--api-key-var", "-k", type=str, default="OPENAI_API_KEY")
+    parser.add_argument("--api-base-url", "-b", type=str, default="https://api.openai.com/v1")
+    parser.add_argument("--num-examples", "-n", type=int, default=5)
+    parser.add_argument("--rollouts-per-example", "-r", type=int, default=3)
+    parser.add_argument("--max-tokens", "-t", type=int, default=1024)
+    parser.add_argument("--temperature", "-T", type=float, default=0.7)
+    args = parser.parse_args()
+    vf_env = load_environment(args.env)
+    client = OpenAI()
+    sampling_args = {
+        "max_tokens": args.max_tokens,
+        "temperature": args.temperature,
+    }
+    results = vf_env.evaluate(
+        client=client, 
+        model=args.model, 
+        num_examples=args.num_examples,
+        rollouts_per_example=args.rollouts_per_example,
+    )
+    rewards = {k: v for k, v in results.items() if 'reward' in k}
+    print(f"--- Evaluation ---")
+    print(f"Environment: {args.env}")
+    print(f"Model: {args.model}")
+    print(f"Provider: {args.api_base_url}")
+    print(f"Examples: {args.num_examples}")
+    print(f"Rollouts per example: {args.rollouts_per_example}")
+
+    print("--- Example ---")
+    print_prompt_completions_sample(results['prompt'], results['completion'], rewards, step=0)
+    print("--- All ---")
+    print("Rewards:")
+    for k, v in results.items():
+        if 'reward' in k:
+            print(f"{k}: avg - {sum(v) / len(v):.3f}, std - {np.std(v):.3f}")
+            n = args.num_examples
+            r = args.rollouts_per_example
+            for i in range(r):
+                trials = [v[(i*r)+j] for j in range(n)]
+                out = f"r{i+1}: {trials}"
+                print(out)
--- a/src/prime_rl/environments/registry.py
+++ b/src/prime_rl/environments/registry.py
@@ -16,12 +16,21 @@ def load_gsm8k_environment(env_args: dict = {}) -> Environment:
         },
         remove_columns=dataset.column_names,  # type: ignore
     )  # type: ignore
+    eval_dataset = load_dataset("openai/gsm8k", "main", split="test")
+    eval_dataset = eval_dataset.map(
+        lambda x: {
+            "question": x["question"],
+            "answer": extract_hash_answer(x["answer"]),
+            "info": {},
+            "task": "gsm8k",
+        },
+        remove_columns=eval_dataset.column_names,  # type: ignore
+    )  # type: ignore
 
     parser = vf.ThinkParser(extract_fn=extract_boxed_answer)  # uses \boxed{...} to parse the answer by default
 
     def correct_answer_reward_func(completion, answer, **kwargs) -> float:
         response = parser.parse_answer(completion) or ""
-        print(response, answer)
         return 1.0 if response == str(answer) else 0.0
 
     rubric = vf.Rubric(
@@ -37,7 +46,13 @@ Think step by step inside <think>...</think> tags.
 
 Provide the final numerical answer inside \\boxed{{...}}."""
 
-    vf_env = vf.SingleTurnEnv(dataset=dataset, system_prompt=system_prompt, parser=parser, rubric=rubric)
+    vf_env = vf.SingleTurnEnv(
+        dataset=dataset,
+        eval_dataset=eval_dataset,
+        system_prompt=system_prompt,
+        parser=parser,
+        rubric=rubric,
+    )
     return vf_env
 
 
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -18,6 +18,9 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+# ???
+core.*
+
```
mikasenghaas, line 23: yeaa, we need to get @mikex86 on this again

README.md
```diff
@@ -149,7 +149,7 @@ uv run rl \
 
 Train a small model (`willcb/DeepSeek-R1-Distill-Qwen-1.5B`) on high-school level math questions. It is recommended to have at least 2xA100-80GB GPUs or more for this experiment.
 
-On two GPUs, run the following command to run the experiment.
+On eight GPUs, run the following command to run the experiment.
```
mikasenghaas: this is  already on main but whatever ig

configs/inference/multiturn_wordle/14b.toml
mikasenghaas, line 1: let's just call this `wordle`. should be clear from context and readme section that this is multiturn.

mikasenghaas: also, is the 14b working? if not, let's not include its configs in this pr?

pyproject.toml
```diff
@@ -29,6 +29,8 @@ dependencies = [
     "rich>=14.0.0",
     "tomli-w",
     "verifiers",
+    "textarena>=0.6.16",
+    "nltk>=3.9.1",
```
mikasenghaas, line 33: uff, this will be gone with pihub right?

scripts/test_env.py
```diff
@@ -18,8 +19,9 @@
     parser.add_argument("--max-tokens", "-t", type=int, default=1024)
     parser.add_argument("--temperature", "-T", type=float, default=0.7)
     args = parser.parse_args()
+    print(args)
```
mikasenghaas: this on purpose?

src/prime_rl/environments/registry.py
```diff
@@ -16,6 +18,16 @@ def load_gsm8k_environment(env_args: dict = {}) -> Environment:
...
+    eval_dataset = load_dataset("openai/gsm8k", "main", split="test")
+    eval_dataset = eval_dataset.map(
+        lambda x: {
+            "question": x["question"],
+            "answer": extract_hash_answer(x["answer"]),
+            "info": {},
+            "task": "gsm8k",
+        },
+        remove_columns=eval_dataset.column_names,  # type: ignore
+    )  # type: ignore
```
mikasenghaas, line 30: ah nice! how does this work? is this pre-work for evals with envs or can this be used during training as well somehow to show eval rewards etc.?

src/prime_rl/environments/registry.py
```diff
@@ -36,22 +48,34 @@ def correct_answer_reward_func(completion, answer, **kwargs) -> float:
...
+        rubric=rubric,
+    )
     return vf_env
 
 
-def load_intellect_math_environment(env_args: dict = {}) -> Environment:
+def load_intellect_math_environment(
+    solve_rate_field: str | None = None,
+    min_solve_rate: float | None = None,
+    max_solve_rate: float | None = None,
```
mikasenghaas, line 64: nice!

src/prime_rl/environments/registry.py
```diff
@@ -309,16 +562,420 @@ def pydantic_adherence_reward_func(completion, answer, **kwargs):
...
+    )
+    return vf_env
+
+
+def load_wordle_environment(
+    num_train_examples: int = 2000, num_eval_examples: int = 20, use_think: bool = True, **kwargs
+) -> Environment:
+    # requires `textarena`, `nltk`
+    # model: willcb/Qwen3-{1.7B,4B}-Wordle
+    vf_env = vf.load_environment(
```
mikasenghaas, line 637: wait so our registry plugs into verifiers registry lol? remind me what the final plan was again for pulling envs?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -175,7 +175,7 @@ async def orchestrate(config: OrchestratorConfig):
         problems_per_batch = config.batch_size // config.rollouts_per_prompt
         start_idx = epoch_step * problems_per_batch
         indices = range(start_idx, start_idx + problems_per_batch)
-        problems = dataset.select(indices).to_list() * config.rollouts_per_prompt
+        problems = [problem for problem in dataset.select(indices).to_list() for _ in range(config.rollouts_per_prompt)]
```
mikasenghaas: again, im not sure why this diff shows? if you rebased this change should already there

configs/inference/multiturn_wordle/14b.toml
```diff
@@ -0,0 +1,6 @@
+[model]
+name = "willcb/Qwen3-14B"
```
samsja: what is different with this qwen3 and the official qwen3 ?

src/prime_rl/environments/registry.py
```diff
@@ -102,7 +183,7 @@ def correct_answer_reward_func(completion, info, **kwargs) -> float:
     return vf_env
 
 
-def load_reverse_environment(env_args: dict = {}) -> Environment:
+def load_reverse_environment(**kwargs) -> Environment:
```
samsja: should we just not take any argument here ? I thinki its better to fail at runtime if user pass args to an env that does not support them

uv.lock
samsja, line 1: I think that there is an issue wit hthe uv.lock, there are too many dependencies update. Can u try to redo  uv sync --al-extras without deleeting the uv.locl ?

pyproject.toml
willccbb: yep, but needed for wordle example for now

configs/inference/multiturn_wordle/14b.toml
willccbb: depends what we mean by working. reward goes up, entropy drifts down but this may be due to the narrow nature of the task and not a serious problem. then eventually we hit the logprob nan issue which crashes the run, which i'm still not sure of the cause of (top-1 doesn't fix it, logs aren't super informative)

scripts/test_env.py
willccbb: yeah, nice to be able to see the defaults, but can remove

src/prime_rl/environments/registry.py
willccbb: reworking evals during training will be a separate PR but all envs can natively support an eval dataset, which is used by default if you call `env.evaluate()`, with the train set as a fallback if no eval dataset is present

src/prime_rl/orchestrator/orchestrator.py
willccbb: think it was a rebase snag, yeah, oops

configs/inference/multiturn_wordle/14b.toml
willccbb: changed the chat template to not remove the <think> section in multi-turn. model is same

src/prime_rl/environments/registry.py
willccbb: sure, can remove kwargs from all these

src/prime_rl/environments/registry.py
willccbb: proposal is:
- most envs for most projects will live on the Hub
- verifiers will have an included mini-registry for examples and official implementations of some standard tasks (e.g popular evals)
- there will be a flow for optionally using a local registry in a project (for dev purposes, offline use, or just to be self-contained). exact shape of this is somewhat in flux; optional whether or not we want to use this in prime-rl, but probably makes sense while on the migration path towards pihub

uv.lock
willccbb: fixed

src/prime_rl/environments/registry.py
willccbb: done

scripts/test_env.py
willccbb: removed

README.md
willccbb: yeah, rebase got a bit messy

willccbb:
@mikasenghaas for token processing, added a `process_chat_format_vllm` method to Environment which handles the accumulation of new tokens, always using the ground truth tokens from assistant generations + just injecting the tool call messages in between (which have clean starts/ends thanks to chat template, assert never gets triggered in my runs)

https://github.com/willccbb/verifiers/blob/c8891902297ea4fd1767843183a6bf4952e8612c/verifiers/envs/environment.py#L714

pyproject.toml
samsja: I guess that's if fine for now

configs/trainer/wordle/1b.toml
```diff
@@ -0,0 +1,13 @@
+
+[monitor.wandb]
+project = "wordle"
+name = "1b-debug"
+
+[model]
+name = "willcb/Qwen3-1.7B-Wordle"
+ac = true
```
samsja: I don't think that we need activation checkpointing for the 1b

configs/orchestrator/wordle/1b.toml
```diff
@@ -0,0 +1,25 @@
+max_steps = 500
+batch_size = 1024
+micro_batch_size = 2
```
samsja:
```suggestion
micro_batch_size = 1
```
maybe this should remove the need for activation checkpointing ?

configs/trainer/wordle/1b.toml
willccbb: any reason not to use it? it let me bump batch size per device

configs/orchestrator/wordle/1b.toml
willccbb: reverted + removed AC

scripts/test_env.py
mikasenghaas: yea in general we try to avoid plain prints. typically a sign in pr review that somebody forgot to delete a debug print lol

src/prime_rl/environments/registry.py
mikasenghaas: hmm, so you actually see a long-term future of all three coexisting? i see your arguments but this does seem like it can get unnessarily complex. at the very least there should be one clear default. 

also, i would argue that pi-hub should be the general source of truth (upstream) and then verifiers can have a local subset for examples/ evals etc. and prime-rl may or may not have that too. this model already seems a lot simpler

configs/inference/multiturn_wordle/14b.toml
mikasenghaas: the think section right?

mikasenghaas:
> @mikasenghaas for token processing, added a `process_chat_format_vllm` method to Environment which handles the accumulation of new tokens, always using the ground truth tokens from assistant generations + just injecting the tool call messages in between (which have clean starts/ends thanks to chat template, assert never gets triggered in my runs)
> 
> https://github.com/willccbb/verifiers/blob/c8891902297ea4fd1767843183a6bf4952e8612c/verifiers/envs/environment.py#L714

nice!

src/prime_rl/orchestrator/config.py
```diff
@@ -216,17 +216,24 @@ class OrchestratorConfig(BaseSettings):
...
+    ] = False
 
-    mask_env_responses: Annotated[
+    zero_truncated_completions: Annotated[
         bool,
         Field(
-            description="Whether to mask environment responses from the loss.",
+            description="Whether to zero-truncate truncated completions from the loss.",
         ),
     ] = True
```
mikasenghaas, line 238: im not sure i understand this arg? what's the difference to `mask_truncated_completions`? could probably be made clearer in description

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -206,13 +207,23 @@ async def orchestrate(config: OrchestratorConfig):
         generate_completions_time = time.time() - generate_completions_start_time
 
         # TODO: Switch parsing prompt+completion tokens/ masks to vf_env.process_env_results once it supports parsing directly from vLLM. For now, this only works for single-turn output results
```
mikasenghaas, line 209: remove this TODO

configs/inference/multiturn_wordle/14b.toml
willccbb: ah yes

src/prime_rl/environments/registry.py
willccbb: i think all serve a purpose, yeah -- thinking about cases where users are building on top of both verifiers + prime-rl, prototyping new envs, or otherwise using verifiers enviroments as part of a larger project (e.g. an eval harness, synth data pipeline, etc)

in that case, you want to be able to run `vf-eval` and point at your dev environments before having to push them to pihub, and having a local registry is super useful there. it's where people try things out before sharing, similarly to how you can save a HF dataset/model locally in your project folder

less certain that a core suite needs to live in the verifiers repo long-term, though for now I think it makes sense as a home for the environment examples which are currently in that repo but shouldn't be a core library component

PR #598, opened 2025-07-15
Log rubric reward functions

mikasenghaas:
This PR adds logging for individual rubric rewards (math reward, format reward) to W&B.

TODO:
- [ ] Make it work with composed environments (e.g. math and code rewards)
- [ ] Log in completions table
- [ ] Add debug logs on orchestrator for these too


mikasenghaas:
will get back to this later

PR #599, opened 2025-07-15
add back loss scale option

samsja:


mikasenghaas (approved):
looks good, config description is not accurate tho

mikasenghaas (approved):
looks good, config description is not accurate tho

PR #600, opened 2025-07-15
fix group loading

justusmattern27:


justusmattern27:
bug in the current version of prime-rl: we're not doing proper per-group normalization because of the way we're duplicating our prompts. We're using a sliding window to get our groups for advantage computation but to duplicate our prompts we're doing [problem_1, problem_2]*num_rollouts, which gives us [problem_1, problem_2, problem_1, problem_2] instead of [problem_1, problem_1, problem_2, problem_2].

It's plausible that basic RL runs still succeed since we're basically just normalizing with an approximated per-batch baseline instead of a per-group baseline (other RL algorithms do this too), but it's not correct GRPO.

samsja (approved):
lgtm lol

PR #602, opened 2025-07-16
Better Wandb logging

faresobeid:
Removed logging of total tokens, and duplicate logging in both orchestrator and trainer
Added logging of logprob error (recompute vs inference logprobs), solve_none, solve_all, ...

### DIFF
--- a/src/prime_rl/environments/registry.py
+++ b/src/prime_rl/environments/registry.py
@@ -39,37 +39,6 @@ Provide the final numerical answer inside \\boxed{{...}}."""
     vf_env = vf.SingleTurnEnv(dataset=dataset, system_prompt=system_prompt, parser=parser, rubric=rubric)
     return vf_env
 
-def load_skywork_math_environment(env_args: dict = {}) -> Environment:
-    import json
-
-    from prime_rl.orchestrator.genesys.math import compute_math_reward
-
-    train_dataset = load_dataset("PrimeIntellect/Skywork-OR1-RL-Data-v1-math-prime-rl-format", split="train").map(
-        lambda x: {"question": x["prompt"], "info": json.loads(x["verification_info"]), "task": "simple-math"}
-    )
-    solve_rate_field = env_args.get("solve_rate_field", None)
-    if solve_rate_field is not None:
-        min_solve_rate = env_args.get("min_solve_rate", None)
-        max_solve_rate = env_args.get("max_solve_rate", None)
-        if min_solve_rate is not None:
-            train_dataset = train_dataset.filter(lambda x: x[solve_rate_field] >= min_solve_rate)
-        if max_solve_rate is not None:
-            train_dataset = train_dataset.filter(lambda x: x[solve_rate_field] <= max_solve_rate)
-    train_dataset = train_dataset.remove_columns(["prompt", "verification_info"])
-
-    def correct_answer_reward_func(completion, info, **kwargs) -> float:
-        completion_text = completion[-1]["content"]
-        return compute_math_reward(completion_text, info)
-
-    rubric = vf.Rubric(
-        funcs=[
-            correct_answer_reward_func,
-        ],
-        weights=[1.0],
-    )
-
-    vf_env = vf.SingleTurnEnv(dataset=train_dataset, rubric=rubric)
-    return vf_env
 
 def load_intellect_math_environment(env_args: dict = {}) -> Environment:
     import json
@@ -501,7 +470,6 @@ REGISTRY = {
     "reverse-text": load_reverse_environment,
     "hendrycks-math": load_hendrycks_math_environment,
     "intellect-math": load_intellect_math_environment,
-    "skywork-math": load_skywork_math_environment,
     "unscramble": load_unscramble_environment,
     "ascii-tree": load_ascii_tree_environment,
     "pydantic-adherence": load_pydantic_adherence_environment,
--- a/src/prime_rl/environments/registry.py
+++ b/src/prime_rl/environments/registry.py
@@ -39,6 +39,37 @@ Provide the final numerical answer inside \\boxed{{...}}."""
     vf_env = vf.SingleTurnEnv(dataset=dataset, system_prompt=system_prompt, parser=parser, rubric=rubric)
     return vf_env
 
+def load_skywork_math_environment(env_args: dict = {}) -> Environment:
+    import json
+
+    from prime_rl.orchestrator.genesys.math import compute_math_reward
+
+    train_dataset = load_dataset("PrimeIntellect/Skywork-OR1-RL-Data-v1-math-prime-rl-format", split="train").map(
+        lambda x: {"question": x["prompt"], "info": json.loads(x["verification_info"]), "task": "simple-math"}
+    )
+    solve_rate_field = env_args.get("solve_rate_field", None)
+    if solve_rate_field is not None:
+        min_solve_rate = env_args.get("min_solve_rate", None)
+        max_solve_rate = env_args.get("max_solve_rate", None)
+        if min_solve_rate is not None:
+            train_dataset = train_dataset.filter(lambda x: x[solve_rate_field] >= min_solve_rate)
+        if max_solve_rate is not None:
+            train_dataset = train_dataset.filter(lambda x: x[solve_rate_field] <= max_solve_rate)
+    train_dataset = train_dataset.remove_columns(["prompt", "verification_info"])
+
+    def correct_answer_reward_func(completion, info, **kwargs) -> float:
+        completion_text = completion[-1]["content"]
+        return compute_math_reward(completion_text, info)
+
+    rubric = vf.Rubric(
+        funcs=[
+            correct_answer_reward_func,
+        ],
+        weights=[1.0],
+    )
+
+    vf_env = vf.SingleTurnEnv(dataset=train_dataset, rubric=rubric)
+    return vf_env
 
 def load_intellect_math_environment(env_args: dict = {}) -> Environment:
     import json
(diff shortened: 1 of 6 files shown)


src/prime_rl/trainer/train.py
```diff
@@ -189,8 +189,18 @@ def train(config: TrainerConfig):
...
 
-                    logprobs = compute_logprobs(logprob_model, input_ids, position_ids, temperature)
-                    micro_batch["logprobs"] = logprobs.to("cpu")
+                    recomputed_logprobs = compute_logprobs(logprob_model, input_ids, position_ids, temperature)
+
+                    original_logprobs = micro_batch["logprobs"].to(recomputed_logprobs.device)
+
+                    diff_mask = micro_batch["loss_mask"][:, 1:].to(recomputed_logprobs.device)
+
+                    abs_diff = (recomputed_logprobs - original_logprobs).abs()
```
samsja:
```suggestion
                    abs_diff_recomputed_logprobs = (recomputed_logprobs - original_logprobs).abs()
```

src/prime_rl/trainer/train.py
```diff
@@ -189,8 +189,18 @@ def train(config: TrainerConfig):
...
+                    abs_diff_sum = (torch.exp(abs_diff) * diff_mask).sum()
```
samsja:
```suggestion
                    abs_diff = 
                    abs_diff_recomputed_logprobs_sum = (torch.exp(recomputed_logprobs - original_logprobs).abs()) * diff_mask).sum()
```

src/prime_rl/trainer/train.py
```diff
@@ -248,6 +258,10 @@ def train(config: TrainerConfig):
             loss_metrics["loss/loss"] += loss.detach().clone().float()
             loss_metrics["loss/entropy"] += entropy.detach().clone().float()
             loss_metrics["loss/importance_ratio"] += importance_ratio.detach().clone().float()
+            loss_metrics["loss/clipped_ratio"] += clip_token_count.detach().clone().float()
+
+            abs_diff_sum = micro_batch.get("abs_logprob_diff_sum", torch.tensor(0.0))
```
samsja:
```suggestion
            abs_diff_recomputed_logprobs_sum = micro_batch.get("abs_logprob_diff_sum", torch.tensor(0.0))
```

samsja (commented):
looks good, I added couple of comments, generaly its better to have longer variable name that are more descritive

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -215,7 +215,7 @@ async def orchestrate(config: OrchestratorConfig):
         rewards = outputs["reward"]  # TODO: Align naming between prime-rl <> verifiers
 
         # Compute advantages
-        advantages = compute_advantages(rewards, config.rollouts_per_prompt)
```
mikasenghaas, line 218: im not a huge fan of returning logging only metrics as a tuple from a function (we have the same for the loss funcs which i think we should also change), this seems like it can quickly become convoluted, but it's more code style so i can change later

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -285,17 +287,20 @@ async def orchestrate(config: OrchestratorConfig):
...
         # Log rewards metrics to monitor
         reward_metrics = {
             "reward/reward": np.mean(rewards),
-            "reward/reward_std": np.std(rewards),
-            "reward/advantage": np.mean(advantages),
-            "reward/advantage_std": np.std(advantages),
-            "reward/advantage_zero_ratio": np.mean(np.array(advantages) == 0.0),
+            "reward/solve_none_ratio": solve_none_ratio,
+            "reward/solve_all_ratio": solve_all_ratio,
+            "reward/effective_batch_size_ratio": effective_batch_size_ratio,
```
mikasenghaas: i think we can drop the ratio suffix here. by default, we always compute ratios over the batch. otherwise, it would also be the `reward_ratio` etc.

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -230,6 +230,9 @@ async def orchestrate(config: OrchestratorConfig):
         throughput = num_tokens / (generate_completions_time)
         avg_seq_length = num_tokens / config.batch_size
 
+        seq_lengths = np.array([len(p) + len(c) for p, c in zip(prompt_tokens, completion_tokens)])
+        problem_avg_seqlens = seq_lengths.reshape(-1, config.rollouts_per_prompt).mean(axis=-1)
+
```
mikasenghaas: i like it. it's more general but then let's remove the `avg_seq_lengh` computation from before and compute it from the seq_lengths array as `avg_seq_length = seq_lengths.mean()`

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -273,7 +276,6 @@ async def orchestrate(config: OrchestratorConfig):
 
         # Log progress metrics to monitor
         progress_metrics = {
-            "progress/orchestrator/total_tokens": progress.total_tokens,
```
mikasenghaas, line 276: oh okay so the point of basically all progress metrics is that we might want to use either of these as our "x-axis". the default is obviously the step, but it might be valuable to also view progress as a function of samples or tokens to determine sample/ compute efficiency of a run, so i think we should keep this in

src/prime_rl/trainer/train.py
```diff
@@ -248,6 +248,8 @@ def train(config: TrainerConfig):
             loss_metrics["loss/loss"] += loss.detach().clone().float()
             loss_metrics["loss/entropy"] += entropy.detach().clone().float()
             loss_metrics["loss/importance_ratio"] += importance_ratio.detach().clone().float()
+            loss_metrics["loss/clipped_ratio"] += clip_token_count.detach().clone().float()
```
mikasenghaas: same here, i would remove the ratio. i think for importance ratio it makes sense to keep tho bc its actually the logprob ratio right?

src/prime_rl/trainer/loss.py
```diff
@@ -91,14 +91,14 @@ def grpo_loss_clip(
...
+    clip_token_count = _masked_sum(is_clipped, loss_mask)
 
     if highest_entropy_percentage < 1.0:
         loss_mask = highest_entropy_mask(logits, loss_mask, highest_entropy_percentage)
 
     loss = _masked_sum(per_token_loss, loss_mask)
-
-    return loss, clip_ratio_tensor
+    ratio = _masked_sum(coef_2, loss_mask)
+    return loss, ratio, clip_token_count
```
mikasenghaas: is the ratio returned here the same as for ratio loss?

src/prime_rl/trainer/train.py
```diff
@@ -319,15 +321,6 @@ def train(config: TrainerConfig):
...
         logger.success(step_message)
 
-        # Log progress metrics
-        progress_metrics = {
-            "progress/train/total_tokens": progress.total_tokens,
-            "progress/train/total_samples": progress.total_samples,
-            "progress/train/step": progress.step,  # Shared W&B axis
-            "step": progress.step,
-        }
-        monitor.log(progress_metrics)
```
mikasenghaas, line 329: okay, so after rethinking, i think this is actually fine to do. my concern before was that the orchestrator and trainer both have their own step definition 0,1,2,... which "point to" different model checkpoints. at async level 2 the orchestrator's steps 0,1,2 are from model checkpoint 0 (base model) and the values are getting logged like this. but the orchestrator also logs the the ckpt step which is correctly at 0,0,0 for the first 2 steps. if not both the trainer and the orchestrator log this key we cannot use it as a shared axis i believe. i think we can switch the axis just for the orchestrator plots tho and then all is aligned so should be good

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -273,7 +276,6 @@ async def orchestrate(config: OrchestratorConfig):
...
             "progress/orchestrator/total_samples": progress.total_samples,
             "progress/orchestrator/epoch": progress.epoch,
```
mikasenghaas: lets delete the orchestrator from these then tho
```suggestion
            "progress/total_samples": progress.total_samples,
            "progress/epoch": progress.epoch,
```

mikasenghaas (approved):

src/prime_rl/trainer/train.py
mikasenghaas: *the ratio suffix

src/prime_rl/trainer/loss.py
faresobeid: wdym?

src/prime_rl/environments/registry.py
```diff
@@ -39,6 +39,37 @@ def correct_answer_reward_func(completion, answer, **kwargs) -> float:
...
+
+    rubric = vf.Rubric(
+        funcs=[
+            correct_answer_reward_func,
+        ],
+        weights=[1.0],
+    )
+
+    vf_env = vf.SingleTurnEnv(dataset=train_dataset, rubric=rubric)
+    return vf_env
```
mikasenghaas: this is a separate pr?

src/prime_rl/trainer/loss.py
mikasenghaas: both the clipped and ratio loss return a ratio, is it equivalent in meaning?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -230,6 +230,9 @@ async def orchestrate(config: OrchestratorConfig):
         throughput = num_tokens / (generate_completions_time)
         avg_seq_length = num_tokens / config.batch_size
```
mikasenghaas: delete this?

src/prime_rl/trainer/loss.py
samsja: Yes it is

src/prime_rl/environments/registry.py
faresobeid: ya mistake, fixed it

samsja (approved):
Lfgtm 🔥

PR #606, opened 2025-07-18
Fix logprob packing

mikasenghaas:
This is what I think is the proper fix to a shape mismatch problem arising from packing logprobs. The trainer computes logprobs with tensor shape `(B, L-1, V)`, but when packing k samples into a batch the orchestratror writes a tensor of shape `(B, L-k, V)`. This PR changes the orchestrator batch packing to assert that for each sample from the orchestrator we have `len(input_ids) == len(logprobs)` and `logprobs[0] == 0`. This maintains the simpler equivariance that for every index `i` `logprob[i]` is the logprob associated to the token ID `input_ids[i]`.

### DIFF
--- a/src/prime_rl/orchestrator/data.py
+++ b/src/prime_rl/orchestrator/data.py
@@ -14,7 +14,7 @@ class Sample(TypedDict):
     position_ids: Int[Tensor, "seq"]
     loss_mask: Int[Tensor, "seq"]
     advantages: Float[Tensor, "seq"]
-    logprobs: Float[Tensor, "seq_minus_1"]
+    logprobs: Float[Tensor, "seq"]
 
 
 def prepare_sample(
--- a/src/prime_rl/orchestrator/data.py
+++ b/src/prime_rl/orchestrator/data.py
@@ -170,7 +170,6 @@ def packed_samples_into_micro_bs(samples: list[Sample], max_seq_len: int) -> lis
             bin_len = sum(len(s["input_ids"]) for s in bin_content)
             # Check if sequence fits in this bin
             if bin_len + len(sample["input_ids"]) <= max_seq_len:
-                sample["logprobs"] = torch.cat([torch.tensor([0.0]), sample["logprobs"]])
                 micro_batches[bin_idx].append(sample)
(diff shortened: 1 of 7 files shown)


src/prime_rl/orchestrator/data.py
```diff
@@ -44,7 +44,7 @@ def prepare_sample(
     # Prepare input_ids, loss_mask, position_ids, logprobs, and advantages
     input_ids = torch.cat([prompt_token_ids, completion_token_ids]).long()
     loss_mask = torch.cat([prompt_token_mask, completion_token_mask]).long()
-    logprobs = torch.cat([torch.zeros(len(prompt_token_ids) - 1), torch.tensor(completion_logprobs)]).float()
+    logprobs = torch.cat([torch.zeros(len(prompt_token_ids)), torch.tensor(completion_logprobs)]).float()
```
samsja, line 47: remind me why we remove the -1 here

mikasenghaas:
**Reverse text**

Screenshot 2025-07-19 at 3 26 44 PM

mikasenghaas:
**Hendrycks Math**

Screenshot 2025-07-20 at 12 11 08 AM

Ran out of disk after 450 steps

src/prime_rl/trainer/model.py
```diff
@@ -79,3 +80,69 @@ def forward(
...
+    zeros = torch.zeros(B, 1, V, device=logits.device, dtype=logits.dtype)  # (B, 1, V)
+    logits = torch.cat([zeros, logits], dim=1)  # (B, L, V)
+    return logits
+
+
+@jaxtyped(typechecker=typechecker)
+def selective_log_softmax(
+    logits: Float[Tensor, "batch seq vocab"], index: Int[Tensor, "batch seq"]
+) -> Float[Tensor, "batch seq"]:
+    """
```
samsja: not clear to me why we moved this to model.py, seems very much independent from model ?

mikasenghaas: moved back

samsja (approved):
lgtm

PR #608, opened 2025-07-19
Add `flash_infer`

mikasenghaas:
Adds `flashinfer-python` for optimized attention kernels for vLLM.

## Performance

Flash inference generally seems to be a bit faster across DP setting, but not by a significant margin. 

*NB: Slight deviations are expected because the sampling is not seeded which may lead to different generations lengths which in turn affect throughput. Also comparison might not be 100% fair because done on different machines.*

Screenshot 2025-07-19 at 6 44 57 PM

*W&B View: https://wandb.ai/primeintellect/intellect-math-bench/workspace?nw=57mfkkyig5g*

## Correctness

Matches baseline run on `hendrycks-math` on main for ~450 steps (run out of disk space for ckpts after)

Screenshot 2025-07-19 at 11 40 16 PM

### DIFF
--- a/.github/workflows/gpu_tests.yaml
+++ b/.github/workflows/gpu_tests.yaml
@@ -27,7 +27,7 @@ jobs:
           enable-cache: true
           cache-dependency-glob: "uv.lock"
       - name: Install dependencies
-        run: uv sync && uv sync --extra fa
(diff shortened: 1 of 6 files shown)


samsja (approved):
lgtm

PR #610, opened 2025-07-19
Add custom advantage computation support + negative clipping to zero

kalomaze:
This PR adds an optional `advantage_style` field that defaults to `drgrpo`.
Included is @justusmattern27's proposal for a variant where negative advantages are clipped to zero (`drgrpo-negclipped`).

image

Shown is the `reverse-text` config on `willcb/Qwen2.5-0.5B-Reverse-SFT`.
The only modifications made were in setting `max_norm = 0.2`, `lr = 5e-6`.

Due to the seemingly more stable gradients of `drgrpo-negclipped`, I'm able to successfully continue training on this task for ~5000 steps without any observed collapse. I use more aggressive clipping (`max_norm = 0.00001`) and a larger learning rate to compensate (~1e-5 to ~4e-5 all seemed functional at this threshold for the model).

image

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -25,10 +25,9 @@ from prime_rl.orchestrator.client import (
 from prime_rl.orchestrator.config import OrchestratorConfig
 from prime_rl.orchestrator.data import prepare_batch
 from prime_rl.orchestrator.logger import setup_logger
+from prime_rl.orchestrator.advantage import compute_advantages
 from prime_rl.orchestrator.utils import (
     process_env_results,
-    compute_advantages,
-    compute_negclipped_advantages,
     wait_for_weight_checkpoint,
     print_benchmark,
 )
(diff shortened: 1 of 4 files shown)


samsja (approved):
Lgtm, gonna let @mikasenghaas and @justusmattern27 review before merging

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -214,10 +215,24 @@ async def orchestrate(config: OrchestratorConfig):
...
+        advantage_fn = advantage_functions.get(config.advantage_style)
+        if advantage_fn is None:
+            raise ValueError(
+                f"Unknown advantage style: {config.advantage_style}. Available options: {list(advantage_functions.keys())}"
+            )
+
+        advantages, solve_none_ratio, solve_all_ratio, effective_batch_size_ratio = advantage_fn(
+            rewards, config.rollouts_per_prompt
+        )
+
```
mikasenghaas: the registry and dispatching logic to diff adv style computations should not in our main script, but somewhere outside. let me propose a quick code style refactor that we can merge, will be logically equivalent tho

mikasenghaas (commented):
very nice! im curious how stable our current training setup (default lr and max norm) is for multiple K of steps. basically: is the neg adv clipping necessary only under aggressive learning rate and grad norm clipping or generally desirable? do you already have a run for this? if not, i think it would be valuable to do one

kalomaze:
> is the neg adv clipping necessary only under aggressive learning rate and grad norm clipping or generally desirable?

Needs more thorough testing on larger models before I'm confident, but in the first run you can see a rising grad norm on ~5e-6 with standard Dr. GRPO.
After this is merged, my plan is to properly implement a sequential sweep utility so I can compare different advantage calculations + other hparameters, and I also plan to fix parallel sweeps for single node training
I was having strange issues with wandb logging seemingly being inaccurate when you try to do two trains at the same time. Perhaps duplicating the venv + repo would work to prevent that?

mikasenghaas:
oh i see. have you seen the section in the readme on running multiple experiments in parallel on one node? this should (hopefully) solve your logging issues. if not, let me know and describe what issues your having.

also, how do you envision doing sequential sweeps? chained cli commands? bash script? skypilot? i think this is generally interesting for the research team so worthwhile discussing

kalomaze:
> oh i see. have you seen the section in the readme on running multiple experiments in parallel on one node? this should (hopefully) solve your logging issues. if not, let me know and describe what issues your having.
> 
> also, how do you envision doing sequential sweeps? chained cli commands? bash script? skypilot? i think this is generally interesting for the research team so worthwhile discussing

skypilot would be ideal for longer seqlen runs (or on runs for bigger models in general) for sure.
sequential is fast enough for demoing the basic sweep methodology tho, and having a basic CLI implementation of that that is then extended to have (optional) i.e. skypilot subprocess support would be best of both worlds imo

samsja (approved):
lgtm

liranringel:
@kalomaze  should it subtract the mean only after the clipping?

PR #611, opened 2025-07-19
Refactor advantage type computation

mikasenghaas:
Code style refactor for #610. Includes:
- Defines advantage type factory in `prime_rl.orchestrator.advantage`. Each variant can be implemented as a simple function with (jaxtype-enforced) type `Float[Tensor["group"]] -> Float[Tensor["group"]]`, i.e. it is only responsible for computing the advantage for each rollout of a group
- The logic for nesting the per problem rewards, iteratively calling the correct advantage type function and collecting statistics such as `solve_none` and `solve_all` is ony handled *once* in `compute_advantages` which is 
- The config arg changed to `advantage_type` and is tied to the registry via a literal `AdvantageType`

Also fixes a bug from #602 where the `solve_none` and `solve_all` ratios are always zero because of a wrong type being passed.

### DIFF
--- a/src/prime_rl/orchestrator/advantage.py
+++ b/src/prime_rl/orchestrator/advantage.py
@@ -1,0 +1,47 @@
+from typing import Callable, Literal
+
+import torch
+from beartype import beartype as typechecker
+from jaxtyping import Float, jaxtyped
+from torch import Tensor
+
+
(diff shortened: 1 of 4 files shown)


samsja (approved):
Lgtm

PR #612, opened 2025-07-19
Checkpoint less frequently on `hendrycks_math`

mikasenghaas:
I have run out of disk way too many times and it's not worth it on `hendrycks-math`


samsja (approved):
lgtm

PR #614, opened 2025-07-20
Add skywork configs

faresobeid:
Added initial Skywork Configs for ablations and replicating Skywork OR1


src/prime_rl/environments/registry.py
```diff
@@ -39,6 +39,38 @@ def correct_answer_reward_func(completion, answer, **kwargs) -> float:
...
     return vf_env
 
+def load_skywork_math_environment(env_args: dict = {}) -> Environment:
+    import json
+
+    from prime_rl.orchestrator.genesys.math import compute_math_reward
+
+    train_dataset = load_dataset("PrimeIntellect/Skywork-OR1-RL-Data-v1-math-prime-rl-format", split="train").map(
+        lambda x: {"question": x["prompt"], "info": json.loads(x["verification_info"]), "task": "simple-math"}
+    )
```
mikasenghaas, line 69: lol is ruff formatter not complaining here?

src/prime_rl/environments/registry.py
```diff
@@ -39,6 +39,38 @@ def correct_answer_reward_func(completion, answer, **kwargs) -> float:
...
+    solve_rate_field = env_args.get("solve_rate_field", None)
+    if solve_rate_field is not None:
+        min_solve_rate = env_args.get("min_solve_rate", None)
+        max_solve_rate = env_args.get("max_solve_rate", None)
+        if min_solve_rate is not None:
+            train_dataset = train_dataset.filter(lambda x: x[solve_rate_field] >= min_solve_rate)
+        if max_solve_rate is not None:
+            train_dataset = train_dataset.filter(lambda x: x[solve_rate_field] <= max_solve_rate)
```
mikasenghaas, line 77: we might wanna consider reusing this logic across envs somehow. but likely question for later once we move to verifiers/ pi-hub registry, cc/ @willccbb

src/prime_rl/trainer/config.py
```diff
@@ -205,6 +205,14 @@ class TrainerConfig(BaseSettings):
         ),
     ] = True
 
+    # NEW CONFIG FLAG
```
mikasenghaas: delete this

src/prime_rl/trainer/train.py
```diff
@@ -166,9 +167,22 @@ def train(config: TrainerConfig):
...
+        # Optionally replace zero-advantage micro batches
+        if config.replace_zero_advantage:
+            non_zero_batches = [mb for mb in micro_batches if mb["advantages"].abs().sum().item() != 0]
+            for idx, mb in enumerate(micro_batches):
+                if mb["advantages"].abs().sum().item() == 0:
+                    replacement = random.choice(non_zero_batches)
+                    micro_batches[idx] = {
+                        k: (v.clone() if torch.is_tensor(v) else deepcopy(v))
+                        for k, v in replacement.items()
+                    }
```
mikasenghaas: wait this seems a bit crazy, can you explain?

src/prime_rl/trainer/train.py
```diff
@@ -217,7 +231,12 @@ def train(config: TrainerConfig):
         batch_size = micro_batch_size * num_micro_batches
 
         # Normalize by the number of unmasked tokens in the batch (per-batch length normalization)
-        loss_scale = sum(micro_batch["loss_mask"].sum() for micro_batch in micro_batches)
+        # Exclude zero advantage micro_batches
+        loss_scale = sum(
+            micro_batch["loss_mask"].sum().item()
+            for micro_batch in micro_batches
+            if micro_batch["advantages"].abs().sum().item() != 0
+        )
```
mikasenghaas: shouldn't we exclude zero advantage samples not micro batches? the latter is a much rarer event when we do packing

mikasenghaas (commented):
nice! is the only difference betwen stage 1 and stage 2 the seq_len?

faresobeid:
> nice! is the only difference betwen stage 1 and stage 2 the seq_len?

No its also the rollouts per prompt and batch size (as well as the dataset but not implemented yet)

src/prime_rl/trainer/train.py
faresobeid: It's the idea from 
https://honorable-payment-890.notion.site/POLARIS-A-POst-training-recipe-for-scaling-reinforcement-Learning-on-Advanced-ReasonIng-modelS-1dfa954ff7c38094923ec7772bf447a1
https://arxiv.org/abs/2507.02592
and some others.
Basically replace zero advantage samples by sampling from the non zero advantage samples in the batch to duplicate. We discussed this and don't see why it works but its been done by enough reliable groups that I think it's worth a try

configs/inference/skywork_math/32b/stage2.toml
```diff
@@ -0,0 +1,6 @@
+[model]
+name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-32B"
+max_model_len = 32768
+
+[parallel]
+dp = 6
```
samsja:
```suggestion
[parallel]
tp = 4
dp = 2

```

src/prime_rl/environments/registry.py
samsja: agree

src/prime_rl/trainer/train.py
mikasenghaas: ah i see but then this should also be at the sample not the micro batch level right? and thus, likely live on the orchestrator

samsja: hmm yeah I agree with mika here, it seems weird to look the micro batch level, a micro batch could have two samples one with 0 adv and one with non 0. The issue happen both in padding and packing

src/prime_rl/trainer/train.py
samsja: we should do either token or sample level calculation here, this would break grad acc right now becuase micro_batch["advantages"].abs().sum() == 0 only if all the sample into it have 0 adv, this would be unlikely for large micro batch and very likely for micro_batch =1

PR #628, opened 2025-07-24
Log distributions as W&B table

mikasenghaas:
Adds functionality for logging distributions (e.g. reward/ advantage distribution) as W&B tables. Adds a call from the orchestrator to `monitor.log_distribution` which will log a W&B table `distributions` (similar logic-wise to the sample logging):
- `step`: The logging step
- `rewards`: Histogram of all rewards (irrespective of group size)
- `advantages`: Histogram of all advantages (irrespective of group size)
- `problem_rewards`: Histogram of the mean reward per problem
- `problem_advantages`: Histogram of the mean advantage per problem (always 0 for `drgrpo`)

Screenshot 2025-07-23 at 5 37 41 PM

## Misc Changes

- Figured out how to log incremental (not immutable) tables with rows coming in on the fly. However, it has a max row limit (for whatever reason), so we also still log a final table at the end of the run
- Changed the config key `--monitor.wandb.log-samples` to `--monitor.wandb.log-extras` which boolean flags for logging samples and distributions, i.e. `--monitor.wandb.log-extras.samples` and `--monitor.wandb.log-extras.distributions`. This is mainly so that both of these tables are logged at the same step interval
- The samples table was renamed from `completions` to `samples`

### DIFF
--- a/configs/orchestrator/hendrycks_math/14b.toml
+++ b/configs/orchestrator/hendrycks_math/14b.toml
@@ -13,7 +13,7 @@ name = "willcb/DeepSeek-R1-Distill-Qwen-14B"
 [environment]
 id = "hendrycks-math"
 
-[monitor.wandb.log_samples]
+[monitor.wandb.log_extras]
 interval = 10
 
 [eval]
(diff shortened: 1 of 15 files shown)


samsja (approved):
lgtm

PR #633, opened 2025-07-24
Log correct start command

mikasenghaas:
This PR ensures that if a run is started from the `rl.py` entrypoint, we log the `uv run rl ...` start command to the trainer and orchestrator W&B runs. This allows us to precisely reproduce a run which was not possible before because we were logging the temporary start command which were pointing to temporary TOML files which were not logged. This PR does not affect the start commands of trainers and orchestrators started manually (i.e. not via the entrypoint).

*Check how the command is `uv run rl` on both the trainer and the orchestrator because they were started from `rl.py`*

Screenshot 2025-07-24 at 2 36 52 PM

Screenshot 2025-07-24 at 2 37 03 PM

*If we start the trainer and orchestrator separate, we log `uv run trainer ...` and `uv run orchestrator ...`, respectively.*

Screenshot 2025-07-24 at 2 39 05 PM

Screenshot 2025-07-24 at 2 39 45 PM

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -34,7 +34,7 @@ from prime_rl.orchestrator.utils import (
 )
 from prime_rl.utils.monitor import setup_monitor
 from prime_rl.utils.pydantic_config import parse_argv
-from prime_rl.utils.utils import clean_exit, to_col_format, format_num, maybe_overwrite_wandb_command
+from prime_rl.utils.utils import clean_exit, to_col_format, format_num
 
 
 @clean_exit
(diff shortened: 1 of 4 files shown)


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -44,6 +44,9 @@ async def orchestrate(config: OrchestratorConfig):
     logger = setup_logger(config.log)
     logger.info("Starting orchestrator")
 
+    # If started from RL entrypoint, overwrite start command to log the correct command to W&B
+    maybe_overwrite_wandb_command("uv run orchestrator")
```
samsja: I would hide this in the config management or wandb init, doesn't feel right to be part of the main code

samsja: Actually let's move it to wandb init, let's remove default value and just use sys.arg otherwise

PR #634, opened 2025-07-24
Improve eval suite

mikasenghaas:
- Allow setting `rollouts_per_prompts` for each benchmark to get more robust eval estimates, especially for small benchmarks like AIME (30 problem)
- Changes defaults in TOML configs (AIME-24/25: 16, MATH-500: 1)
- Rename `score` -> `avg@k`

### DIFF
--- a/configs/orchestrator/hendrycks_math/14b.toml
+++ b/configs/orchestrator/hendrycks_math/14b.toml
@@ -19,3 +19,4 @@ interval = 10
 [eval]
 interval = 50
 benchmarks = ["math500", "aime24", "aime25"]
+rollouts_per_prompt = [1, 16, 16]
--- a/configs/orchestrator/hendrycks_math/1b.toml
+++ b/configs/orchestrator/hendrycks_math/1b.toml
@@ -19,6 +19,7 @@ interval = 10
 [eval]
 interval = 50
 benchmarks = ["math500", "aime24", "aime25"]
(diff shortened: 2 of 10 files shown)


faresobeid:
maybe making the default on math-500 around 4 is better

mikasenghaas:
> maybe making the default on math-500 around 4 is better

really? then we have 4x500 prompts and we get bottlenecked by that. aime24/25 is 30x16=480 which is nicely close to 500. it seems like the math-500 plots from prev runs dont have a lot of variance anyways?

faresobeid (approved):

configs/orchestrator/hendrycks_math/14b.toml
```diff
@@ -19,3 +19,4 @@ interval = 10
 [eval]
 interval = 50
 benchmarks = ["math500", "aime24", "aime25"]
+rollouts_per_prompt = [1, 16, 16]
```
samsja, line 22: Isn't making eval way longer ? Why two times 16

mikasenghaas: the rollouts per prompt is per benchark, so math-500 = 1, aime24 = 16 and aime25 = 16. eval time should not be affected much because we do these in parallel and aime24/25 only have 30 questions each, i.e. 30*16 =480 480 requests which is roughly the same as math-500

PR #636, opened 2025-07-25
More performant buffer re-sampling 

mikasenghaas:
This PR sets more performant defaults for buffer types which do not guarantee that `sample_rollouts(n)` will actually return rollouts for `n` problems, i.e. because they have some constraint on which rollouts can be sampled (e.g. the `online-difficulty` buffer only allows problems where the average reward is in range `[min_reward, max_reward]`). 

The previous logic is: While not enough rollouts are sampled to fill the batch, sample `problems_per_batch` problems, generate rollouts and fill the batch.

Clearly, this is quite wasteful, e.g. if we are missing only one problem we request another full batch which we immediately throw away. Thus, we change the logic to:

While not enough rollouts are available to fill the batch, sample and generate rollouts for as many problems as are left, subject to a default oversampling factor of 1.5.

Ideally, the oversampling factor is set exactly right so that (a) its not too low, i.e. we almost never run the inner orchestratror loop and (b) its not too high, i.e. we generate rollouts for more completions than we can use.

From small experimentation the new defaults are more performant than the old, but the oversampling factor is a hyperparameter that depends on the likelihood of problems falling within the sampling criteria. 

Screenshot 2025-07-25 at 11 14 56 AM

For the future we can consider two more things to optimize more:
- "Adaptive oversampling factor": This is based on the observation that the sampling criteria of online difficulty filtering is harder to fulfill in earlier steps (too many hard problems) and later steps (too many easy problems) but easy to fulfill in the meantime. Ideally, we wanna oversample more on the problematic earlier and later steps but not so much in the intermediate steps
- "Reuse old rollouts": This is more of a research question: To what extend can we use/ mix rollouts from older generations. For now we use the conservative default that we cannot use them, and hence throw everything away. 

Would be curious to hear what you think @samsja @faresobeid

### DIFF
--- a/src/prime_rl/orchestrator/buffer.py
+++ b/src/prime_rl/orchestrator/buffer.py
@@ -398,7 +398,7 @@ class OnlineDifficultyBuffer(Buffer):
 
         if len(sampled_problem_ids) < n:
             self.logger.warning(
-                f"Only {len(sampled_problem_ids)} (<{n}) problems with rollouts available ({num_too_easy=}, {num_too_hard=})"
+                f"Only {len(sampled_problem_ids)} (<{n}) valid problems with rollouts available ({num_too_easy=}, {num_too_hard=})"
             )
 
         return sampled_rollouts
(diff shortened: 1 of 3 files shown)


samsja (approved):
lgtm

PR #637, opened 2025-07-25
Add skywork configs

faresobeid:
Added skywork configs to reproduce the skywork OR1 7B and 32B models (without code and data filtering from the previous stage)
Added skywork 7B config as a good testing ground for ablations

### DIFF
--- a/src/prime_rl/environments/registry.py
+++ b/src/prime_rl/environments/registry.py
@@ -152,6 +152,44 @@ def load_intellect_math_vf_environment(
     )
     return vf_env
 
+def load_skywork_math_environment(
+    solve_rate_field: str | None = None,
+    min_solve_rate: float | None = None,
+    max_solve_rate: float | None = None,
+    **kwargs,
+) -> Environment:
+    import json
+
+    from prime_rl.orchestrator.genesys.math import compute_math_reward
+
+    train_dataset = load_dataset("PrimeIntellect/Skywork-OR1-RL-Data-v1-math-prime-rl-format", split="train").map(
+        lambda x: {
+            "question": x["prompt"],
+            "info": json.loads(x["verification_info"]),
+            "task": "simple-math",
+        }
+    )
+    if solve_rate_field is not None:
+        if min_solve_rate is not None:
+            train_dataset = train_dataset.filter(lambda x: x[solve_rate_field] >= min_solve_rate)
+        if max_solve_rate is not None:
+            train_dataset = train_dataset.filter(lambda x: x[solve_rate_field] <= max_solve_rate)
+    train_dataset = train_dataset.remove_columns(["prompt", "verification_info"])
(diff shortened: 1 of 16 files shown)


src/prime_rl/environments/registry.py
```diff
@@ -152,6 +152,44 @@ def correct_answer_reward_func(completion, answer, **kwargs) -> float:
...
+) -> Environment:
+    import json
+
+    from prime_rl.orchestrator.genesys.math import compute_math_reward
+
+    train_dataset = load_dataset("PrimeIntellect/Skywork-OR1-RL-Data-v1-math-prime-rl-format", split="train").map(
+        lambda x: {
+            "question": x["prompt"],
+            "info": json.loads(x["verification_info"]),
+            "task": "simple-math",
```
mikasenghaas, line 169: should be `skywork-math`? altho i dont think this is ever used anywhere tbf

src/prime_rl/environments/registry.py
```diff
@@ -152,6 +152,44 @@ def correct_answer_reward_func(completion, answer, **kwargs) -> float:
...
 
+def load_skywork_math_environment(
+    solve_rate_field: str | None = None,
+    min_solve_rate: float | None = None,
+    max_solve_rate: float | None = None,
+    **kwargs,
+) -> Environment:
+    import json
+
+    from prime_rl.orchestrator.genesys.math import compute_math_reward
```
mikasenghaas, line 163: skywork is `math-verify` right?

mikasenghaas (approved):
lgtm, once we have final ablation results, we should have remove the intellect-math configs and replace the section in the readme with skywork ablation commands

src/prime_rl/environments/registry.py
faresobeid: ya fair wasn't sure if it was best to keep it like this as its the same task as the other math datasets

src/prime_rl/environments/registry.py
faresobeid: true will need to fix that but i ran my ablation like this. Would say this isn't a skywork specific thing, more so that we should move to math-verify entirely

faresobeid:
Initial Skywork 7B ablation results, shows some improvements although probably needs a stage 2 to really shine

Base:
- MATH-500 = 91%
- AIME-24     = 50%
- AIME-25     = 34%

Step 200:
- MATH-500 = 92%
- AIME-24     = 54%
- AIME-25     = 38%

Step 400:
- MATH-500 = 92%
- AIME-24     = 51%
- AIME-25     = 40%

Step 600:
- MATH-500 = 93%
- AIME-24     = 53%
- AIME-25     = 41%

mikasenghaas:
can you link the w&b run here as well for completeness

faresobeid:
https://wandb.ai/primeintellect/skywork_7b_ablation?nw=z82h5ukb2bc

PR #638, opened 2025-07-26
Length penalty

faresobeid:
Added a length penalty by giving the shortest rollout a bonus reward, if all rollouts for that group are correct

### DIFF
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -338,6 +338,25 @@ class OrchestratorConfig(BaseSettings):
         ),
     ] = True
 
+    # ---------------------------------------------------
+    # Reward shaping options
+    # ---------------------------------------------------
+
+    apply_shortest_correct_bonus: Annotated[
+        bool,
+        Field(
+            description="If True, add an extra reward to the shortest correct answer in fully correct rollout groups.",
+        ),
+    ] = False
+
+    shortest_correct_bonus_value: Annotated[
+        float,
+        Field(
+            ge=0,
+            description="Bonus value to add when apply_shortest_correct_bonus is enabled.",
+        ),
+    ] = 0.5
+
     # TODO(Mika): This should be automatic from the number of ZMQ connections
     num_train_workers: Annotated[
         int,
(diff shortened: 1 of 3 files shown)


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -221,8 +222,19 @@ async def orchestrate(config: OrchestratorConfig):
...
+            if config.apply_shortest_correct_bonus:
+                completion_lengths = [len(tokens) for tokens in results["completion_ids"]]
+                train_rewards = apply_shortest_correct_bonus(
+                    outputs["reward"],
+                    completion_lengths,
+                    group_size=config.rollouts_per_prompt,
+                    bonus=config.shortest_correct_bonus_value,
+                )
+            else:
+                train_rewards = outputs["reward"]
```
mikasenghaas, line 234: lets avoid if..else when we can. here, you should just need the if branch and make your function overwrite the previous rewards in the output dict, i.e. `outputs["reward"] = applye_shortest_correct_bonus(...)`

src/prime_rl/orchestrator/config.py
```diff
@@ -338,6 +338,21 @@ class OrchestratorConfig(BaseSettings):
...
+        ),
+    ] = False
+
+    shortest_correct_bonus_value: Annotated[
+        float,
+        Field(
+            ge=0,
+            description="Bonus value to add when apply_shortest_correct_bonus is enabled.",
+        ),
+    ] = 0.5
```
mikasenghaas: lets merge this into one config key `shortest_correct_bonus: float | None` where if None, nothing happens and else its a value >=0 that we use for length penalty

src/prime_rl/orchestrator/utils.py
```diff
@@ -108,6 +108,28 @@ def compute_rewards(
     return rewards
 
 
+def apply_shortest_correct_bonus(
+    rewards: list[float],
+    completion_lengths: list[int],
+    group_size: int,
```
mikasenghaas: lets call this `rollouts_per_prompt` do be consistent

mikasenghaas (commented):
logic looks good! a couple minor code style comments

src/prime_rl/orchestrator/orchestrator.py
faresobeid: issue is though that imo the actual rewards shouldn't be modified (when logging and elsewhere), so we are still aware that this group of rollouts is all correct, but just adjust the advantages

PR #639, opened 2025-07-26
Filter wrong rollouts for hard prompts 

faresobeid:
Added the ability to filter out wrong rollouts for prompts with only one correct rollout (emphasise it)


mikasenghaas (commented):
very interesting! where is this idea coming from? let's validate it on the hendrycks 5k runs and see if it does better than `drgrpo-negclipped` lol. we will only merge if we have some empirical evidence that it's good tho

PR #640, opened 2025-07-26
RLOO and OPO baselines

faresobeid:
Added support for RLOO (used by DeepSWE) and OPO (shown promising results for stability) baselines in advantage computation


samsja (approved):
lgtm

PR #643, opened 2025-07-28
Use discriminated union for loss config

mikasenghaas:
Previously, loss types were not set up as Pydantic discrimainted unions, so Pydantic would not be able to validate config variants correctly, leading to errors when trying to switch to deviate from the default config (e.g. use `clip` loss)

```
pydantic_core._pydantic_core.ValidationError: 2 validation errors for RLConfig
trainer.loss.variant.ClippingConfig
  Input should be a valid dictionary or instance of ClippingConfig [type=model_type, input_value='clip', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/model_type
trainer.loss.variant.RatioConfig
  Input should be a valid dictionary or instance of RatioConfig [type=model_type, input_value='clip', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/model_type
```

This PR slightly restructures the loss configs to correctly use Pydantic's disciminated unions to correctly type checking config variants with differing keys (e.g. `ClipLossConfig` has the epsilon params, whereas the `RatioLossConfig` does not have this). Now we can do:

Specify to run with `RatioLossConfig` (default):

```bash
uv run trainer ... --loss.type ratio
```

Run with clip loss config

```bash
uv run trainer ... --loss.type clip
```

Also adds a info log print on trainer on the loss config used

Screenshot 2025-07-28 at 1 48 32 PM

### DIFF
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -1,7 +1,7 @@
 from pathlib import Path
-from typing import Annotated, Literal, TypeAlias, Union
+from typing import Annotated, Literal, TypeAlias
 
-from pydantic import Field, model_validator
+from pydantic import BaseModel, Field, model_validator
 
 from prime_rl.utils.config import LogConfig, MultiMonitorConfig
 from prime_rl.utils.pydantic_config import BaseConfig, BaseSettings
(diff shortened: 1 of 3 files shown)


samsja (approved):
Lgtm

PR #644, opened 2025-07-28
Add multi-node DP

mikasenghaas:
> This PR is based on #696 and should not be merged before

This PR adds docs for multi-node DP inference using vLLM, following the [vLLM Internal Load Balancing](https://docs.vllm.ai/en/stable/serving/data_parallel_deployment.html) docs.

## Multi-Node DP

> ⚠️ Because we require VPN setup, multi-node inference cannot run in containerized environments. Make sure you own the nodes or run a VM.

> 🚨 In my experiments I have found multi-node DP to be quite brittle. For example, for some reason only one of my nodes was able to be the "head" node, the other direction wouldn't set up correctly. Also, for some reason the CUDA graph capture is hanging in my experiments, and the only way to resolve it was to enforce eager mode..

Setup VPN using `tailscale` . Then, find the IP from the `tailscale0` interface

```bash
ip a
```

Configure GLOO and NCCL to use `tailscale` network interface

```bash
export GLOO_SOCKET_IFNAME=tailscale0
export NCCL_SOCKET_IFNAME=tailscale0
```

Ensure nothing is running on port `13345`

```bash
fuser -k 13345/tcp
```

Start multi-node DP

```bash
# Node 0  (With IP 100.107.120.57)
uv run inference \
	--data-parallel-size 2 \
	--data-parallel-size-local 1 \
	--data-parallel-address 100.107.120.57 \
	--data-parallel-rpc-port 13345 \
        --enforce-eager
```

```bash
# Node 1
uv run inference \
	--data-parallel-size 2 \
	--data-parallel-size-local 1 \
	--data-parallel-address 100.107.120.57 \
	--data-parallel-rpc-port 13345 \
	--data-parallel-start-rank 1 \
	--headless \
        --enforce-eager
```

## Co-located Multi-Node DP (Testing)

```bash
CUDA_VISIBLE_DEVICES=0 uv run inference \
	--data-parallel-size  2 \
	--data-parallel-size-local 1 \
	--data-parallel-address localhost \
	--data-parallel-rpc-port 13345
```

```bash
	--data-parallel-size  2 \
	--data-parallel-size-local 1 \
	--data-parallel-address localhost \
	--data-parallel-rpc-port 13345 \
	--data-parallel-start-rank 1 \
	--headless
```

---

**GitHub Issue**: #752 
**Linear Issue**: Resolves PRIMERL-26


samsja:
couple of question:

* why do we need tailscale ? Multi node should be able to talk to each other via private IP
* why do we use enforce_eager in the command ? Is it only to make the boottime faster for testing ?

JannikSt:
> couple of question:
> 
> * why do we need tailscale ? Multi node should be able to talk to each other via private IP
> * why do we use enforce_eager in the command ? Is it only to make the boottime faster for testing ?

+1 on the tailscale question - on co-located multi node machines tailscale should not be required.

Jackmin801 (dismissed):
nice! yea same question about requiring tailscale as it is only needed if the nodes arent on private network already.

mikasenghaas:
oh yea so we just need nodes to be on the same private subnet. i was testing on non-colocated 1xH100 from different providers, so I needed VPN. if the nodes are in the same datacenter and share a subnet, this is indeed not required. i also mention this in the README

> First, ensure that your nodes are in the same private network and can reach each other. If not, a simple solution is to set up a VPN using [Tailscale](https://tailscale.com). (...)

the `enforce_eager` was (for some reason i don't know) needed, otherwise the engines would get stuck starting, seemingly while trying to do the graph capture. super not clear to me why this happens. that's why i say the vllm multi-node is quite a cutting edge feature and i don't expect everything to be super smooth just yet.

README.md
```diff
@@ -177,8 +177,61 @@ uv run rl \
...
+```bash
+# Node 1
+uv run inference \
+	--data-parallel-size 4 \
+	--data-parallel-size-local 2 \
+	--data-parallel-address <reachable-ip> \
+	--data-parallel-rpc-port 13345 \
+	--data-parallel-start-rank 2 \
+	--headless
+```
```
Jackmin801, line 227: these dont have enforce eager though. should they be added?

samsja:
> oh yea so we just need nodes to be on the same private subnet. i was testing on non-colocated 1xH100 from different providers, so I needed VPN. if the nodes are in the same datacenter and share a subnet, this is indeed not required. i also mention this in the README
> 
> > First, ensure that your nodes are in the same private network and can reach each other. If not, a simple solution is to set up a VPN using [Tailscale](https://tailscale.com). (...)
> 
> the `enforce_eager` was (for some reason i don't know) needed, otherwise the engines would get stuck starting, seemingly while trying to do the graph capture. super not clear to me why this happens. that's why i say the vllm multi-node is quite a cutting edge feature and i don't expect everything to be super smooth just yet.

okay then seems that we need to fix the enforce eager thiny before merging, quite important for perf

mikasenghaas:
> okay then seems that we need to fix the enforce eager thiny before merging, quite important for perf

mmh yea, it might be quite hard to track down where and why it hangs because its vllm internal. even if we find why we would have to probably do some ugly patch. im not sure time invested is 

alternatively, we could merge as is with a warning that we have found it to be not fully stable and suggest running with `enforce_eager` for now and hope that it gets fixed upstream soon. it's defo smth they are actively developing on

samsja:
> > okay then seems that we need to fix the enforce eager thiny before merging, quite important for perf
> 
> mmh yea, it might be quite hard to track down where and why it hangs because its vllm internal. even if we find why we would have to probably do some ugly patch. im not sure time invested is
> 
> alternatively, we could merge as is with a warning that we have found it to be not fully stable and suggest running with `enforce_eager` for now and hope that it gets fixed upstream soon. it's defo smth they are actively developing on

I would wait then tbh

mikasenghaas:
Closed + continued in #805

PR #645, opened 2025-07-29
Add GSPO

faresobeid:
Added GSPO implementation 

---

Linear: Resolves PRIMERL-27

### DIFF
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -179,7 +179,7 @@ class TrainerConfig(BaseSettings):
     weights: WeightCheckpointConfig = WeightCheckpointConfig()
 
     # The loss configuration
-    loss: LossConfig = Field(discriminator="type", default=RatioLossConfig())
+    loss: LossConfig = Field(discriminator="type", default=GRPOClipLossConfig())
 
     # The logging configuration
     log: LogConfig = LogConfig()
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -119,7 +119,21 @@ class RatioLossConfig(BaseLossConfig):
     clip_ratio: Annotated[float, Field(ge=0)] = 8.0
 
 
-LossConfig: TypeAlias = ClippingLossConfig | RatioLossConfig
+class GSPOClipLossConfig(BaseLossConfig):
+    """Configures the GSPO clip loss."""
+
+    type: Literal["clip_gspo"] = "clip_gspo"
+    epsilon_low: Annotated[float, Field(ge=0)] = 3e-4
+    epsilon_high: Annotated[float, Field(ge=0)] = 4e-4
+    clip_ratio: Annotated[float, Field(ge=0)] = 1.1
+
+class GSPORatioLossConfig(BaseLossConfig):
+    """Configures the GSPO ratio loss."""
+
+    type: Literal["ratio_gspo"] = "ratio_gspo"
+    clip_ratio: Annotated[float, Field(ge=0)] = 1.1
+
+LossConfig: TypeAlias = ClippingLossConfig | RatioLossConfig | GSPORatioLossConfig | GSPOClipLossConfig
 
 
 class FakeDataLoaderConfig(BaseConfig):
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -106,7 +106,7 @@ class BaseLossConfig(BaseModel):
 class GRPOClipLossConfig(BaseLossConfig):
     """Configures the clipping loss."""
 
-    type: Literal["clip"] = "clip-grpo"
+    type: Literal["clip-grpo"] = "clip-grpo"
     epsilon_low: Annotated[float, Field(ge=0)] = 0.2
     epsilon_high: Annotated[float, Field(ge=0)] = 0.2
     clip_ratio: Annotated[float, Field(ge=0)] = 4.0
@@ -115,14 +115,14 @@ class GRPOClipLossConfig(BaseLossConfig):
 class GRPORatioLossConfig(BaseLossConfig):
     """Configures the ratio loss."""
 
-    type: Literal["ratio"] = "ratio-grpo"
+    type: Literal["ratio-grpo"] = "ratio-grpo"
     clip_ratio: Annotated[float, Field(ge=0)] = 8.0
 
 
 class GSPOClipLossConfig(BaseLossConfig):
     """Configures the GSPO clip loss."""
 
-    type: Literal["clip_gspo"] = "clip-gspo"
+    type: Literal["clip-gspo"] = "clip-gspo"
     epsilon_low: Annotated[float, Field(ge=0)] = 3e-4
     epsilon_high: Annotated[float, Field(ge=0)] = 4e-4
     clip_ratio: Annotated[float, Field(ge=0)] = 1.1
@@ -130,7 +130,7 @@ class GSPOClipLossConfig(BaseLossConfig):
 class GSPORatioLossConfig(BaseLossConfig):
     """Configures the GSPO ratio loss."""
 
-    type: Literal["ratio_gspo"] = "ratio-gspo"
+    type: Literal["ratio-gspo"] = "ratio-gspo"
     clip_ratio: Annotated[float, Field(ge=0)] = 1.1
 
 LossConfig: TypeAlias = GRPOClipLossConfig | GRPORatioLossConfig | GSPORatioLossConfig | GSPOClipLossConfig
@@ -179,7 +179,7 @@ class TrainerConfig(BaseSettings):
     weights: WeightCheckpointConfig = WeightCheckpointConfig()
 
     # The loss configuration
-    loss: LossConfig = Field(discriminator="type", default=GRPOClipLossConfig())
+    loss: LossConfig = Field(discriminator="type", default=GRPORatioLossConfig())
 
     # The logging configuration
     log: LogConfig = LogConfig()
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -103,26 +103,26 @@ class BaseLossConfig(BaseModel):
     max_norm: Annotated[float, Field(ge=0, description="Maximum gradient norm to clip.")] = 1.0
 
 
-class ClippingLossConfig(BaseLossConfig):
+class GRPOClipLossConfig(BaseLossConfig):
     """Configures the clipping loss."""
 
-    type: Literal["clip"] = "clip"
+    type: Literal["clip"] = "clip-grpo"
     epsilon_low: Annotated[float, Field(ge=0)] = 0.2
     epsilon_high: Annotated[float, Field(ge=0)] = 0.2
     clip_ratio: Annotated[float, Field(ge=0)] = 4.0
 
 
-class RatioLossConfig(BaseLossConfig):
+class GRPORatioLossConfig(BaseLossConfig):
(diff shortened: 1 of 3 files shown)


src/prime_rl/trainer/config.py
```diff
@@ -203,7 +214,7 @@ class TrainerConfig(BaseSettings):
         Field(
             description="Whether to recompute the logprobs. If True, will always recompute logprobs and overwrite those found in the training batch.",
         ),
-    ] = True
+    ] = False
```
mikasenghaas: should not be part of this pr

src/prime_rl/trainer/config.py
```diff
@@ -118,14 +118,26 @@ class RatioConfig(BaseGRPOVariantConfig):
...
+    epsilon_low: Annotated[float, Field(ge=0)] = 4e-4
+    epsilon_high: Annotated[float, Field(ge=0)] = 4e-4
+
+class GSPORatioConfig(BaseGRPOVariantConfig):
+    """Configures the GSPO ratio loss."""
+
+    type: Literal["gspo_ratio"] = "gspo_ratio"
+    clip_ratio: Annotated[float, Field(ge=0)] = 1.1
+
+GRPOVariantsConfig: TypeAlias = Union[ClippingConfig, RatioConfig, GSPOClipConfig, GSPORatioConfig]
```
mikasenghaas: this will need a slight refactor because of #643

mikasenghaas: should be very straight-forward tho

src/prime_rl/trainer/loss.py
```diff
@@ -42,6 +42,29 @@ def grpo_loss(
             clip_ratio=grpo_loss_config.clip_ratio,
             highest_entropy_percentage=grpo_loss_config.highest_entropy_ratio_loss,
         )
+    elif isinstance(grpo_loss_config, GSPOClipConfig):
```
mikasenghaas: this is also affected by #643

src/prime_rl/trainer/loss.py
```diff
@@ -111,19 +134,96 @@ def grpo_loss_ratio(
...
+    advantages: Float[Tensor, "batch seq"],
+    original_logprobs: Float[Tensor, "batch seq"],
+    loss_mask: Int[Tensor, "batch seq"],
+    temperature: float,
+    epsilon_low: float,
+    epsilon_high: float,
+    highest_entropy_percentage: float,
+) -> tuple[Tensor, Tensor, Tensor]:
+    """
+    DeepSeek Math Loss: https://arxiv.org/abs/2402.03300
```
mikasenghaas: copy paste?

src/prime_rl/trainer/loss.py
```diff
@@ -111,19 +134,96 @@ def grpo_loss_ratio(
...
+
+    Args:
+        policy_logprobs: Log probabilities from the policy model
+        ref_logprobs: Log probabilities from the reference model
+        advantages: Advantages for each token
+        beta: KL penalty coefficient
+        epsilon: Clipping parameter for PPO
+        ignore_index: Specifies a target value that is ignored and does not contribute to the loss
```
mikasenghaas: doesnt match args

src/prime_rl/trainer/loss.py
```diff
@@ -111,19 +134,96 @@ def grpo_loss_ratio(
...
+
+    raw_ratio = ((per_token_logps - original_logprobs) * loss_mask).sum(dim=-1) / loss_mask.sum(dim=-1)
+    raw_ratio = raw_ratio.exp()
 
     ratio = torch.clamp(raw_ratio, 0, clip_ratio)
-    loss = -ratio * advantages
+    loss = -ratio * advantages[:, 0]
 
     if highest_entropy_percentage < 1.0:
         loss_mask = highest_entropy_mask(shifted_logits, loss_mask, highest_entropy_percentage)
```
mikasenghaas: we removed this right?

src/prime_rl/trainer/loss.py
```diff
@@ -111,19 +134,96 @@ def grpo_loss_ratio(
...
 
-    loss = _masked_sum(loss, loss_mask)
-    ratio = _masked_sum(ratio, loss_mask)
+    loss = loss.sum()
+    clipped_count = (raw_ratio > clip_ratio).float().sum()
+    raw_ratio = raw_ratio.sum()
```
mikasenghaas: it's a bit confusing imo that we call this raw_ratio again, maybe we can find smth more descriptive

src/prime_rl/trainer/train.py
```diff
@@ -222,9 +222,10 @@ def train(config: TrainerConfig):
         batch_size = micro_batch_size * num_micro_batches
 
         # Normalize by the number of unmasked tokens in the batch (per-batch length normalization)
-        loss_scale = sum(micro_batch["loss_mask"].sum() for micro_batch in micro_batches)
+        entropy_scale = sum(micro_batch["loss_mask"].sum() for micro_batch in micro_batches)
+        loss_scale = entropy_scale if (config.loss.variant.type == "gspo_clip" or config.loss.variant.type == "gspo_ratio") else batch_size
```
mikasenghaas: this changes the `loss_scale` compared for non-gspo right?

src/prime_rl/trainer/config.py
faresobeid: ya just had set for running experiments

src/prime_rl/trainer/loss.py
faresobeid: oops

src/prime_rl/trainer/loss.py
faresobeid: oops

src/prime_rl/trainer/loss.py
faresobeid: good point, this change is just to log the importance ratios before clipping to be easier to debug/tune the clipping ratio. Should prob also have a different pr to log the max/min ratios

src/prime_rl/trainer/train.py
faresobeid: oh oops this is wrong

src/prime_rl/trainer/train.py
```diff
@@ -259,10 +260,10 @@ def train(config: TrainerConfig):
...
 
             # Accumulate unnormalized local metrics
-            loss_metrics["loss/loss"] += loss.detach().float()
-            loss_metrics["loss/entropy"] += entropy.detach().float()
-            loss_metrics["loss/importance_ratio"] += importance_ratio.detach().float()
-            loss_metrics["loss/clipped_ratio"] += clipped_token_count.detach().float()
+            loss_metrics["loss/loss"] += loss.detach().float() / loss_scale
+            loss_metrics["loss/entropy"] += entropy.detach().float() / entropy_scale
+            loss_metrics["loss/importance_ratio"] += importance_ratio.detach().float() / loss_scale 
+            loss_metrics["loss/clipped_ratio"] += clipped_token_count.detach().float() / loss_scale
```
samsja: why the change here ?

faresobeid: To log them properly

src/prime_rl/trainer/config.py
```diff
@@ -119,7 +119,21 @@ class RatioLossConfig(BaseLossConfig):
...
+    epsilon_high: Annotated[float, Field(ge=0)] = 4e-4
+    clip_ratio: Annotated[float, Field(ge=0)] = 1.1
+
+class GSPORatioLossConfig(BaseLossConfig):
+    """Configures the GSPO ratio loss."""
+
+    type: Literal["ratio_gspo"] = "ratio_gspo"
+    clip_ratio: Annotated[float, Field(ge=0)] = 1.1
+
+LossConfig: TypeAlias = ClippingLossConfig | RatioLossConfig | GSPORatioLossConfig | GSPOClipLossConfig
```
mikasenghaas: Let's call `ClippingLossConfig` -> `GRPOClipLossConfig` and `RatioLossConfig` -> `GRPORatioLossConfig` for consistency

src/prime_rl/trainer/config.py
```diff
@@ -119,7 +119,21 @@ class RatioLossConfig(BaseLossConfig):
     clip_ratio: Annotated[float, Field(ge=0)] = 8.0
 
 
-LossConfig: TypeAlias = ClippingLossConfig | RatioLossConfig
+class GSPOClipLossConfig(BaseLossConfig):
+    """Configures the GSPO clip loss."""
+
+    type: Literal["clip_gspo"] = "clip_gspo"
```
mikasenghaas: let's use `-` here instead of `_` (same as e.g. for advantage types)

mikasenghaas: so then the general types would be `grpo-clip`, `grpo-ratio`, `gspo-clip` and `gspo-ratio`

src/prime_rl/trainer/loss.py
```diff
@@ -89,16 +111,76 @@ def grpo_loss_ratio(
...
+    shifted_logits: Float[Tensor, "batch seq vocab"],
+    input_ids: Int[Tensor, "batch seq"],
+    advantages: Float[Tensor, "batch seq"],
+    original_logprobs: Float[Tensor, "batch seq"],
+    loss_mask: Int[Tensor, "batch seq"],
+    temperature: float,
+    epsilon_low: float,
+    epsilon_high: float,
+    clip_ratio: float,
+) -> tuple[Tensor, Tensor, Tensor]:
```
mikasenghaas, line 136: can we add a docstring here

src/prime_rl/trainer/loss.py
```diff
@@ -89,16 +111,76 @@ def grpo_loss_ratio(
...
+@jaxtyped(typechecker=typechecker)
+def gspo_loss_ratio(
+    shifted_logits: Float[Tensor, "batch seq vocab"],
+    input_ids: Int[Tensor, "batch seq"],
+    advantages: Float[Tensor, "batch seq"],
+    original_logprobs: Float[Tensor, "batch seq"],
+    loss_mask: Int[Tensor, "batch seq"],
+    temperature: float,
+    clip_ratio: float,
+) -> tuple[Tensor, Tensor, Tensor]:
```
mikasenghaas, line 169: and here

src/prime_rl/trainer/loss.py
```diff
@@ -89,16 +111,76 @@ def grpo_loss_ratio(
...
+    
+    shifted_logits = shifted_logits / temperature
+    per_token_logps = selective_log_softmax(shifted_logits, input_ids)
+
+    raw_ratio = ((per_token_logps - original_logprobs) * loss_mask).sum(dim=-1) / loss_mask.sum(dim=-1)
```
mikasenghaas: this seems to be the critical line, can you add some comments explaining the tensors shapes and how it translates to the math in the paper

src/prime_rl/trainer/train.py
```diff
@@ -281,7 +282,10 @@ def train(config: TrainerConfig):
 
         # Normalize all loss metrics globally before reporting
         for key, value in loss_metrics.items():
-            loss_metrics[key] = value / loss_scale
+            if key == "loss/entropy":
+                loss_metrics[key] = value / total_tokens
+            else:
+                loss_metrics[key] = value / loss_scale
```
mikasenghaas: argh this looks like code smell. seems like an easy footgun in the future. can we not move the normalization back in the grad accumulation loop? then the normalization factor is defined for each value explicitly which seems less likely to fail silently

mikasenghaas:
Duplicate of #801. Reopen if mistake

PR #646, opened 2025-07-29
Log average seqlens for eval benchmarks

faresobeid:


src/prime_rl/eval/utils.py
```diff
@@ -74,6 +74,9 @@ async def run_benchmark(
     generate_completions_time = time.time() - generate_completions_start_time
     logger.debug(f"Generated completions in {generate_completions_time:.2f}s")
 
+    completion_lengths = [len(parse_completion_tokens(c)) for c in chat_completions]
+    avg_seq_len = sum(completion_lengths) / len(completion_lengths)
```
mikasenghaas: on orchestrator that same key is used for num input + output tokens iirc

src/prime_rl/eval/utils.py
```diff
@@ -114,10 +117,11 @@ async def run_benchmark(
     if could_be_binary:
         for pass_rate, pass_rate_score in pass_at_k.mean().items():
             message += f", {capitalize(pass_rate)}: {pass_rate_score:.2f}"
+    message += f", AvgSeqLen: {avg_seq_len:.2f}"
```
mikasenghaas: let's use `Seq. Length` for the message, as we do on the orchestrator. average is always implied if not specified otherwise

src/prime_rl/eval/utils.py
faresobeid: wdym

src/prime_rl/eval/utils.py
```diff
@@ -114,10 +117,11 @@ async def run_benchmark(
     if could_be_binary:
         for pass_rate, pass_rate_score in pass_at_k.mean().items():
             message += f", {capitalize(pass_rate)}: {pass_rate_score:.2f}"
+    message += f", Seq. Len: {avg_completion_len:.2f}"
     logger.success(message + ")")
 
     # Log statistics to monitor
-    eval_metrics = {f"avg@{k}": float(sample_stats.reward.mean())}
+    eval_metrics = {f"avg@{k}": float(sample_stats.reward.mean()), "avg_completion_len": float(avg_completion_len)}
```
mikasenghaas: acc, let's call this just `completion_len` in wandb. avg is always implied. if we use another point stat we do e.g. `completion_len/max` (similar to `seq/` section on orchestrator)

mikasenghaas (commented):
one last naming change, then merge!

samsja (approved):
lgtm but lets wait for @mikasenghaas review before merging

mikasenghaas (approved):

PR #648, opened 2025-07-29
improve importance ratio metrics

samsja:
# Improve importance ratio metrics

this PR add a bunch of new metrics regarding importance ratio, namely the max and min, the raw and unraw ratio, as well as the error sum.

It also refactors the way we handle the importance ratio metrics and disentangles it from the loss metrics. It should make it easier if we want to log the full distribution of the logprob ratio. 

screenshot-20250731-142745

### DIFF
--- a/src/prime_rl/trainer/loss.py
+++ b/src/prime_rl/trainer/loss.py
@@ -1,4 +1,7 @@
+from dataclasses import dataclass
+
 import torch
+import torch.distributed as dist
 from beartype import beartype as typechecker
 from jaxtyping import Float, Int, jaxtyped
 from torch import Tensor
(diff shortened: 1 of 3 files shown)


src/prime_rl/trainer/train.py
```diff
@@ -309,7 +326,7 @@ def train(config: TrainerConfig):
         # Maybe clean up weight checkpoint
         weight_ckpt_manager.maybe_clean(progress.step)
 
-        # Optionally, dump memory snapshot
+        # Optionally, dump memorfy snapshot
```
Jackmin801: whos memorfy?

src/prime_rl/trainer/loss.py
```diff
@@ -182,3 +217,67 @@ def shift_logits(logits: Float[Tensor, "batch seq vocab"]) -> Float[Tensor, "bat
     zeros = torch.zeros(B, 1, V, device=logits.device, dtype=logits.dtype)  # (B, 1, V)
     logits = torch.cat([zeros, logits], dim=1)  # (B, L, V)
     return logits
+
+
+class ImportanceRatioMetrics:
```
mikasenghaas, line 222: i think we might wanna generalize this pattern, seems very bulky "only" for loggin importance ratio metrics. maybe smth similar in spirit to the metrics averager, just better?

samsja: I would just keep it like this for now and generalize later, agree that refactor could be nice but not need for now

mikasenghaas (approved):

PR #649, opened 2025-07-30
Reorganize environments into standalone modules

willccbb:
Putting up for initial review, ported registry to `environments/` folder, bumped vf commit with pydantic fixes + native tool-call support

Found type issue in `Rollout` class I think, rewards + advantages should be `list[float]` not `list[list[float]]`

Hendrycks math:
image

image

ref:
image
image

Verifiers updates for Pydantic objects: https://github.com/willccbb/verifiers/blob/main/verifiers/envs/environment.py

For a follow-up PR:
- add native tool call support to vLLM config, add experiment config for tool use (search-R1 and/or math-python)


configs/endpoints.py
```diff
@@ -0,0 +1,42 @@
+ENDPOINTS = {
+    "my-model": {
```
samsja: not sure if configs is the best place for this

pyproject.toml
```diff
@@ -54,13 +56,20 @@ prerelease = "allow"
 
 [tool.uv.sources]
 torch = [{ index = "pytorch-cu128" }]
-verifiers = { git = "https://github.com/willccbb/verifiers", rev = "90c06b2" }
+verifiers = { git = "https://github.com/willccbb/verifiers", rev = "75c7df8" }
+vf-intellect-math = { workspace = true }
```
samsja, line 62: out of curiosity what does uv do with  this workspace = true ?

samsja (commented):
This looks great to me :) Added a couple of minor comments on the PR. Could we add a section in the readme that explains a bit how we deal with environments as sub package and that they are stored inside the environments folder? (just a short description is enough imo)

Also we should run an hendrycks training to make sure nothing breaks before merging

configs/endpoints.py
willccbb: anywhere else you'd propose? imo it's super useful to have an endpoint registry file like this somewhere for managing multiple API sources when doing `vf-eval` from command line without needing to list out full args

pyproject.toml
willccbb: used in conjunction with this:

```toml
[tool.uv.workspace]
members = [
    "environments/vf_intellect_math",
]
```
it tells uv to install the module from the "environments/vf_intellect_math" directory


auto-populated when you do

```bash
uv add "vf-intellect-math @ ./environments/vf_intellect_math"
```

configs/endpoints.py
mikasenghaas: yea, defo not in configs. im not sure i understand the purpose of this file in general. do we not expect ppl to run `vf-eval` within `verifiers` and not `prime-rl`? (assuming that this is gettin an initial reward dist for a new env from a couple of different models)

environments/vf_ascii_tree/README.md
mikasenghaas, line 1: oh yea i like the environments registry, very clearly structured and user-friendly

pyproject.toml
mikasenghaas: oh do i have to change our main `pyproject.toml` whenever i want to use a new env? if this is true, I don't like it

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -77,7 +77,7 @@ async def orchestrate(config: OrchestratorConfig):
     ckpt_step = 0
     if config.ckpt and config.ckpt.resume_step:
         logger.info(f"Resuming training from checkpoint step `{config.ckpt.resume_step}`")
-        ckpt_manager.load(progress, step=config.ckpt.resume_step)
+        ckpt_manager.load(progress, step=config.ckpt.resume_step)  # type: ignore
```
mikasenghaas: what's the type error? can we fix it? after long discussion with sami, we agreed to not have type ignores in our code in general

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -108,7 +108,7 @@ async def orchestrate(config: OrchestratorConfig):
         if config.ckpt and config.ckpt.interval and not is_first_step and progress.step % config.ckpt.interval == 0:
             logger.info(f"Saving checkpoint at step {progress.step}")
             save_ckpt_start_time = time.time()
-            ckpt_manager.save(progress, step=progress.step)
+            ckpt_manager.save(progress, step=progress.step)  # type: ignore
```
mikasenghaas: same here

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -210,10 +210,10 @@ async def orchestrate(config: OrchestratorConfig):
             generate_completions_time = time.time() - generate_completions_start_time
 
             results = vf_env.process_env_results_vllm(
-                prompts=outputs["prompt"],
-                completions=outputs["completion"],
-                states=outputs["state"],
-                rewards=outputs["reward"],
+                prompts=outputs.prompt,
```
mikasenghaas, line 225: beautiful

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -381,7 +381,7 @@ async def orchestrate(config: OrchestratorConfig):
     # Write final checkpoint
     if config.ckpt:
         logger.info("Writing final checkpoint")
-        ckpt_manager.save(progress, step=progress.step)
+        ckpt_manager.save(progress, step=progress.step)  # type: ignore
```
mikasenghaas: and here too

mikasenghaas (commented):
new registry looks really nice. i hope there is another way to install deps from a env than changing our main lockfile

environments/vf_ascii_tree/README.md
willccbb: the hope is that no one is ever running anything "in verifiers" as part of the user typical workflow, unless you're PRing core functionality (i.e. not just a environment, but a feature which touches lower-level classes or the verifiers trainer). it's designed to be used as a library, and ships with some scripts like vf-eval which would be executed by the user in a downstream project where verifiers is installed (such as prime-rl). 

endpoints.py in my mind *is* a convenience config file for the user, but is optional. the research flow i'm imagining for us, as we develop more environments designed to be trained with prime-rl, is that we do `vf-init` (and/or `prime env init`) to create a template in the prime-rl repo's environments folder, write env code, install env to project with one of:
```bash
vf-install my-new-env # wrapper for uv pip install -e ./environments/...
prime env install # similar
uv add "my-new-env @ ./path/to/my_new_env"
```
and then initially test "step 0" by running with an API model, before then making proper training configs for running `rl.py`.

Testing with `endpoints.py` configured: 
```bash
vf-eval my-new-env -n 10 -r 5 -m a3b
```
without:
```bash
vf-eval my-new-env -n 10 -r 5 -m Qwen/Qwen3-30B-A3B-Instruct-2507 -b "http://0.0.0.0:8000/v1" -k "EMPTY"
```
This gets annoying quickly and it's nice to be able to configure your personal endpoints globally somewhere. you can use `-e path/to/endpoints.py` to configure if desired. 

it could live in `scripts`, but it's not a script. Another option could be at the root level, but that feels cluttered. Maybe root + add to .gitignore? Though once we have more inference offerings, we'll probably want to have some of our own endpoints preconfigured

willccbb: `vf-eval` is exactly the same as the old `scripts/test_env.py` script in terms of functionality, but IMO the "global" version of that should live in verifiers. if we want other more prime-rl specific eval scripts, we can add those to scripts too, but I think there should be a global verifiers one as a first entrypoint.

willccbb:
> oh do i have to change our main pyproject.toml whenever i want to use a new env? if this is true, I don't like it

Nope, you can also just do `uv pip install ...` or (`vf-install`) for a temporary install, adding to global pyproject.toml is optional, but IMO is what we want to do for environments that we always want to have pre-installed in prime-rl (like our core debug ones), rather than listing those deps directly. For example, we can remove the `textarena` and `nltk` deps for the wordle env if we want to have the wordle env itself listed in pyproject.toml. Not doing either gets annoying quickly, as you have to remember to temporarily install any env you want to use, and it'll get wiped every time you do `uv sync`. 

Maybe would be convenient to have this done programmatically via `rl.py`? Or the global setup script (this will miss any new ones added in the current session)? Environments can have arbitrary dependencies and thus shouldn't all be in pyproject.toml, but if not, we have to install them somehow. Up to us what we want the typical flow to be.

src/prime_rl/orchestrator/orchestrator.py
willccbb: will remove the type ignores. comes up because it thinks `ckpt_manager might be None. how do we feel about asserts?

willccbb: fixed. issue was that it could be *unbound*, not None, fixed by initializing it to None outside the if block where it gets initialized.

pyproject.toml
samsja: > oh do i have to change our main `pyproject.toml` whenever i want to use a new env? if this is true, I don't like it
but I don't see how we can avoid it if we treat each new env as a python package. I am fine with it

samsja:
> Maybe would be convenient to have this done programmatically via `rl.`

I would try to avoid this for now, will bite us later. I see no problem with updating the pyproject.toml each time we add an env. There is also no other way around imo, if an env install a new dependency that we want to lock we have to update pyproject and lock file, that's how it is imo

pyproject.toml
willccbb: IMO we should have core ones in pyproject.toml, most will be up to user to install, we can add individual ones to setup script if we want to conveniently mass-install without forcing into pyproject.toml

mikasenghaas (approved):

samsja (approved):
lgtm !!

PR #650, opened 2025-07-30
[feat] expert parallel trainer

Jackmin801:


src/prime_rl/trainer/model.py
```diff
@@ -18,18 +18,24 @@
 
 from prime_rl.trainer.config import ModelConfig
 
-Model: TypeAlias = LlamaForCausalLM | Qwen2ForCausalLM | Qwen3ForCausalLM
+Model: TypeAlias = LlamaForCausalLM | Qwen2ForCausalLM | Qwen3ForCausalLM | nn.Module
```
samsja: we can also just use nn.Module everywhere

mikasenghaas: yes, we should do that!

src/prime_rl/inference/config.py
```diff
@@ -69,6 +69,13 @@ class ModelConfig(BaseConfig):
         ),
     ] = False
 
+    trust_remote_code: Annotated[
+        bool,
+        Field(
+            description="Whether to trust remote code. Passed to vLLM engine init",
+        ),
+    ] = False
```
mikasenghaas, line 77: can you explain the need for trust remote code everywhere?

willccbb: lots of model architectures will require it if the modeling code comes from HF / isn't part of standard transformers

willccbb: for example Moonlight-16B needs it, which we'll def want to use for Muon ablations ([configuration_deepseek.py](https://huggingface.co/moonshotai/Moonlight-16B-A3B-Instruct/blob/main/configuration_deepseek.py))

src/prime_rl/trainer/model.py
Jackmin801: ook. eh then we dont need the type alias I think. ill just put nn.Module where that used to exist

Jackmin801:
ci looks ok
image

src/prime_rl/trainer/sft/data.py
```diff
@@ -14,6 +14,7 @@
 
 
 class Sample(TypedDict):
+    epoch: int  # TODO: Argh, can we find a way to export epoch metainformation in a nicer way?
```
samsja: is it duplicate ?

Jackmin801: ah yea maybe this is an uncaught merge conflict

Jackmin801: ook removed one

Jackmin801:
> should we add sft toml also for both ?

oh but we dont have model specific for the sft.toml I think. I can only find task specific

mikasenghaas:
> should we add sft toml also for both ?

i dont think we do tbh, goes back to our scaling number of config files discussion

src/prime_rl/trainer/model.py
```diff
@@ -11,44 +10,57 @@
...
 def setup_tokenizer(config: ModelConfig) -> PreTrainedTokenizer:
-    tokenizer = AutoTokenizer.from_pretrained(config.name)
+    tokenizer = AutoTokenizer.from_pretrained(config.name, trust_remote_code=config.trust_remote_code)
     tokenizer.pad_token_id = tokenizer.eos_token_id
     return tokenizer
 
 
-def setup_fsdp(model: Model, config: ModelConfig):
+def setup_fsdp(model: nn.Module, config: ModelConfig):
+    dist.init_process_group()
```
mikasenghaas: ah does this also solve #747?

mikasenghaas: if yes, we could remove the warning filters

src/prime_rl/trainer/weights.py
```diff
@@ -4,19 +4,77 @@
...
+            ][j]
+            state_dict[f"model.layers.{i}.mlp.experts.{j}.down_proj.weight"] = state_dict[
+                f"model.layers.{i}.mlp.experts.w2"
+            ][j]
+            state_dict[f"model.layers.{i}.mlp.experts.{j}.up_proj.weight"] = state_dict[
+                f"model.layers.{i}.mlp.experts.w3"
+            ][j]
+        del state_dict[f"model.layers.{i}.mlp.experts.w1"]
+        del state_dict[f"model.layers.{i}.mlp.experts.w2"]
+        del state_dict[f"model.layers.{i}.mlp.experts.w3"]
```
mikasenghaas, line 75: nit: hm is there a better file for this than `weights.py`? looks like model or general utils to me, i.e. `utils.py` or `model.py` on trainer?

mikasenghaas: ah nvm this is just for ckpt conversion, my bad. and we only need this in weights bc that's written to vllm, while full checkpoints are not affected right?

src/prime_rl/trainer/model.py
Jackmin801: oh im not sure what those warning filters were about

src/prime_rl/trainer/weights.py
Jackmin801: yeap

Jackmin801:
image

PR #652, opened 2025-07-30
Multi-turn alphabetic sorting task

kalomaze:
## Alphabet Sort (`alphabet-sort`)

This task requires the model to maintain and update an alphabetically sorted list of names across multiple conversation turns, with new names being tagged appropriately. The dataset uses real author names from arXiv papers, with 1-3 turns per conversation and 2-5 total names (the turn and name counts are randomized during the data creation process by default).

**Example Input (Turn 1):**

```
Sort these names in alphabetical order by FIRST name: Johnson, Alice, Bob, Charlie

Use exactly this format:
<alphabetical_sorted>
Name1
Name2
Name3
Name4
</alphabetical_sorted>
```

**Expected Output (Turn 1):**

```
<alphabetical_sorted>
Alice
Bob
Charlie
Johnson
</alphabetical_sorted>
```

**Example Input (Turn 2):**

```
Now sort ALL of these names alphabetically by FIRST name: David, Emily

These are in addition to the prior list. Mark any NEW names (that weren't in the prior list) with `// new name!` at the end.

Use exactly this format:
<combined_alphabetical_sorted>
Name1
Name2
Name3 // new name!
Name4
Name5 // new name!
Name6
</combined_alphabetical_sorted>
```

**Expected Output (Turn 2):**

```
<combined_alphabetical_sorted>
Alice
Bob
Charlie
David // new name!
Emily // new name!
Johnson
</combined_alphabetical_sorted>
```

**Example Input (Turn 3):**

```
Now sort ALL of these names alphabetically by FIRST name: Frank

These are in addition to the prior list. Mark any NEW names (that weren't in the prior list) with `// new name!` at the end. Follow the same format as before.
```

**Expected Output (Turn 3):**

```
<combined_alphabetical_sorted>
Alice
Bob
Charlie
David
Emily
Frank // new name!
Johnson
</combined_alphabetical_sorted>
```

The reward function uses difflib to calculate sequence similarity between predicted and expected outputs, with the final score raised to the nth power (`similarity_power`, defaults to 4) to emphasize precision.

A sample training configuration for `Qwen/Qwen2.5-3B-Instruct` is provided that sets `min_turns` and `max_turns` (two of the various optional environment args) to 2.

image

image


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -356,6 +360,7 @@ async def orchestrate(config: OrchestratorConfig):
             "perf/problem_requests": problem_requests,
             "perf/completion_requests": completion_requests,
             "perf/calls_to_generate": calls_to_generate,
+            "perf/effective_batch_size_tokens": effective_batch_size_tokens,
```
samsja: makes sense, should we rename effective batch size to effective batch size ratio ?

src/prime_rl/environments/registry.py
```diff
@@ -704,6 +706,242 @@ def load_wordle_nothink_environment(num_train_examples: int = 2000, num_eval_exa
...
+def load_alphabet_sorting_environment(
+    max_turns: int = 3,
+    min_turns: int = 1,
+    single_turn_min_names: int = 2,
+    single_turn_max_names: int = 5,
+    first_turn_min_names: int = 2,
+    first_turn_max_names: int = 3,
+    subsequent_turn_min_names: int = 1,
+    subsequent_turn_max_names: int = 3,
+    similarity_power: int = 4,
```
samsja: I will let @willccbb review this part

src/prime_rl/orchestrator/orchestrator.py
kalomaze: it might make more sense to have the metric reflect what % of groups have at least one sample with non-zero advantages instead of naming it something ambigious like "effective batch size"

src/prime_rl/environments/registry.py
willccbb: discussed in discord but would propose collapsing single_turn with first_turn, and adding asserts to ensure that anything like "min_turns > max_turns" fails loudly at the start

samsja (approved):
lgtm

src/prime_rl/environments/registry.py
willccbb: @kalomaze let's fix this before merge?

mikasenghaas: collpase was handled by @kalomaze, i added some basic arg validation. should be good to go, can also refine later as well

mikasenghaas:
adapted a readme from your pr description @kalomaze, feel free to refine if you see the need to. @willccbb, i think there should probably be a template for an env readme, being opinioated about how users describe their environments concisely. have we thought abt this yet?

samsja (approved):
lgtm

PR #654, opened 2025-07-30
add skywork math-verify impl

faresobeid:


mikasenghaas (commented):
this seems like it should be a pr into `verifiers`? cc/ @willccbb

willccbb:
if it's a "canonical environment" then yes, i'd vote PR-ing it into verifiers

bespoke PI research things like kalo's unscramble/ascii stuff make sense here

faresobeid:
its not a canonical environment, just used the skywork implementation of math-verify. It's for math verification in general

PR #655, opened 2025-07-30
Sami/recompute logprob gpu

samsja:


src/prime_rl/trainer/train.py
```diff
@@ -205,10 +204,10 @@ def train(config: TrainerConfig):
                     micro_batch["recomputed_logprob_error"] = recomputed_logprob_error.to("cpu")
                     micro_batch["logprobs"] = recomputed_logprobs.to("cpu")
 
-            # here we sepcifically don't save the tensor offloaded, they are alreay consumed and we will never use it again.
-            # this avoid having to make sure we don't keep too much tensor offloaded in cpu memory
             reshard_module(logprob_model)
-            offload_model_to_cpu(logprob_model)
+
+            # we don't need the weight of this policy anymore so release it from memory to avoid memory leak
+            del tensor_offloaded_repository[infer_step]
```
mikasenghaas: we already do this in line 189-190

mikasenghaas: but shouldn't effect correctness

src/prime_rl/trainer/train.py
```diff
@@ -309,7 +325,7 @@ def train(config: TrainerConfig):
         # Maybe clean up weight checkpoint
         weight_ckpt_manager.maybe_clean(progress.step)
 
-        # Optionally, dump memory snapshot
+        # Optionally, dump memorfy snapshot
```
mikasenghaas, line 325: i like this change

PR #656, opened 2025-07-31
LR schedulers

kalomaze:


src/prime_rl/trainer/lr_scheduler.py
```diff
@@ -0,0 +1,64 @@
+from torch.optim import Optimizer
+from torch.optim.lr_scheduler import CosineAnnealingLR, LinearLR, LRScheduler, SequentialLR
+
+from prime_rl.trainer.config import TrainerConfig
+
+
+def create_lr_scheduler(optimizer: Optimizer, config: TrainerConfig) -> LRScheduler | None:
+    """Create learning rate scheduler based on config."""
+    if config.optim.scheduler == "constant":
+        return None
```
samsja: I would rather do 

scheduler = create_lr_scheduler(..) if config.optim.scheduler else None

src/prime_rl/trainer/lr_scheduler.py
```diff
@@ -0,0 +1,64 @@
+from torch.optim import Optimizer
+from torch.optim.lr_scheduler import CosineAnnealingLR, LinearLR, LRScheduler, SequentialLR
+
+from prime_rl.trainer.config import TrainerConfig
+
+
+def create_lr_scheduler(optimizer: Optimizer, config: TrainerConfig) -> LRScheduler | None:
```
samsja: lets only pass the OptimizerConfig here and no the full config

src/prime_rl/trainer/config.py
```diff
@@ -41,13 +42,28 @@ class ModelConfig(BaseConfig):
...
+        int, Field(ge=0, description="Number of warmup steps for the learning rate scheduler.")
+    ] = 0
+
+    n_final_decay: Annotated[
+        int | None,
+        Field(
+            ge=1,
+            description="Number of steps to decay the learning rate during the final portion of training. If None, will use max_steps from trainer config.",
+        ),
+    ] = None
```
samsja: any chance you can allign the config name with the one form torchtitan? https://github.com/pytorch/torchtitan/blob/cf30b2902718790cbe91900414c3201b6d7680b0/torchtitan/config/job_config.py#L135

samsja (commented):
I added couple of comment but looks great  :)

src/prime_rl/trainer/lr_scheduler.py
mikasenghaas, line 1: rename to `scheduler.py`

src/prime_rl/trainer/lr_scheduler.py
```diff
@@ -0,0 +1,64 @@
...
+    """Create learning rate scheduler based on config."""
+    if config.optim.scheduler == "constant":
```
mikasenghaas: is there a way to have a constant LR scheduler that leaves LR unchanged on `.step`? if so, would prefer the return type of this function to be just `LRScheduler` instead of `LRScheduler | None`

src/prime_rl/trainer/lr_scheduler.py
```diff
@@ -0,0 +1,64 @@
...
+        # Fallback: decay for remaining steps after warmup
+        n_final_decay = config.max_steps - warmup_steps
+
+    if n_final_decay <= 0:
+        raise ValueError(f"n_final_decay must be positive, got {n_final_decay}")
+
+    if warmup_steps + n_final_decay > config.max_steps:
+        raise ValueError(
+            f"Warmup steps ({warmup_steps}) + final decay steps ({n_final_decay}) exceeds max_steps ({config.max_steps})"
+        )
```
mikasenghaas: we could (re)move a lot of this arg validation to pydantic if we had separate scheduler configs for each type. easier to read imo but would need a bit of refactor

src/prime_rl/trainer/train.py
```diff
@@ -332,13 +344,15 @@ def train(config: TrainerConfig):
...
-        step_message = f"Step {progress.step} | Time: {step_time:.2f}s | Loss: {loss_metrics['loss/loss']:.2f} | Entropy: {loss_metrics['loss/entropy']:.2f} | Importance Ratio: {loss_metrics['loss/importance_ratio']:.2f} | Throughput: {throughput:.0f} tokens/s | MFU: {mfu:.1f}%"
+        current_lr = optimizer.param_groups[0]["lr"]
+        step_message = f"Step {progress.step} | Time: {step_time:.2f}s | Loss: {loss_metrics['loss/loss']:.2f} | Entropy: {loss_metrics['loss/entropy']:.2f} | Importance Ratio: {loss_metrics['loss/importance_ratio']:.2f} | LR: {current_lr:.2e} | Throughput: {throughput:.0f} tokens/s | MFU: {mfu:.1f}%"
         logger.success(step_message)
 
         # Log performance metrics
         perf_metrics = {
             "perf/train/throughput": throughput,
             "perf/train/mfu": mfu,
+            "optim/lr": current_lr,
```
mikasenghaas: separate section should log separately

src/prime_rl/trainer/lr_scheduler.py
samsja: wut i wrote this exact same comment where did it go

src/prime_rl/trainer/lr_scheduler.py
samsja: its here

samsja (commented):
added commebt

src/prime_rl/trainer/config.py
```diff
@@ -40,14 +40,70 @@ class ModelConfig(BaseConfig):
...
+    """Configuration for linear learning rate scheduler."""
+    
+    scheduler: Literal["linear"] = "linear"
+    decay_steps: Annotated[
+        int | None,
+        Field(
+            ge=1,
+            description="Number of steps to decay the learning rate during the final portion of training. If None, will use remaining steps after warmup.",
+        ),
+    ] = None
```
samsja, line 65: wondering if it could not be easier to do a decay percentage like in torchtitan, but no strong opinion

kalomaze: I just think it's neat to have warmup and decay standardized as being based on steps rather than ratio over total steps, IMO it is fine like this

samsja: sure !

src/prime_rl/trainer/config.py
```diff
@@ -40,14 +40,70 @@ class ModelConfig(BaseConfig):
     ] = True
 
 
+class BaseSchedulerConfig(BaseModel):
+    """Base configuration for learning rate schedulers."""
+    
+    warmup_steps: Annotated[int, Field(ge=0, description="Number of warmup steps for the learning rate scheduler.")] = 0
```
mikasenghaas, line 46: constant lr scheduler should not have warmup steps, ie. this arg should not be in a base class

src/prime_rl/trainer/config.py
```diff
@@ -40,14 +40,70 @@ class ModelConfig(BaseConfig):
...
+        int | None,
+        Field(
+            ge=1,
+            description="Number of steps to decay the learning rate during the final portion of training. If None, will use remaining steps after warmup.",
+        ),
+    ] = None
+    
+    @model_validator(mode="after")
+    def validate_decay_steps(self):
+        if self.scheduler == "linear" and self.decay_steps is not None and self.decay_steps <= 0:
```
mikasenghaas, line 69: linear is a constant in this config class and will always be true?

mikasenghaas: wait acc we can remove this entire check, you already have a `ge=1` enforced on the field which does the same thing but more neatly

src/prime_rl/trainer/config.py
```diff
@@ -40,14 +40,70 @@ class ModelConfig(BaseConfig):
...
+            ge=1,
+            description="Number of steps to decay the learning rate during the final portion of training. If None, will use remaining steps after warmup.",
+        ),
+    ] = None
+    
+    @model_validator(mode="after")
+    def validate_decay_steps(self):
+        if self.scheduler == "cosine" and self.decay_steps is not None and self.decay_steps <= 0:
+            raise ValueError(f"decay_steps must be positive for cosine scheduler, got {self.decay_steps}")
+        return self
```
mikasenghaas, line 90: same here, can be removed

src/prime_rl/trainer/config.py
```diff
@@ -40,14 +40,70 @@ class ModelConfig(BaseConfig):
...
-    """Configures the Adam optimizer."""
+    """Configures the Adam optimizer and learning rate scheduler."""
 
     lr: Annotated[float, Field(ge=0)] = 4e-4
     weight_decay: Annotated[float, Field(ge=0)] = 0.01
     betas1: Annotated[float, Field(ge=0)] = 0.9
     betas2: Annotated[float, Field(ge=0)] = 0.99
 
+    # LR Scheduler configuration
+    scheduler_config: SchedulerConfig = Field(discriminator="scheduler", default=ConstantSchedulerConfig())
```
mikasenghaas, line 105: let's use consistent naming here. `scheduler_config` -> `scheduler` and the discriminator field should be `type` (to not have the ugly `scheduler.scheduler`

mikasenghaas:
nvm lol, will redo in pr

PR #657, opened 2025-07-31
refactor logprob

samsja:
screenshot-20250801-191423

### DIFF
--- a/src/prime_rl/trainer/loss.py
+++ b/src/prime_rl/trainer/loss.py
@@ -8,7 +8,6 @@ from torch import Tensor
 from torch.nn import functional as F
 
 from prime_rl.trainer.config import LossConfig
-from prime_rl.trainer.model import Model, forward
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


mikasenghaas (commented):
i think i like the refactor idea, but should probably do a couple of steps on hendrycks before merging, also fix loss unit test

src/prime_rl/trainer/loss.py
```diff
@@ -177,17 +160,15 @@ def selective_log_softmax(
...
-) -> Float[Tensor, "batch seq"]:
-    logits = forward(model, input_ids, position_ids).contiguous()
+) -> tuple[Float[Tensor, "batch seq"], Float[Tensor, "batch seq vocab"]]:
     shifted_logits = shift_logits(logits)
     shifted_logits = shifted_logits / temperature
     logprobs = selective_log_softmax(shifted_logits, input_ids)
-    del logits, shifted_logits
-    return logprobs
+    del logits
+    return logprobs, shifted_logits
```
mikasenghaas, line 171: i dont like that we are changing the return type of this function. this looks like a "side effect" to me. if the caller needs `shifted_logits` it can just call on `shift_logits` on the `logits` which we know he already has from the func args

samsja: fair

PR #663, opened 2025-08-01
fix vf dependencies

samsja:
what this pr do 

* adding all the environment as python deps
* group all vf related into a dependency group


mikasenghaas (approved):
oh yea i like this!

PR #670, opened 2025-08-02
Reorganize configuration files

kalomaze:
## Reorganize config file structure for better usability

**Before:** Configs were scattered across separate `inference/`, `orchestrator/`, and `trainer/` directories with duplicate subfolder structures, making it difficult to work with complete configuration sets.

**After:** Unified structure where each configuration variant gets its own folder containing standardized `infer.toml`, `orch.toml`, and `train.toml` files.

### Structure Changes:
- `configs/inference/skywork_math/7b/stage1.toml` + similar files across 3 dirs
- → `configs/skywork_math_7b/stage1/` with all 3 config types

This makes it trivial to grab a complete config set for any experiment variant without hunting across multiple directories.


README.md
```diff
@@ -69,26 +69,26 @@ uv run python -c "import flash_attn"
...
 ```bash
-uv run trainer @ configs/trainer/debug.toml
+uv run trainer @ configs/main/debug/train.toml
 ```
 
 4. Check that you can run the orchestrator against an inference server (*this requires 1 GPU*)
 
 ```bash
-uv run inference @ configs/inference/debug.toml
+uv run inference @ configs/main/debug/infer.toml
```
mikasenghaas: i am not sure i like the `main` directory. can we not just move `debug` and `reverse_text` one level up?

kalomaze: yes we can, they just weren't folders originally so my script moved them into a shared folder

PR #676, opened 2025-08-04
Refactor batch collate

samsja:
this pr move the loss_scale calculation to the orchestartor, allowing to support multiple scaling.

It also notabily fix a bug where multi gpu training was not syncing the loss_scale and doing it per gpu

### DIFF
--- a/src/prime_rl/orchestrator/batch.py
+++ b/src/prime_rl/orchestrator/batch.py
@@ -125,14 +125,19 @@ def prepare_batch_padding(
 
         batches_per_gpu.append(batches)
 
+    num_completion_tokens_full_batch = sum(
+        micro_batch["num_completion_tokens"] for batch in batches_per_gpu for micro_batch in batch
+    )
+
     if loss_scale_type == "count_completion_tokens":
-        loss_scale = sum(micro_batch["num_completion_tokens"] for batch in batches_per_gpu for micro_batch in batch)
+        loss_scale = num_completion_tokens_full_batch
     elif loss_scale_type == "fixed":
         loss_scale = sum(len(batches) for batches in batches_per_gpu) * micro_batch_size * seq_len
 
     for batch in batches_per_gpu:
         for micro_batch in batch:
             micro_batch["loss_scale"] = loss_scale
+            micro_batch["num_completion_tokens_full_batch"] = num_completion_tokens_full_batch
 
     return batches_per_gpu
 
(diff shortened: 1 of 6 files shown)


src/prime_rl/orchestrator/batch.py
```diff
@@ -16,6 +17,7 @@ class BatchSample(TypedDict):
     loss_mask: Int[Tensor, "seq"]
     advantages: Float[Tensor, "seq"]
     logprobs: Float[Tensor, "seq"]
+    completion_tokens_count: int
```
mikasenghaas: let's call this `num_completion_tokens` for consistency reasons

src/prime_rl/orchestrator/batch.py
```diff
@@ -119,6 +125,17 @@ def prepare_batch_padding(
 
         batches_per_gpu.append(batches)
 
+    if loss_scale_type == "count_completion_tokens":
+        loss_scale = sum(micro_batch["completion_tokens_count"] for batch in batches_per_gpu for micro_batch in batch)
+    elif loss_scale_type == "fixed":
+        loss_scale = sum(len(batches) for batches in batches_per_gpu) * micro_batch_size * seq_len
+    else:
+        raise ValueError(f"Invalid loss scale type: {loss_scale_type}")
```
mikasenghaas: else not necesary bc pydantic enforces literal

mikasenghaas (commented):
i think it's directionally correct but im not a big fan of the current batching implementation. even before this pr it seemed a bit brittle (clashing types, uncessary separation into sample, micro batch and batch, nested function calls, loops and copies, etc.). bottomline is that im not super comfortable merging this in the current state. i will make a proposal pr that is hopefully more robust but with the same idea of making the loss scale an orchestrator concern

samsja:
> i think it's directionally correct but im not a big fan of the current batching implementation. even before this pr it seemed a bit brittle (clashing types, uncessary separation into sample, micro batch and batch, nested function calls, loops and copies, etc.). bottomline is that im not super comfortable merging this in the current state. i will make a proposal pr that is hopefully more robust but with the same idea of making the loss scale an orchestrator concern

We shouldn't refactor this part for now, lets just fix quicky the loss_scale to see if it make a difference in term of stability

mikasenghaas:
Closing for now because not super high priority, will pick up again in a bit

PR #678, opened 2025-08-04
Add Deepscaler and AceReason-Math envs

faresobeid:
Add deepscaler 1.5B and Acereason 7B configs for ablations and baselines


configs/deepscaler_math/stage1/train.toml
```diff
@@ -0,0 +1,12 @@
+
+[monitor.wandb]
+
+[model]
+name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B"
+ac = true
```
samsja: I don't think you need activation checkpointing with 1.5b

configs/deepscaler_math/stage2/orch.toml
```diff
@@ -0,0 +1,29 @@
+max_steps = 1000
+batch_size = 1024
+micro_batch_size = 1
+seq_len = 16384
+rollouts_per_prompt = 8
+mask_truncated_completions = false
+
+[sampling]
+temperature = 0.6
```
samsja, line 9: right now we have a bug with temperature that is not 1.0 afaik because vllm change their behavior (they do not divide by temp anymore when giving us the logprob)

samsja (approved):

PR #679, opened 2025-08-05
Refactor temperature handling

samsja:
Since vLLM v1 the exported logprobs are raw (i.e. no temperature, min_p, etc.) is applied. This PR applies temperature scaling to the vLLM logprobs before writing the batches to the trainer. Also, it removes any other sampling altering args from the sampling config on the orchestrator to avoid distribution drift (essentially assert that the only sampling arg can be set during training is temperature). Also isolates the sampling config used for eval.

### DIFF
--- a/src/prime_rl/trainer/loss.py
+++ b/src/prime_rl/trainer/loss.py
@@ -70,7 +70,6 @@ def grpo_loss_clip(
     clip_ratio: float,
 ) -> tuple[Tensor, RatioInfo]:
     assert shifted_logits.dtype == torch.float32, "shifted_logits must be float32"
-    shifted_logits = shifted_logits / temperature
     per_token_logps = selective_log_softmax(shifted_logits, input_ids)
 
     raw_ratio = torch.exp(per_token_logps - original_logprobs)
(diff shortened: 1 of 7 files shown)


mikasenghaas (commented):
do you also wanna remove temperature from the batch written by orchestrator all together then?

src/prime_rl/trainer/loss.py
```diff
@@ -70,7 +70,8 @@ def grpo_loss_clip(
     clip_ratio: float,
 ) -> tuple[Tensor, RatioInfo]:
     assert shifted_logits.dtype == torch.float32, "shifted_logits must be float32"
-    shifted_logits = shifted_logits / temperature
+    if temperature != 1.0:
+        shifted_logits = shifted_logits / temperature
```
mikasenghaas: why this?

samsja: just save a bit of memory

samsja:
> what do we do with stuff like `top_p`, `top_k`, etc.? do we just disallow those as training sampling args? if so, we should remove them from the orchestrator sampling config

yes I would say we disable them

mikasenghaas:
> > what do we do with stuff like `top_p`, `top_k`, etc.? do we just disallow those as training sampling args? if so, we should remove them from the orchestrator sampling config
> 
> yes I would say we disable them

okay, yea i agree i think. we can always bring back if we decide to switch to this project which exports non-raw logprobs from vllm

mikasenghaas:
Screenshot 2025-08-06 at 4 50 22 PM

PR #680, opened 2025-08-05
More detailed W&B logs + tensor metrics + loss refactor

mikasenghaas:
This PR refactors some of the W&B logs, especially from the trainer, to have a canonical way to log diverse point statistics (min/max/mean) from tensors (like `probs`, `old_probs`, `importance_ratio`, `entropy`, ...) for more observability during runs. To do this properly, it also required a slight refactor of the loss interface.

- Introduces the `Tensors` class with is used to accumulate local tensors and on a call to `compute_stats` all-gathers the tensors across ranks and steps to compute point metrics (min, max, mean, median, std)
- Refactors input and output of `compute_loss`. It now takes `logprobs` (computed in the main trainer loop), `old_logprobs` as arguments, computes and returns the per-token loss (to be reduced in main trainer loop), as well as a flexible `loss_tensors` dictionary which exports additional tensors for logging (e.g. `importance_ratio`, `clipped_importance_ratio` etc.) which are accumulated into `tensors: Tensors`
- Removed `compute_logprobs` function, instead use `forward`, `shift_logits` and `select_log_softmax` in main loop for logprob recomputation and grad acc loop
- More informative debug micro step logs

Also adds some extra and updates logging keys

- Removes `train`, `orchestrator`, `infer` from any logging keys (e.g.  (e.g. `perf/train/throughput` and `perf/infer/throughput` -> `perf/throughput`); we really don't need them - it's actually nicer to see the same throughput from trainer and orchestrator in one plot
- Adds logging for prompt and completion tokens with keys `prompt_len` and `completion_len`; sequence length is now `seq_len`. Logs `min`, `max`, `mean` and `std` (over full batch)
- Adds logging for `reward` and `advantage` including `min`, `mean` and `max` (over full batch)

## Overview

[W&B View](https://wandb.ai/primeintellect/mika-testing/workspace?nw=10dic2utkcp) with new logging keys.

## Hendryck's

*Run from #688 which is baed on #685 which is based on this PR*

Screenshot 2025-08-07 at 4 55 59 PM

## TODOs

- [x] Implement computing and logging tensor point estimates in a standardized way
- [x] We do not have `train`/ `infer` information in any of our keys, e.g. if there is a duplicate key both the trainer and orchestrator will simply log to that key (e.g. `perf/train/throughput` and `perf/infer/throughput` -> `perf/throughput`)
- [x] We add logging for prompt and completion tokens with keys `prompt_len` and `completion_len`
- [x] Check equivalence with 1 vs 2 trainer
- [x] Remove `sum` and `numel` from logging
- [x] Accumulate tensors on CPU and all-gather, then reduce on master rank
- [x] Use `batch`, `seq` and `vocab` for jaxtyping across repo

### DIFF
--- a/src/prime_rl/utils/monitor.py
+++ b/src/prime_rl/utils/monitor.py
@@ -174,8 +174,17 @@ class WandbMonitor(Monitor):
 
             if config.log_extras.distributions:
                 self.last_log_distributions_step = -1
-                # Incremental table is initialized dynamically in `log_distributions`
-                self.distributions_table = None
+                self.distributions_cols = [
+                    "step",
+                    "rewards",
+                    "advantages",
+                    "problem_rewards",
+                    "problem_advantages",
+                ]
+                self.distributions_table = wandb.Table(
+                    columns=self.distributions_cols,
+                    log_mode="INCREMENTAL",
+                )
                 self.distributions = []
 
     def _maybe_overwrite_wandb_command(self) -> None:
@@ -269,7 +278,9 @@ class WandbMonitor(Monitor):
         self.last_log_samples_step = step
         self.logger.debug(f"Logged samples at step {step} to W&B table in {time.time() - start_time:.2f}s")
 
-    def log_distributions(self, distributions: dict[str, list[float]], step: int) -> None:
+    def log_distributions(
+        self, rewards: list[float], advantages: list[float], rollouts_per_problem: int, step: int
+    ) -> None:
         if (
             not self.config.log_extras
             or not self.config.log_extras.distributions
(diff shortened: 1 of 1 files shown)


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -347,44 +348,76 @@ async def orchestrate(config: OrchestratorConfig):
...
+            "reward/min": rewards.min().item(),
+            "reward/max": rewards.max().item(),
             "step": progress.step,
         }
         monitor.log(reward_metrics)
 
+        advantage_metrics = {
+            "advantage/mean": advantages.mean().item(),
+            "advantage/min": advantages.min().item(),
+            "advantage/max": advantages.max().item(),
```
faresobeid: we can remove these, unnecessary

mikasenghaas: why?

faresobeid:
Looks good, just 3 remarks 
1. Should we log the probs instead of logprobs (more clear imo)
2. is the loss min/max (over all tokens) useful?
3. Is the entropy min/max useful?
4. Could we only log the clipped_importance_ratio and remove the importance_ratio (will lose info on how big the clipped tokens were)?
@samsja

src/prime_rl/trainer/loss.py
```diff
@@ -1,146 +1,106 @@
...
 @jaxtyped(typechecker=typechecker)
-def grpo_loss(
-    shifted_logits: Float[Tensor, "batch seq vocab"],
-    input_ids: Int[Tensor, "batch seq"],
-    advantages: Float[Tensor, "batch seq"],
-    original_logprobs: Float[Tensor, "batch seq"],
-    loss_mask: Int[Tensor, "batch seq"],
-    temperature: float,
+def compute_loss(
+    logprobs: Float[Tensor, "B L"],
```
samsja:
```suggestion
    logprobs: Float[Tensor, "batch seq"],
```
lets stick to batch seq everywhere

src/prime_rl/trainer/loss.py
```diff
@@ -1,146 +1,106 @@
...
-    raw_ratio = (raw_ratio.detach() - 1) * loss_mask
-    ratio = (coef_2.detach() - 1) * loss_mask
-
-    return loss, RatioInfo(
-        ratio_sum=ratio.sum().float(),
-        clipped_token_count=clipped_token_count,
-        raw_ratio_sum=raw_ratio.sum().float(),
-        raw_ratio_max=raw_ratio.max().float() + 1,
-        raw_ratio_min=raw_ratio.min().float() + 1,
-        raw_ratio_abs_sum=raw_ratio.abs().sum().float(),
```
samsja, line 99: where are we doing the +1 thingy now ?

src/prime_rl/trainer/loss.py
```diff
@@ -1,146 +1,106 @@
...
-    )
+    return summed_loss, {
+        "loss": loss.detach(),
+        "importance_ratio": importance_ratio.detach(),
+        "coef_1": coef_1.detach(),
+        "coef_2": coef_2.detach(),
+        "is_clipped": is_clipped.detach(),
+    }
```
samsja: so we are going to accumulate all this tensor in memory and reduce them later ?

mikasenghaas: no, three things:
- these tensors should be garbage collected after each micro step
- all tensors are (B, L) shape, so small
- the accumulation happens only on scalars that live on cpu
so bottomline is that nothing should change in terms of mem footprint

src/prime_rl/trainer/loss.py
mikasenghaas: its not necessary? feel free to check the w&b view but i didn't see any numerical issues with any reduced value. might be because we are doing fp32 logits now?

src/prime_rl/trainer/loss.py
mikasenghaas: will do this tmr!

samsja:
> Looks good, just 3 remarks
> 
> 1. Should we log the probs instead of logprobs (more clear imo)
> 2. is the loss min/max (over all tokens) useful?
> 3. Is the entropy min/max useful?
> 4. Could we only log the clipped_importance_ratio and remove the importance_ratio (will lose info on how big the clipped tokens were)?
>    @samsja

I don't see the point of logging less thing tbh. Maybe makes sense to log only the clip one indeed

PR #686, opened 2025-08-07
Use all gather in tensor metrics

samsja:
move from the all_reduce approach to a gather one where rank 0 gather every tensor over grad acc and gpu and then do distribution logging like max,min, mean but also median and quantile, can also do wandb histogram if we wanted to

### DIFF
--- a/src/prime_rl/trainer/loss.py
+++ b/src/prime_rl/trainer/loss.py
@@ -12,7 +12,6 @@ def compute_loss(
     logprobs: Float[Tensor, "B L"],
     old_logprobs: Float[Tensor, "B L"],
     advantages: Float[Tensor, "B L"],
-    loss_mask: Int[Tensor, "B L"],
     loss_config: LossConfig,
 ) -> tuple[Tensor, dict[str, Tensor]]:
     if loss_config.type == "clip":
@@ -20,7 +19,6 @@ def compute_loss(
             logprobs=logprobs,
             old_logprobs=old_logprobs,
             advantages=advantages,
-            loss_mask=loss_mask,
             epsilon_low=loss_config.epsilon_low,
             epsilon_high=loss_config.epsilon_high,
             clip_ratio=loss_config.clip_ratio,
@@ -30,7 +28,6 @@ def compute_loss(
             logprobs=logprobs,
             old_logprobs=old_logprobs,
             advantages=advantages,
-            loss_mask=loss_mask,
             clip_ratio=loss_config.clip_ratio,
         )
 
@@ -40,7 +37,6 @@ def grpo_loss_clip(
     logprobs: Float[Tensor, "B L"],
     old_logprobs: Float[Tensor, "B L"],
     advantages: Float[Tensor, "B L"],
-    loss_mask: Int[Tensor, "B L"],
     epsilon_low: float,
     epsilon_high: float,
     clip_ratio: float,
@@ -58,15 +54,9 @@ def grpo_loss_clip(
     loss = torch.max(loss_1, loss_2)
     is_clipped = (loss_1 < loss_2).float()
 
-    # Sum-reduce the loss for all unmasked tokens
-    summed_loss = (loss * loss_mask).sum()
-
-    return summed_loss, {
-        "loss": loss.detach(),
-        "importance_ratio": importance_ratio.detach(),
-        "coef_1": coef_1.detach(),
-        "coef_2": coef_2.detach(),
-        "is_clipped": is_clipped.detach(),
+    return loss, {
+        "importance_ratio/importance_ratio": coef_2,
+        "importance_ratio/is_clipped": is_clipped,
     }
 
 
@@ -75,7 +65,6 @@ def grpo_loss_ratio(
     logprobs: Float[Tensor, "B L"],
     old_logprobs: Float[Tensor, "B L"],
     advantages: Float[Tensor, "B L"],
-    loss_mask: Int[Tensor, "B L"],
     clip_ratio: float,
 ) -> tuple[Tensor, dict[str, Tensor]]:
     assert logprobs.dtype == torch.float32, "logprobs must be float32"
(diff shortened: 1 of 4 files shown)


src/prime_rl/trainer/loss.py
```diff
@@ -89,13 +78,10 @@ def grpo_loss_ratio(
...
-    summed_loss = (loss * loss_mask).sum()
 
-    return summed_loss, {
-        "loss": loss.detach(),
-        "importance_ratio": importance_ratio.detach(),
-        "clipped_importance_ratio": clipped_importance_ratio.detach(),
-        "is_clipped": is_clipped.detach(),
+    return loss, {
+        "importance_ratio/importance_ratio": clipped_importance_ratio,
+        "importance_ratio/is_clipped": is_clipped,
```
mikasenghaas: wait why are we not exporting the other tensors? now we are not logging `loss/{min,mean,max}`

mikasenghaas: also i think we should embrace the sections and have each tensor that we log have its own section otherwise it will get messy soon with the amount of point metrics that we are logging now, so no nesting like `importance_ratio/...` imo

src/prime_rl/trainer/loss.py
```diff
@@ -138,7 +124,6 @@ def selective_log_softmax(logits: Float[Tensor, "B L V"], index: Int[Tensor, "B
 
 
 @jaxtyped(typechecker=typechecker)
-@torch.no_grad()
```
mikasenghaas, line 141: oh i liked this, looks much cleaner in trainer code. any reason for this?

src/prime_rl/trainer/train.py
```diff
@@ -189,7 +188,7 @@ def train(config: TrainerConfig):
         # Optionally, compute the logprobs for the training batch
         compute_logprobs_time = 0
         num_micro_batches = len(micro_batches)
-        recomputed_logprob_errors = [torch.zeros_like(mb["logprobs"], device="cuda") for mb in micro_batches]
+        recomputed_logprob_errors = [torch.ones_like(mb["logprobs"], device="cuda") for mb in micro_batches]
```
mikasenghaas, line 191: i actually somewhat liked that the default is zero simply to show that we did not recompute. but yea maybe that's the more sane default given that we are logging a ratio

src/prime_rl/trainer/loss.py
```diff
@@ -12,15 +12,13 @@ def compute_loss(
     logprobs: Float[Tensor, "B L"],
     old_logprobs: Float[Tensor, "B L"],
     advantages: Float[Tensor, "B L"],
-    loss_mask: Int[Tensor, "B L"],
     loss_config: LossConfig,
 ) -> tuple[Tensor, dict[str, Tensor]]:
```
mikasenghaas: let's jaxtype the return type as well now that it is a B L tensor?

src/prime_rl/trainer/train.py
```diff
@@ -260,65 +259,33 @@ def train(config: TrainerConfig):
...
+            loss_tensors["importance_ratio/recomputed_logprob_error"] = recomputed_logprob_errors[micro_step]
+            loss_tensors["importance_ratio/old_proba"] = torch.exp(old_logprobs)
+            loss_tensors["importance_ratio/proba"] = torch.exp(logprobs)
+            loss_tensors["entropy"] = entropy
+            loss_tensors["loss"] = per_token_loss
 
-            if micro_step == 0:
-                logger.debug("Logging loss tensors at first micro step")
-                for k, v in loss_tensors.items():
-                    logger.debug(f"\t{k}={lt.lovely(v)}")
```
mikasenghaas, line 317: RIP

mikasenghaas: why don't you like this? im thinking it might be nice to log the micro statistics to console in debug mode at least, o/w we have literlaly no way of knowing what is going on in the micro steps. will not be shown by default in info log mode ofc

src/prime_rl/trainer/utils.py
```diff
@@ -114,64 +115,75 @@ def print_benchmark(history: dict[str, list[Any]]) -> None:
...
+            metrics[f"{k}/std"] = tensor.std().item()
+            metrics[f"{k}/min"] = tensor.min().item()
+            metrics[f"{k}/max"] = tensor.max().item()
+            metrics[f"{k}/median"] = torch.median(tensor).item()
+            metrics[f"{k}/1%"] = torch.quantile(tensor, 0.01).item()
+            metrics[f"{k}/5%"] = torch.quantile(tensor, 0.05).item()
+            metrics[f"{k}/10%"] = torch.quantile(tensor, 0.1).item()
+            metrics[f"{k}/90%"] = torch.quantile(tensor, 0.9).item()
+            metrics[f"{k}/95%"] = torch.quantile(tensor, 0.95).item()
+            metrics[f"{k}/99%"] = torch.quantile(tensor, 0.99).item()
```
mikasenghaas: this seems like a LOT. im all for observability but having 11 plots for every tensor that we log (especially given that they all go in the same section atm) seems too much. also we seem to have lost the ability to specify which reductions we are interested, but ig we can also make the logging more verbose

src/prime_rl/trainer/loss.py
samsja: idk why it was here tbh

src/prime_rl/trainer/train.py
samsja: hmmmm idk because the perfect score is 1 so would makes sense that we use 1 here

src/prime_rl/trainer/train.py
samsja: feel free to add it back, was just me playing around and it was hard to read haha

src/prime_rl/trainer/utils.py
samsja: Its defo a lot but also the more concise, tbh max and min are useless most of the time. 

I think we should find a way to have a junk section where we put everything that we don't want to immediatily look at

PR #687, opened 2025-08-07
Log tensor stats into two sections

mikasenghaas:
On top of #680, but logs all mean stats into section `tensor/...` and all others into `tensor_details/...`


mikasenghaas:
Close because team favors #680 5:1

PR #690, opened 2025-08-08
add muon

samsja:


pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     "textarena>=0.6.16",
     "nltk>=3.9.1",
     "math-verify>=0.8.0",
+    "muon_fsdp2>=0.3.0"
```
mikasenghaas, line 35: is it that much code that it needs a package?

src/prime_rl/trainer/config.py
```diff
@@ -80,6 +80,8 @@ class CosineSchedulerConfig(BaseModel):
 
 SchedulerConfig: TypeAlias = ConstantSchedulerConfig | LinearSchedulerConfig | CosineSchedulerConfig
 
+OptimizerType: TypeAlias = Literal["AdamW", "Muon"]
```
mikasenghaas: can we keep convention here and do `adamw` and `muon`

src/prime_rl/trainer/config.py
```diff
@@ -89,6 +91,8 @@ class OptimizerConfig(BaseConfig):
     betas1: Annotated[float, Field(ge=0)] = 0.9
     betas2: Annotated[float, Field(ge=0)] = 0.999
 
+    name: Annotated[OptimizerType, Field(description="The optimizer to use. By default use AdamW")] = "AdamW"
+
```
mikasenghaas: i think we should probably already move to the discriminated union type pattern, because not all optimizers will have the same configs, e.g. sgd does not have the betas

src/prime_rl/trainer/optim.py
```diff
@@ -0,0 +1,45 @@
+import torch
+from muon_fsdp2 import Muon
+from torch.optim import Optimizer
+
+from prime_rl.trainer.config import OptimizerConfig
+from prime_rl.trainer.model import Model
+
+
+def create_optimizer(config: OptimizerConfig, model: Model) -> Optimizer:
```
mikasenghaas: setup?

pyproject.toml
samsja: want to make it reusable by other yeah

src/prime_rl/trainer/config.py
samsja: sure

src/prime_rl/trainer/config.py
```diff
@@ -78,19 +78,27 @@ class CosineSchedulerConfig(BaseModel):
     ] = None
 
 
-SchedulerConfig: TypeAlias = ConstantSchedulerConfig | LinearSchedulerConfig | CosineSchedulerConfig
+SchedulerType: TypeAlias = ConstantSchedulerConfig | LinearSchedulerConfig | CosineSchedulerConfig
```
mikasenghaas: nah this is `SchedulerConfig` - `SchedulerType` would be just the literal

src/prime_rl/trainer/config.py
```diff
@@ -78,19 +78,27 @@ class CosineSchedulerConfig(BaseModel):
...
+class MuonConfig(BaseModel):
+    type: Literal["muon"] = "muon"
+    lr: Annotated[float, Field(ge=0)] = 4e-4
+    weight_decay: Annotated[float, Field(ge=0)] = 0.01
+
+    adam_betas1: Annotated[float, Field(ge=0)] = 0.9
+    adam_betas2: Annotated[float, Field(ge=0)] = 0.999
+
+
+OptimizerType: TypeAlias = AdamWConfig | MuonConfig
```
mikasenghaas: this is also `OptimizerConfig`

mikasenghaas (approved):

src/prime_rl/trainer/sft/config.py
```diff
@@ -50,14 +50,10 @@ class SFTTrainerConfig(BaseSettings):
...
-    optim: Annotated[
-        OptimizerType, Field(discriminator="type", description="The optimizer to use. By default use AdamW")
-    ] = AdamWConfig()
+    optim: Annotated[OptimizerConfigType, Field(discriminator="type")] = AdamWConfig()
 
     # The learning rate scheduler configuration
-    scheduler: Annotated[
-        SchedulerType, Field(discriminator="type", description="The learning rate scheduler to use.")
-    ] = ConstantSchedulerConfig()
+    scheduler: Annotated[SchedulerConfigType, Field(discriminator="type")] = ConstantSchedulerConfig()
```
samsja, line 56: shoudl we call this lr_scheduler, less obivous what it is now that its outside the optimizer

mikasenghaas: mmh then i would rather move it back in, we have shared attrs anyways with `lr` and `max_norm` in #716

PR #696, opened 2025-08-08
Bump vLLM to v0.10.0

mikasenghaas:

PR #697, opened 2025-08-08
Fix temperature scaling of vLLM logprobs

mikasenghaas:
> This PR is based on #696 and should not be merged before

This PR fixes a bug in our temperature-scaling of raw vLLM logprobs. Previously we did `logprobs /= temperature` for scaling, which is just plain wrong. Temperature scaling is applied before the softmax, but because we only have access to chosen token logprobs, we cannot reverse this process. Luckily, v0.10.0 allows us to specify processed logprobs via `--logprob-mode processed_logprobs` which we now set as a default, and raise if any other value is set . Funnily, vLLM's definition of "processed" does not include temperature scaling so we also have to monkeypatch the `Sampler` class to apply temperature scaling before (not after) exporting the `processed_logprobs` . With this we get the expected behavior of the importance ratio at lower temperature levels for `hendrycks-math`

Screenshot 2025-08-08 at 5 39 06 PM

All runs are 10 steps on `hendrycks-math` with 1B model with no recompute, ie. using vLLM logprobs:
- Green: How the importance ratio should look (around 1) at temperature 1.0 (baseline)
- Pink: What we have on current main, completely wrong temperature scaling (we scaled logprobs, not logits oops)
- Blue: After bump to v0.10.0 exproting `processed_logprobs` (new vllm feature), still off because "processed" does not include temperature scaling in vllm's mind (wtf?)
- Orange: After a one-line patch to the sampler we are finally good

### DIFF
--- a/src/prime_rl/trainer/train.py
+++ b/src/prime_rl/trainer/train.py
@@ -213,7 +213,6 @@ def train(config: TrainerConfig):
                     shifted_logits = shift_logits(logits)
                     shifted_logits = shifted_logits / temperature
                     recomputed_logprobs = selective_log_softmax(shifted_logits, input_ids)
-                    del logits, shifted_logits
 
                     # Compute the recomputed logprob error
                     recomputed_logprob_error = torch.exp(recomputed_logprobs - logprobs)
--- a/src/prime_rl/inference/config.py
+++ b/src/prime_rl/inference/config.py
@@ -107,4 +107,7 @@ class InferenceConfig(BaseSettings):
             value = rgetattr(self, key.replace("-", "_"))
             rsetattr(namespace, to_vllm.get(key, key), value)
 
+        # Set `logprobs_mode` to `processed_logprobs` by default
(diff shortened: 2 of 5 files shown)


src/prime_rl/trainer/train.py
```diff
@@ -213,7 +213,6 @@ def train(config: TrainerConfig):
                     shifted_logits = shift_logits(logits)
                     shifted_logits = shifted_logits / temperature
                     recomputed_logprobs = selective_log_softmax(shifted_logits, input_ids)
-                    del logits, shifted_logits
```
Jackmin801, line 216: why dont we need this anymore? would this not save memory?

mikasenghaas: yes but i figured we don't really care in the logprob recomputation loop because our mem spike is during .backward() in the main grad acc loop, and we use the same variables `logits` and `shifted_keys` which we del before backward

mikasenghaas:
> lgtm! would be nice to PR moving the apply_temperature up to vllm. guessing this is useful for many

You mean PRing this into vLLM?

PR #701, opened 2025-08-09
Add CI test for ckpt resuming

mikasenghaas:
As advertised. There's quite some intricate logic in the ckpt resuming and I would like to at least catch stupid bugs in CI. We now do 20 steps and write final ckpt for the first run, and then another 20 resumed in CI for a total of steps (before we did 30 steps). Given that server startup takes most of the time and we persist the inference server across the whole session, the total testing time should not be affected much.

Also fixes a minor bug where we write the final checkpoint twice (once inside the loop) and once outside if `ckpt.interval_steps` divides `max_steps`. Now adds a guard that checks that we are not on the last step inside the loop.

Screenshot 2025-08-09 at 11 50 29 AM

Screenshot 2025-08-09 at 11 50 32 AM

### DIFF
--- a/configs/reverse_text/orch.toml
+++ b/configs/reverse_text/orch.toml
@@ -2,7 +2,7 @@ batch_size = 128
 micro_batch_size = 16
 rollouts_per_prompt = 16
 seq_len = 128
-max_steps = 30
+max_steps = 20
 mask_truncated_completions = false
 
 [model]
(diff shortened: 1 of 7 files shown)


Jackmin801 (approved):
lgtm! the test just makes sure it didnt crash and we check the wandb to see that the resume curve looks ok?

mikasenghaas:
yepp. ig we could make it more elaborate and parse log files and check final reward or smth, but idk if really necessary for now

mikasenghaas:
@cursor review

PR #705, opened 2025-08-10
Implement checkpoint cleanup logic

mikasenghaas:
Implement checkpoint cleanup logic with a new `--ckpt.keep` option to retain a specified number of recent step checkpoints.

This PR adds an optional `--ckpt.keep` configuration key (int | None) to the `CheckpointManager` for both trainer and orchestrator. If set to an integer, it will automatically delete older `step_{x}` checkpoint directories, ensuring only the `keep` most recent ones are preserved. This addresses the lack of cleanup logic and helps manage disk space. The default behavior (`None`) is to keep all checkpoints. Each local trainer and orchestrator is responsible for managing its own checkpoint. This also means that we never delete the parent step directory, but this should be fine imo

Here is an example of only retaining two checkpoints:

```bash
uv run rl  \
  --trainer @ configs/reverse_text/train.toml \
   --ckpt.keep 2 \
   --ckpt.interval 2 \
   --log.level debug \
   --outputs-dir /workspace/outputs
```

Screenshot 2025-08-10 at 4 17 42 PM

---
Linear Issue: [RES-784](https://linear.app/primeintellect/issue/RES-784/implement-cleaning-checkpoint-logic)

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,7 +412,7 @@ checkpoints
     └── trainer.pt
 ```
 
-Checkpointing is configured by the `CheckpointConfig`, with the config key `--ckpt`. One can specify the interval (`--ckpt.interval`, defaults to `50`) and whether to save checkpoints asynchronoously  (`--ckpt.save-async`, defaults to `False`)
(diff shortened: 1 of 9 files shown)


mikasenghaas:
Screenshot 2025-08-10 at 4 23 34 PM
Screenshot 2025-08-10 at 4 23 43 PM

Resume still functional in CI

PR #707, opened 2025-08-11
Make no recompute the default

faresobeid:
No recompute is giving us very stable runs and allows us to use vllm v1.
Also is technically recompute but with the mismatch fix from https://fengyao.notion.site/off-policy-rl


mikasenghaas (approved):
we love to see it

samsja (approved):
Lgtm

PR #708, opened 2025-08-11
Add SFT support

mikasenghaas:
This PR adds SFT support into `prime-rl` as a trainer submodule.

TODOs:
- [x] Add logging for accuracy
- [x] Validate gradient accumulation loop
- [x] Make multi-GPU compatible
- [x] Log epoch information
- [x] Make multi-node compatible
- [x] Write checkpoint to disk allowing to directly start and RL run from the SFT-warmed up model
- [x] Check resuming from ckpt
- [x] Verify full pipeline with SFT warmup + RL on `reverse-text`
- [x] Add small-scale SFT dataset + align RL dataset format
- [x] Add SFT CI run
- [x] Implement packing
- [x] Write docs
- [x] Fix epoch information wrong in `packing`

Check out the runs linked in [SFT W&B View](https://wandb.ai/primeintellect/mika/workspace?nw=m4yhbnmn7sb), [RL W&B View](https://wandb.ai/primeintellect/mika/workspace?nw=6ygwdqfmrxa)

## Basic Training

We train `Qwen/Qwen3-0.6B` on SFT data from `willcb/R1-reverse-wikipedia-paragraphs-v1-1000` with batch size `16`, sequence length `4096`, constant learning rate of `2e-5` over 100 steps, checkpointing at step 50 and 100. Loss, norm and accuracy curves seem to ~match the curves from HF's SFTTrainer.

```bash
uv run torchrun --nproc-per-node 8 src/prime_rl/trainer/sft/train.py --max-steps 100 --data.batch-size 16 --data.micro-batch-size 1 --data.seq-len 4096 --optim.lr 2e-5 --ckpt.interval 50 --log.level debug
```

Screenshot 2025-08-13 at 12 57 16 PM

## Validating Correctness

We verify that training dynamics are invariant to the number of gradient accumulation steps, here micro batch size `1` vs. `4`

Screenshot 2025-08-13 at 12 57 37 PM

And here with a singe, four or eight trainers. Also the throughput increases linearly with the number of devices.

Screenshot 2025-08-13 at 12 57 00 PM

## Padding and Packing

Both work but packing is just better in every regard. Better MFU, faster convergence, less memory and faster epoch iterations

Screenshot 2025-08-13 at 1 04 09 PM

## Resuming Training

We can resume a run using the same component as our RL trainer. Here, we are resuming from 25 steps.

Screenshot 2025-08-13 at 1 00 32 PM

## Starting RL from SFT warmup

Finally, we check that the SFT warmup works for RL. As we can see the base model never gets any reward, resulting in no RL training signal and stagnation. Our 100 step SFT model from the `prime-rl` trainer has a lower initial reward but enough for the RL training to get started. Overall, the training dynamic is very similar to @willccbb's model which we are currently using in CI.

### DIFF
--- a/tests/integration/test_rl.py
+++ b/tests/integration/test_rl.py
@@ -51,8 +51,8 @@ def wandb_project(username: str) -> str:
 
 
 @pytest.fixture(scope="module")
-def outputs_dir(tmp_path: Path) -> Path:
-    return tmp_path / "outputs"
+def outputs_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
+    return tmp_path_factory.mktemp("outputs")
 
 
 @pytest.fixture(scope="module")
--- a/tests/integration/test_sft.py
+++ b/tests/integration/test_sft.py
@@ -32,8 +32,8 @@ def wandb_project(username: str) -> str:
 
 
 @pytest.fixture(scope="module")
-def outputs_dir(tmp_path: Path) -> Path:
-    return tmp_path / "outputs"
+def outputs_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
+    return tmp_path_factory.mktemp("outputs")
 
 
 @pytest.fixture(scope="module")


src/prime_rl/trainer/sft/config.py
```diff
@@ -0,0 +1,124 @@
...
+
+from prime_rl.trainer.config import CheckpointConfig, ModelConfig, OptimizerConfig, WeightCheckpointConfig
+from prime_rl.utils.config import LogConfig, MultiMonitorConfig
+from prime_rl.utils.pydantic_config import BaseConfig, BaseSettings
+
+
+class FakeDataConfig(BaseConfig):
+    """Configures a fake dataset for debugging."""
+
+    n: Annotated[int, Field(ge=1)] = 1000
```
samsja: is n the number of fake sample ?

src/prime_rl/trainer/sft/train.py
```diff
@@ -0,0 +1,284 @@
...
+    monitor = setup_monitor(config.monitor, outputs_dir=config.outputs_dir, run_config=config)
+
+    # Set precision and cuda device
+    torch.set_float32_matmul_precision("high")
+    torch.cuda.set_device(world.rank)
+
+    # Initialize the model and tokenizer
+    logger.info(f"Initializing model and tokenizer ({config.model})")
+    model = setup_model(config.model)
+    tokenizer = get_tokenizer(config.model)
```
samsja, line 54: should we rename get_tokenizer setup_tokenizer ( not in this pr but name read off now)

src/prime_rl/trainer/sft/train.py
```diff
@@ -0,0 +1,284 @@
...
+    logger.info(f"Initializing optimizer ({config.optim})")
+    optimizer = torch.optim.AdamW(
+        params=model.parameters(),
+        lr=config.optim.lr,
+        weight_decay=config.optim.weight_decay,
+        betas=(config.optim.betas1, config.optim.betas2),
+    )
+
+    # Set up the learning rate scheduler
+    scheduler = create_lr_scheduler(optimizer, config.optim.scheduler, config.max_steps)
```
samsja, line 66: naming wise should we stick with setup ? 

setup_lr_scheduler ?

(also should rename in rl.py, doesn't have to be in this pr)

src/prime_rl/trainer/sft/config.py
mikasenghaas: yep, we need it here because we enforce that the SFT trainer is epoch aware so the dataset needs to be finite (unlike the RL trainer which can train infinitely)

src/prime_rl/trainer/sft/train.py
mikasenghaas: yea, i like that. agree that not in this pr tho

src/prime_rl/trainer/sft/train.py
mikasenghaas: yes setup everywhere, except if we have a singleton pattern. then we should have setup_x and get_x (like logger and monitor)

mikasenghaas:
> I cannot find whre you split the data by rank

distributed sampler makes this super easy

src/prime_rl/trainer/rl/config.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+class RatioLossConfig(BaseLossConfig):
+    """Configures the ratio loss."""
+
+    type: Literal["ratio"] = "ratio"
+    clip_ratio: Annotated[float, Field(ge=0)] = 8.0
+
+
+LossConfig: TypeAlias = ClippingLossConfig | RatioLossConfig
```
samsja, line 31:
```suggestion
LossConfigType: TypeAlias = ClippingLossConfig | RatioLossConfig
```
:)

mikasenghaas: separate pr?

src/prime_rl/trainer/sft/train.py
```diff
@@ -0,0 +1,285 @@
...
+            dataloader = get_dataloader(dataset, tokenizer, config=config.data)
+            epoch += 1 if progress.step > 0 else 0  # Only increment epoch if we are not at the first step
+
+        if config.profile_path and world.rank == 0:
+            torch.cuda.memory._record_memory_history()
+
+        step_start_time = time.time()
+        forward_backward_start_time = time.time()
+        tensors = Tensors()  # Used to accumulate tensor statistics across grad acc and ranks for logging
+        grad_accum_steps = config.data.batch_size // (config.data.micro_batch_size * world.world_size)
```
samsja, line 123: did we assert that batch_size is divisble by micro_batch_size * world.world_size somewhere ?

mikasenghaas: we do `drop_last` on the micro batch data loader which ensures this, but can add a manual assert to be 100% sure

mikasenghaas: oh wait you're talking abt a different assert, yea we check this in pydantic

samsja: you check world size in pydantic ?

src/prime_rl/trainer/sft/data.py
```diff
@@ -0,0 +1,181 @@
...
+                    packed_sample["loss_mask"].extend(sample["loss_mask"])
+                    packed_sample["position_ids"].extend(sample["position_ids"])
+                    packed_sample_found = True
+                    break
+
+            # If no suitable packed sample found, create a new packed sample
+            if not packed_sample_found:
+                packed_samples.append(sample)
+
+        return packed_samples
```
samsja: hmm this is dangerous, this means we don't have the same original distribution when doing padding and packing, as it is right now we will start training on longer sequence first in the epoch.

willccbb: shuffle after?

src/prime_rl/trainer/sft/train.py
```diff
@@ -0,0 +1,285 @@
...
+    )
+
+    # Set up the dataset (optionaly, use a fake dataset for debugging)
+    logger.info(f"Initializing dataset ({config.data})")
+    dataset = get_dataset(tokenizer, config=config.data)
+    logger.info(f"Initialized {dataset.__class__.__name__} for `{config.data.name}` with {len(dataset)}{' packed' if config.data.collate_mode == 'packing' else ''} samples")
+
+    logger.info(f"Starting training loop ({config.max_steps=})")
+    is_first_step = True
+    steps_per_epoch = len(dataset) // config.data.batch_size
```
samsja: this won't work for sequence packing

src/prime_rl/orchestrator/config.py
```diff
@@ -117,7 +117,7 @@ def validate_rollouts_per_prompt(self):
 class CheckpointConfig(BaseConfig):
     """Configures checkpointing the orchestrator."""
 
-    interval: Annotated[int, Field(ge=1, description="Interval at which to save the checkpoint.")] = 50
+    interval: Annotated[int, Field(ge=1, description="Interval at which to save the checkpoint.")] = None
```
Jackmin801:
```suggestion
    interval: Annotated[int | None, Field(ge=1, description="Interval at which to save the checkpoint.")] = None
```

src/prime_rl/trainer/config.py
```diff
@@ -89,14 +87,17 @@ class OptimizerConfig(BaseConfig):
...
+
     # LR Scheduler configuration
     scheduler: SchedulerConfig = Field(discriminator="type", default=ConstantSchedulerConfig())
 
 
 class CheckpointConfig(BaseConfig):
     """Configures checkpointing the full model, optimizer and training state for resuming training."""
 
-    interval: Annotated[int, Field(ge=1, description="Interval at which to save the checkpoint.")] = 50
+    interval: Annotated[int, Field(ge=1, description="Interval at which to save the checkpoint. If None, will only checkpoint at the end of training.")] = None
```
Jackmin801:
```suggestion
    interval: Annotated[int | None, Field(ge=1, description="Interval at which to save the checkpoint. If None, will only checkpoint at the end of training.")] = None
```

samsja (approved):

PR #710, opened 2025-08-12
Implement length bonus

faresobeid:
Give an extra reward to the shortest rollout if all rollouts for that prompt are correct
Some early results with a 0.5 extra reward on acereason stage 1, can see the reasoning length is much shorter and eval performance is not bad (might be lagging behind normal training but could be down to many other factors)
image
image
image
image


src/prime_rl/orchestrator/config.py
```diff
@@ -305,6 +305,13 @@ class OrchestratorConfig(BaseSettings):
         ),
     ] = False
 
+    apply_shortest_correct_bonus: Annotated[
+        float | None,
```
samsja: I can tell @mikasenghaas won't be happy with this name

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -223,9 +228,19 @@ async def orchestrate(config: OrchestratorConfig):
...
 
+            if config.length_bonus is not None:
+                train_rewards = apply_length_bonus(
+                    rewards=processed_outputs.rewards,
+                    completion_lengths=list(map(len, processed_outputs.completion_ids)),
+                    rollouts_per_prompt=config.rollouts_per_prompt,
+                    bonus=config.length_bonus,
+                )
+            else:
+                train_rewards = processed_outputs.rewards
```
mikasenghaas: would prefer if made a more general function `process_rewards: list[float] -> list[float]` which for now only does length bonus and is an noop if not set. then we can remove the ugly if else here

src/prime_rl/orchestrator/utils.py
```diff
@@ -124,6 +124,27 @@ def compute_rewards(
...
+) -> list[float]:
+    """Return a new reward list where the shortest *correct* rollout(s) in each
+    fully correct group receive a bonus."""
+
+    assert len(rewards) == len(completion_lengths), "Rewards and lengths must align"
+
+    new_rewards = list(rewards)
+    for start in range(0, len(rewards), rollouts_per_prompt):
+        group_rewards = new_rewards[start : start + rollouts_per_prompt]
+        if sum(group_rewards) == rollouts_per_prompt:
```
mikasenghaas, line 66: this seems to assume that perfect reward == 1.0 which might not be true. what if we dont have binary rewards? or we compose multiple rewards?

src/prime_rl/orchestrator/utils.py
```diff
@@ -124,6 +124,27 @@ def compute_rewards(
...
+            group_lengths = completion_lengths[start : start + rollouts_per_prompt]
+            min_len = min(group_lengths)
+            for idx, length in enumerate(group_lengths):
+                if length == min_len:
+                    new_rewards[start + idx] += bonus
```
mikasenghaas: this also means that potentially more than one seq gets the bonus, in the case where multiple sequences share the same min length right? is this intended?

mikasenghaas (commented):
i actually like the name `length_bonus`. couple of comments but almost there

src/prime_rl/orchestrator/utils.py
faresobeid: Oh good point will fix although i would think its pretty unlikely. Not sure though if it is a bad feature

src/prime_rl/orchestrator/utils.py
faresobeid: Ya it is only intended for binary, and in training all rewards should be binary 1 or 0 imo

samsja (approved):
lgtm

mikasenghaas (approved):

PR #722, opened 2025-08-14
Misc cleanup

mikasenghaas:
Miscellaneous cleanup, including:
- Moving to proper singleton pattern for logger
- Setup function for `WeightCkptManager` and `CkptManager`
- Fixed unbound variable type errors
- Fixed some wrong type hints here and there

### DIFF
--- a/src/prime_rl/trainer/weights.py
+++ b/src/prime_rl/trainer/weights.py
@@ -7,7 +7,7 @@ import torch
 from torch import Tensor
 from torch.distributed.checkpoint.state_dict import _get_fqns as get_fqns
 from torch.distributed.tensor import DTensor
-from transformers import AutoTokenizer
+from transformers.tokenization_utils import PreTrainedTokenizer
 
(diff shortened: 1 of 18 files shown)


src/prime_rl/trainer/weights.py
```diff
@@ -139,3 +141,12 @@ def maybe_clean(self, step: int):
...
             self._maybe_clean(step)
+
+
+def setup_weight_ckpt_manager(
+    outputs_dir: Path,
+    weight_ckpt_config: WeightCheckpointConfig,
+    ckpt_config: CheckpointConfig | None,
+    async_level: int,
+) -> WeightCheckpointManager:
+    return WeightCheckpointManager(outputs_dir, weight_ckpt_config, ckpt_config, async_level=async_level)
```
samsja, line 152: I don't see the point here

mikasenghaas: just convention

PR #695, opened 2025-08-08
Use verifiers for evals

willccbb:
This PR adds support for doing online evals from the orchestrator and offline evals in `eval` entrypoint using `verifiers` environments with initial support for AIME24, AIME25 and MATH-500. 

```bash
uv run inference  --model.name willcb/DeepSeek-R1-Distill-Qwen-1.5B
```

**prime-rl**

```bash
uv run eval --model.name willcb/DeepSeek-R1-Distill-Qwen-1.5B --environment-ids vf-math500 --rollouts-per-example 4 --sampling.temperature 0.7 --sampling.max-tokens 2048
```

```txt
Evaluated vf-math500 in 281.53s (Avg@4=0.36, Pass@4: 0.55, Seq. Len: 1832.04, Max Seq. Len: 2048.00, Min Seq. Len: 350.50)
```

**vf-eval**

```bash
vf-eval vf-math500 -b http://localhost:8000/v1 -m willcb/DeepSeek-R1-Distill-Qwen-1.5B -n 500 -r 4 -t 2048
```

```txt
reward: avg - 0.368, std - 0.482
```

Misc changes:
- Changed `rollouts_per_prompt` -> `rollouts_per_example` for alignment w/ verifiers

TODOs:
- [x] Change verifiers pin to main after merging [this](https://github.com/willccbb/verifiers/pull/213)
- [x] Re-run `hendrycks-math` and `intellect-math` baselines with new math rubric

---

**GitHub Issue**: #754 
**Linear Issue**: Resolves PRIMERL-28

PR #724, opened 2025-08-15
Improve benchmark suite

mikasenghaas:
Fixes and improvements to the benchmark suite on the SFT trainer (some also apply to RL trainer):
- Fix `print_benchmark` trying to parse outdates logging keys
- Also print MFU in benchmark table
- Bring back layer-wise partial activation checkpointing with `--model.ac.freq` which is an integer specifying that ever `freq`-th layer will be ACd. Confirmed that in specific settings checkpointing half the layers gives best MFU.
- More sophisticated fake data loader for benchmarking with mode `fixed` (all samples will be exactly `seq_len`) and `variable` (all samples have uniformly random length in interval [0, seq_len])

### DIFF
--- a/src/prime_rl/trainer/sft/config.py
+++ b/src/prime_rl/trainer/sft/config.py
@@ -100,6 +100,10 @@ class SFTTrainerConfig(BaseSettings):
     def auto_setup_bench(self):
         if self.bench:
             self.max_steps = 4  # 1 Warmup + 3 Benchmark
+            if self.monitor.wandb: # Do not log extras
+                self.monitor.wandb.log_extras = None
+            if self.ckpt: # Do not checkpoint
+                self.ckpt = None
         return self
 
     @model_validator(mode="after")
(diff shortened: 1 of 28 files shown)


src/prime_rl/trainer/sft/config.py
```diff
@@ -29,7 +44,12 @@ class DataConfig(BaseConfig):
     seq_len: Annotated[int, Field(ge=1)] = 128
     shuffle: Annotated[bool, Field(description="Whether to shuffle the dataset at the beginning of each epoch.")] = True
 
-    fake: Annotated[bool, Field(description="Whether to use a fake dataset.")] = False
+    fake: Annotated[
+        FakeDataConfigType | None,
+        Field(
+            description="How to generate fake data, mostly used for benchmarking. If None, will use the regular dataset."
+        ),
+    ] = None
```
samsja: this seems a bit 
```suggestion
    fake: Annotated[
        Literal["fixed","variabe"] | None,
        Field(
            description="How to generate fake data, mostly used for benchmarking. If None, will use the regular dataset."
        ),
    ] = None
```

samsja (approved):

willccbb:
can we standardize naming to "evals" instead of "benchmarks" to match with #695?

mikasenghaas:
> can we standardize naming to "evals" instead of "benchmarks" to match with #695?

oh benchmark in this pr actually refers to performance benchmarking, tps, mfu etc.

Issue #744, opened 2025-08-16
Write eval artifacts


mikasenghaas:
We should make use of `verifiers` `Environment.make_dataset` functionality to save the eval results (e.g. prompt, completion and rewards for all examples) as a HF dataset to disk. This will make it easier to manually assess the eval results during offline and online evals. Hence, this should likely be implemented inside the util functon to run evals.

PR #767, opened 2025-08-17
Add eval artifacts

mikasenghaas:
This PR adds the option to save eval artifacts, including a HF dataset of the eval samples (useful to push to hub to use conversation view to manually inspect the completions) and a JSON-dump of the computed metrics.

- Added `--outputs-dir` to `OfflineEvalConfig` which defaults to `outputs` as everywhere
- Added boolean `--save` to `EvalConfig` used in both online and offline evals, which defaults to `True`

The path is auto-generated as `{output_dir}/evals/step_{x}/{eval_id}` where `eval_id` is the ID of the environment we eval against. 

Misc:
- I removed all the completion length logs from W&B and moved to the offline report to declutter the `eval` W&B log section. We should only see scores there by default, especially as the number of our evals grows

---

**GitHub Issue**: #744 
**Linear Issue**: Resolves PRIMERL-18

### DIFF
--- a/src/prime_rl/eval/config.py
+++ b/src/prime_rl/eval/config.py
@@ -30,13 +30,6 @@ class OfflineEvalConfig(EvalConfig, BaseSettings):
         ),
     ] = Path("outputs")
 
-    save: Annotated[
-        bool,
-        Field(
-            description="Whether to save the evaluation artifacts to the outputs directory.",
-        ),
-    ] = False
-
     use_tqdm: Annotated[
         bool,
         Field(
--- a/src/prime_rl/eval/config.py
+++ b/src/prime_rl/eval/config.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from typing import Annotated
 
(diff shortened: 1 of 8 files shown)


samsja (approved):

src/prime_rl/eval/config.py
```diff
@@ -22,6 +23,13 @@ class OfflineEvalConfig(EvalConfig, BaseSettings):
     # The logging configuration
     log: LogConfig = LogConfig()
 
+    outputs_dir: Annotated[
```
willccbb, line 26: would slightly prefer if we call this `output_dir` instead of `outputs_dir` for consistency with the convention in the HF ecosystem

willccbb: Can handle this in separate issue

https://github.com/PrimeIntellect-ai/prime-rl/issues/769

LGTM

willccbb (approved):

Issue #768, opened 2025-08-17
Support evaling all checkpoints


mikasenghaas:
Currently, one can only eval a single model checkpoint (e.g. HF model name or path) and has to re-start the inference server everytime. When eval'ing local checkpoints written during a SFT or RL run into `weights/step_{x}`, it would be more convenient to only specify the weight directory to eval which would automatically eval all step subdirectories. We should use similar logic as on the orchestrator for calling `update_weights` dynamically.

PR #774, opened 2025-08-18
Support evaling all checkpoints

mikasenghaas:
Adds `--weights-dir` to `OfflineEvalConfig` to allow specifying a weight directory to search for checkpoints during a run. If set, will find any subdirectory `step_{x}` and eval the checkpoints in ascending order. Can be set either with or without evaling the base model as well. All weight checkpoints have to have the same architecture as the base model specified with `--model.name`, as the weights are just hot-swapped in the inference server that is doing the evals (similar to how online  evals are done on the orchestrator).

Example of how to eval only the weight checkpoints in `outputs/weights` for `Qwen/Qwen2.5-7B`

```bash
uv run eval \
  --model.name Qwen/Qwen2.5-7B \
  --environment-ids aime2025 \
  --rollouts-per-example 64 \
  --outputs-dir qwen2.5-7b-sft \
  --weights-dir outputs/weights \
  --no-eval-base \
  --save
```

---

**GitHub Issue**: #768 
**Linear Issue**: Resolves PRIMERL-42

### DIFF
--- a/src/prime_rl/eval/eval.py
+++ b/src/prime_rl/eval/eval.py
@@ -74,7 +74,7 @@ async def eval(config: OfflineEvalConfig):
         logger.info(f"Evaluating weight checkpoints in {config.weights_dir}")
         ckpt_steps = sorted([int(step_path.name.split("_")[-1]) for step_path in config.weights_dir.glob("step_*")])
         logger.info(f"Found {len(ckpt_steps)} weight checkpoints (steps: {', '.join(map(str, ckpt_steps))})")
-        for ckpt_step in ckpt_steps:
+        for ckpt_step in ckpt_steps[::-1]:
             # Update the weights
             logger.info(f"Evaluating weight checkpoint {ckpt_step}")
             await update_weights(client, config.weights_dir, ckpt_step)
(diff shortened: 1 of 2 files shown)


src/prime_rl/eval/eval.py
```diff
@@ -46,26 +47,58 @@ async def eval(config: OfflineEvalConfig):
...
+                )
+            ]
+        )
+
+    # If specified, evaluate all checkpoints found in the weights directory
+    if config.weights_dir is not None:
+        logger.info(f"Evaluating weight checkpoints in {config.weights_dir}")
+        ckpt_steps = sorted([int(step_path.name.split("_")[-1]) for step_path in config.weights_dir.glob("step_*")])
+        logger.info(f"Found {len(ckpt_steps)} weight checkpoints (steps: {', '.join(map(str, ckpt_steps))})")
+        for ckpt_step in ckpt_steps:
```
samsja:
```suggestion
        for ckpt_step in ckpt_steps[::-1]:
```

proposal to go reverse here, if I am in a hurry I would rather always get the latest ckpt results first, or like if my node die on me I would be mad if I only get the eval for the first ckpt

PR #780, opened 2025-08-19
Implementation for other loss scales

faresobeid:
Performs loss normalization at the token level and allows for different types (sequence, group, and batch level)

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -315,6 +315,13 @@ async def orchestrate(config: OrchestratorConfig):
             torch.tensor([len(c) for c in completion_tokens]).float().reshape(-1, config.rollouts_per_example)
         )
         seq_lens = prompt_lens + completion_lens
+        if config.loss_scale == "seq":
+            loss_scale = completion_lens.reshape(-1) * config.batch_size
+        elif config.loss_scale == "group":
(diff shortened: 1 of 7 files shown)


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -315,6 +315,13 @@ async def orchestrate(config: OrchestratorConfig):
             torch.tensor([len(c) for c in completion_tokens]).float().reshape(-1, config.rollouts_per_example)
         )
         seq_lens = prompt_lens + completion_lens
+        if config.loss_scale == "seq":
+            loss_scale = completion_lens.reshape(-1) * config.batch_size
+        elif config.loss_scale == "group":
+            loss_scale = completion_lens.sum(-1).repeat_interleave(config.rollouts_per_example) * config.batch_size / config.rollouts_per_example
+        elif config.loss_scale == "batch":
+            loss_scale = completion_lens.sum().repeat_interleave(config.batch_size)
```
mikasenghaas: this as util function? maybe even abstracted like the advantage computation type?

mikasenghaas (commented):
yep, let's try to test the most extreme example that we would run in in production, i.e. quite high batch size and seq len and normalize by global tokens

also, we have to be mindful tha loss scales that should be *global* actually need to be *local* for us. this is a limitation imposed by fsdp because the all-reduce avg and not all-reduce sum gradients

PR #784, opened 2025-08-20
use proper prime environmentshub index instead of hardcoded urls

JannikSt:
we're switching over to using our proper hub index for environments which uses: 
https://hub.primeintellect.ai/username/simple for all packages. 

Added fixed: `alphabet-sort = { index = "primeintellect" }` to ensure this only pulls the wheels from our own hub instead of falling back to pypi.

The wheel structure in the lock file also reveals the new content hash based approach for fetching versions instead of using @latest. We need this because the hub (similar to docker hub) allows you to push updates without an explicit tag. 
```
wheels = [
    { url = "https://hub.primeintellect.ai/primeintellect/reverse-text/@f4f1589b/reverse_text-0.1.1-py2.py3-none-any.whl", hash = "sha256:fb2861707faa2e6b4fcc74f695c76265edd3391d1417ed1af72aa7d2e374845c" },
]
```

On the lock file I'm a bit surprised that the upload time was removed, hence the following does not exist anymore:
```
{ url = "https://files.pythonhosted.org/packages/28/d9/7150d5cf9163e05081f1c5c64a0cdf3c32d2f56e2ac95db2a28fe90eca69/aiohttp-3.12.13-cp312-cp312-macosx_10_13_x86_64.whl", hash = "sha256:b103a7e414b57e6939cc4dece8e282cfb22043efd0c7298044f6594cf83ab347", size = 475104, upload-time = "2025-06-14T15:14:01.691Z" },
```


samsja (commented):
nice this is way cleaner like this, wondering if why the uv.lock has the uploaded-time change, it seems to not be harmfull but still worried about the changes

samsja (approved):
lgtm, uv lock seems hughe, I did reproduced the same tho locally so I would say its not an issue

samsja (changes requested):
actually I mange to fix the uv lock

PR #785, opened 2025-08-20
Sami/imp/prime envshub usage

samsja:
trying to fix https://github.com/PrimeIntellect-ai/prime-rl/pull/784
---

**GitHub Issue**: [Issue ID]


uv.lock
```diff
@@ -1,6 +1,6 @@
 version = 1
-revision = 3
-requires-python = "==3.12.*"
+revision = 2
+requires-python = ">=3.12.0, <3.13"
```
mikasenghaas, line 3: i guess it means the same but why/ how did this change?

mikasenghaas (approved):

uv.lock
samsja: no clue tbh I would say its fine like this ?

PR #787, opened 2025-08-20
[Fix] MoE pyproject

Jackmin801:
The MoE pyproject wasnt synced with the recent hub fixes that landed in main pyproject. This PR fixes that.

Also added a unit test to make sure this gets caught in the future

---

**GitHub Issue**: [Issue ID]

### DIFF
--- a/tests/unit/test_venv.py
+++ b/tests/unit/test_venv.py
@@ -1,3 +1,14 @@
+"""
+This test is used to ensure that the moe venv is up to date with the main venv.
+This split is currently needed because torchtitan moe requires torch 2.8 but vllm requires torch 2.7.
+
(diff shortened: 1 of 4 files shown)


mikasenghaas (approved):
beautiful

tests/unit/test_venv.py
```diff
@@ -0,0 +1,29 @@
+import tomllib
+
+MAIN_PYPROJECT = "pyproject.toml"
+MOE_PYPROJECT = "deps/moe/pyproject.toml"
+
+
```
samsja, line 17: maybe add comment why we need this tests

samsja: also add instruction for what to do when this test is failing

Jackmin801: doone

PR #789, opened 2025-08-21
[feat] max vio in sft

Jackmin801:
This PR adds logic to log the max_vio expert balance stat. This is `(tokens_per_expert.max() - tokens_per_expert.mean()) /  tokens_per_expert.mean()` as detailed in eq 4 of https://arxiv.org/pdf/2408.15664v1

image

---

**GitHub Issue**: #790 
**Linear Issue**: Resolves PRIMERL-48

### DIFF
--- a/src/prime_rl/trainer/rl/train.py
+++ b/src/prime_rl/trainer/rl/train.py
@@ -24,6 +24,8 @@ from prime_rl.trainer.model import (
     setup_tokenizer,
     reshard_module,
     setup_model,
+    is_tt_moe_model,
+    get_load_balance_stats,
 )
 from prime_rl.trainer.perf import get_perf_counter
 from prime_rl.trainer.utils import (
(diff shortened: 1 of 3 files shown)


Jackmin801:
Moe logging
image

Doesnt break non moe logging
image

SFT
image

src/prime_rl/trainer/rl/train.py
```diff
@@ -274,15 +276,21 @@ def train(config: RLTrainerConfig):
...
             # Add loss tensors to tensor dict for logging purposes
             for key, loss_tensor in loss_tensors.items():
                 loss_tensor = loss_tensor.detach()[loss_mask].detach().to("cpu")
                 tensors[key].append(loss_tensor)
 
             # Debug log with *local, micro step* stats
-            logger.debug(
-                f"Micro Step {micro_step} | Loss: {tensors['loss'][-1].mean().item():.4f} | Entropy: {tensors['entropy'][-1].mean().item():.4f} | Importance Ratio: {tensors['importance_ratio'][-1].mean().item():.4f}"
-            )
+            _log_str = f"Micro Step {micro_step} | Loss: {tensors['loss'][-1].mean().item():.4f} | Entropy: {tensors['entropy'][-1].mean().item():.4f} | Importance Ratio: {tensors['importance_ratio'][-1].mean().item():.4f}"
```
mikasenghaas: call this `micro_step_message` to be consistent with below?

src/prime_rl/trainer/sft/train.py
```diff
@@ -159,9 +166,10 @@ def train(config: SFTTrainerConfig):
             loss.backward()
 
             # Debug log with *local, micro step* stats
-            logger.debug(
-                f"Micro Step {micro_step} | Loss: {tensors['loss'][-1].mean().item():.4f} | Accuracy: {tensors['accuracy'][-1].mean().item():.4f}"
-            )
+            _log_str = f"Micro Step {micro_step} | Loss: {tensors['loss'][-1].mean().item():.4f} | Accuracy: {tensors['accuracy'][-1].mean().item():.4f}"
```
mikasenghaas: same here

mikasenghaas (commented):
ah nice, so we are logging an average over num layers here, right? also, how does this work? does tt keep the tokens routed to each expert as a layer attribute? or is this smth you added:)

Jackmin801:
> ah nice, so we are logging an average over num layers here, right? also, how does this work? does tt keep the tokens routed to each expert as a layer attribute? or is this smth you added:)

Yup. Its average over the layers because thats the multiple of slowdown this would cause in EP.

Yep, theres a buffer that stores the num tokens per expert that is added to at each fwd. Its needed for the loss free load balancing that deepseek does. Actually that reminded me I have to reset it to 0 after logging otherwise it will just keep accumulating.

Jackmin801:
image

image

mikasenghaas (approved):
cleannn, i really like it

PR #795, opened 2025-08-21
Simple implementation of different loss scaling

faresobeid:


src/prime_rl/trainer/rl/train.py
```diff
@@ -224,9 +225,12 @@ def train(config: RLTrainerConfig):
         batch_size = micro_batch_size * num_micro_batches
 
         # Normalize by the local number of unmasked tokens in the batch (per-batch length normalization)
-        loss_scale = torch.tensor(
-            sum(micro_batch["loss_mask"].sum().item() for micro_batch in micro_batches), device="cuda"
-        )
+        if config.loss.norm_type == "token":
+            loss_scale = torch.tensor(
```
mikasenghaas, line 230: would align the variable name to config key here somehow. not sure if `norm` or `loss_scale` is the better naming, but should be consistent. feel free to decide.

src/prime_rl/trainer/rl/train.py
```diff
@@ -246,19 +250,40 @@ def train(config: RLTrainerConfig):
...
+                )
+                new_loss = (new_loss * loss_mask).sum()
+                if config.loss.norm_type == "sequence":
+                    new_loss = new_loss / torch.clamp_min(loss_mask.sum(), 1)
+                loss += new_loss
+                for key, new_loss_tensor in new_loss_tensors.items():
+                    if key not in loss_tensors:
+                        loss_tensors[key] = new_loss_tensor
+                    else:
+                        loss_tensors[key] += new_loss_tensor
```
mikasenghaas: defo abstract this. also smth like `per_sequence_loss` would be more descriptive than `new_loss` i think

src/prime_rl/trainer/rl/train.py
```diff
@@ -246,19 +250,40 @@ def train(config: RLTrainerConfig):
...
+            for logprob, old_logprob, advantage, loss_mask in zip(
+                logprobs_unpacked, old_logprobs_unpacked, advantages_unpacked, loss_mask_unpacked
+            ):
+                new_loss, new_loss_tensors = compute_loss(
+                    logprobs=logprob,
+                    old_logprobs=old_logprob,
+                    advantages=advantage,
+                    loss_config=config.loss,
+                )
+                new_loss = (new_loss * loss_mask).sum()
```
mikasenghaas: somewhat unrelated to this pr, but i think we should move towards `loss[loss_mask]` pattern. since `loss_mask` is now guaranteed to be a bool tensor

mikasenghaas: i think we could just do it in this pr

src/prime_rl/trainer/rl/train.py
```diff
@@ -246,19 +250,40 @@ def train(config: RLTrainerConfig):
...
-                logprobs=logprobs,
-                old_logprobs=old_logprobs,
-                advantages=advantages,
-                loss_config=config.loss,
-            )
+            response_lengths = get_response_lengths(position_ids)
+            loss = 0
+            loss_tensors = {}
+            # Split tensors by response lengths for per-sequence processing
+            logprobs_unpacked = logprobs.split(response_lengths)
```
mikasenghaas: does this also work for padding (ie. if we have a 2d tensor) or no? also, what happens with padding tokens (we may still have some even in the packing case)

mikasenghaas (commented):
so we are saying goodbye to the idea of writing per-token loss scales on orch bc of numerical issue? did we confirm this in experiments?

ig this impl is nice bc it's a pure trainer concern now, which feels better given that we define the loss here as well. it's not super super elegant with the double nested loop and potential double norm, but i also dont have concrete proposal to write it better so i think we should just go with it

faresobeid:
Well not validated but this does seem like the way that other frameworks do this. Also makes life easier with gspo as we have to unpack for it either way

src/prime_rl/trainer/rl/train.py
faresobeid: well would say this is already good, the norm type makes the loss scale different

src/prime_rl/trainer/rl/train.py
samsja: I would make this whole section a function for the loss

src/prime_rl/trainer/rl/loss.py
```diff
@@ -134,3 +132,62 @@ def shift_logits(logits: Float[Tensor, "batch seq vocab"]) -> Float[Tensor, "bat
     zeros = torch.zeros(batch, 1, vocab, device=logits.device, dtype=logits.dtype)  # (batch, 1, vocab)
     logits = torch.cat([zeros, logits], dim=1)  # (batch, seq, vocab)
     return logits
+
+
+def compute_packed_sequence_loss(
+    logprobs, old_logprobs, advantages, loss_mask, response_lengths, loss_config, loss_scale
+):
```
samsja: lets add jaxtyping here

samsja: also lets make this function take positions_ids and compute the resp lenght inside it

configs/debug/train.toml
```diff
@@ -3,6 +3,6 @@ async_level = 5
 
 [data.fake]
 batch_size = 16
-micro_batch_size = 8
+micro_batch_size = 1
```
mikasenghaas, line 6: why this? imo good in debug to test batch_size != micro_batch_size != 1. better code coverage

src/prime_rl/orchestrator/batch.py
```diff
@@ -51,15 +50,6 @@ def prepare_sample(
             f"Number of tokens {len(input_ids)} is greater than sequence length {seq_len}. This should not happen."
         )
 
-    # Pad the sequence to the sequence length
```
mikasenghaas, line 54: loving this

src/prime_rl/orchestrator/batch.py
```diff
@@ -83,45 +73,6 @@ def prepare_micro_batch(samples: list[MicroBatch], temperature: float):
     return micro_batch
 
 
-def prepare_batch_padding(
```
mikasenghaas, line 86: loving this even more

src/prime_rl/orchestrator/batch.py
```diff
@@ -233,31 +183,16 @@ def prepare_batch(
...
-                temperature,
-                tokenizer,
-                batch_size,
-                micro_batch_size,
-                seq_len,
-                num_train_workers,
-            )
-        case _:
-            raise ValueError(f"Invalid collate mode: {collate_mode}")
+    return prepare_batch_packing(
```
mikasenghaas: mh, let's remove the dispatch function if we don't have dispatches. i.e. `batch.py` exposes `prepare_batch` which implies that it prepares a batch with packing

src/prime_rl/orchestrator/utils.py
```diff
@@ -50,11 +50,8 @@ def wait_for_weight_checkpoint(path: Path, step: int, interval: int = 1, log_int
 
 
 def apply_length_bonus(
-    rewards: list[float], 
-    completion_lengths: list[int], 
-    rollouts_per_prompt: int, 
-    length_bonus: float
-    ) -> list[float]:
+    rewards: list[float], completion_lengths: list[int], rollouts_per_prompt: int, length_bonus: float
```
mikasenghaas, line 53: get your ruff config right lol

src/prime_rl/trainer/rl/loss.py
```diff
@@ -64,23 +64,22 @@ def grpo_loss_clip(
...
-    clipped_importance_ratio = torch.clamp(importance_ratio, 0, clip_ratio)  # (B, L)
-    loss = -clipped_importance_ratio * advantages  # (B, L)
-    is_clipped = (importance_ratio > clip_ratio).float()  # (B, L)
+    importance_ratio = torch.exp(logprobs - old_logprobs)
+    clipped_importance_ratio = torch.clamp(importance_ratio, 0, clip_ratio)
+    loss = -clipped_importance_ratio * advantages
+    is_clipped = (importance_ratio > clip_ratio).float()
 
     return loss, {
-        "loss": loss,
```
mikasenghaas, line 83: why this?

mikasenghaas (commented):
nice work! really like the implementation. tentative approve but defo think we should wait for confirmation of equal curves on hendrycks math at least as the pr touches quite a lot of core logic

configs/debug/train.toml
faresobeid: micro_batch_size > 1 made the tests fail bcs this sets it on the trainer too

src/prime_rl/trainer/rl/loss.py
faresobeid: seemed redundant

PR #801, opened 2025-08-21
Implement GSPO and deprecate ppo-style clipping

faresobeid:


src/prime_rl/trainer/rl/config.py
```diff
@@ -26,24 +26,10 @@ class BaseLossConfig(BaseModel):
         ),
     ] = "token"
 
+    type: Annotated[Literal["gspo", "grpo"], Field(description="Type of loss to use.")] = "grpo"
```
samsja, line 29: maybe add a comment with the link to the gspo paper

src/prime_rl/trainer/rl/loss.py
```diff
@@ -168,35 +90,48 @@ def compute_packed_sequence_loss(
...
+            log_importance_ratio = torch.clamp(log_importance_ratio, max=10.0)
+
+        importance_ratio = torch.exp(log_importance_ratio)
+        clipped_importance_ratio = torch.clamp(
+            importance_ratio, loss_config.clip_ratio_low, loss_config.clip_ratio_high
         )
+        loss = -clipped_importance_ratio * advantages
+        is_clipped_high = (importance_ratio > loss_config.clip_ratio_high).float()
+        is_clipped_low = (importance_ratio < loss_config.clip_ratio_low).float()
+        is_clipped = is_clipped_high + is_clipped_low
```
samsja, line 114: I would keep this in a separate function for more modularity

src/prime_rl/trainer/rl/config.py
faresobeid: in the config or the actual impl?

src/prime_rl/trainer/rl/loss.py
```diff
@@ -168,35 +90,48 @@ def compute_packed_sequence_loss(
...
 
         # Apply loss mask and sum
-        masked_loss = (per_sequence_loss[loss_mask]).sum()
+        loss = (loss[loss_mask]).sum()
```
samsja: I would keep this in a separate function for more modularity

src/prime_rl/trainer/rl/config.py
samsja: in the config like just a comment above the type

samsja:
> that's a lot of if..else and diff types in one function now. not sure if im a big fan of that. i kinda like the abstraction that this function is doing the unpacking, compute loss per sequence and is responsible for normalizing properly.and then each loss type has its own isolated implementation. less convolution of logic = less risk of bugs imo.
> 
> i wouldnt hate adopting nemo-rl's loss protocol. https://github.com/NVIDIA-NeMo/RL/blob/223bfa86b2d4267395928573c900e246d5f5708a/nemo_rl/algorithms/interfaces.py. clean and simple interface and also aligned with torch conventions.

idk I think its fine for now I would rather merge and refactror later

src/prime_rl/trainer/utils.py
```diff
@@ -27,6 +27,7 @@ def get_response_lengths(position_ids: torch.Tensor) -> list[int]:
     Returns:
         List of sequence lengths
     """
+    position_ids = position_ids.flatten()
```
samsja, line 30: why flatten here ?

PR #807, opened 2025-08-22
[fix] dont error on mixed dense moe models

Jackmin801:
Some MoE models like DeepSeekV3 and Moonlight have dense MLPs at the beginning. This will cause the load balancing stats function to throw an error on finding the tokens_per_expert attr. This fixes it by skipping the dense MLPs when collecting load balance stats.

---

**GitHub Issue**: [Issue ID]

### DIFF
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -1,3 +1,5 @@
+import warnings
+
 import torch
 import torch.distributed as dist
 import torch.nn as nn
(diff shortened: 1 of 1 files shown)


src/prime_rl/trainer/model.py
```diff
@@ -23,12 +25,18 @@ def is_tt_moe_model(model: nn.Module) -> bool:
...
+        if not hasattr(transformer_block.mlp, "tokens_per_expert"):
+            continue
         tokens_per_expert = transformer_block.mlp.tokens_per_expert
         balanced_load = tokens_per_expert.mean()
         max_vio = (tokens_per_expert.max() - balanced_load) / balanced_load
         per_layer_max_vio.append(max_vio.item())
         if reset_stats:
             tokens_per_expert.zero_()
+    if len(per_layer_max_vio) == 0:
+        warnings.warn("No load balance stats to report")
```
mikasenghaas: use our logger for this?

mikasenghaas (approved):
approve after comment

src/prime_rl/trainer/model.py
Jackmin801: ah good catch! done

PR #809, opened 2025-08-22
Better pass@k metric

faresobeid:
Currently the way pass@k is measured is naive and very noisy, this way we compute pass@k where k is half the rollouts per response we actually use.
This method is also used in https://arxiv.org/abs/2506.13284


mikasenghaas (approved):
gotta love some bootstrapping

PR #819, opened 2025-08-24
Git clone https in `install.sh`

rasdani:
switch to `git clone` with https in install script to avoid setting up SSH keys. packages in `pyproject.toml` use https with Github anyway.

will mark for review after test install on fresh machine.

UPDATE: works on fresh machine.

cloned this way the git remote is setup with https, so ppl preferring ssh need to update remote once before pushing.
but as install script is meant for usage and manual install instructions are meant for dev'ing this is fine.


mikasenghaas:
as mentioned internally, i do see the point of having https, but would prefer it as a backup not the primary option

mikasenghaas (commented):
as mentioned internally, i do see the point of having https, but would prefer it as a backup not the primary option

mikasenghaas:
Closed with #820

Issue #818, opened 2025-08-24
Add docs on setting up GH SSH key before using `install.sh`


mikasenghaas:

Mrigank118:
Hi,
I would like to work on it, please assign it to me.

mikasenghaas:
@Mrigank118 for context, this is to make sure that our install script works for everyone as we use SSH to clone the repository. the docs should not be too intrusive but should make sure that if ppl run into issues they have a way to resolve

mikasenghaas:
@Mrigank118 please wait with this one. we are discussing internally what the best option is to support both ssh and https

PR #820, opened 2025-08-24
HTTPS fallback in install script

mikasenghaas:
Improves the install script in the following way:
- Probes SSH remote. If not succesful, falls back to HTTPS for cloning instead of failing (#818)
- Makes adding GitHub to known hosts more robust
- Removes GCP installation as we don't use it

---

**GitHub Issue**: #818 
**Linear Issue**: Resolves PRIMERL-62


mikasenghaas:
Tested working on remote machine w/ SSH and HTTPS

Issue #748, opened 2025-08-16
Implement serialization to make buffer checkpoint-compliant


mikasenghaas:
Part of the checkpointing of the orchestrator should be to write the state of the data buffer (including any metadata like current difficulty label, average reward, etc.) as a HF dataset to disk on a call to ckpt_manager.save. When resuming from a step the data buffer should be initialized from the checkpoint, instead of using the base dataset used. I can see using either basic torch serialization (essentially using a same pattern as on the trainer) or serializing this as a HF dataset which can be passed as an optional arg to the env in which case the loading/ resuming logic would only point to a different HF dataset.

semioz:
@mikasenghaas hi, i'd like to take this.

mikasenghaas:
@semioz go for it!

Issue #821, opened 2025-08-24
Use `safetensors` instead of `torch` for saving/ loading weight checkpoints


mikasenghaas:
Currently we use the unsafe and less optimized `torch.load` and `torch.save` for saving weight checkpoints inside `WeightCheckpointManager`. We should use `safetensors.save` and `safetensors.load` instead. 

[Safetensors](https://huggingface.co/docs/safetensors/en/index)

samsja:
I would delay this feature for now. We used to have safetensor support but this was breaking the async saving, hence we moved to pickle. We could instead do smaller pickle chunk, but I would say no prio for now

PR #823, opened 2025-08-24
Implement serialization to make buffer checkpoint-compliant

semioz:
Resuming data buffer properly from a checkpoint. When saving, it stores the rollout buffer, metadata, and the dataset to disk. Whether to restore from a checkpoint or start fresh is controlled by a simple config flag just like we had with **resume_step**.

---

**GitHub Issue**: https://github.com/PrimeIntellect-ai/prime-rl/issues/748

### DIFF
--- a/src/prime_rl/orchestrator/buffer.py
+++ b/src/prime_rl/orchestrator/buffer.py
@@ -1,10 +1,10 @@
+import json
 import random
 from abc import ABC, abstractmethod
 from collections import Counter, defaultdict
-from dataclasses import dataclass
+from dataclasses import asdict, dataclass
 from pathlib import Path
 
-import torch
 from datasets import Dataset
 
 from prime_rl.orchestrator.config import (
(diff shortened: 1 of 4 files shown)


mikasenghaas:
sorry, we gon fix the w&b api key so the integration test runs proper.

semioz:
@mikasenghaas was about to test but should i also convert my torch.save's to safetensors as you mentioned in this issue? https://github.com/PrimeIntellect-ai/prime-rl/issues/821

mikasenghaas:
@semioz nope, wait with this for now. we will rewrite from `torch.save` to `safetensors.save` in a separate pr. depending on whether this or the other pr goes in first, you may have to rebase

src/prime_rl/orchestrator/buffer.py
```diff
@@ -96,6 +98,15 @@ def __init__(self, dataset: Dataset):
...
         self.metadata: dict[int, dict] = {problem_id: {} for problem_id in self.problem_ids}
 
+    def save(self, path: Path):
+        """Saves the buffer state to a directory."""
+        path.parent.mkdir(parents=True, exist_ok=True)
+
+        torch.save(self.rollout_buffer, path / "rollout_buffer.pt")
+        torch.save(self.metadata, path / "metadata.pt")
+
+        self.dataset.save_to_disk(path / "buffer_dataset")
```
mikasenghaas: hm, i can see how this is the easiest way to implement this but i wonder if somehow we could save/load from a single dataset instance. eg. metadata could easily be a column in the dataset. rollout buffer is a bit trickier tho.. i think it's fine like this for now but it feels like a single hf dataset might be the most clean way of serializing

src/prime_rl/orchestrator/buffer.py
```diff
@@ -411,10 +422,26 @@ def sample_rollouts(self, n: int) -> list[Rollout]:
         return sampled_rollouts
 
 
+def load_buffer(path: Path, buffer_config: DataBufferConfigType) -> Buffer:
```
mikasenghaas: can we make this function in-place like our other checkpoint logic?

src/prime_rl/orchestrator/ckpt.py
```diff
@@ -80,12 +84,15 @@ def save(
         self,
         progress: RLProgress | SFTProgress,
         step: int,
+        buffer: Buffer,
```
mikasenghaas, line 95: `buffer` positional arg before `step` please

src/prime_rl/orchestrator/ckpt.py
```diff
@@ -80,12 +84,15 @@ def save(
...
     ) -> None:
         """Saves the full checkpoint state for a specified step."""
         step_path = self._get_step_path(step)
         step_path.mkdir(parents=True, exist_ok=True)
         ckpt_path = self._get_ckpt_path(step)
         self._save_to_path(ckpt_path, step, progress)
+        buffer_path = step_path / "buffer"
+        buffer.save(buffer_path)
```
mikasenghaas: this should be inside `_save_to_path` for async checkpointing to work properly

src/prime_rl/orchestrator/config.py
```diff
@@ -220,6 +220,13 @@ class CheckpointConfig(BaseConfig):
         ),
     ] = None
 
+    resume_buffer_from_checkpoint: Annotated[
+        bool,
+        Field(
+            description="Whether to resume the data buffer from the checkpoint.",
+        ),
+    ] = True
```
mikasenghaas: hm, i wonder whether we want to make this configurable. @faresobeid wdyt? if we do want to make it configurable it should live in the buffer config as `config.buffer.resume`. the default should probably be `True`

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -92,18 +96,22 @@ async def orchestrate(config: OrchestratorConfig):
...
         await update_weights(client, get_weights_dir(config.output_dir), ckpt_step)
+        
+        # Load buffer from checkpoint
+        if config.ckpt.resume_buffer_from_checkpoint:
+            logger.info("Resuming buffer from checkpoint")
+            buffer_path = ckpt_manager.get_ckpt_path(config.ckpt.resume_step) / "buffer"
+            buffer = load_buffer(str(buffer_path), config.buffer)
+        else:
+            logger.info("Initializing buffer from scratch")
+            buffer = setup_buffer(dataset, config.buffer)
```
mikasenghaas: if we make the suggested changes to the buffer load function and the checkpoint, there should be very minimal changes to the orchestrator code

mikasenghaas (commented):
couple of comments, but on a good way! nice job

semioz:
@mikasenghaas thanks. converted the load/save process to fully hf via columns and moved the resume option to buffer configs(lmk if you guys don't wanna make it configurable at all) also other changes you requested. ready for your review again i guess.

mikasenghaas:
Closing as continued in #839

Issue #840, opened 2025-08-26
Change shared config key from `--model-name` to `--model.name` to align with module's configs


mikasenghaas:
Even though it's a bit silly to make a whole config module `ModelConfig` just to hold `name` in `rl.py`, I think it's the right thing to do because all other module's define the model name in this way, so the current solution is a bit inconsistent.

baonudesifeizhai:
https://github.com/PrimeIntellect-ai/prime-rl/pull/848

PR #848, opened 2025-08-27
fix: Change shared config key from --model-name to --model.name to align with module's configs  #840

baonudesifeizhai:
- Add ModelConfig class for shared model settings
- Replace model_name field with model.name structure in RLConfig
- Update auto_setup_model validator to use model.name
- Maintain backward compatibility with existing configs
- All tests pass (167 unit tests)

### DIFF
--- a/src/prime_rl/rl.py
+++ b/src/prime_rl/rl.py
@@ -86,6 +86,17 @@ class CheckpointConfig(BaseSettings):
     ] = None
 
 
+class ModelConfig(BaseSettings):
+    """Configures shared model settings."""
+
+    name: Annotated[
+        str | None,
+        Field(
+            description="The name of the model to use. If None, will fallback to the model names specified on submodule configs."
+        ),
+    ] = None
+
+
 class RLConfig(BaseSettings):
     """Configures an RL training run."""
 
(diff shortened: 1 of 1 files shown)


src/prime_rl/rl.py
```diff
@@ -86,6 +86,17 @@ class CheckpointConfig(BaseSettings):
     ] = None
 
 
+class ModelConfig(BaseSettings):
+    """Configures shared model settings."""
+
+    name: Annotated[
+        str | None,
```
mikasenghaas: should be `str`

src/prime_rl/rl.py
```diff
@@ -141,12 +152,12 @@ class RLConfig(BaseSettings):
         ),
     ] = None
 
-    model_name: Annotated[
-        str | None,
+    model: Annotated[
+        ModelConfig,
```
mikasenghaas: should be `ModelConfig | None`

src/prime_rl/rl.py
```diff
@@ -141,12 +152,12 @@ class RLConfig(BaseSettings):
...
         Field(
-            description="The name of the model to use. If None, will fallback to the model names specified on submodule configs."
+            description="Shared model configs. If None, will fallback to the model configs specified on submodule configs."
         ),
-    ] = None
+    ] = ModelConfig()
```
mikasenghaas: should be `None` by default

src/prime_rl/rl.py
```diff
@@ -286,11 +297,11 @@ def auto_setup_bench(self):
     @model_validator(mode="after")
     def auto_setup_model(self):
         # Use the same model for trainer, orchestrator and inference
-        if self.model_name:
-            self.trainer.model.name = self.model_name
-            self.orchestrator.model.name = self.model_name
+        if self.model.name:
```
mikasenghaas: should check for `self.model is not None`

mikasenghaas (changes requested):

mikasenghaas (approved):
lgtm! thanks

mikasenghaas:
Tests failing on missing API key (should be fixed with #851 in the future). All tests pass locally.

Issue #745, opened 2025-08-16
Checkpoint SFT dataloader


mikasenghaas:
Currently, we don't checkpoint the data loader properly in an SFT run. This means that if we resume we will start from a fresh data loader instance and will see redundant samples

baonudesifeizhai:
https://paste.ubuntu.com/p/vdr4VvCtKf/   the  test script 
res here:   
<img width="540" height="750" alt="Image" src="https://github.com/user-attachments/assets/f7cfd852-d2e5-4b77-ae80-53e0588d5d49" />  
my gpu is 4060 6gb  i try to run with real model like qwen 0.5b 1.5b but all oom  ....

baonudesifeizhai:
for single gpu 
<img width="999" height="229" alt="Image" src="https://github.com/user-attachments/assets/ceee6979-55a6-4f14-8932-86ba369b64ee" /> 
  for two gpu 

<img width="979" height="446" alt="Image" src="https://github.com/user-attachments/assets/0bef17e5-1c35-41f5-836c-28cb2f54af1c" />

PR #849, opened 2025-08-28
Implement SFT dataloader checkpointing

mikasenghaas:
> Don't be scared by locs - most of it is unit testing the stateful dataset and dataloader:)) the loc change outside of tests dir is actually super small

This PR makes the SFT dataset and dataloader stateful to allow for distributed checkpointing without any changes to the `CheckpointManager`.

This is done by enforcing any SFT dataset to implement the `Stateful` protocol, which means it has to expose methods `state_dict` and `load_state_dict`:
- We define the base class `StatefulIterableDataset`. It contains logic for setting up the data rank and world size based on the number of data workers (local processes) and data ranks (DP world size) and initialize the dataset state, which only contains a `step` (how many samples have been seen globally) and `epoch` (which epoch are we in)
- Both the `FakeDataset` and `SFTDataset` inherit from this class and define a `__iter__` function as before but update the state in place
- The `PackingDataset` which is used to wrap the regular dataset simply passes the dataset state on. This means we count a sample as "used" even if we truncate it
- We replace the regular `torch.utils.data.DataLoader` with `torchdata.stateful_dataloader.StatefulDataloader` ([docs](https://github.com/pytorch/data/tree/main/torchdata/stateful_dataloader#saving-custom-state-with-iterable-style-datasets)). Because the iterable dataset we pass to it is rank- and worker-aware state, so is the dataloader
- We add extensive unit tests to ensure the logic is correct in `tests/unit/train/sft/test_{dataset,dataloader}` and show example runs in the PR description

Misc changes:
- Renamed test directory `tests/unit/training` -> `tests/unit/train`
- Fix correctly counting number of samples when packing (before we were incrementing by `batch_size` on each step which is only correct for padding which we deprecated)
- Moved some unit tests from `tests/unit/train` to `tests/unit/train/sft` and `tests/unit/train/rl` if the test is specific to that module
- Add the `num_examples` keyword for running SFT on a subset of data, here used for unit testing
- Allow `FakeData` to generate `fixed` or `variable` length inputs and `random` or `increasing` `input_ids` (the latter means that the first sample will be a tensor full of 0s, then 1s, etc.) - which is nice for unit testing the dataloader checkpointing
- Use discriminated union for `FakeDataConfig` and `SFTDataConfig` because the fields 

TODO:
- [x] Add fake data toy example (1 rank)
- [x] Add real data toy example (1 rank)
- [x] Add real data toy example (2 ranks)
- [x] Show checkpoint resuming on actual run

## Fake Data Toy Example

We do four steps on fake data with (micro) batch size 1 where the inputs IDs are tensors full of 0, 1, 2, ... (increasing). In this setup the dataloader step equals the fake input IDs.

```bash
uv run sft @ configs/reverse_text/sft.toml --log.level debug --data.type fake --data.length fixed --data.input_ids increasing --data.batch-size 1 --data.micro-batch-size 1 --max-steps 4 --ckpt
```

Screenshot 2025-08-28 at 2 25 46 PM

```bash
uv run sft @ configs/reverse_text/sft.toml --log.level debug --data.type fake --data.length fixed --data.input_ids increasing --data.batch-size 1 --data.micro-batch-size 1 --max-steps 8 --ckpt.resume-step 4
```

Screenshot 2025-08-28 at 2 26 45 PM

## Real Data Toy Example (1 Rank)

We start a tiny SFT run on real, packed data on a single rank for 4 steps and checkpoint at the very end.

```bash
uv run sft @ configs/reverse_text/sft.toml --log.level debug --max-steps 4 --ckpt
```

Screenshot 2025-08-28 at 2 20 54 PM

```bash
uv run sft @ configs/reverse_text/sft.toml --log.level debug --max-steps 8 --ckpt.resume-step 4
```

Screenshot 2025-08-28 at 2 21 38 PM

As we can see, we  from checkpoint and dataloader step 0. With each step we increment the dataloader step by ~20  which makes sense given that in each step we see `batch_size (16) * seq_len (128) = 2048` tokens, which is roughly 2048/20≈102 tokens/per sample which is true for this dataset. Upon restart , we resume with step 4 and and dataloader step 89, as expected.

## Real Data Toy Example (2 Ranks)

We start a tiny SFT run on real, packed data on a two ranks rank for 4 steps and checkpoint at the very end.

```bash
uv run torchrun --nproc-per-node 2 src/prime_rl/trainer/sft/train.py @ configs/reverse_text/sft.toml --log.level debug --max-steps 4 --ckpt
```

Screenshot 2025-08-28 at 2 42 19 PM

```bash
uv run torchrun --nproc-per-node 2 src/prime_rl/trainer/sft/train.py @ configs/reverse_text/sft.toml --log.level debug --max-steps 8 --ckpt.resume-step 4
```

Screenshot 2025-08-28 at 2 41 47 PM

## Real Data Example

[W&B View](https://wandb.ai/primeintellect/mika?nw=z5gu3zpiq4p)

Screenshot 2025-08-28 at 2 46 46 PM

```bash
uv run torchrun --nproc-per-node 2 src/prime_rl/trainer/sft/train.py @ configs/reverse_text/sft.toml --log.level debug --monitor.wandb.project mika --monitor.wandb.name pr-849 --max-steps 50 --ckpt
```

```bash
uv run torchrun --nproc-per-node 2 src/prime_rl/trainer/sft/train.py @ configs/reverse_text/sft.toml --log.level debug --monitor.wandb.project mika --monitor.wandb.name pr-849-resume --max-steps 100 --ckpt.resume-step 50
```

---

**GitHub Issue**: #745 
**Linear Issue**: Resolves PRIMERL-19


mikasenghaas:
Thanks @baonudesifeizhai for pioneering this!

src/prime_rl/trainer/sft/data.py
```diff
@@ -29,41 +31,85 @@ class Batch(TypedDict):
...
+        # -1 because the stateful dataloader expects uses 1-based counting while we start at 0
+        self.step = state_dict["step"] - 1
+        self.epoch = state_dict["epoch"]
+
+    def _setup_world_info(self):
+        worker_info = get_worker_info()
+        worker_id, num_workers = 0, 1
+        if worker_info is not None:
+            worker_id = worker_info.id
+            num_workers = worker_info.num_workers
```
samsja:
```suggestion
        if worker_info is not None:
            worker_id = worker_info.id
            num_workers = worker_info.num_workers
        else: 
              worker_id, num_workers = 0, 1

```

mikasenghaas: spacing looked off here, committed separate

Jackmin801 (approved):
letsgoo

Jackmin801 (approved):
nice! test looks like it should test the correct behavior. ship it!

Issue #855, opened 2025-08-29
Fix parsing `is_truncated` for multi-turn envs


mikasenghaas:
This is a bug in the `parse_truncated_completions` from the state returned by verifiers multi-turn environments. We, wrongly, append truncation information for each (possible many) responses within a rollout leading to the total length of the `is_truncated` list to be larger than the batch size.

PR #856, opened 2025-08-29
Fix parsing `is_truncated` for multi-turn envs

mikasenghaas:
> Based on #854, do not merge before.

Previously, the `parse_truncated_completions`, wrongly, appends truncation information for each response within a multi-turn interaction (e.g. many chat interactions, tool calls, etc.). This leads to a shape mismatch because the `len(is_truncated) > batch_size` which is unexpected.

This PR proposes the following logic for computing `is_truncated` for a multi-turn interaction: If any response within the multi-turn completion is truncated, the entire rollout is truncated. Note, that non-LLM responses like user messages or tool call responses, are not counted here (they do not have the `response.finish_reason` field and as such are always *not truncated*).

I verified that the multi-turn `alphabet-sort` environment is functional again (it is broken on current main)

Screenshot 2025-08-29 at 10 57 04 AM

Screenshot 2025-08-29 at 10 57 13 AM

---

**GitHub Issue**: #855 
**Linear Issue**: Resolves PRIMERL-81

### DIFF
--- a/src/prime_rl/orchestrator/utils.py
+++ b/src/prime_rl/orchestrator/utils.py
@@ -31,24 +31,32 @@ def parse_completion_tokens(states: list[State]) -> list[list[int]]:
             assert chat_completion.choices[0].logprobs.content is not None, (
                 "Logprob content should not be None. Make sure to set logprobs=True in the extra body when making the request to /v1/chat/completions"
             )
-            completion_tokens.append([int(token.token.split(":")[-1]) for token in chat_completion.choices[0].logprobs.content])
+            completion_tokens.append(
+                [int(token.token.split(":")[-1]) for token in chat_completion.choices[0].logprobs.content]
+            )
     return completion_tokens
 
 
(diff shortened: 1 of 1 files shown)


src/prime_rl/orchestrator/utils.py
```diff
@@ -31,24 +31,32 @@ def parse_completion_tokens(states: list[State]) -> list[list[int]]:
...
         )
-        for chat_completion in state["responses"]:
-            assert len(chat_completion.choices) == 1, "Response should always have one choice"
-            if chat_completion.choices[0].finish_reason == "length":
-                is_truncated.append(True)
-            else:
-                is_truncated.append(False)
-    return is_truncated
+        is_truncated = False
+        for response in state["responses"]:
```
willccbb, line 53: Ah yep good catch. LGTM

willccbb (approved):

willccbb (approved):

Issue #852, opened 2025-08-28
rl.py script fails if installing the repo as a package


faustomilletari:
Hello,

if I install prime-rl as a package (Eg. uv pip install git+https://github.com/PrimeIntellect-ai/prime-rl.git) invoking `uv run rl ...` results in an error due to the path "src/prime_rl/trainer/rl/train.py" not existing in the current directory.

The line of code creating the issue is here https://github.com/PrimeIntellect-ai/prime-rl/blob/main/src/prime_rl/rl.py#L542

I fixed it by changing to 

```
trainer_cmd = [
            "uv",
            "run",
            "torchrun",
            f"--rdzv-endpoint=localhost:{get_free_port()}",
            f"--rdzv-id={uuid.uuid4().hex}",
            "--nproc-per-node",
            str(config.trainer_gpus),
            "-m",
            "prime_rl.trainer.rl.train",
            "@",
            trainer_file.as_posix(),
        ]
```

mikasenghaas:
@faustomilletari ah yea, that's right - the project is not really developed to be used as a package rn. curious tho: what is your workflow atm when using prime-rl as a package? how is it better for you compared to running from source? trying to figure out if we should push towards packaging soon

cc @samsja, opinions?

mikasenghaas:
also, is this the only change you've had to make to running it as a package? in that case, it's a pretty non-intrusive change and we could just support it

samsja:
> also, is this the only change you've had to make to running it as a package? in that case, it's a pretty non-intrusive change and we could just support it

we should defo suppport this usecase. the src/ flag in the rl.py seems anyway pretty restrictive, imo makes more sense to rather use the __file__ attribute from the prime-rl package that should point to the src/.. when developed locally but also to the site-packages folder when install via pip.

We could also support using 'torchtitan -m prime_rl.yl" as en entrypoint instead of the file

faustomilletari:
yes, that's the only change. I prefer to install things as packages, especially the things that I consider self contained like this project. 

I think I have made a PR with my changes https://github.com/PrimeIntellect-ai/prime-rl/pull/857

PR #857, opened 2025-08-29
Allow usage of rl when prime-rl is installed as a package

faustomilletari:
make it so that rl can be invoked from any directory if the user has installed the package, otherwise make it return an error if the user hasn't installed it AND the relative path src/prime_rl/trainer/rl/train.py does not exist.

I did not have time to check the execution from sources.

---

**GitHub Issue**: [#852]

### DIFF
--- a/src/prime_rl/rl.py
+++ b/src/prime_rl/rl.py
@@ -1,5 +1,6 @@
 import json
 import os
+import pathlib
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


src/prime_rl/rl.py
```diff
@@ -530,7 +531,21 @@ def rl(config: RLConfig):
         trainer_file = get_temp_toml_file()
         with open(trainer_file, "wb") as f:
             tomli_w.dump(config.trainer.model_dump(exclude_none=True, mode="json"), f)
-
+        
+        # Check if prime_rl.trainer.rl.train can be imported
+        try:
+            import prime_rl.trainer.rl.train
```
mikasenghaas: im not sure we need this check?  shouldn't `prime_rl` be available if installed as package or from source. wondering if this pr can be the one-line change that you initially proposed in the issue

src/prime_rl/rl.py
```diff
@@ -539,7 +554,7 @@ def rl(config: RLConfig):
             f"--rdzv-id={uuid.uuid4().hex}",
             "--nproc-per-node",
             str(config.trainer_gpus),
-            "src/prime_rl/trainer/rl/train.py",
+        ] + rl_cmd + [
```
mikasenghaas: this is not the `rl_cmd` (rl is the trainer, orch and infer all together). we can skip the headache of naming this variable by just putting the module path inline

mikasenghaas (commented):
thanks! i can test this tn so we can get it in

src/prime_rl/rl.py
faustomilletari: totally right.

src/prime_rl/rl.py
```diff
@@ -539,7 +540,7 @@ def rl(config: RLConfig):
             f"--rdzv-id={uuid.uuid4().hex}",
             "--nproc-per-node",
             str(config.trainer_gpus),
-            "src/prime_rl/trainer/rl/train.py",
+            os.path.join(rl_module.__path__[0], 'train.py'),
```
mikasenghaas: argh i think this is slightly ugly, can we not just do exactly what you proposed in your issue, so replacing file path to module path?

mikasenghaas (commented):
sorry to have you do this again..:) should be the last time now!

PR #858, opened 2025-08-29
add support for Liger Kernel

cosmicoptima:
Adds a `model.liger_kernel` option to the trainer config. When enabled, `AutoModelForCausalLM` is swapped out with `AutoLigerKernelForCausalLM` and GRPO loss computation uses `triton_grpo_loss`.

Liger Kernel doesn't support GSPO, so if you try to simultaneously set `model.liger_kernel = true` and `loss.type = "gspo"` you'll get a validation error.

### DIFF
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -51,18 +51,12 @@ def get_model(config: ModelConfig) -> nn.Module:
     )
     config_model.use_cache = False
     
-    if config.liger_kernel:
-        model = AutoLigerKernelForCausalLM.from_pretrained(
-            pretrained_model_name_or_path=config.name,
-            config=config_model,
-            trust_remote_code=config.trust_remote_code,
-        )
-    else:
-        model = AutoModelForCausalLM.from_pretrained(
-            pretrained_model_name_or_path=config.name,
-            config=config_model,
-            trust_remote_code=config.trust_remote_code,
-        )
+    model_cls = AutoLigerKernelForCausalLM if config.liger_kernel else AutoModelForCausalLM
+    model = model_cls.from_pretrained(
+        pretrained_model_name_or_path=config.name,
+        config=config_model,
+        trust_remote_code=config.trust_remote_code,
+    )
 
     return model
 
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -5,14 +5,11 @@ import torch.distributed as dist
 import torch.nn as nn
 from beartype import beartype as typechecker
 from jaxtyping import Float, Int, jaxtyped
+from liger_kernel.transformers import AutoLigerKernelForCausalLM
 from torch import Tensor
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import checkpoint_wrapper
 from torch.distributed.fsdp import FSDPModule, MixedPrecisionPolicy, fully_shard
-from transformers import (
-    AutoConfig,
-    AutoModelForCausalLM,
-    AutoTokenizer,
-)
+from transformers import AutoConfig, AutoTokenizer
 from transformers.tokenization_utils import PreTrainedTokenizer
 
 from prime_rl.trainer.config import ActivationCheckpointConfig, ModelConfig
@@ -53,11 +50,20 @@ def get_model(config: ModelConfig) -> nn.Module:
         config.name, attn_implementation=config.attn, trust_remote_code=config.trust_remote_code
     )
     config_model.use_cache = False
-    model = AutoModelForCausalLM.from_pretrained(
-        pretrained_model_name_or_path=config.name,
-        config=config_model,
-        trust_remote_code=config.trust_remote_code,
-    )
+    
+    if config.liger_kernel:
+        model = AutoLigerKernelForCausalLM.from_pretrained(
+            pretrained_model_name_or_path=config.name,
+            config=config_model,
+            trust_remote_code=config.trust_remote_code,
+        )
+    else:
+        from transformers import AutoModelForCausalLM
+        model = AutoModelForCausalLM.from_pretrained(
+            pretrained_model_name_or_path=config.name,
+            config=config_model,
+            trust_remote_code=config.trust_remote_code,
+        )
 
     return model
 
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -9,7 +9,7 @@ from liger_kernel.transformers import AutoLigerKernelForCausalLM
 from torch import Tensor
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import checkpoint_wrapper
 from torch.distributed.fsdp import FSDPModule, MixedPrecisionPolicy, fully_shard
-from transformers import AutoConfig, AutoTokenizer
+from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
 from transformers.tokenization_utils import PreTrainedTokenizer
 
 from prime_rl.trainer.config import ActivationCheckpointConfig, ModelConfig
@@ -58,7 +58,6 @@ def get_model(config: ModelConfig) -> nn.Module:
             trust_remote_code=config.trust_remote_code,
         )
     else:
-        from transformers import AutoModelForCausalLM
         model = AutoModelForCausalLM.from_pretrained(
             pretrained_model_name_or_path=config.name,
             config=config_model,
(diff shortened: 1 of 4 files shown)


samsja:
hey thank you for the great work. Do you have any measurement of the benefit of lieger kernel on the performance  ?

src/prime_rl/trainer/rl/train.py
```diff
@@ -246,33 +246,74 @@ def train(config: RLTrainerConfig):
...
+                    ref_logp=None,
+                    completion_ids=input_ids,
+                    advantages=advantages.squeeze(),
+                    completion_mask=loss_mask.squeeze().int(),
+                    temperature=temperature,
+                    beta=0.0,
+                    eps_low=1.0 - config.loss.clip_ratio,
+                    eps_high=config.loss.clip_ratio - 1.0,
+                    inplace=True,
+                )
```
samsja: hmmm unfortunatty I don't think that the grpo loss is equivalent to ours

cosmicoptima:
```
no liger
2048 mbs1: INFO Memory (Standard): 29.482 GB allocated, 43.439 GB reserved
4096 mbs1: INFO Memory (Standard): 40.356 GB allocated, 51.398 GB reserved
2048 mbs2: INFO Memory (Standard): 39.775 GB allocated, 56.367 GB reserved
4096 mbs2: OOM >.<

liger
2048 mbs1: INFO Memory (Liger): 25.808 GB allocated, 37.496 GB reserved
4096 mbs1: INFO Memory (Liger): 33.056 GB allocated, 41.676 GB reserved
2048 mbs2: INFO Memory (Liger): 33.085 GB allocated, 41.713 GB reserved
4096 mbs2: INFO Memory (Liger): 47.763 GB allocated, 68.705 GB reserved
```

qwen 2.5 3b on...i think 2 trainer h100s

also the loss computation change might not be as important, at least the memory usage with just the `AutoLigerKernelForCausalLM` change was basically exactly the same on 2048 mbs1

cosmicoptima:
Screenshot 2025-08-29 at 16 24 15

also i don't remember the configuration (except it also being qwen2.5 3b) but these runs are the same except liger kernel vs no liger kernel, the faster trainer here is with liger kernel

i didn't check if the loss part vs model part is responsible for that

also group size should be 16 for all of the aforementioned runs, batch size 16x64

src/prime_rl/trainer/rl/train.py
cosmicoptima: i've just removed the use of liger kernel for loss, since some bug in it is affecting the model very significantly and causing the model not to learn

i could make another PR in the future to add that but for now i am not really interested in debugging it when the model-level change provides ~all of the reduction in memory usage. (but it looks like the increase in trainer speed was from loss computation)

src/prime_rl/trainer/rl/train.py
```diff
@@ -291,7 +294,9 @@ def train(config: RLTrainerConfig):
                 tensors[key].append(loss_tensor)
 
             # Debug log with *local, micro step* stats
-            micro_step_message = f"Micro Step {micro_step} | Loss: {tensors['loss'][-1].mean().item():.4f} | Entropy: {tensors['entropy'][-1].mean().item():.4f} | Importance Ratio: {tensors['importance_ratio'][-1].mean().item():.4f}"
+            micro_step_message = f"Micro Step {micro_step} | Loss: {tensors['loss'][-1].mean().item():.4f} | Entropy: {tensors['entropy'][-1].mean().item():.4f}"
+            if "importance_ratio" in tensors:
+                micro_step_message += f" | Importance Ratio: {tensors['importance_ratio'][-1].mean().item():.4f}"
```
mikasenghaas: why this change?

src/prime_rl/trainer/rl/train.py
```diff
@@ -340,7 +345,10 @@ def train(config: RLTrainerConfig):
         # Log step metrics
         step_time = time.time() - step_start_time
         current_lr = optimizer.param_groups[0]["lr"]
-        step_message = f"Step {progress.step} | Time: {step_time:.2f}s | Loss: {tensor_stats['loss/mean']:.4f} | Entropy: {tensor_stats['entropy/mean']:.4f} | Importance Ratio: {tensor_stats['importance_ratio/mean']:.4f} | Grad. Norm: {grad_norm:.4f} | LR: {current_lr:.2e} | Throughput: {throughput:.0f} tokens/s | MFU: {mfu:.1f}%"
+        step_message = f"Step {progress.step} | Time: {step_time:.2f}s | Loss: {tensor_stats['loss/mean']:.4f} | Entropy: {tensor_stats['entropy/mean']:.4f}"
+        if "importance_ratio/mean" in tensor_stats:
+            step_message += f" | Importance Ratio: {tensor_stats['importance_ratio/mean']:.4f}"
+        step_message += f" | Grad. Norm: {grad_norm:.4f} | LR: {current_lr:.2e} | Throughput: {throughput:.0f} tokens/s | MFU: {mfu:.1f}%"
```
mikasenghaas: also here

mikasenghaas (commented):
nice work! wondering: is the package already installed as sub-dependency of something else? also, i think we should defo do a medium scale run with liger kernel to validate that it doesn't fuck training dynamics. you can leave that to us:)

src/prime_rl/trainer/rl/train.py
cosmicoptima: oops that's a holdover from an old version I'll remove it one sec

cosmicoptima: i may also need to remove the GSPO validation

cosmicoptima:
is what package already installed as sub-dependency of something else? liger-kernel? I'm not sure but it was listed as an explicit dependency before I made any changes, the dependency just wasn't being used

src/prime_rl/trainer/model.py
```diff
@@ -5,14 +5,11 @@
...
+from liger_kernel.transformers import AutoLigerKernelForCausalLM
 from torch import Tensor
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import checkpoint_wrapper
 from torch.distributed.fsdp import FSDPModule, MixedPrecisionPolicy, fully_shard
-from transformers import (
-    AutoConfig,
-    AutoModelForCausalLM,
-    AutoTokenizer,
-)
+from transformers import AutoConfig, AutoTokenizer
```
samsja:
```suggestion
from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer, 
```

src/prime_rl/trainer/model.py
```diff
@@ -53,11 +50,20 @@ def get_model(config: ModelConfig) -> nn.Module:
...
+    
+    if config.liger_kernel:
+        model = AutoLigerKernelForCausalLM.from_pretrained(
+            pretrained_model_name_or_path=config.name,
+            config=config_model,
+            trust_remote_code=config.trust_remote_code,
+        )
+    else:
+        from transformers import AutoModelForCausalLM
+        model = AutoModelForCausalLM.from_pretrained(
```
samsja:
```suggestion
```

src/prime_rl/trainer/model.py
```diff
@@ -53,11 +50,20 @@ def get_model(config: ModelConfig) -> nn.Module:
...
+            pretrained_model_name_or_path=config.name,
+            config=config_model,
+            trust_remote_code=config.trust_remote_code,
+        )
```
samsja: Since both class have the same interface what about doing 

```python
model_cls = AutoLigerKernelForCausalLM if config.liger_kernel else AutoModelForCausalLM 

model = model_cls.from_pretrained(
            pretrained_model_name_or_path=config.name,
            config=config_model,
            trust_remote_code=config.trust_remote_code,
        )

src/prime_rl/trainer/model.py
cosmicoptima: done

cosmicoptima:
going to check the GSPO thing first

cosmicoptima:
ok removing the GSPO validation seems fine

src/prime_rl/trainer/model.py
cosmicoptima: sure

cosmicoptima: done

mikasenghaas:
> is what package already installed as sub-dependency of something else? liger-kernel? I'm not sure but it was listed as an explicit dependency before I made any changes, the dependency just wasn't being used

oh my bad then, nvm. yea i think we have liger kernel in at some point but it didn't survive the refactor lol

mikasenghaas (approved):

samsja (approved):
lgtm, this is really helpful thank you :)

Issue #853, opened 2025-08-29
Support multiple workers in SFT dataloader


mikasenghaas:
Should be clear. Need to make 100% sure that checkpointing the dataloader is still functional.

baonudesifeizhai:
The current data_rank calculation already takes workers into account, but multiple workers are not actually enabled?

mikasenghaas:
yea as mentioned in #859 we were scared of it breaking the ckpt logic and decided to postpone this feature to later bc data loading is not bottlenecking us atm

baonudesifeizhai:
cat > configs/debug/single_worker_test.toml << 'EOF'
max_steps = 20

[data]
type = "fake"
batch_size = 16
micro_batch_size = 4
seq_len = 128
num_workers = 4

[ckpt]
interval = 5
keep = 3

[weights]
interval = 5
save_async = false

[monitor.wandb]
EOF

 
uv run sft @ configs/debug/single_worker_test.toml --output-dir outputs/single_worker_test ---when num_worker goes to 4 its comes an error with  KeyError: 'dataset_state'... i will fix it ...

baonudesifeizhai:
<img width="1160" height="328" alt="Image" src="https://github.com/user-attachments/assets/3ef298e9-1fcf-4de8-b86f-504f0272187b" /> multi-woker struct changed...

 
And i met the mulit worker cuda problem.... i think that the real problem..

baonudesifeizhai:
https://github.com/PrimeIntellect-ai/prime-rl/pull/871

mikasenghaas:
im afraid we are likely not going to merge this anytime soon because we are not bottlenecked by dataloading and prefer not breaking the ckpt logic!

PR #859, opened 2025-08-30
Support multiple dataloader workers in StatefulDataLoader

semioz:
Support multiple dataloader workers in StatefulDataLoader. For reference: https://docs.pytorch.org/data/beta/torchdata.stateful_dataloader.html

**GitHub Issue**: https://github.com/PrimeIntellect-ai/prime-rl/issues/853


mikasenghaas:
can you make a compelling case that this doesn't break dataloader checkpointing? we were kinda scared to add this in and decided to postpone to later bc we are not bottlenecked by dataloading atm

PR #863, opened 2025-08-31
Move infer.max_model_len config values to orch.sampling.max_tokens

willccbb:
Changes configs to control request sequence length via orch.sampling.max_tokens (passed per-request via sampling_args thru environment) instead of infer.max_model_len (hard limit for vLLM server).

As a general rule, we should probably not assume that prompt lengths are known in advance. While this is possible in single-turn cases, it generally isn't for multi-turn, as the environment doesn't know max_model_len or how many tokens are in intermediate env responses. This can lead to crashes in the middle of the run:

```
openai.BadRequestError: Error code: 400 - {'object': 'error', 'message': "This model's maximum context length is 4096 tokens. However, you requested 4344 tokens (3320 in the messages, 1024 in the completion). Please reduce the length of the messages or completion. None", 'type': 'BadRequestError', 'param': None, 'code': 400}
```

Relying on the default max_model_len is a good safeguard, as this is generally quite large, and not really any efficiency savings in an absolute sense by limiting it. IMO `max_tokens` is the best place to control how many tokens we want a model to be able to generate, though we should also probably add better error handling for the case where this goes beyond max_model_len anyway (separate PR to verifiers, need to think about the best non-hacky way to do it -- naive way is parsing the error message).

### DIFF
--- a/configs/acereason_math/stage1/infer.toml
+++ b/configs/acereason_math/stage1/infer.toml
@@ -1,6 +1,5 @@
 [model]
 name = "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B"
-max_model_len = 8192
 
 [parallel]
 dp = 6
--- a/configs/acereason_math/stage1/orch.toml
+++ b/configs/acereason_math/stage1/orch.toml
@@ -7,6 +7,7 @@ mask_truncated_completions = false
 
 [sampling]
 temperature = 0.6
+max_tokens = 8192
 
 [model]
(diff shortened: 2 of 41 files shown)


mikasenghaas (commented):
we didn't remove from inference config yet? would do this to properly show that setting it, is unexpected behavior

mikasenghaas (approved):

willccbb:
I think could be a separate PR maybe where we decide exactly what to expose + where? this is just changing the configs, would be good to get it in now, the multi-turn configs are currently breaking.

PR #864, opened 2025-09-01
Run evals on final checkpoint

mikasenghaas:
Previously, if `max_steps` was not a multiple of `eval.interval`, then we would not run evals on the final checkpoint which is kind of annoying. This PR fixes this.

---

**GitHub Issue**: #835 
**Linear Issue**: Resolves PRIMERL-72


samsja (approved):
lgtm

PR #869, opened 2025-09-02
Allow evaling API models

mikasenghaas:
Previously, some of the logic in the eval script assumed that the OAI client is talking to our custom vLLM server (e.g. it would hit the `/health` or `/upload_weight` routes or try parse completion token IDs). This PR adds the necessary alternative paths to allow evaling API models, e.g. GPT-5 from `https://api.openai.com/v1`. All of the changes are non-breaking for the orchestrator and RL training - here we simply assert that the inference server type is `vllm` (can later be extended by `sglang`) because we require weight reload for RL.

The specific changes are:
- Use `client_config.base_url` instead of `client_config.host` and `client_config.port` - **this is a breaking config change!**
- Pass `client_config.api_key` as OAI key, similar to how done it `verifiers` (more secure than passing this in plain-text)
- Have fallbacks for non-standard routes (e.g. `/health` and `/reload_weights`)
- Add auto-inferred `client.config.server_type` (either `vllm` or `openai` for conditional execution)
- Do not attempt to compute completion length and truncation info from OAI models (because we only get text and unclear how to tokenize this)

## Example

Eval `gpt-5-mini` on three environments with one command

```bash
export OPENAI_API_KEY=...
```

```bash
uv run eval \
  --client.base_url https://api.openai.com/v1 \
  --client.api_key OPENAI_API_KEY \
  --model.name gpt-5-mini-2025-08-07  \
  --sampling.max-tokens 65536 \
  --environment_ids math500,aime2024,aime2025 \
  --environment_args '{"math500": {"use_think": false}, "aime2024": {"use_think": false}, "aime2025": {"use_think": false}, "gpqa": {"use_diamond": true, "use_think": true}}'
```

---

**GitHub Issue**: #867 
**Linear Issue**: Resolves PRIMERL-84

### DIFF
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -7,6 +7,8 @@ from prime_rl.orchestrator.advantage import AdvantageType
 from prime_rl.utils.config import LogConfig, ModelConfig, MultiMonitorConfig
 from prime_rl.utils.pydantic_config import BaseConfig, BaseSettings
 
+ServerType = Literal["vllm", "openai"]
+
 
 class ClientConfig(BaseConfig):
     """Configures the client to be used for inference."""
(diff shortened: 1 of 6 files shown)


samsja (approved):
lgtm sir

src/prime_rl/orchestrator/config.py
```diff
@@ -18,26 +20,32 @@ class ClientConfig(BaseConfig):
...
         Field(
-            description="Host to use for the OpenAI API. By default, it is set to a local inference server.",
+            description="Base URL to use for the OpenAI API. By default, it is set to None, which means ",
         ),
-    ] = "localhost"
+    ] = "http://localhost:8000/v1"
 
-    port: Annotated[
-        int,
+    api_key: Annotated[
```
willccbb: maybe want to call this `api_key_var` to make clearer downstream that it's not the actual key?

willccbb (dismissed):

src/prime_rl/orchestrator/config.py
mikasenghaas: done

JohannesHa (approved):

PR #873, opened 2025-09-06
Add LoRA support

cosmicoptima:
This PR adds support for LoRA. It adds a `model.lora` config section with the fields `enabled` `rank` `alpha` `dropout` `target_modules` (modules adapted by the LoRA) `trainable_modules` (modules kept fully trainable).

The LoRA is merged into the base model on save for now because vLLM doesn't support loading a LoRA that keeps some modules fully trained. Support for it is being added though: https://github.com/vllm-project/vllm/pull/11714.

---

**GitHub Issue**: #796

### DIFF
--- a/src/prime_rl/trainer/lora.py
+++ b/src/prime_rl/trainer/lora.py
@@ -174,7 +174,6 @@ def apply_lora_to_model(model: nn.Module, config: LoRAConfig) -> None:
         )
         raise RuntimeError("Cannot apply LoRA to FSDP-wrapped model. Apply LoRA before setup_fsdp().")
 
-
     target_modules = _find_target_modules(model, config.target_modules)
 
     if not target_modules:
@@ -287,8 +286,6 @@ def restore_lora_weights_inplace(model: nn.Module, original_lora_state: Dict[str
             delta_weight = (module.lora_B @ module.lora_A) * module.scaling
             module.base_layer.weight.data.sub_(delta_weight)
             restored_count += 1
-
-
 
 
 def get_lora_state_dict(model: nn.Module) -> Dict[str, torch.Tensor]:


cosmicoptima:
oops one sec im fixing the style check (edit: it's good now)

cosmicoptima:
i'm able to fit 14b parameters into 2 trainer h100s with lora enabled, and 32b parameters into 3 trainer h100s with lora + liger kernel

kalomaze:
some notes:
- if we want to standardize naming conventions with PEFT then trainable_modules should be modules_to_save
- this implementation is less invasive in terms of changing stuff at the expense of not being any faster at saving to disk (entire model is saved, not just the lora adapter), this is perfectly fine tho bc you cannot keep some layers like the output head or embeddings fully trainable otherwise because of how vllm's lora implementation works
- this implementation also implicitly adds support for freezing arbitrary layers or module types via regex, i.e someone could specify no target modules and leave only attention projs in trainable_modules, which would keep the MLPs frozen. this is slightly less direct than having an explicit configuration for freezing, but IMO is fine

cosmicoptima:
> if we want to standardize naming conventions with PEFT then trainable_modules should be modules_to_save

good idea

src/prime_rl/trainer/config.py
```diff
@@ -24,6 +24,62 @@ class ActivationCheckpointConfig(BaseModel):
     ] = 1
 
 
+class LoRAConfig(BaseModel):
+    """Configuration for LoRA (Low-Rank Adaptation)."""
+
+    enabled: Annotated[
+        bool,
+        Field(description="Whether to enable LoRA training."),
+    ] = False
```
mikasenghaas: this seems redundant with `lora: LoRAConfig | None`, should prob remove the `enabled` key

src/prime_rl/trainer/model.py
```diff
@@ -101,6 +102,11 @@ def apply_ac(model: nn.Module, ac_config: ActivationCheckpointConfig):
 
 def setup_model(config: ModelConfig, parallel_dims: ParallelDims) -> nn.Module:
     model = get_model(config)
+
+    # Apply LoRA before FSDP setup
+    if config.lora is not None and config.lora.enabled:
```
mikasenghaas: redundant check

src/prime_rl/trainer/weights.py
```diff
@@ -193,9 +272,6 @@ def _maybe_clean(self, step: int):
             <= self.async_level
         )
         if not (keep_for_eval or keep_for_ckpt):
-            self._logger.debug(
```
mikasenghaas: why delete this?

src/prime_rl/trainer/weights.py
```diff
@@ -131,9 +213,6 @@ def _save_to_path(self, cpu_state: dict[str, Tensor], model: nn.Module, tokenize
         step_path = self._get_step_path(step)
         step_path.mkdir(parents=True, exist_ok=True)
 
-        self._logger.debug(f"Saving weight checkpoint to {step_path}")
```
mikasenghaas: and this?

cosmicoptima:
the gpu tests pass on my runpod and it looks like there has been the same OOM for unit tests on some other recent checks outside of this PR

src/prime_rl/trainer/weights.py
cosmicoptima: that shouldn't be removed fixed

src/prime_rl/trainer/weights.py
cosmicoptima: same here, that shouldn't be removed, fixed

src/prime_rl/trainer/config.py
cosmicoptima: yep

mikasenghaas:
ah yea, looks like a dangling process on our ci machine, cleaned it up and re-ran your tests

samsja:
btw @cosmicoptima I forget to thank you for the great PR. We were not ready at the time to merge it. We are working on porting it here https://github.com/PrimeIntellect-ai/prime-rl/pull/1059 and will make sure you get the correct attribution for your work :)

cosmicoptima:
cool to know it's being worked on @samsja ! also found a little trick whose implementation i don't think i put in this PR, and that i don't think i mentioned here--which is initializing a new lora every step and merging it into the base model at the end of the step

thereby approximating a higher rank over time while saving the same amount of compute

mikasenghaas:
tysm for your contribution @cosmicoptima, super clean pr. we updated with latest main, ran some experiments to validate and merged this as an experimental feature now!

mikasenghaas:
closing this now:)

PR #878, opened 2025-09-08
add dion

samsja:
# update muon optimizer

this PR update the muon optimizer to use the dion implementation that is working better over multi node that the previous one

Screenshot from 2025-09-08 22-26-46


mikasenghaas (commented):
lgtm! need to copy the changes to the MOE pyproject

PR #884, opened 2025-09-10
Log all reward functions belonging to vf_env.rubric independently

kalomaze:
A simple way to log all the individual reward functions that belong to the Rubric (along with the existing mean) for logging additional metrics, such as those logging any "perfect" attempts, tracking per-turn accuracies independently, etc.

image

We should also remove the hardcoded logic for `solve_none` and `solve_all` in the orchestrator, and directly incorporate this logic into the environments where these are interpretable metrics. Also, there isn't any coherent interpretation of these metrics for continuous rewards (such as BTRM / Bradley-Terry reward model scalars), but they are also incoherent if we are to change our rewards to be negative.

## Side Note

There is potential value in having rewards exist as negatives where maximization is driving the rewards towards zero due to Dr. GRPO's normalization not being scale invariant, and because this ensures that relative differences get smaller as the reward increases (rather than the magnitude growing according to the scale, which may exaggerate small relative differences depending on the problem setup).

image

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -82,6 +82,20 @@ async def orchestrate(config: OrchestratorConfig):
     vf_env = load_environment(config.environment.id, **config.environment.args)
     dataset = vf_env.get_dataset(seed=config.seed)
 
+    # DEBUG: Log environment structure
+    logger.info(f"Environment type: {type(vf_env)}")
+    logger.info(f"Environment has rubric: {hasattr(vf_env, 'rubric')}")
+    if hasattr(vf_env, 'rubric'):
+        logger.info(f"Rubric type: {type(vf_env.rubric)}")
+        logger.info(f"Rubric has reward_funcs: {hasattr(vf_env.rubric, 'reward_funcs')}")
+        if hasattr(vf_env.rubric, 'reward_funcs'):
+            logger.info(f"Number of reward functions: {len(vf_env.rubric.reward_funcs)}")
+            for i, func in enumerate(vf_env.rubric.reward_funcs):
+                logger.info(f"Reward function {i}: {func.__name__}")
+        logger.info(f"Rubric has reward_weights: {hasattr(vf_env.rubric, 'reward_weights')}")
+        if hasattr(vf_env.rubric, 'reward_weights'):
+            logger.info(f"Weights: {vf_env.rubric.reward_weights}")
+
     # Setup buffer
     logger.info(f"Setting up buffer ({config.buffer})")
     buffer = setup_buffer(dataset, config.buffer)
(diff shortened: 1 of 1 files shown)


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -82,6 +82,20 @@ async def orchestrate(config: OrchestratorConfig):
     vf_env = load_environment(config.environment.id, **config.environment.args)
     dataset = vf_env.get_dataset(seed=config.seed)
 
+    # DEBUG: Log environment structure
+    logger.info(f"Environment type: {type(vf_env)}")
+    logger.info(f"Environment has rubric: {hasattr(vf_env, 'rubric')}")
+    if hasattr(vf_env, 'rubric'):
```
samsja: not of fan the hasatr here

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -82,6 +82,20 @@ async def orchestrate(config: OrchestratorConfig):
     vf_env = load_environment(config.environment.id, **config.environment.args)
     dataset = vf_env.get_dataset(seed=config.seed)
 
+    # DEBUG: Log environment structure
+    logger.info(f"Environment type: {type(vf_env)}")
+    logger.info(f"Environment has rubric: {hasattr(vf_env, 'rubric')}")
```
samsja: need to remove this log or use logger.debug

src/prime_rl/orchestrator/orchestrator.py
kalomaze: as in this should be unconditional right?

src/prime_rl/orchestrator/orchestrator.py
mikasenghaas: +1, would remove

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -82,6 +82,20 @@ async def orchestrate(config: OrchestratorConfig):
...
+    if hasattr(vf_env, 'rubric'):
+        logger.info(f"Rubric type: {type(vf_env.rubric)}")
+        logger.info(f"Rubric has reward_funcs: {hasattr(vf_env.rubric, 'reward_funcs')}")
+        if hasattr(vf_env.rubric, 'reward_funcs'):
+            logger.info(f"Number of reward functions: {len(vf_env.rubric.reward_funcs)}")
+            for i, func in enumerate(vf_env.rubric.reward_funcs):
+                logger.info(f"Reward function {i}: {func.__name__}")
+        logger.info(f"Rubric has reward_weights: {hasattr(vf_env.rubric, 'reward_weights')}")
+        if hasattr(vf_env.rubric, 'reward_weights'):
+            logger.info(f"Weights: {vf_env.rubric.reward_weights}")
```
mikasenghaas: also all of this, we can assume these exist given the base structure of envs

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -257,6 +271,32 @@ async def orchestrate(config: OrchestratorConfig):
...
+                    
+                    for i in range(len(processed_outputs.prompt_ids)):
+                        try:
+                            # Call individual reward function
+                            reward = func(
+                                prompt=generate_outputs.prompt[i],
+                                completion=generate_outputs.completion[i], 
+                                answer=inputs["answer"][i],
+                                state=generate_outputs.state[i]
+                            )
```
mikasenghaas: wait wait wait, we don't need to call the function again (and we absolutely should not!). these should already be exported in the generate outputs iirc (can you confirm @willccbb). if not, we should make that change in verifiers before tackling this feature in prime-rl. computing a reward may take a long time and we don't want to do that work twice. this is 100% a verifiers concern, prime-rl should just consume and log

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -431,11 +471,18 @@ async def orchestrate(config: OrchestratorConfig):
...
+        # Log reward metrics to monitor (composite + individual)
         reward_metrics = {
             "reward/mean": rewards.mean().item(),
             "step": progress.step,
         }
+        
+        # Add individual reward function metrics
+        for func_name, func_rewards in individual_reward_outputs.items():
+            reward_metrics[f"reward/{func_name}_mean"] = func_rewards.mean().item()
+            reward_metrics[f"reward/{func_name}_std"] = func_rewards.std().item()
```
mikasenghaas: do we need std for each reward func?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -431,11 +471,18 @@ async def orchestrate(config: OrchestratorConfig):
...
+            logger.info(f"Logging reward/{func_name}_mean: {func_rewards.mean().item():.6f}")
```
mikasenghaas: still not convinced what the best naming convention here is. i think i would prefer `reward/{func_name}/mean` for consistency

mikasenghaas (changes requested):
we absolutely cannot recompute the reward functions, this is verifiers concern. otherwise agree with sami, a bit less logging (never thought i'd say this). nice job otherwise!!

src/prime_rl/orchestrator/orchestrator.py
kalomaze: nope!

mikasenghaas (commented):
nice!! this is how i imagined it. can you add a screenshot to how it looks in wandb? after that, good to go imo

kalomaze:
screenshot updated!

mikasenghaas (approved):

PR #885, opened 2025-09-10
Support tool call SFT data

mikasenghaas:
This PR adds the necessary logic in the SFT dataset to correctly insert tool call information for any model whose chat template follows the OAI spec. This required two main changes:
- We assume that the list of available tools is JSON-serialized and as such we have to deserialize it before tokenizing the example. This serialization step is necessary because PyArrow would complain if we tried uploading a non-serialized version of the tool list
- We have to deserialize the tool call arguments (the arguments to the function the LLM scores). In standard OAI spec these arguments are serialized but chat templates such as Qwen3's which our chat template is based on require this field to already be deserialized to correctly format the tool calls of the model. Hence we introduce the (somewhat ugly) helper function `deserialize_tool_calls` which we apply to both the `prompt` and `completion`.

I have started a run with `Qwen3-8B` (similar tool calling functionality as our chat template) on the debug SFT datasets which includes 1K samples from each split to demonstrate that we correctly tokenize the tool calls.

Check the [W&B run](https://wandb.ai/primeintellect/intellect3-sft?nw=rr02lxlz3jb) here. Run looking solid a couple steps in. Note, that we expect high accuracy very quickly because we only have 6K samples in total so we will quickly do epochs here.

Screenshot 2025-09-10 at 6 02 39 PM

### DIFF
--- a/src/prime_rl/trainer/sft/data.py
+++ b/src/prime_rl/trainer/sft/data.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from typing import Iterator, TypedDict
 
 import torch
+import json
 from datasets import Dataset, concatenate_datasets, load_dataset
 from jaxtyping import Bool, Int
 from torch import Tensor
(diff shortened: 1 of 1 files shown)


samsja (approved):
lgtm

PR #886, opened 2025-09-10
Add repetition penalty sampler

kalomaze:
Only adding this because Qwen2.5 instruct models come with a default chat template that is pretty terrible/unprincipled for RL training, and vllm will default to the generation configuration setting for the penalty if it isn't specified outright.

In some contexts, the differences can be significant. (Orange line is vanilla probability distribution, blue line is vanilla probability distribution except with 1.05 repetition penalty from the Qwen Instruct chat template).

image


samsja (approved):
lgtm

mikasenghaas (approved):
lgtm

PR #894, opened 2025-09-11
Reduce memory usage during accuracy computation in SFT trainer

mikasenghaas:
> Should merge either this PR or #887 

Previously, we compute accuracy on each micro batch we not only materialize the `logits` but also the `probs` tensor, leading to unnecessary GPU memory usage which slightly decreases throughput.

This PR **changes the accuracy** computation such that we don't materialize the `probs` tensor in memory, reducing the peak GPU memory usage. However, compared to #887, we do not gain much in throughput because we still do the accuracy computation.

**Benchmarking command**

```bash
uv run sft @ configs/debug/sft.toml \
  --data.type fake \
  --data.length variable \
  --data.seq-len 4096 \
  --data.batch-size 4 \
  --data.micro-batch-size 1 \
  --bench
```

**Current**

Peak GPU mem: 26.20 GB

Screenshot 2025-09-11 at 1 24 17 PM

**This PR**

Peak GPU mem: 23.88 GB

Screenshot 2025-09-11 at 1 32 16 PM

**Comparison**

Screenshot 2025-09-11 at 1 32 38 PM

### DIFF
--- a/src/prime_rl/trainer/sft/train.py
+++ b/src/prime_rl/trainer/sft/train.py
@@ -5,7 +5,7 @@ from contextlib import nullcontext
 # ruff: noqa: I001
 
 import torch
-from torch.nn.functional import cross_entropy
+from torch.nn.functional import cross_entropy, softmax
 from torch.distributed.tensor.experimental import context_parallel
 from loguru import logger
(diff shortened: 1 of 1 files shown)


mikasenghaas:
Closed in favor of #887

Issue #792, opened 2025-08-21
Remove general monitor abstraction


mikasenghaas:
*TBD (For OSS: Do not pick this up yet).* We may not need the general `Monitor` abstraction anymore, because we don't have use for file logging, API logging or socket logging anymore. The only thing we need is the `WandbMonitor`. If that is true, we can simplify the monitor setup and configs, e.g. only specify `--wandb.project …` directly instead of `--monitor.wandb.project …`

PR #895, opened 2025-09-11
Deprecate monitor abstraction

mikasenghaas:
Deprecates a bunch of monitor classes like `FileMonitor`, `APIMonitor` and `SocketMonitor` which we are not using. Since the only monitor class that we have left is `WandbMonitor`, the `MultiMonitor` is also obsolete, so we remove. Could bring back at some point if we want to support other remote logging solutions apart from W&B but not on the roadmap for now. This means means that the config keys for launching a run using the`rl` endpoint and launching the individual components finally aligns and then just have a `--wandb` config key, instead of `--monitor.wandb`. Otherwise, nothing changes to the configuration and logic.

**Launch using RL entrypoint**

```bash
 uv run rl  \
  --trainer @ configs/reverse_text/train.toml   \
  --inference @ configs/reverse_text/infer.toml \
  --wandb.project mika \
  --wandb.name pr-895-1
```

**Launch individually**

```bash
CUDA_VISIBLE_DEVICES=1 uv run trainer @ configs/reverse_text/train.toml --wandb.project mika --wandb.name pr-895-2-trainer
```

```bash
uv run orchestrator @ configs/reverse_text/orch.toml --wandb.project mika --wandb.name pr-895-2-orchestrator
```

```bash
uv run inference @ configs/reverse_text/infer.toml
```

Screenshot 2025-09-11 at 2 45 16 PM

---

**GitHub Issue**: #792 
**Linear Issue**: Resolves PRIMERL-49


samsja (approved):
lgtm

PR #899, opened 2025-09-12
Do not cast to FP32 on SFT trainer

samsja:
# fix memory issue

This remove the float32 casting from sft. It fix memory usage and we don't need fp32 logits for sft only for RL

**Command**

```bash
uv run sft --model.name Qwen/Qwen3-0.6B --model.compile --model.liger-kernel --data.type fake --data.seq-len 2048 --data.batch-size 64 --data.micro-batch-size 1 --bench 
```

**Previous**

Screenshot 2025-09-12 at 12 02 51 PM

**Now**

Screenshot 2025-09-12 at 12 00 56 PM

### DIFF
--- a/src/prime_rl/trainer/sft/train.py
+++ b/src/prime_rl/trainer/sft/train.py
@@ -15,7 +15,6 @@ from prime_rl.trainer.logger import setup_logger
 from prime_rl.trainer.optim import setup_optimizer
 from prime_rl.trainer.scheduler import setup_scheduler
 from prime_rl.trainer.model import (
-    forward,
     setup_tokenizer,
     setup_model,
     is_tt_moe_model,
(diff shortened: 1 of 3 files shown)


src/prime_rl/trainer/sft/train.py
```diff
@@ -170,7 +169,7 @@ def train(config: SFTTrainerConfig):
 
             with maybe_context_parallel:
                 # Forward pass
-                logits = forward(model, input_ids, position_ids).contiguous()
+                logits = model(input_ids=input_ids, position_ids=position_ids).logits
```
mikasenghaas: the reason we used a helper func here is to have jaxtype catch type and shape errors (remember position ids not being long). can we maybe change signature to have a boolean flag for whether to cast or not. or not cast by default and then do the cast optionally on the caller

mikasenghaas (commented):
nice!! small comment then good to go

mikasenghaas:
added back the fp32 cast on rl trainer to avoid possible regression there

samsja:
> added back the fp32 cast on rl trainer to avoid possible regression there

oh good catch

PR #900, opened 2025-09-12
Only do mean reduction on SFT trainer

samsja:
This pr remove the pattern that we have from  RL where we save all tensor in memory and do reduce computation at the end. During sft we only care about mean so we can remove the move to cpu. 

**Start command**

```bash
uv run sft --model.name Qwen/Qwen3-0.6B --model.compile --model.liger-kernel --data.type fake --data.seq-len 2048 --data.batch-size 64 --data.micro-batch-size 1
```

Running this from `main` (`dd200a6`) against `sami-fix` (sami's latest commit) and then `sami-fix+custom-reduction` (my latest commit), we see that we get the expected bump in throughput (MFU) and do not have a memory leak anymore. My version uses slightly more (but constant!) CPU memory because we materialize the full loss tensor and reduce with the loss mask instead of doing the reduction in the `cross_entropy` function directly.

Screenshot 2025-09-12 at 2 45 07 PM

Screenshot 2025-09-12 at 2 47 18 PM

### DIFF
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -114,4 +114,4 @@ def setup_model(config: ModelConfig, parallel_dims: ParallelDims) -> nn.Module:
 def forward(
     model: nn.Module, input_ids: Int[Tensor, "batch seq"], position_ids: Int[Tensor, "batch seq"]
 ) -> Float[Tensor, "batch seq vocab"]:
-    return model(input_ids=input_ids, position_ids=position_ids).logits
+    return model(input_ids=input_ids, position_ids=position_ids).logits.float()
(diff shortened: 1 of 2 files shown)


mikasenghaas (commented):
should we not just fix the leak in the tensors class? seems like the more elegant solution

mikasenghaas:
brought back the `none` reduction and then reducing with `loss[loss_mask].mean()` to avoid the issue of not learning EOS token if EOS=Pad token ID. this comes at no performance penalty.

mikasenghaas:
Indeed, will need @Jackmin801's input on how to reduce the `max_vio` statistics

PR #901, opened 2025-09-12
Refactor tensor class API

mikasenghaas:
This PR is a change the API of the `Tensors` class. Previously, we would add back the accumulated distribution and then grab it as `tensors[key][-1]`. This is somewhat unelegant, so now `compute_stats` clears all accumulated tensors and returns the distributions as a flattened list alongside the point statistics

Also, we do not log distributions by default anymore if wandb is enabled.

### DIFF
--- a/src/prime_rl/trainer/utils.py
+++ b/src/prime_rl/trainer/utils.py
@@ -203,10 +203,11 @@ class Tensors(defaultdict):
         assert dist.is_initialized(), "Tensors requires a distributed environment"
         super().__init__(list)
 
-    def compute_stats(self) -> dict[str, float | int]:
+    def compute_stats(self) -> tuple[dict[str, list[float | int]], dict[str, float | int]]:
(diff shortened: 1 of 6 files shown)


src/prime_rl/trainer/utils.py
```diff
@@ -203,10 +203,11 @@ def __init__(self):
         assert dist.is_initialized(), "Tensors requires a distributed environment"
         super().__init__(list)
 
-    def compute_stats(self) -> dict[str, float | int]:
+    def compute_stats(self, log_distributions: bool) -> tuple[dict[str, list[float | int]], dict[str, float | int]]:
```
samsja, line 206: the type hint are getting out of hand lol, maybe lets just say tuple[list,dict] or just nothing ?

mikasenghaas: but its not a tuple of list and dict, its a dict with values being list of numbers or a number, quite useful to know imo

PR #902, opened 2025-09-12
Fix memory profiler step accumulation

mikasenghaas:
This PR fixes incrementing the step counter in the  `MemoryProfiler` class:
- Actually increment it, e.g. `self.step_num += 1` instead of `self.step_num + 1`
- Increment at the end of `.step()` to align with 0-based trainer step definition


samsja (approved):
lgtm

PR #905, opened 2025-09-12
[fix] use max for max vio

Jackmin801:
the idea with max vio is it measures distance from balanced. 0 means balanced. 0.5 means it has an expert with 50% more token than the balanced amount which theoretically makes u 50% slower.

across layers its thus avg. if i have a layer thats 100% slower and one that is 0%, i am 50% slower

across dp its max because all the dps wait for each other so thats the slowdown. if one is balanced and one is 20% slower, everyone is 20% slower

---

**GitHub Issue**: [Issue ID]

### DIFF
--- a/src/prime_rl/trainer/sft/train.py
+++ b/src/prime_rl/trainer/sft/train.py
@@ -180,6 +180,7 @@ def train(config: SFTTrainerConfig):
 
                 if is_tt_moe_model(model):
                     max_vio = get_load_balance_stats(model)["max_vio"] / grad_accum_steps
+                    dist.all_reduce(max_vio, op=dist.ReduceOp.MAX)
                     batch_max_vio += max_vio
 
                 loss = loss[loss_mask].mean()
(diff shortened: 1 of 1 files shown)


mikasenghaas (commented):
we should also do max in grad acc loop no?

samsja:
> we should also do max in grad acc loop no?

yes

Jackmin801:
Oh I think grad acc loop avg makes sense. Because grad acc steps are sequential like layers. Only parallel like experts and dp use max since they are bottlenecked by the straggler

src/prime_rl/trainer/sft/train.py
```diff
@@ -180,7 +180,7 @@ def train(config: SFTTrainerConfig):
 
                 if is_tt_moe_model(model):
                     max_vio = get_load_balance_stats(model)["max_vio"] / grad_accum_steps
-                    # TODO(sami): Check with Jackmin if we should do a max or avg here
+                    dist.all_reduce(max_vio, op=dist.ReduceOp.MAX)
                     batch_max_vio += max_vio
```
mikasenghaas: are we not missing a divide by `grad_acc_steps` here to compute the avg over micro steps?

mikasenghaas (approved):
send ittt

src/prime_rl/trainer/sft/train.py
Jackmin801: we had it above but ur right that its probably clearer to have it here

PR #909, opened 2025-09-13
Fix micro step logging

mikasenghaas:
```bash
uv run sft @ configs/debug/sft.toml --log.level debug
```

**Previous**

We reported the scaled micro step loss

Screenshot 2025-09-13 at 9 22 59 PM

**Now**

We report the unscaled micro step loss, which matches the batch loss scale

Screenshot 2025-09-13 at 9 24 41 PM

---

**GitHub Issue**: #908 
**Linear Issue**: Resolves PRIMERL-99

### DIFF
--- a/src/prime_rl/trainer/sft/train.py
+++ b/src/prime_rl/trainer/sft/train.py
@@ -187,7 +187,7 @@ def train(config: SFTTrainerConfig):
                 loss = loss[loss_mask].mean()
 
                 # Accumulate average loss over gradient accumulation steps
-                batch_loss += loss / grad_accum_steps
+                batch_loss += loss.detach() / grad_accum_steps
 
                 # Delete logits before backward pass to avoid memory spike
                 del logits
--- a/src/prime_rl/trainer/sft/train.py
+++ b/src/prime_rl/trainer/sft/train.py
@@ -183,24 +183,22 @@ def train(config: SFTTrainerConfig):
                     dist.all_reduce(max_vio, op=dist.ReduceOp.MAX)
                     batch_max_vio += max_vio / grad_accum_steps
 
+                # Compute average loss over unmasked tokens
                 loss = loss[loss_mask].mean()
 
-                # Scale loss by number of gradient accumulation steps
-                loss /= grad_accum_steps
-
-                # Accumulate loss
-                batch_loss += loss
+                # Accumulate average loss over gradient accumulation steps
+                batch_loss += loss / grad_accum_steps
 
                 # Delete logits before backward pass to avoid memory spike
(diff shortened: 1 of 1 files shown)


Jackmin801 (approved):
makes sense. lgtm!

src/prime_rl/trainer/sft/train.py
```diff
@@ -183,24 +183,22 @@ def train(config: SFTTrainerConfig):
...
+                # Compute average loss over unmasked tokens
                 loss = loss[loss_mask].mean()
 
-                # Scale loss by number of gradient accumulation steps
-                loss /= grad_accum_steps
-
-                # Accumulate loss
-                batch_loss += loss
+                # Accumulate average loss over gradient accumulation steps
+                batch_loss += loss / grad_accum_steps
```
samsja:
```suggestion
                batch_loss += loss.detach() / grad_accum_steps
```
otherwise we will keep this in the graph but never backward on it

samsja (changes requested):
hmm not we can do this more elegantly.

```python

loss = loss / grad_acc
loss.backward()

batch_loss += loss.detach()

mikasenghaas:
> hmm not we can do this more elegantly.
> 
> ```python
> loss = loss / grad_acc
> loss.backward()
> 
> batch_loss += loss.detach()
> ```

yea i thought about doing this but then when reporting you have to do `loss.item() * grad_acc_steps` which i'd argue is even uglier. in the current setup we maintain the definition that `loss` (and `max_vio`) are the *micro-step metrics*, i.e. they have the same scale as the `batch_loss` and `batch_max_vio` at all times so we can use them for reporting as is. we only scale when we need to, e.g. when accumulating into batch metrics or before calling .backward() which we do inline. i acc prefer it this way but if you have really strong opinions we can change ofc

samsja:
> > hmm not we can do this more elegantly.
> > ```python
> > loss = loss / grad_acc
> > loss.backward()
> > 
> > batch_loss += loss.detach()
> > ```
> 
> yea i thought about doing this but then when reporting you have to do `loss.item() * grad_acc_steps` which i'd argue is even uglier. in the current setup we maintain the definition that `loss` (and `max_vio`) are the _micro-step metrics_, i.e. they have the same scale as the `batch_loss` and `batch_max_vio` at all times so we can use them for reporting as is. we only scale when we need to, e.g. when accumulating into batch metrics or before calling .backward() which we do inline. i acc prefer it this way but if you have really strong opinions we can change ofc

okay fair, with the detach like you added then its good

samsja (approved):
lgtm

PR #922, opened 2025-09-16
Debug up vllm version

Jackmin801:
This draft is just to get a diff from the main PR


src/prime_rl/inference/vllm/sampler.py
```diff
@@ -0,0 +1,279 @@
...
+        for processor in sampling_metadata.logitsprocs.argmax_invariant:
+            logits = processor.apply(logits)
+
+        # Apply top_k and/or top_p.
+        random_sampled, _ = self.topk_topp_sampler(
+            logits,
+            sampling_metadata.generators,
+            sampling_metadata.top_k,
+            sampling_metadata.top_p,
+        )
```
Jackmin801, line 182: This makes the export happen right after temperature applied and before the argmax invariant and topktopp lines

mikasenghaas:
i think this can be closed? if wrong, feel free to reopen

Issue #896, opened 2025-09-11
Fix Pydantic warnings


mikasenghaas:
Fix these pydantic warnings appearing at the start of RL training

```txt
/root/prime-rl/.venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py:2225: UnsupportedFieldAttributeWarning: The 'repr' attribute with value False was provided to the `Field()` function, which has no effect in the context it was used. 'repr' is field-specific metadata, and can only be attached to a model field using `Annotated` metadata or by assignment. This may have happened because an `Annotated` type alias using the `type` statement was used, or if the `Field()` function was attached to a single member of a union type.
  warnings.warn(
/root/prime-rl/.venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py:2225: UnsupportedFieldAttributeWarning: The 'frozen' attribute with value True was provided to the `Field()` function, which has no effect in the context it was used. 'frozen' is field-specific metadata, and can only be attached to a model field using `Annotated` metadata or by assignment. This may have happened because an `Annotated` type alias using the `type` statement was used, or if the `Field()` function was attached to a single member of a union type.
  warnings.warn(
```

mikasenghaas:
Done in #928

PR #919, opened 2025-09-15
[feat] Up vllm version

Jackmin801:
This should make us no longer need to have `deps/moe` as torch2.8 works for all components as well as no need for custom sampler because the correct processed logprobs post temp are exported.

---

**GitHub Issue**: https://github.com/PrimeIntellect-ai/prime-rl/issues/918
**Linear Issue**: Resolves [Issue ID]

PR #928, opened 2025-09-16
Fix Pydantic warnings

mikasenghaas:
This PR fixes the annoying Pydantic warnings reported in #896 and #925 

```txt
/root/prime-rl/.venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py:2225: UnsupportedFieldAttributeWarning: The 'repr' attribute with value False was provided to the `Field()` function, which has no effect in the context it was used. 'repr' is field-specific metadata, and can only be attached to a model field using `Annotated` metadata or by assignment. This may have happened because an `Annotated` type alias using the `type` statement was used, or if the `Field()` function was attached to a single member of a union type.
  warnings.warn(
/root/prime-rl/.venv/lib/python3.12/site-packages/pydantic/_internal/_generate_schema.py:2225: UnsupportedFieldAttributeWarning: The 'frozen' attribute with value True was provided to the `Field()` function, which has no effect in the context it was used. 'frozen' is field-specific metadata, and can only be attached to a model field using `Annotated` metadata or by assignment. This may have happened because an `Annotated` type alias using the `type` statement was used, or if the `Field()` function was attached to a single member of a union type.
  warnings.warn(
```

Tracked it down to the `wandb` (ofc) import which throws this version to the Pydantic version `v.2.12.1.a1` which was locked before. Minimal reproduction

```bash
uv venv
source .venv/bin/activate
uv pip install wandb pydantic==2.12.0a1
uv run python -c "import wandb"
```

Fixed by locking the latest stable Pydantic version `v2.11.9` (@samsja how/ when did we upgrade to beyond a stable release?!). Running SFT debug

```bash
uv run sft @ configs/debug/sft.toml
```

**Before**

Screenshot 2025-09-16 at 2 27 56 PM

**After**

Screenshot 2025-09-16 at 2 27 24 PM

---

**GitHub Issue**: #896, #925 
**Linear Issue**: Resolves PRIMERL-105, PRIMERL-105

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2878,7 +2878,7 @@ wheels = [
 
 [[package]]
 name = "pydantic"
-version = "2.12.0a1"
+version = "2.11.9"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "annotated-types" },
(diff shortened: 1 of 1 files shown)


samsja:
also need to update the moe deps?

mikasenghaas:
> also need to update the moe deps?

we should merge #919 first and then rebase this imo

PR #929, opened 2025-09-16
Fix W&B logging on multi-node

mikasenghaas:
We should only log from true master rank (`RANK==0`), not each local master, as before.


samsja (approved):
lgtm

Issue #933, opened 2025-09-16
Removing checkpoints fails due to attempting to unlink directory rather than file


faustomilletari:
```
packages/prime_rl/orchestrator/ckpt.py", line 109, in maybe_clean
    ckpt_path.unlink(missing_ok=True)
  File "/home/user/.local/share/uv/python/cpython-3.12.11-linux-x86_64-gnu/lib/python3.12/pathlib.py", line 1342, in unlink
    os.unlink(self)
IsADirectoryError: [Errno 21] Is a directory: 'outputs/checkpoints/step_1000/orchestrator'

```

May be fixed by

                - ckpt_path.unlink(missing_ok=True)

                + if ckpt_path.is_dir():
                +   shutil.rmtree(ckpt_path)
                +else:
                +   ckpt_path.unlink(missing_ok=True)

mikasenghaas:
oh yea, feel free to open a pr for this! i think we don't need the if..else here, just shutil rmtree it

PR #934, opened 2025-09-16
Fix deletion of older checkpoints

faustomilletari:
Fixed issue 933.

Removing older checkpoint caused failure due to attempting to delete a directory rather than a file. Modified code so that deletion can happen with no issue even when kept_path is a directory.

---

**GitHub Issue**: #933


mikasenghaas:
lgtm! just quick fix on ruff then good to go, thanks!!

PR #937, opened 2025-09-17
[Fix] meta loading

Jackmin801:
There are two issues with loading in meta device then dcp loading from HF right now
1. Non-persistent buffers will not be initialised correctly - We thus need to initialise them after materialisation
2. If the word embs are tied, the LM head weight gets copied by the way we apply FSDP which will lead them to be the wrong value.
---

**GitHub Issue**: [Issue ID]

### DIFF
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -77,13 +77,6 @@ def setup_fsdp(model: nn.Module, config: ModelConfig, parallel_dims: ParallelDim
     # TODO: Support dp_replicate
     hsdp_mesh = parallel_dims.world_mesh["dp_shard_cp"]
 
-    # fully_shard(
-    #    model.model.embed_tokens,
-    #    mesh=hsdp_mesh,
-    #    mp_policy=mp_policy,
-    #    reshard_after_forward=config.reshard_after_forward,
-    # )
-
     for transformer_block in model.model.layers:
         fully_shard(
             transformer_block,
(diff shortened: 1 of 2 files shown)


Jackmin801:
The reward issue seems to be fixed by fixing the 2 issues:
1. Non-persistent buffers were not be initialised correctly - We thus needed to initialise them after materialisation
2. If the word embs are tied, the LM head weight gets copied by the way we applied FSDP which led them to be the wrong value.

image
image

src/prime_rl/trainer/model.py
```diff
@@ -110,8 +114,34 @@ def load_dcp_from_hf(model: nn.Module, config: ModelConfig):
...
+    fix_model_post_empty(model)
+
+
+def can_load_dcp_from_hf(model: nn.Module):
+    """Whether the model will be loaded correctly by load_dcp_from_hf.
+
+    The main issue is with anything that is not in the checkpoint.
+    This is usually any non-persistent buffers.
+    """
+    buffer_names = [name for name, _ in model.named_buffers()]
```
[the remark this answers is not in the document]
samsja: nul nul comment

src/prime_rl/utils/tensor_hashing.py
```diff
@@ -0,0 +1,42 @@
...
+        return f"{a.dtype}{a.shape}<meta>"
+
+    if a.numel() < TENSOR_SIG_SAMPLE_SIZE:
+        b = a.as_strided(size=(a.numel(),), stride=(1,))
+    else:
+        step_size = a.numel() // TENSOR_SIG_SAMPLE_SIZE
+        b = a.as_strided(size=(TENSOR_SIG_SAMPLE_SIZE,), stride=(step_size,))
+    element_str = "".join([f"{x:.3e}" for x in b])
+    element_hash = hashlib.md5(element_str.encode("utf-8")).hexdigest()
+    return f"{a.dtype}{a.shape}{a.stride()}<{element_hash}>"
```
samsja, line 29: THE KING IS BACK

src/prime_rl/trainer/model.py
```diff
@@ -77,26 +77,30 @@ def setup_fsdp(model: nn.Module, config: ModelConfig, parallel_dims: ParallelDim
...
             reshard_after_forward=config.reshard_after_forward,
         )
-    fully_shard(
-        [model.lm_head, model.model.norm],
-        mesh=hsdp_mesh,
-        mp_policy=mp_policy,
-        reshard_after_forward=config.reshard_after_forward,
-    )
+
+    if hasattr(model, "config") and not model.config.tie_word_embeddings:
```
samsja, line 88: do we need the hasattr here ?

samsja (approved):
lgtm, good work !

src/prime_rl/trainer/model.py
Jackmin801: im not sure if model always has config

Jackmin801: if not, I guess we be conservative and not do the optimization

Issue #915, opened 2025-09-15
Fix wrong log


mikasenghaas:
This warning log from the wandb monitor is wrong

```
14:02:03 WARNING Skipping WandbMonitor initialization from non-master rank (0) [monitor.py::37]                                                        
```

PR #948, opened 2025-09-18
Minor fixes in `WandbMonitor`

mikasenghaas:
This PR fixes an issue where we wrongly log `Skipping WandbMonitor initialization from non-master rank (0)` from the master rank if W&B is disabled. Also fixes one type issue and makes the conditions to skip logging based on configs easier to read.

---

**GitHub Issue**: #915 
**Linear Issue**: Resolves PRIMERL-100


samsja (approved):
lgtm

PR #949, opened 2025-09-18
Docs overhaul

mikasenghaas:
This PR is a major overhaul of our docs. The main changes are:
- Moves most docs into subpages in the `/docs` directory to make the main README more lean and also allow to automatically host using MDX in our public docs
- Separates configs into `/examples` (public, make an effort to maintain and referenced in docs) and `/configs` (internal usage, no guarantees on functionality)
- Brushes up end-to-end training examples in two environments: `reverse-text` (single-turn, debug) and `wordle` (multi-turn, medium-size) with SFT, RL and evals, with documented start command and W&B logs. They now live in the W&B project [`examples`](https://wandb.ai/primeintellect/examples) and the artifacts in the [PRIME-RL HF collection](https://huggingface.co/collections/PrimeIntellect/prime-rl-687560f61d766320424a7df3)
- Restructure config file structure. To differentiate RL and SFT trainer we use `sft/train.toml`and `rl/train.toml`. The inference and orchestrator configs also move into the `rl` subdirectory. For an example, see `examples/reverse-text` and `examples/wordle` (see https://x.com/wondering_camel/status/1967600198726860816)

In terms of the docs itself we now document:
- More details on doing evals, especially against API model, self-hosted models and checkpoints
- More details on multi-node deployments, especially the SFT and RL trainer
- Add docs on single-GPU training
- Add docs on `torchrun` logging capability

Also, fixes a couple of minor things:
- Our RL benchmark command was broken because the fake data loader would yield micro batch size > 1 which is not supported by our RL trainer anymore. This is fixed by flattening the tensors similar to how we do when packing.
- Ensures that the log directory exists before starting an RL training to prevent a file not found error
- Moves the trainer checkpoints from `outputs/ckpt` into `outputs/ckpt/trainer` to be consistent with checkpoints from other components

---

**GitHub Issue**: #874, #903, #893 
**Linear Issue**: Resolves PRIMERL-86, PRIMERL-95, PRIMERL-97


samsja (approved):
lgtm

PR #954, opened 2025-09-19
rename stack and cat to padd and pack

samsja:
Rename sft data arugments:

stack --> packing
cat -> padding


Jackmin801:
quite against this rename. both are forms of packing and padding. stack and cat have clear semantics in torch and I think following those semantics makes more sense.

PR #955, opened 2025-09-19
Revert "Minor fixes in `WandbMonitor` (#948)"

samsja:
hot fix wandb that create one wandb per rank

Screenshot from 2025-09-18 18-48-17

reverting the last commit fix it


Jackmin801 (approved):
probably claudes fault smh

Issue #951, opened 2025-09-18
Loading from meta device with torch compile doesnt work


Jackmin801:
It probably changes the fqdn somewhere and then dcp is confused

mikasenghaas:
closed?

PR #957, opened 2025-09-19
[Fix] meta loading with compile

Jackmin801:
If we do `model.layers[i] = torch.compile`, we get a module with `_orig_mod` in the fqn. This breaks the dcp loading logic.
However, if we do `model.layers[i].compile` we compile the fwd call inplace and the dcp loading logic is happy.

---

**GitHub Issue**: https://github.com/PrimeIntellect-ai/prime-rl/issues/951


samsja (approved):
lgtm, good catch

PR #959, opened 2025-09-19
Sami/half layer

samsja:
example usage

```
torchrun --nproc_per_node=2 src/prime_rl/trainer/sft/train.py @ configs/debug/sft.toml  --model.num_layers 2
```


src/prime_rl/trainer/model.py
```diff
@@ -60,6 +60,10 @@ def get_model(
     config_model.use_cache = False
     config_model.use_grouped_mm = config.moe_use_grouped_mm
 
+    if config.num_layers is not None:
+        get_logger().info(f"Setting num_layers to {config.num_layers}")
+        config_model.num_hidden_layers = min(config.num_layers, config_model.num_hidden_layers)
```
Jackmin801, line 65: maybe have a warning if it gets clipped to lower? we could also technically allow higher, all that will happen is later layers are all zero

src/prime_rl/trainer/model.py
```diff
@@ -70,7 +74,9 @@ def get_model(
                 model_cls = AutoModelForCausalLMPrimeRL
 
         if device == torch.device("meta"):
+            get_logger().info(f"model num_layers random init: {config_model.num_hidden_layers}")
             model = model_cls.from_config(config_model, trust_remote_code=config.trust_remote_code, dtype=dtype)
+            get_logger().info(f"model num_layers random init: {len(model.model.layers)}")
```
Jackmin801, line 79: only later layers are random though

Jackmin801 (approved):
nice! I think this is helpful. left some nit comments

PR #962, opened 2025-09-19
Allow arbitrary seq len in `StackDataset`

mikasenghaas:
Previously, we asserted the `max_area` that can be yielded from the `StackDataset` to be a power of 2. This PR defines three fixed-sized buckets that we sort samples into. We then yield if the bucket is full or hasn't been released after some timeout number of steps. If a timeout is reached we merge a smaller bucket with a larger bucket.

- Bucket 0: `[0, max_area//4]`
- Bucket 1: `]max_area//4, max_area//2]`
- Bucket 2: `]max_area//2, max_area]`

This PR also adds debug logs for dataset indices which can help to pin down "bad samples", as we experienced with some of our tool call samples. This has to live in the `CatDataset` and `StackDataset` as these "pack" the batches.

---

### DIFF
--- a/src/prime_rl/trainer/sft/data.py
+++ b/src/prime_rl/trainer/sft/data.py
@@ -356,10 +356,11 @@ class StackDataset(StatefulIterableDataset):
         self.dataset = dataset
         self.max_area = max_area
         self.current_seq_len = 0
-        self.buckets = [[] for _ in range(int(math.log2(self.max_area)) + 1)]
         # TODO: Can we steal step from dataset?
         self.step = 0
-        self.bucket_sizes = [2**i for i in range(int(math.log2(self.max_area)))] + [self.max_area]
+        assert self.max_area % 256 == 0
+        self.bucket_sizes = [max_area // 4, max_area // 2, max_area]
+        self.buckets = [[] for _ in range(len(self.bucket_sizes))]
         self.bucket_timers: list[int | None] = [None] * len(self.buckets)
         self.bucket_timeout = STACKING_DATASET_BUCKET_TIMEOUT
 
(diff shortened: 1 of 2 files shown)


Jackmin801 (approved):
Should be ok as far as I can tell

samsja (changes requested):
chated with @Jackmin801 about it offline.

We should rather go with the following algorithm:

Instead of relying on power of two for the bucket we should rather take the max seq len and create bucket of size seq_len / 2 and seq_len / 4. This assume the max seq len is divisible by 4, in practice we actually want it to be divisible by smth like 256 so that we still push power of two input to the matmul.

We also care a lot about having the exact same amount of numel per batch for load balancing reason and avoid out of memory issue.

(it being exact same is actually a bit wrong and we should actually take into account how slower it is the run at larger seq len and how bigger are the activation of the large seq len, but I will let this as exercise for the reader)

src/prime_rl/trainer/sft/data.py
```diff
@@ -338,22 +344,25 @@ def __iter__(self) -> Iterator[Sample]:
                 for key, value in packed_samples.items():
                     if isinstance(value, list):
                         packed_samples[key] = value[: self.seq_len]
+                self.logger.debug(f"Yield batch with dataset indices={indices}")
                 yield packed_samples
-                packed_samples, seq_len = defaultdict(list), 0
+                packed_samples, seq_len, indices = defaultdict(list), 0, []
```
samsja, line 349: should we add this indices to the state dict ?

samsja (approved):
lgtm

src/prime_rl/trainer/sft/data.py
mikasenghaas: no i dont think we need it. it's just for logging

Jackmin801 (approved):
nice! lgtm!

PR #965, opened 2025-09-19
add llama and glm custom

samsja:
# Add custom modelling

This pr open the door to custom modelization outside of HF.

It currently add llama and glm custom model.

To turn on the custom implementing we just have to do

```bash
--model.impl custom
```

to use default hf

```bash
--model.impl hf
```

and to use lieger kernel 

```bash
--model.impl liger_kernel
```

for now we still overfit to hf forward definition but might also move away from it if we have to do (for torch compile or TP reason)

### DIFF
--- a/src/prime_rl/trainer/custom_models/glm.py
+++ b/src/prime_rl/trainer/custom_models/glm.py
@@ -1,0 +1,620 @@
+# coding=utf-8
+# Copyright 2025 The ZhipuAI Inc. team and HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


src/prime_rl/trainer/custom_models/glm.py
```diff
@@ -0,0 +1,620 @@
...
+        >>> prompt = "Hey, are you conscious? Can you talk to me?"
+        >>> inputs = tokenizer(prompt, return_tensors="pt")
+
+        >>> # Generate
+        >>> generate_ids = model.generate(inputs.input_ids, max_length=30)
+        >>> tokenizer.batch_decode(generate_ids, skip_special_tokens=True, clean_up_tokenization_spaces=False)[0]
+        "Hey, are you conscious? Can you talk to me?\nI'm not conscious, but I can talk to you."
+        ```"""
+        if position_ids is not None:
+            raise NotImplementedError("Position IDs are not for custom glm4_moe for now")
```
Jackmin801, line 806: hrmm wondering if we should support it and call flash attn instead. but dont allow CP with it

samsja: we can also do this yeah :)

Jackmin801 (approved):
lfgtm!

PR #971, opened 2025-09-20
Allow RL on single GPU from `rl` entrypoint

mikasenghaas:
This PR changes the hardware allocation patterns for the`rl` entrypoint to allow running RL training run on a single GPU. For example, `reverse-text` can be run on a single GPU with as little as ~48GB like so

```bash
uv run rl \
  --trainer @ configs/reverse_text/train.toml \
  --inference @ configs/reverse_text/infer.toml \
  --trainer-gpu-ids 0 \
  --inference-gpu-ids 0 \
  --inference.gpu-memory-utilization 0.5 \
  --inference.model.max-model-len 256
```

Change `gpu_memory_utilization` so that you use roughly ~8GB on inference for model weights and KV cache.

The changes necessary are:
- The trainer and inference GPU IDs are now directly configured, with overlap allowed
- We add `gpu_memory_utilization` into the inference config so that it can be controlled from `rl.py`
- Remove the strict check that orchestrator seq len = inference max model len

---

**GitHub Issue**: #970 
**Linear Issue**: Resolves PRIMERL-123

### DIFF
--- a/src/prime_rl/inference/config.py
+++ b/src/prime_rl/inference/config.py
@@ -84,7 +84,6 @@ class ModelConfig(BaseConfig):
     ] = False
 
     tool_call_parser: Annotated[
-
         str,
         Field(
             description="The tool call parser to use. Passed to vLLM as `--tool-call-parser`",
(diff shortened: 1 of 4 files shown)


samsja (approved):
lgtm, I would even add some docs how to do this for 1 gpu

mikasenghaas:
> lgtm, I would even add some docs how to do this for 1 gpu

will do in #949

PR #972, opened 2025-09-20
Fix wrong logs

mikasenghaas:
This PR fixes two wrong logs:
- We log `Skipping WandbMonitor initialization from non-master rank` when on master rank but wandb is not enabled
- We log `Logging final distributions to W&B table` even if we don't do that (e.g. because wandb is not enabled)

---

**GitHub Issue**: #915 
**Linear Issue**: Resolves PRIMERL-100


samsja (approved):
lgtm

PR #975, opened 2025-09-22
fix ci moe

samsja:
fix ci that was not running an moe training


Jackmin801 (approved):
oh so the previous specified config didnt exist and it only passed because the wrong cmd was used? haha lgtm!

samsja:
> oh so the previous specified config didnt exist and it only passed because the wrong cmd was used? haha lgtm!

yeah lol, double mistake turns out to work

PR #978, opened 2025-09-22
Mask SFT loss by message type + skip samples without trainable tokens

mikasenghaas:
Previously, the `SFTDataset` would would train on all completion tokens and only mask out prompt tokens. This is the HF `SFTTrainer` default but not flexible enough, especially for multi-turn interactions with tool calls. This PR implements building the `loss_mask` for the SFT trainer incrementally by message type, e.g. one can specify whether to include the system message (`--data.loss_mask.system`, default: `False`), user message (`--data.loss_mask.user`, default: `False`) tool message (`--data.loss_mask.tool`, default: `False`), and assistant message (`--data.loss_mask.assistant`, default: `True`). By default, we only train on assistant messages.

**Misc. Changes**:
-  We include a new utility `print_sample` which prints a colored output indicating which tokens are masked and which are not. If `log.level=debug`, we print out the first micro batch using this helper
- We slightly change the signature of the `SFTDataset` to make it easier to unit test. Loading of the dataset now happens outside the dataset. We add some unit tests for the processing of multi-turn tool call data
- We skip samples without trainable tokens within the context window to prevent NaN losses
- Switch `Jackmin108/glm-0.5b` -> `PrimeIntellect/GLM-0.5B` with updated EOS token for SFT integration test to pass

## Multi-Turn w/ Parallel Tool Calls

Screenshot 2025-09-22 at 12 17 09 PM

---

**GitHub Issue**: #974 
**Linear Issue**: Resolves PRIMERL-124

### DIFF
--- a/tests/unit/train/test_model.py
+++ b/tests/unit/train/test_model.py
@@ -106,7 +106,7 @@ def test_model_with_sequence_packing(model, correct_position_ids):
 
 
 def test_moe_custom_impl():
-    config = ModelConfig(name="Jackmin108/glm-0.5B", attn="sdpa", impl="custom", moe_use_grouped_mm=False)
+    config = ModelConfig(name="PrimeIntellect/GLM-0.5B", attn="sdpa", impl="custom", moe_use_grouped_mm=False)
     model = get_model(config)
     model = model.to("cuda")
     with torch.autocast("cuda", dtype=torch.bfloat16):
--- a/configs/debug/sft_moe.toml
+++ b/configs/debug/sft_moe.toml
@@ -1,7 +1,7 @@
 max_steps = 5
 
 [model]
-name = "Jackmin108/glm-0.5B" 
+name = "PrimeIntellect/GLM-0.5B" 
 moe_use_grouped_mm = false
 impl = "custom"
 attn = "sdpa"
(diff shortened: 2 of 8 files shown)


tests/unit/train/test_model.py
```diff
@@ -106,7 +106,7 @@ def test_model_with_sequence_packing(model, correct_position_ids):
 
 
 def test_moe_custom_impl():
-    config = ModelConfig(name="Jackmin108/glm-0.5B", attn="sdpa", impl="custom", moe_use_grouped_mm=False)
+    config = ModelConfig(name="PrimeIntellect/GLM-0.5B", attn="sdpa", impl="custom", moe_use_grouped_mm=False)
```
samsja, line 109: can u also update the debug_moe  with this model ?

samsja (approved):
added small comment, looks good otherwise, I am wondering if we can generalize the print_sample, like it could be very nice to have a way to say --show-sample during sft and we show some sample in the stdout or in a file

tests/unit/train/test_model.py
mikasenghaas: i did no?

mikasenghaas:
> I am wondering if we can generalize the print_sample, like it could be very nice to have a way to say --show-sample during sft and we show some sample in the stdout or in a file

yes, should we do this in diff pr tho to get this in asap? can make a ticket for it

samsja:
> > I am wondering if we can generalize the print_sample, like it could be very nice to have a way to say --show-sample during sft and we show some sample in the stdout or in a file
> 
> yes, should we do this in diff pr tho to get this in asap? can make a ticket for it

yeah different, should not block merging this one

PR #982, opened 2025-09-22
Batched evals

faresobeid:
For large evals or difficulty filtering, using the entire dataset at once hangs, so do evals in batches


mikasenghaas:
what, why does it hang? can you give an example? would rather like to fix the hanging properly than doing this hotfix

faresobeid:
Fair will get an example but also nice to batch for difficulty filtering to get a sense of accuracy along the eval

mikasenghaas:
then we should handle this async, like async completion of a task updates some running avg metrics that are being displayed or smth. batching = sync barriers = perf deg. by default we never wanna avoid anything that looks like a batch. full "aerial" mode

Issue #968, opened 2025-09-20
Error when `warmup_steps == max_steps`


mikasenghaas:
<img src="https://uploads.linear.app/8847b5c2-f01d-4611-9116-0dcd988da0a6/80341d3c-5f77-49e1-95d9-e50848558130/ebf53138-c2bf-4c96-ae1f-cd3365457a60?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzg4NDdiNWMyLWYwMWQtNDYxMS05MTE2LTBkY2Q5ODhkYTBhNi84MDM0MWQzYy01Zjc3LTQ5ZTEtOTVkOS1lNTA4NDg1NTgxMzAvZWJmNTMxMzgtYzJiZi00Yzk2LWFlMWYtY2QzMzY1NDU3YTYwIiwiaWF0IjoxNzU4MzU4NjAxLCJleHAiOjMzMzI4OTE4NjAxfQ.l6DlLGHseh0k89w7DxWS6o1fOqATgN77xEKk_yl2u6k " alt="Screenshot 2025-09-20 at 10.55.46 AM.png" width="919" data-linear-height="62" />

We crash when all steps are scheduler warmup steps

PR #983, opened 2025-09-22
Implement WSD without cooldown

mikasenghaas:
Refactors the setup of LR schedulers a little bit with the main intention to support WSD without cooldown with infinite training. Also fixes a reporting issue where we call `scheduler.step()` before logging the current learning rate (i.e. we report the LR one step off). Also fixes a bug from before when `max_steps == warmup_steps`.

Check the schedules on [W&B](https://wandb.ai/primeintellect/mika/workspace?nw=0rkxwwyka9es).

Screenshot 2025-09-22 at 11 38 50 PM

---

**GitHub Issue**: #968, #985
**Linear Issue**: Resolves PRIMERL-121, PRIMERL-129

### DIFF
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -145,15 +145,19 @@ class LinearSchedulerConfig(BaseModel):
 
     type: Literal["linear"] = "linear"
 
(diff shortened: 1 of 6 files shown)


samsja (approved):
lgtm

PR #988, opened 2025-09-23
fix remove log sample on debug and add real glag

samsja:
remove logging sample on debug and add a real flag for it


Jackmin801 (approved):
should make future data issues much easier to debug. lgtm!

PR #989, opened 2025-09-23
remove force eos in a sequence

samsja:
remove current behavior where we force the eos if we don't find an as it could be harmful when doing sft a lower seq len that the max seq len of the dataset


mikasenghaas:
we force eos on the full seq in the sft dataset so the order is tokenize + maybe add eos + maybe truncate which is fine

PR #997, opened 2025-09-23
add debug model config

samsja:
this pr add to feature that are usefull for debugging:

* all to load only some layer of the model
* allow random init when using meta loading to avoid long downloading for large model on a fresh machine


src/prime_rl/trainer/model.py
```diff
@@ -127,9 +135,13 @@ def setup_fsdp(model: nn.Module, config: ModelConfig, parallel_dims: ParallelDim
 def load_dcp_from_hf(model: nn.Module, config: ModelConfig):
     from huggingface_hub import snapshot_download
     from torch.distributed.checkpoint import DefaultLoadPlanner, HuggingFaceStorageReader
+    model.to_empty(device="cuda")
 
+    if config.debug.random_init:
+        get_logger().warning("Random initializing model. skipping HF checkpoint loading.")
```
Jackmin801: its technically not random and is actually zero here

src/prime_rl/trainer/config.py
```diff
@@ -148,6 +170,14 @@ def trust_remote_code_only_with_hf(self):
             if self.impl != "hf":
                 raise ValueError("Trust remote code is only supported with the HF implementation.")
         return self
+    
+    @model_validator(mode="after")
+    def random_init_only_with_meta(self):
+        """Random initialize is only supported with the custom implementation."""
+        if self.debug.random_init:
+            if self.impl != "custom":
```
Jackmin801: oh I think it should work as long as meta load is used. even if the impl is the HF or trust remote code on

src/prime_rl/trainer/model.py
```diff
@@ -70,7 +76,9 @@ def get_model(
                 model_cls = AutoModelForCausalLMPrimeRL
 
         if device == torch.device("meta"):
+            get_logger().info(f"model num_layers random init: {config_model.num_hidden_layers}")
             model = model_cls.from_config(config_model, trust_remote_code=config.trust_remote_code, dtype=dtype)
+            get_logger().info(f"model num_layers random init: {len(model.model.layers)}")
```
Jackmin801: these two logs should log the same thing. useful for debug but maybe not info?

Jackmin801 (approved):
yeep lgtm! added some nits

PR #1007, opened 2025-09-24
Use add_column instead of map to add SFT dataset indices

mikasenghaas:
Adding the unshuffled dataset indices using `.map` was too slow for large datasets. This PR changes it to use `.add_column` which should be a lot quicker, see screenshot of a timing comparison on 10K samples

Screenshot 2025-09-24 at 12 55 55 PM

---

**GitHub Issue**: #1006 
**Linear Issue**: Resolves PRIMERL-135


samsja (approved):
lgtm

PR #1012, opened 2025-09-24
Fix incorrect environment link in Wordle example

anakin87:
👋 

Updated the link to the Wordle environment in the README.

@mikasenghaas I also want to report that Wandb links are not working.


mikasenghaas:
ah, great - thanks for the fix and thanks for pinging! i thought i made the wandb project public.. let me check the wandb settings

Screenshot 2025-09-24 at 10 02 35 PM

Issue #947, opened 2025-09-18
`AssertionError: Token prefix mismatch` on Qwen3-30B-A3B


Danau5tin:
Hello!

So this happens at the end of the first RL step when training the Qwen3-30B-A3B model.

I believe Will fixed this for other Qwen3 models [here](https://huggingface.co/collections/willcb/qwen3-68434f4883925bfdb4570ee5), but I remember there is some mapping of the models by @Jackmin801 in prime-rl, so perhaps the model on HF needs to updated with the working chat template?

```
Traceback (most recent call last):                                                                         
  File "/mnt/storage/prime-rl/.venv/bin/orchestrator", line 10, in <module>                                
    sys.exit(main())                                                                                       
             ^^^^^^                                                                                        
  File "/mnt/storage/prime-rl/src/prime_rl/orchestrator/orchestrator.py", line 522, in main                
    asyncio.run(orchestrate(parse_argv(OrchestratorConfig)))                                               
  File "/home/ubuntu/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/asyncio/runners.
py", line 195, in run                                                                                      
    return runner.run(main)
           ^^^^^^^^^^^^^^^^                                                                      [757/1973]
  File "/home/ubuntu/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/asyncio/runners.
py", line 118, in run                                                                                      
    return self._loop.run_until_complete(task)                                                             
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                             
  File "/home/ubuntu/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/lib/python3.12/asyncio/base_eve
nts.py", line 691, in run_until_complete                                                                   
    return future.result()                                                                                 
           ^^^^^^^^^^^^^^^                                                                                 
  File "/mnt/storage/prime-rl/.venv/lib/python3.12/site-packages/loguru/_logger.py", line 1283, in catch_wr
apper                                                                                                      
    return await function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                       [745/1973]
  File "/mnt/storage/prime-rl/src/prime_rl/orchestrator/orchestrator.py", line 237, in orchestrate
    processed_outputs: ProcessedOutputs = vf_env.process_env_results_vllm(
[...]

PR #1021, opened 2025-09-25
map any model with supported MoEs to the fast version

Jackmin801:
This should fix the chat template issue in https://github.com/PrimeIntellect-ai/prime-rl/issues/947

### DIFF
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -7,8 +7,9 @@ from prime_rl.utils.pydantic_config import BaseConfig
 AttnImplementation: TypeAlias = Literal["sdpa", "flash_attention_2"]
 
 MOE_MODEL_MAPS = {
-    "Qwen/Qwen3-30B-A3B": "Jackmin108/Qwen3-30B-A3B-Fast",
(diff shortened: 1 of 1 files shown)


src/prime_rl/trainer/config.py
```diff
@@ -7,8 +7,9 @@
 AttnImplementation: TypeAlias = Literal["sdpa", "flash_attention_2"]
 
 MOE_MODEL_MAPS = {
-    "Qwen/Qwen3-30B-A3B": "Jackmin108/Qwen3-30B-A3B-Fast",
-    "moonshotai/Moonlight-16B-A3B-Instruct": "Jackmin108/Moonlight-16B-A3B-Instruct-Fast",
+    "Qwen3-30B-A3B": "Jackmin108/Qwen3-30B-A3B-Fast",
+    "Qwen3-30B-A3B-Base": "Jackmin108/Qwen3-30B-A3B-Fast",
```
mikasenghaas, line 11: why do these 2 map to the same model?

Jackmin801: oops. nice catch!

Jackmin801:
I think we will go towards supporting custom mode impl rather than trust_remote_code and thus this isnt needed anymore

Issue #1023, opened 2025-09-26
Make AutoModelForCausalLMPrimeRL moe and attn impls modular


Jackmin801:
Right now, every model has its own definitions. Make them modular. Maybe make the support positional args instead of kwargs while we're at it to aid a future TP impl

PR #1027, opened 2025-09-27
[Feat] better custom

Jackmin801:
Address https://github.com/PrimeIntellect-ai/prime-rl/issues/1023 &  https://github.com/PrimeIntellect-ai/prime-rl/issues/1024


mikasenghaas:
oh, while we're at it, do you think we can rename the module from `custom_models` to just `models`. i think it will be clear that these are custom implementations anyways and this way we keep having nice clean single-word files and modules?xdd

Jackmin801:
The loss curve seems to be different from the original. This seems to be a bug in the old torchtitan moe impl. Switching from our custom MoE impl which is a copy of torchtitan latest to the old pinned MoE impl reproduces the old curve
image

I guess since the new code is better than the old on we should merge. Not sure if its worth it to find out what was different with the old one

samsja (approved):
lgtm

samsja (approved):
lgtm

PR #1031, opened 2025-09-28
make default keep value None

samsja:
having 1 has default keeping value is too dangerous, should be None by defauly


mikasenghaas (approved):
it was good while it lasted

PR #1035, opened 2025-09-29
Fix distributed dataloader checkpointing

mikasenghaas:
This PR fixes an inconsistency in the SFT dataloader checkpointing logic on multi-GPU. The fix is to save the local dataloader checkpoint outside of the DCP to correctly resume the dataloader state.

## Debug Runs

```bash
uv run torchrun --nproc-per-node 2 \
  src/prime_rl/trainer/sft/train.py \
  @ configs/reverse_text/sft/train.toml \
  --no-data.shuffle \
  --log.level debug \
  --ckpt.interval 10 # Change this to `--ckpt.resume-step 10` to restart
```

### Before

Resuming a training on 1 GPU perfectly matches curves, but not on multi-GPU.

Screenshot 2025-09-29 at 5 03 07 PM

Screenshot 2025-09-29 at 5 05 37 PM

Checking the debug logs, we can understand why. Before checkpointing, we see the the following dataset indices in the last micro batch

```bash
Rank 0: [202, 204, 206, 208, 210, 212, 214, 216, 218, 220, 222] -> Dataloader step 223
Rank 1: [201, 203, 205, 207, 209, 211, 213, 215, 217, 219] -> Dataloader step 220
```

And on the next micro batch after checkpointing, we get the expected indices

```bash
Rank 0: [224, 226, 228, 230, 232, 234, 236, 238, 240, 242, 244]
Rank 1: [221, 223, 225, 227, 229, 231, 233, 235, 237, 239, 241]
```

However, on the resumed run both ranks load the dataloader checkpoint from rank 0, i.e. `dataloader_state={'step': 220, 'epoch': 0}` which leads to the following samples in the first micro upon after restarting

```bash
Rank 0: [*220*, *222*, 224, 226, 228, 230, 232, 234, 236, 238, 240]
Rank 1:  [221, 223, 225, 227, 229, 231, 233, 235, 237, 239, 241]
```

We see that the first two samples of rank 0 are duplicated because a wrong data loader step was loaded. Here it is fairly small offset because all samples in this particular dataset all samples are roughly equal length and we are not deep into the training. However, because different ranks essentially operate on entirely different shards of the data, it is, in theory, possible that the dataloader steps *vastly diverge*. I argue in the discussion why our current sharding and packing strategy does not allow for perfectly deterministic training resumption in dynamic worlds.

## Now

Running the same command as in the previous section but for single and multi-GPU gives matching curves again.

Screenshot 2025-09-29 at 5 04 15 PM

Screenshot 2025-09-29 at 5 04 40 PM

## Large Run

```bash
# Full run
uv run torchrun --local-ranks-filter 0 --nproc-per-node 8 \
  src/prime_rl/trainer/sft/train.py \
  --model.name Qwen/Qwen3-0.6B \
  --data.name PrimeIntellect/INTELLECT-3-SFT-10K \
  --data.splits math,code \
  --data.seq-len 16384 \
  --data.batch-size 64 \
  --data.micro-batch-size 1 \
  --data.pack-function cat \
  --log.level debug \
  --max-steps 100 \
  --optim.lr 1e-4 \
  --scheduler.type linear \
  --scheduler.warmup-steps 10 \
  --ckpt.interval 50 # Change this to `--ckpt.resume-step 50` to restart
```

### Before

Screenshot 2025-09-29 at 5 30 46 PM

### Now

Screenshot 2025-09-29 at 4 36 02 PM

## Discussion

The current solution comes at the cost of not allowing to restart in dynamic world sizes, i.e. restarting with less GPUs will lead to wrong results and error with more GPUs. In fact, I think its impossible to support both **deterministic restarts** and **dynamic world sizes**, given our current batching + truncation logic. Note, that both the `CatDataset` and `StackDataset`, in one way or the other, assemble rank-specific samples into a batches until they overflow and truncate the overflowing tokens to match the seq len.

Consider this simple minimal example as an example: Let `micro_batch_size=2` and `seq_len=4096` and consider a dataset with $n$ samples $i=0,...,n-1$ where all sample with even indices are 3k tokens long, and all odd indices 1k tokens. We assume no shuffling and the `CatDataset` (wlg):

On a single GPU:
- MB0_0:  [0,1] (sample 0 has 2k tokens and sample 1 has 1k tokens, so no truncation)
- MB1_0: [2,3] (sample 2 has 2k tokens and sample 3 has 1k tokens, so no truncation)

On 2 GPUs:
- MB0_0: [0,2] (sample 0 has 2k tokens and sample 2 has 2k tokens, so sample 2 gets truncated)
- MB0_1: [1,3,5,7] (all samples have 1k tokens length, so no truncation)

On 1 GPU truncation never occurs but on 2 GPUs truncation. The two runs can never have the same loss curves because they do not see the same tokens. This is true even without restarts, so there is really no hope to get exactly matching loss curves in dynamic worlds. 

That said, we may be able to do a "best-effort" approach. If we believe that instabilities upon restart occur because samples are seen twice we could all-reduce the maximum dataloader step across all ranks for checkpointing. This would ensure that we do not see a sample redundantly but comes at the cost the we skip some samples unnecessarily.

---

**GitHub Issue**: #1036 
**Linear Issue**: Resolves PRIMERL-143

### DIFF
--- a/src/prime_rl/trainer/sft/data.py
+++ b/src/prime_rl/trainer/sft/data.py
@@ -158,7 +158,7 @@ class SFTDataset(StatefulIterableDataset):
 
             # Update stored epoch if new epoch is reached, optionall shuffle
             if epoch > self.epoch:
-                dataset = self.dataset.shuffle(seed=self.epoch) if self.config.shuffle else self.dataset
+                dataset = self.dataset.shuffle(seed=epoch) if self.config.shuffle else self.dataset
                 self.epoch = epoch
 
             assert "prompt" in example and "completion" in example, (
--- a/src/prime_rl/trainer/sft/data.py
+++ b/src/prime_rl/trainer/sft/data.py
@@ -309,18 +309,15 @@ class CatDataset(StatefulIterableDataset):
         self.logger = get_logger()
         self.dataset = dataset
         self.seq_len = seq_len
-        # Public state attributes for checkpointing
-        self.packed_samples = defaultdict(list)
-        self.current_seq_len = 0
 
     def state_dict(self) -> dict:
-        return self.dataset.state_dict()
+        return {"dataset": self.dataset.state_dict()}
 
     def load_state_dict(self, state_dict: dict):
-        self.dataset.load_state_dict(state_dict)
+        self.dataset.load_state_dict(state_dict["dataset"])
 
     def __iter__(self) -> Iterator[Sample]:
-        packed_samples, seq_len, indices = self.packed_samples, self.current_seq_len, []
+        packed_samples, seq_len, indices = defaultdict(list), 0, []
         for sample in self.dataset:
             # Add sample to packed samples
             for key, value in sample.items():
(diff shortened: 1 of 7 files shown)


src/prime_rl/trainer/sft/data.py
```diff
@@ -357,20 +348,26 @@ def __init__(self, dataset: StatefulIterableDataset, max_area: int):
...
-        self.bucket_timeout = STACKING_DATASET_BUCKET_TIMEOUT
 
     def state_dict(self) -> dict:
-        return self.dataset.state_dict()
+        return {
+            "dataset": self.dataset.state_dict(),
+            "step": self.step,
+            "buckets": self.buckets,
+            "bucket_timers": self.bucket_timers,
+        }
```
samsja, line 364: need the same for StackDataset

mikasenghaas: you mean for `CatDataset` right? i dont think we need to. there is no state persisted in between yielding micro batches (i.e. on each yield we reset these values). since we only checkpoint in between steps, the values we checkpoint would always be the default values so we might as well not do it

seanxwzhang: great fix! It seems that the `step` attribute is needed for [this line](https://github.com/PrimeIntellect-ai/prime-rl/blob/main/src/prime_rl/trainer/sft/train.py#L235) in train.py, so we might need the same thing for `CatDataset`

mikasenghaas: this step is actually the step of the underlying dataset, i.e. `FakeDataset` or `SFTDataset` which is still present but i moved it to `{"dataset": {"step": ...}}` to allow potentially checkpointing state from the packing wrapper datasets:) hadn't updated the accessing in train.py yet. fixed in [d14e361](https://github.com/PrimeIntellect-ai/prime-rl/pull/1035/commits/d14e361169cc439b6a39e6313277f0c6094c79d7)

PR #1037, opened 2025-09-29
support setting torch dist timeout in config

Jackmin801:
Support setting torch distributed timeout

### DIFF
--- a/src/prime_rl/trainer/utils.py
+++ b/src/prime_rl/trainer/utils.py
@@ -1,6 +1,7 @@
 import pickle
 import time
 from collections import defaultdict
+from datetime import timedelta
 from itertools import chain
 from pathlib import Path
 from typing import Any, TypeAlias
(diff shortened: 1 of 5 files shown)


src/prime_rl/trainer/utils.py
```diff
@@ -20,10 +21,16 @@
 from prime_rl.utils.logger import get_logger
 from prime_rl.utils.utils import format_num, format_time
 
+DEFAULT_TIMEOUT = timedelta(seconds=600)
```
mikasenghaas, line 24: do we need this default if our config alr has a default and the function itself?

mikasenghaas (commented):
imo it'd be fine to just hardcode a larger timeout and not make configurable but also fine as is

samsja:
> imo it'd be fine to just hardcode a larger timeout and not make configurable but also fine as is

very large timeout also create some issue, like any nccl error would take X minutes to be reported. So we defo want to have ti configurable imo

samsja (approved):
lgtm

PR #1040, opened 2025-09-29
add seed data

samsja:
#  Adding seed data

this pr allow to add a seed for the shuffling, we still keep the deterministic behavior of shuffle after each epoch in a deterministic way by adding the epoch to the seed.

PR is backward compatible as the default seed is 0

### DIFF
--- a/src/prime_rl/trainer/sft/config.py
+++ b/src/prime_rl/trainer/sft/config.py
@@ -64,6 +64,7 @@ class SFTDataConfig(BaseDataConfig):
     )
     splits: Annotated[list[str], Field(description="Splits to use from the HF dataset.")] = ["train"]
     shuffle: Annotated[bool, Field(description="Whether to shuffle the dataset at the beginning of each epoch.")] = True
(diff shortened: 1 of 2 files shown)


src/prime_rl/trainer/sft/config.py
```diff
@@ -64,6 +64,7 @@ class SFTDataConfig(BaseDataConfig):
     )
     splits: Annotated[list[str], Field(description="Splits to use from the HF dataset.")] = ["train"]
     shuffle: Annotated[bool, Field(description="Whether to shuffle the dataset at the beginning of each epoch.")] = True
+    seed: Annotated[int, Field(description="Random seed to use for shuffling the dataset. We also shuffle at the end of each epoch by adding epoc count to the seed.")] = 0
```
mikasenghaas:
```suggestion
    seed: Annotated[int, Field(description="Random seed to use for shuffling the dataset. We also shuffle at the end of each epoch by adding epoch count to the seed.")] = 0
```

mikasenghaas (approved):
yep, that works

PR #1044, opened 2025-09-30
Implementation of sequence level importance ratio without length normalisation

faresobeid:
Adds implementation of sequence level importance ratio without length normalisation, used in deepseek R1 and https://yingru.notion.site/When-Speed-Kills-Stability-Demystifying-RL-Collapse-from-the-Training-Inference-Mismatch-271211a558b7808d8b12d403fd15edda

### DIFF
--- a/src/prime_rl/trainer/custom_models/layers/moe.py
+++ b/src/prime_rl/trainer/custom_models/layers/moe.py
@@ -61,6 +61,7 @@ class FeedForward(nn.Module):
         for linear in (self.w2, self.w3):
             nn.init.trunc_normal_(linear.weight, mean=0.0, std=init_std)
 
+
 class BCFeedForward(nn.Module):
     def __init__(
         self,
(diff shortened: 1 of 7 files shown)


mikasenghaas (commented):
code looks fine, i think? im losing the overview a bit on what loss types and norm we support at this point. can we do anything to make this a little clearer? ideally explicitly in code, and docs as last resort

faresobeid:
> code looks fine, i think? im losing the overview a bit on what loss types and norm we support at this point. can we do anything to make this a little clearer? ideally explicitly in code, and docs as last resort

Ya so basically now we would support using a token-level or sequence-level importance ratio. The sequence-level importance ratio would have the option to length normalise or not (GSPO length normalises while others don't). The token-level importance ratio now would have no option to normalise on a sequence-level, only by batch/token-level (most widely used)

mikasenghaas (approved):

PR #1047, opened 2025-09-30
Implement global std norm

faresobeid:
Normalise advantage by global std (instead of group level std in normal GRPO). Proposed in https://arxiv.org/abs/2508.08221


faresobeid:
https://wandb.ai/primeintellect/acereason_tests/workspace?nw=u5r4iq1nh3 (this is from an older commit)
Screenshot 2025-09-30 at 03 04 07
Screenshot 2025-09-30 at 03 04 25
Screenshot 2025-09-30 at 03 04 37
Screenshot 2025-09-30 at 03 04 51
Screenshot 2025-09-30 at 03 05 05

samsja (approved):
lgtm

src/prime_rl/orchestrator/config.py
```diff
@@ -453,6 +453,13 @@ class OrchestratorConfig(BaseSettings):
         ),
     ] = "drgrpo"
 
+    global_std_norm: Annotated[
```
mikasenghaas: why is this called `global_...`? i think we can find a more explicit name here

mikasenghaas (commented):
so this is back to regular grpo adv estimation, right? any reason for this. can't clearly tell if with or without is better from the ablations

src/prime_rl/orchestrator/config.py
mikasenghaas: ah nvm

mikasenghaas:
> so this is back to regular grpo adv estimation, right? any reason for this. can't clearly tell if with or without is better from the ablations

ah nvm, didn't read closely enough.. makes sense then. same comment as in #1048 that i might be worth to break out the adv related configs into a `AdvantageConfig` config class. o/w looks good!

src/prime_rl/orchestrator/advantage.py
```diff
@@ -1,88 +1,59 @@
...
-@jaxtyped(typechecker=typechecker)
-def compute_advantage_rloo(rewards: Float[Tensor, "group"], _: Int[Tensor, "group"]) -> Float[Tensor, "group"]:
-    """
-    Computes RLOO (rescaled leave-one-out) advantages for a single group by
-    scaling the standard DR.GRPO advantage by the factor G / (G - 1).
-    """
-    group_size = rewards.shape[0]
-    advantages = rewards - rewards.mean()
-    return advantages * group_size / (group_size - 1)
+class AdvantageConfig(BaseConfig):
```
mikasenghaas: move to orch `config.py`

PR #1048, opened 2025-09-30
Implement GFPO

faresobeid:
https://www.arxiv.org/abs/2508.09726
Basically to bias the model to generate shorter reasoning traces, it basically only keeps the shortest N rollouts in a group. This way the other rollouts are not penalised, but the model is just naturally pushed to generate shorter traces

### DIFF
--- a/src/prime_rl/orchestrator/advantage.py
+++ b/src/prime_rl/orchestrator/advantage.py
@@ -73,6 +73,7 @@ def compute_advantages(
     completion_lengths: list[int],
     samples_per_problem: int,
     advantage_type: AdvantageType,
+    keep_n_shortest: int | None = None,
 ) -> list[float]:
     """
     Computes advantages and statistics for logging from a flattened list of rewards for a given advantage type.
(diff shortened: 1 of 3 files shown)


samsja (approved):

mikasenghaas:
put to draft until we have ablations

src/prime_rl/orchestrator/advantage.py
```diff
@@ -96,7 +98,21 @@ def compute_advantages(
     for group_rewards, group_lengths in zip(all_group_rewards, all_group_lengths):
         group_rewards_tensor = torch.tensor(group_rewards)
         group_lengths_tensor = torch.tensor(group_lengths)
-        group_advantages_tensor = compute_advantage(group_rewards_tensor, group_lengths_tensor)
+
+        if keep_n_shortest is not None:
+            k = min(max(keep_n_shortest, 0), group_rewards_tensor.numel())
```
mikasenghaas, line 103: could also fail loudly if the value is not in [0,G] imo, rather than doing these fallbacks but up to you

src/prime_rl/orchestrator/config.py
```diff
@@ -453,6 +453,14 @@ class OrchestratorConfig(BaseSettings):
         ),
     ] = "drgrpo"
 
+    advantage_keep_n_shortest: Annotated[
```
mikasenghaas, line 456: wondering if we should break this out into a `AdvantageCnfig` which has a type and `keep_n_shortest`

mikasenghaas (commented):
nice! some minor comments re configs. are we planning any ablations for this one?

PR #1053, opened 2025-10-02
[Feat] ckpt convert scripts

Jackmin801:
A script to convert dcp model ckpts to sharded safetensor format that can be loaded by vllm and a script to convert torchtitan moe model from hub back to hf format.


samsja (commented):
code lgtm but can we add documentation around it ? Maybe a moe section ?

PR #1059, opened 2025-10-02
Third attempt update lora pr with upstream

kalomaze:


src/prime_rl/trainer/model.py
```diff
@@ -13,7 +13,8 @@
 from transformers.tokenization_utils import PreTrainedTokenizer
 
 from prime_rl.trainer.config import ActivationCheckpointConfig, CompileConfig, ModelConfig
-from prime_rl.trainer.models import AutoModelForCausalLMPrimeRL
+from prime_rl.trainer.custom_models import AutoModelForCausalLMPrimeRL
```
Jackmin801, line 16: I believe this import is now from trainer.models now

src/prime_rl/trainer/lora.py
```diff
@@ -0,0 +1,323 @@
...
+            bias=self.base_layer.bias is not None,
+            device=self.base_layer.weight.device,
+            dtype=self.base_layer.weight.dtype,
+        )
+
+        merged_layer.weight.data = self.base_layer.weight.data + delta_weight
+        if self.base_layer.bias is not None:
+            merged_layer.bias.data = self.base_layer.bias.data.clone()
+
+        return merged_layer
```
Jackmin801, line 73: do we need a @torch.no_grad decorator here?

PR #1072, opened 2025-10-06
Feature: Add multi-env eval support using verifiers.

d42me:


configs/deepscaler/eval.toml
```diff
@@ -1,5 +1,8 @@
 environment_ids = ["math500", "aime2024"]
-rollouts_per_example = [1, 16]
+rollouts_per_example = 1
+
+[rollouts_per_example_per_env]
```
mikasenghaas: yea i dont like this way of configuring.. seems like the wrong nesting, see discord

examples/eval_with_api_push.toml
```diff
@@ -0,0 +1,27 @@
...
+environment_ids = ["reverse-text"]
+rollouts_per_example = 8
+num_examples = 100
+
+# Enable pushing evaluation results to Prime Hub
+# Requires prime-cli to be installed and configured with API credentials
+# Install: pip install prime-cli
+# Configure: prime config set-api-key YOUR_API_KEY
+# When enabled, displays the evals viewer URL after each push
+push_to_hub = true
```
mikasenghaas: should prob make clear that this is env hub, not hf hub. will not be clear for most external users o/w

examples/eval_with_api_push.toml
```diff
@@ -0,0 +1,27 @@
...
+
+# Optional: Save to disk (default: true)
+save_to_disk = true
```
mikasenghaas: would also like to use this pr bundle to use the same config keys between verifiers and prime-rl for ones that are shared

src/prime_rl/eval/eval.py
```diff
@@ -1,18 +1,238 @@
...
+                hub_results.append(result_entry)
+            
+            eval_name = f"{env_model}-{eval_id}-step{ckpt_step}"
+            push_eval_to_prime_hub(
+                eval_name=eval_name,
+                model_name=env_model,
+                dataset=eval_id,
+                metrics=hub_metrics,
+                metadata=hub_metadata,
+                results=hub_results if hub_results else None,
```
mikasenghaas, line 254: i might be blind but are we not pushing the rollouts itself? that's the most important to see

src/prime_rl/eval/eval.py
```diff
@@ -53,14 +273,11 @@ async def eval(config: OfflineEvalConfig):
     # Run benchmarks on base model
     if config.eval_base:
         logger.info(f"Evaluating model {config.model.name}")
-        await run_evals(
+        await _run_evals_with_prime_rl_features(
```
mikasenghaas: no need to change func name here imo

examples/eval_with_api_push.toml
d42me: Renamed to save_to_env_hub

configs/deepscaler/eval.toml
d42me: Updated now in verifiers

src/prime_rl/eval/eval.py
d42me: Reverted

src/prime_rl/eval/eval.py
d42me: Updated

PR #1073, opened 2025-10-06
PEFT-compatible LoRA adapter saving

kalomaze:
- adds a `save_adapter_separately` variable to the configuration save_lora_config() produces a peft-compatible lora adapter config JSON, extracts + detects both the target lora modules (target_modules) & fully trainable modules (modules_to_save)
- weights.py has `_get_adapter_state_dict`, it only triggers when: `save_adapter_separately=True`, the process is master rank, and lora layers are found. this function transforms/serializes the keys to PEFT format. we specifically must add the .weight suffix to the A/B lora matrices, while the `base_model.model.*` prefix is added to everything that requires_grad (all trainable parameters) and is not None in `named_parameters()`
- adapters are saved to `{step}/lora_adapters/adapter_model.bin` and `adapter_config.json` via `_save_lora_adapters()`
- we raise ValueError if save adapter is set to True but LoRA is disabled
- both train.py scripts in SFT and RL paths pass the lora config to `setup_weight_ckpt_manager()`
- removed obsolete/unused stray function `get_lora_state_dict()`

### DIFF
--- a/src/prime_rl/trainer/weights.py
+++ b/src/prime_rl/trainer/weights.py
@@ -13,7 +13,6 @@ from transformers.tokenization_utils import PreTrainedTokenizer
 from prime_rl.trainer.config import CheckpointConfig, LoRAConfig
 from prime_rl.trainer.lora import (
     clean_lora_state_dict,
-    get_lora_state_dict,
     has_lora_layers,
     merge_lora_weights_inplace,
     restore_lora_weights_inplace,
--- a/src/prime_rl/trainer/weights.py
+++ b/src/prime_rl/trainer/weights.py
@@ -239,9 +239,11 @@ class WeightCheckpointManager:
         has_lora = has_lora_layers(model)
 
         # Save LoRA adapters separately if configured
-        if self.config.save_adapter_separately and has_lora and self._is_master:
-            lora_state = self._get_adapter_state_dict(model)
-            self._save_lora_adapters(lora_state, model, step)
+        if self.config.save_adapter_separately and has_lora:
+            if self._is_master:
+                lora_state = self._get_adapter_state_dict(model)
+                self._save_lora_adapters(lora_state, model, step)
+            torch.distributed.barrier()
 
         cpu_state = self._gather_weights(model, dtype, has_lora_layers=has_lora)
         if _has_tt_moe_layers(cpu_state):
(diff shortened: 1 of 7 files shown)


src/prime_rl/trainer/weights.py
```diff
@@ -191,6 +238,11 @@ def save(
         """Save a HF-compatible weight-only checkpoint for a given step."""
         has_lora = has_lora_layers(model)
 
+        # Save LoRA adapters separately if configured
+        if self.config.save_adapter_separately and has_lora and self._is_master:
+            lora_state = self._get_adapter_state_dict(model)
+            self._save_lora_adapters(lora_state, model, step)
+
         cpu_state = self._gather_weights(model, dtype, has_lora_layers=has_lora)
```
Jackmin801, line 248: I think rn the gather still tries to merge even if we want to save the adapter separately. It shouldnt affect the result since the intended flow is that we only update adapters and therefore it doesnt matter what the cpu state is but might be good to not waste cycles merging

kalomaze: We still merge ephemerally for full vllm shape compatibility due to limitations with vllm lora support, and my understanding was that refactoring / doing a custom collective_rpc that merges in memory for vllm was out of scope.

src/prime_rl/trainer/weights.py
```diff
@@ -249,7 +301,17 @@ def setup_weight_ckpt_manager(
...
 ) -> WeightCheckpointManager | None:
     if weight_ckpt_config is None:
         return None
-    return WeightCheckpointManager(output_dir, weight_ckpt_config, ckpt_config, async_level=async_level)
+    
+    if weight_ckpt_config.save_adapter_separately and lora_config is None:
+        raise ValueError(
+            "save_adapter_separately=True requires LoRA to be enabled. "
+            "Set model.experimental.lora or disable save_adapter_separately."
+        )
```
mikasenghaas: should put this into pydantic config as a model validator. general rule is that config errors should be caught at the time of pydantic validation so once we enter the main loop we can assert that the set of configs makes sense

src/prime_rl/trainer/weights.py
```diff
@@ -191,6 +238,11 @@ def save(
         """Save a HF-compatible weight-only checkpoint for a given step."""
         has_lora = has_lora_layers(model)
 
+        # Save LoRA adapters separately if configured
+        if self.config.save_adapter_separately and has_lora and self._is_master:
```
mikasenghaas: isnt it a little weird that we save the lora adapters in the gather weights function? shouldn't this be done in the save function?

mikasenghaas (commented):
do i understand correctly that, if enabled, we save the lora adapters *and* the full weight checkpoint?

src/prime_rl/trainer/weights.py
kalomaze: done!

src/prime_rl/trainer/weights.py
kalomaze: i think you misread the diff lol it is within save()

mikasenghaas (approved):
lgtm!

PR #1079, opened 2025-10-08
Support loading dataset subsets and dynamic up/downsampling in SFT

mikasenghaas:
This PR allows more configurability to the SFT data mix:
- Specify training on multiple subsets (`--data.subsets`) and/or splits (`--data.splits`). If None is specified, we default to loading the default train split. If one is specified we default to `subset=None` and `split="train"` for the other. If both are specified, we assume that `splits[i]` is the split for `subsets[i]` and so we check for a length match.
- Dynamically define the ratio of each subset/split via `--data.probabilities`. We rely on `datasets.interleave_datasets` to achieve this, here is a minimal example of how it works

```python
from datasets import Dataset, interleave_datasets

a = Dataset.from_dict({"x": ["a1"]})
b = Dataset.from_dict({"x": ["b1", "b2", "b3", "b4", "b5", "b6", "b7", "b8", "b9"]})
ds = [a, b]
d = interleave_datasets(ds, probabilities=[0.5, 0.5], stopping_strategy="all_exhausted")
print("a: ", a["x"])
print("b: ", b["x"])
print("\nSampling order: ", end="")
for x in d:
    print(x["x"], end=" ")
```

```
a:  Column(['a1'])
b:  Column(['b1', 'b2', 'b3', 'b4', 'b5'])

Sampling order: a1 b1 a1 a1 a1 b2 a1 a1 b3 a1 a1 a1 a1 b4 a1 b5 b6 a1 a1 b7 b8 a1 b9
```

Misc changes:
- We use `interleave_datasets` instead of `concatenate_datasets` to support dynamic up- and downsampling via `probabilities`
- The above required some refactoring of our `iter` function because epoch borders are now not specified by the total number of examples (sum of all splits) but by the randomness of the sampling. We now define an epoch border precisely as "the dataset has run out of samples"
- A bunch of updates and improvements to thoroughly test the dataset and dataloader (checkpointing) behavior
- Deprecates `num_examples` on fake data. It's always an infinite dataloader
- Do not pass config object into base data classes to make them easier to unit test

---

**GitHub Issue**: #1077,  #1078
**Linear Issue**: Resolves PRIMERL-155, PRIMERL-156

### DIFF
--- a/src/prime_rl/trainer/sft/config.py
+++ b/src/prime_rl/trainer/sft/config.py
@@ -62,12 +62,27 @@ class SFTDataConfig(BaseDataConfig):
     name: Annotated[str, Field(description="Name or path of the HF dataset to use.")] = (
         "PrimeIntellect/Reverse-Text-SFT"
     )
+    subsets: Annotated[list[str] | None, Field(description="Subsets to use from the HF dataset.")] = None
     splits: Annotated[list[str], Field(description="Splits to use from the HF dataset.")] = ["train"]
     shuffle: Annotated[bool, Field(description="Whether to shuffle the dataset at the beginning of each epoch.")] = True
-    seed: Annotated[int, Field(description="Random seed to use for shuffling the dataset. We also shuffle at the end of each epoch by adding epoch count to the seed.")] = 0
+    seed: Annotated[
+        int,
+        Field(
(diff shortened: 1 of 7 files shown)


mikasenghaas:
Screenshot 2025-10-09 at 8 50 54 PM

made sure that we still have deterministic ckpt resume (tested on single rank but pretty sure multi-rank works as well)

PR #1081, opened 2025-10-08
Sliding Window Attention Support

adi-kmt:
Sliding window Attention implementation
-------------------------------------------------
Saw a TODO on implementing sliding window attention. Tried to implement it !!
Also we can pass window in the Flash attention, let me know if that was the intended purpose of the TODO

Ran some shape tests in this [colab](https://colab.research.google.com/drive/1EI0150c9cJPIPMwouGfgfwpmnoNEupbj#scrollTo=unI2uDjvMm5w)


samsja:
Hey, thanks for the PR, for now we don't need to impl sliding window, at least no model we support need it. Lets talk about adding this pr back when we need it it ?

adi-kmt:
Cool, no worries

PR #1087, opened 2025-10-10
Align dataset state definition, fix progress logging and multi-rank resume

mikasenghaas:
#1085 introduced reporting `progress/num_samples` and `progress/num_tokens` directly from the statistics the dataset would keep. However, these are local metrics and so we are reporting the wrong thing. To revert to reporting a proxy of the global number of samples trained on (without introducing additional comm) was to revert the dataset step definition to be strictly increasing (technically, this is not the true number of samples trained on, but "seen" but ideally these two are close to each other). This PR also fixes an inconsistency in how we handled step and epoch increments. Now, we increment the epoch counter *at the beginning of entering a loop* (instead of at the end) to make it checkpoint compatible.

---

### DIFF
--- a/src/prime_rl/trainer/sft/data.py
+++ b/src/prime_rl/trainer/sft/data.py
@@ -37,7 +37,7 @@ class StatefulIterableDataset(Stateful, IterableDataset):
     """SFT dataset are iterable (infinite) and stateful (can be checkpointed)."""
 
     def __init__(self):
-        self.step, self.epoch = -1, 0
+        self.step, self.epoch = 0, 0
         self.num_samples = defaultdict(int)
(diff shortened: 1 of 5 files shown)


mikasenghaas:
[W&B Project](https://wandb.ai/primeintellect/mika?nw=wxrfxdqb02d) 

Single-rank resume still functional

Screenshot 2025-10-10 at 2 40 29 PM

mikasenghaas:
Also in the middle of a run

Screenshot 2025-10-10 at 2 46 37 PM

Screenshot 2025-10-10 at 2 47 21 PM

mikasenghaas:
And for multi-rank

Screenshot 2025-10-10 at 3 31 11 PM

PR #1101, opened 2025-10-13
Double sided masking

faresobeid:


src/prime_rl/trainer/rl/config.py
```diff
@@ -24,7 +24,8 @@ class LossConfig(BaseModel):
         bool, Field(description="Whether to normalize the importance ratio by the sequence length.")
     ] = False
 
-    clip_ratio: Annotated[float, Field(ge=0)] = 8.0
+    mask_ratio_high: Annotated[float, Field(ge=0)] = 8.0
+    mask_ratio_low: Annotated[float, Field(ge=0)] = 0.0
```
mikasenghaas: this is to be backward compatible right? but we're alr not backward compatible bc we mask instead of clip correct? if so, would prob go with default being your gut feeling for what is best, which is prob double sided no?

src/prime_rl/trainer/rl/loss.py
```diff
@@ -82,51 +82,63 @@ def compute_loss(
...
-
-        if loss_config.ratio_type == "token":
-            is_clipped = is_clipped[loss_mask]
-
-        # Apply loss mask and sum
-        loss = (loss[loss_mask]).sum()
+        is_masked_low = importance_ratio < loss_config.mask_ratio_low
+        is_masked_high = importance_ratio > loss_config.mask_ratio_high
+        is_masked = is_masked_low | is_masked_high
+        keep_mask = loss_mask & ~is_masked
```
mikasenghaas, line 111: rough

src/prime_rl/trainer/rl/loss.py
```diff
@@ -82,51 +82,63 @@ def compute_loss(
...
+    total_masked_mismatch_kl = []
+    total_unmasked_mismatch_kl = []
+    total_is_masked = []
+    total_is_masked_low = []
+    total_is_masked_high = []
 
     for trainer_logprobs, inference_logprobs, advantages, loss_mask in zip(
         trainer_logprobs, inference_logprobs, advantages, loss_mask
     ):
-        total_tokens = torch.clamp_min(loss_mask.sum(), 1)
```
mikasenghaas, line 91: why remove this? seeing `torch.clamp_min(loss_mask.sum(), 1)` being used multiple times below

mikasenghaas (commented):
yea logic looks fine to me, altho im still a little lost on all the loss flags and conditionals hahah. if you could do a quick run on one of our examples (wordle or hendrycks math), probably would prefer with double not one-sided masking because afaiu this is sota now, and put the wandb logs in the pr, then happy to merge

src/prime_rl/trainer/rl/loss.py
faresobeid: ya but only used twice so felt cleaner, can change

src/prime_rl/trainer/rl/loss.py
faresobeid: ya lol, ended up on this design but im sure it can be made cleaner

src/prime_rl/trainer/rl/config.py
faresobeid: ya i guess we still don't have a good value (will push the icepop authors again for this), can set to 0.1 which seems reasonable

faresobeid:
https://wandb.ai/primeintellect/acereason_tests/workspace?nw=u5r4iq1nh3 
Logs look good so far, also using mask low 0.1 u can see more tokens are being clipped below than above

src/prime_rl/trainer/rl/config.py
mikasenghaas: ah let's make it matching tho, hahah (1/10,10) or (1/8/8) unless you have a strong opinion on why we should mask more aggressively to one side

src/prime_rl/trainer/rl/loss.py
mikasenghaas: all good

src/prime_rl/trainer/rl/config.py
faresobeid: don't have a strong opinion although the recent meta paper and CISPO both don't even mask from below so would've guessed it should be slightly less aggressive. Not sure tho, ideally the icepop authors would share their default

samsja (approved):
lgtm

PR #1103, opened 2025-10-14
Strip content to make incremental tokenization work

mikasenghaas:
If an assistant message starts with a newline we might fail building the loss mask, because tokens get merged weirdly.


samsja (approved):
lgtm

PR #1105, opened 2025-10-14
add data debug script

samsja:
# Add script for debugging

This script is a drop in replacement for the train.py that only run the data, useful for debugging the data pipeline.

Example usage:

lets assume your training run with this command :
  ```
  uv run torchrun --local-ranks-filter 0 --nproc_per_node=2 src/prime_rl/trainer/sft/train.py @ configs/debug/sft/train.toml
  ```
    
and that you want to debug the data pipeline, you can run this command to only run the data:
```
    uv run torchrun --local-ranks-filter 0 --nproc_per_node=2 src/prime_rl/trainer/sft/data_debug.py @ configs/debug/sft/train.toml
```
    
notice how the only change is the script name. The API is the name but only tests the data.

### DIFF
--- a/src/prime_rl/trainer/sft/data_debug.py
+++ b/src/prime_rl/trainer/sft/data_debug.py
@@ -1,0 +1,136 @@
+"""
+This script is a drop in replacement for the train.py that only run the data, useful for debugging the data pipeline.
+
+Example usage:
+
+lets assume your training run with this command :
+    uv run torchrun --local-ranks-filter 0 --nproc_per_node=2 src/prime_rl/trainer/sft/train.py @ configs/debug/sft/train.toml
+    
+and that you want to debug the data pipeline, you can run this command to only run the data:
+
+    uv run torchrun --local-ranks-filter 0 --nproc_per_node=2 src/prime_rl/trainer/sft/data_debug.py @ configs/debug/sft/train.toml
+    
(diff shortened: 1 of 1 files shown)


src/prime_rl/trainer/sft/data_debug.py
```diff
@@ -0,0 +1,136 @@
...
+        raise ValueError(
+            f"The number of micro batches ({num_micro_batches}) must be divisible by the world size ({world.world_size})."
+        )
+
+    # Optionally, resume training from a checkpoint
+    progress = Progress()
+
+    while True:
+
+
```
mikasenghaas, line 81: is ruff not deleting those whitespaces haah?

mikasenghaas (commented):
i see how it's a very useful script but i don't like that we have a whole "data debug" script in our src/ folder right next to the actual entrypoints. is there any way we can more clearly distinguish that this is a "special debugging entrypoint"

samsja:
> i see how it's a very useful script but i don't like that we have a whole "data debug" script in our src/ folder right next to the actual entrypoints. is there any way we can more clearly distinguish that this is a "special debugging entrypoint"

we can put in scripts/ but then it not clear that its only for sft, or we can have a sft/debug

mikasenghaas:
i dont think this is the right way of adding this functionality, closing for now

PR #1107, opened 2025-10-14
Misc QoL SFT changes

mikasenghaas:
- Increase default NCCL timeout to build initial process group from 600s -> 1200s
- Build as many buckets as possible to reduce number of pad tokens in `StackDataset`
- More logs around the model and data loading for better understanding which operations take long (and where stuff might hang)


samsja (approved):
lgtm

PR #1110, opened 2025-10-15
add nan count

samsja:
# handling nan loss

CP randomly return nan loss that doesn't seems to impact the training, at relavilty large batch you just need one nan per batch to make the whole batch reporting no loss, reducing visibility on what is happening in the training.

This pr skip the nan loss in the loss report, it also count the number of nan


mikasenghaas (approved):
looking good

PR #1111, opened 2025-10-16
fix: eliminate weight update bottleneck with dedicated admin client

AmeenP:
## Problem
During RL training with high concurrent rollouts, weight update operations (`update_weights`, `reload_weights`) experienced 8-30 second delays before the server received requests due to shared HTTP connection pool saturation.

## Root Cause
- Shared httpx client pool saturates with streaming inference requests
- HTTP/1.1 requires connections to be IDLE for reuse
- Thousands of active streaming connections = 0 IDLE connections
- Weight update requests queue waiting for available connections

## Solution
Create dedicated HTTP client for admin endpoints that bypasses the saturated shared pool. Uses `Connection: close` header for one-shot connections.

## Implementation
- Add `_admin_client()` helper function
- Update `update_weights()` to use dedicated client
- Update `reload_weights()` to use dedicated client

## Impact
- **Weight updates: 8.58s → 55ms** (155x faster under 4k concurrent streams)
- Zero queue delay under production load
- No impact on streaming workload
- Minimal code change (20 lines added)

## Testing
Validated on 8xH100 with 4,000 concurrent streams matching production configuration.

### DIFF
--- a/src/prime_rl/eval/eval.py
+++ b/src/prime_rl/eval/eval.py
@@ -6,6 +6,7 @@ from prime_rl.orchestrator.client import (
     check_has_model,
     check_health,
     reload_weights,
+    setup_admin_client,
     setup_client,
     update_weights,
 )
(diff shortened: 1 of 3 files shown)


mikasenghaas (commented):
very nice! let's fix the failing test and then merge

samsja:
> very nice! let's fix the failing test and then merge

failing test was about zombie on ci

samsja (approved):
lgtm, great work :)

PR #1112, opened 2025-10-16
add skip clone in install.sh

samsja:
this pr add support for not cloning when using install.sh

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -11,6 +11,9 @@ log_warn()  { echo -e "${YELLOW}[WARN]${NC} $*"; }
 
 REPO_ID="prime-rl"
 
+# Flag defaults
+SKIP_CLONE=0
+
 has_ssh_access() {
     # Probe SSH auth to GitHub without prompting; treat any nonzero as "no ssh"
     # We try a quick ls-remote to avoid cloning on failure.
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -47,18 +58,22 @@ main() {
...
+            log_warn "SSH auth to GitHub not available. Cloning via HTTPS."
+            git clone https://github.com/PrimeIntellect-ai/${REPO_ID}.git
+        fi
+
+        log_info "Entering project directory..."
+        cd ${REPO_ID}
     fi
 
-    log_info "Entering project directory..."
-    cd ${REPO_ID}
```
mikasenghaas, line 60: i dont think we can remove this if we do not skip clone? later commands depend on us being in the prime-rl repo, no?

scripts/install.sh
```diff
@@ -35,6 +38,14 @@ ensure_known_hosts() {
...
 
 main() {
+    # Parse args
+    for arg in "$@"; do
+        case "$arg" in
+            --skip-clone)
+                SKIP_CLONE=1
+                ;;
+        esac
+    done
```
mikasenghaas: format looks weird here?

mikasenghaas: why is there so many lines?

samsja: > why is there so many lines?

entirely vibe coded

mikasenghaas (approved):
lfg

PR #1116, opened 2025-10-17
Multi-client support to increase inference throughput

mikasenghaas:
This PR introduces the ability to communicate with multiple vLLM (DP) server replicates from the orchestrator and eval entrypoint. It makes the following changes:
- The init functions `setup_client` and `setup_admin_client` function become `setup_clients` and `setup_admin_clients`. For now, they are identical clients but point to a unique `base_url`, e.g. this means we could specify `--client.base-url url1,url2` which would create too unique clients which share all arguments (e.g. API key, timeout, etc.) but point to unique inference servers
- The admin functions `check_health`, `update_weights` and `reload_weights` are sent from all clients into each inference server and synchronously awaited with an asyncio gather.
- The OAI client functions (namely using OAI clients to make chat completions requests) are reworked to do round-robin at the example level. For example, if we have a batch of four examples, the first and third problem will be processed by the first inference server through the first clients, while the second and forth problem will be processed by the second inference server through the second client. Since all rollouts for a unique prompt are processed on one server we maximize cache hits, while load balancing across the servers

Other changes:
- Moved client utils into top-level `src/prime_rl/utils/client.py` as their are used by both the orchestrator and eval script
- Moved some verifiers utilities into `src/prime_rl/utils/vf.py` 
- Use the example ID that is part of the `verifiers` env dataset directly instead of creating our own in eval script and buffer

## Benchmarks

We validate that we get linear scaling when deploying multiple independent vLLM instances. We host `zai-org/GLM-4.5-Air` on one or more 8xH200 node(s). We use the first 256 examples of the `math500` environment and run with `rollouts_per_example` to match the batch size `B`. We generate precisely 512 tokens per sequence and measure end-to-end throughput.

#Nodes | B=1K | B=2K | B=4K
-- |  -- | -- | -- 
1 |  12153.0 tok/sec | 12501.6 tok/sec | 13038.7 tok/sec
2 |  12501.6 tok/sec | 22322.5 tok/sec | 24102.6 tok/sec
4 |  14950.7 tok/sec | 25541.4 tok/sec | 40084.8 tok/sec

We see near linear scaling, as expected. Because each inference server can process at most 1024 sequences in parallel (default for H200), a single server's throughout plateaus at 1K batch size, two servers plateau at 2K batch size and 4 servers plateau at 4K batch size.

## Eval Speedup

We also run a quick real-world eval suite to check how much speedup we can expect when there are long-tail generations.

```bash
uv run eval \
    --client.base-url $BASE_URL \
    --model.name $MODEL_NAME \
    --environment-ids math500,aime2024,aime2025,gpqa \
    --num-examples 500,30,30,198 \
    --rollouts-per-example 1,16,16,2 \
    --sampling.max-tokens 32768
```

#Nodes | Eval Time | Speedup
-- |  -- | --
1 | 01h:37m:55s | -
2 | 00h:59m:57s | 1.6x
4 | 00h:31m:53s | 3.1x


mikasenghaas:
Benchmark Script


```python
import argparse
import asyncio
import time

from verifiers import load_environment

from prime_rl.utils.client import setup_clients
from prime_rl.utils.config import ClientConfig
from prime_rl.utils.vf import generate_batch


async def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--base-url", "-b", nargs="+", type=str)
    parser.add_argument("--num-examples", "-n", type=int, default=5)
    parser.add_argument("--rollouts-per-example", "-r", type=int, default=3)
    parser.add_argument("--max-tokens", "-t", type=int, default=128)
    parser.add_argument("--max-concurrent", "-c", type=int, default=-1)
    args = parser.parse_args()

    env = load_environment("math500", use_think=True)
    client_config = ClientConfig(base_url=args.base_url)
    clients = setup_clients(client_config)
    print(
        f"Evaluating {args.num_examples=}, {args.rollouts_per_example=}, {args.max_tokens=}, {args.max_concurrent=}..."
    )
    start_time = time.time()
    await generate_batch(
        clients=clients,
        env=env,
        model_name="zai-org/GLM-4.5-Air",
        problems=env.get_eval_dataset(n=args.num_examples).to_list(),
        sampling_args=dict(max_tokens=args.max_tokens),
        rollouts_per_example=args.rollouts_per_example,
        max_concurrent=args.max_concurrent,
    )
    end_time = time.time()
    time_taken = end_time - start_time
    tokens_processed = args.num_examples * args.rollouts_per_example * args.max_tokens
    print(f"Throughput: {tokens_processed / time_taken:.1f} tok/sec ({tokens_processed} in {time_taken:.1f} seconds)")


if __name__ == "__main__":
    asyncio.run(main())
```

```bash
uv run bench.py -b http://10.20.0.56:8000/v1 -n 256 -r 4 -t 512
```
```bash
uv run bench.py -b http://10.20.0.56:8000/v1 http://10.20.0.57:8000/v1 -n 256 -r 8 -t 512
```
```bash
uv run bench.py -b http://10.20.0.56:8000/v1 http://10.20.0.57:8000/v1 http://10.20.0.58:8000/v1 http://10.20.0.59:8000/v1 -n 256 -r 16 -t 512
```

mikasenghaas:
Screenshot 2025-10-18 at 5 50 04 PM

Screenshot 2025-10-18 at 5 50 07 PM

Issue #926, opened 2025-09-16
Remove pynvml fron uv.lock


samsja:
we should just remove pynvml from deps to avoid this warning

```

  import pynvml  # type: ignore[import]
/data/prime-rl/.venv/lib/python3.12/site-packages/torch/cuda/__init__.py:63: FutureWarning: The pynvml package is deprecated. Please install nvidia-ml-py instead. If you did not install pynvml directly, please report this to the maintainers of the package that installed pynvml for you.


```

PR #1121, opened 2025-10-19
Remove `flashinfer`

mikasenghaas:
We do not need the performance boost and remove a potential source of trainer-inference mismatch. Tested performance against 4x8xH200 nodes with the following debug suite

```bash
export BASE_URL=...
uv run eval \
    --client.base-url $BASE_URL \
    --model.name zai-org/GLM-4.5-Air \
    --environment-ids math500,aime2024,aime2025,gpqa \
    --num-examples 500,30,30,198\
    --rollouts-per-example 1,16,16,2 \
    --sampling.max-tokens 32768 \
```

vLLM | Eval Time | Diff. |
-- |  -- | -- |
With `flashinfer` | 00h:31m:53s | - |
Without `flashinfer` | 00h:32m:39s | +0.7% |

---

**GitHub Issue**: #926 
**Linear Issue**: Resolves PRIMERL-106

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,6 @@ eval = "prime_rl.eval.eval:main"
 
 [project.optional-dependencies]
 flash-attn = ["flash-attn>=2.8.3"]
-flash-infer = ["flashinfer-python>=0.2.8rc1"]
 
 [dependency-groups]
 dev = [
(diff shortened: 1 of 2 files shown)


samsja (approved):
Lgtm

PR #1127, opened 2025-10-20
WIP: Feat multi step rollout2

S1ro1:
WIP


src/prime_rl/orchestrator/config.py
```diff
@@ -502,6 +502,28 @@ class OrchestratorConfig(BaseSettings):
...
+
+    inflight_problems_factor: Annotated[
+        float,
+        Field(
+            gt=0,
+            description="Only used if `areal=True`. Max fraction of the batch that is being generated at once. Defaults to 1.5, which means at most 1.5x the batch size is being generated at once.",
+        ),
+    ] = 1.5
+
+    difficulty_filtering: Annotated[
```
```suggestion
    filter_advantage_zero: Annotated[
```

src/prime_rl/orchestrator/buffer.py
```diff
@@ -231,6 +231,15 @@ def sample_problems(self, n: int) -> list[dict]:
 
         return sampled_problems
 
+    def sample_problem(self) -> tuple[list[int], list[dict]]:
```
mikasenghaas, line 234: i dont get why we need this method? why not call `buffer.sample_problems(n=1)`?

mikasenghaas: ig in this function we have a check that the problem cannot already be in the rollout buffer - but i dont see why this would be needed for areal specifically, just seems like another flavor the the simple buffer

src/prime_rl/orchestrator/config.py
```diff
@@ -388,6 +388,23 @@ class AdvantageConfig(BaseConfig):
...
+class ArealConfig(BaseConfig):
+    inflight_problems_factor: Annotated[
+        float,
+        Field(
+            gt=0,
+            description="Maximum fraction of the batch that is being generated at once. Defaults to 1.5, which means at most 1.5x the batch size is being generated at once.",
+        ),
+    ] = 1.5
+
+    filter_advantage_zero: Annotated[
```
mikasenghaas, line 400: i dont think this arg is/ should be areal concern. this is the buffer's concern. the buffer exposes `sample_rollouts` which samples from the completed rollouts. e.g. we could easily make a buffer that ensures that zero advantage groups/ rollouts are never returned. but this functionality should be available whether or not we are in areal mode

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -194,35 +229,93 @@ async def orchestrate(config: OrchestratorConfig):
...
-            sampling_args["extra_body"]["repetition_penalty"] = sampling_args.pop("repetition_penalty")
-
-            # Generate completions + rewards with verifiers
-            generate_completions_start_time = time.time()
-            generate_outputs: GenerateOutputs = await generate_batch(
+        async def areal_loop():
+            nonlocal filtered
+
+            def generate_call():
+                problem_ids, problem = buffer.sample_problem()
```
mikasenghaas, line 239: would replace this with `buffer.sample_problems(n=1)`. also each problem dict is guaranteed to have an ID, so we can use that instead of explicitly returning a tuple

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -194,35 +229,93 @@ async def orchestrate(config: OrchestratorConfig):
...
+
+                task = generate_batch(
+                    clients=clients,
+                    env=vf_env,
+                    model_name=config.model.name,
+                    problems=problem,
+                    rollouts_per_example=config.rollouts_per_example,
+                    sampling_args=sampling_args,
+                    use_tqdm=False,
+                )
```
mikasenghaas, line 251: the helper `generate_group` (also in `vf.py`) should do the same?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -194,35 +229,93 @@ async def orchestrate(config: OrchestratorConfig):
...
+                    new_task = asyncio.create_task(coro)
+                    await asyncio.sleep(0)
+                    inflight_tasks.append(new_task)
+                    task_to_problem_id[new_task] = new_problem_ids
+
+                    candidate_generate_outputs: GenerateOutputs = task.result()
+
+                    rewards = candidate_generate_outputs.reward
+
+                    if config.areal.filter_advantage_zero and all(reward == rewards[0] for reward in rewards):
```
mikasenghaas, line 285: yea it really feels like this should live inside `sample_rollouts` of the buffer. basically, i think both the areal and default loop should have a call to `buffer.update_rollouts` and then `buffer.sample_rollouts`, this way we can immediately be compatible with any buffer type

src/prime_rl/utils/vf.py
```diff
@@ -104,25 +104,63 @@ async def generate_batch(
     rollouts_per_example: int,
     sampling_args: dict,
     max_concurrent: int = -1,
+    use_tqdm: bool = True,
 ) -> GenerateOutputs:
     """Asynchronously generate and score rollouts for a list of problems."""
-    from tqdm import tqdm
+    if use_tqdm:
```
mikasenghaas, line 110: we shouldn't need the tqdm toggle if we use `generate_group`

mikasenghaas (commented):
nice work! i think we're almost there. i think the main thing that's missing is to clearly separate out the interface of the buffer and the "areal"/'default" scheduler (not sure this is the right name). but basically, i think of areal being a scheduling block that wraps a buffer to schedule rollouts reqs to build batches. it should by default be compatible with any buffer.

default scheduling (current main): we sample `batch_size // rollouts_per_example` problems, await all, update buffer with rollouts, sample `batch size` rollouts. if less than `batch_size` rollouts are valid, loop. else, exit

areal scheduling: sample `batch_size // rollouts_per_examples` (+maybe some oversampling factor) times a single problem from the buffer. schedule and await them asynchronously, update the buffer with the returned rollouts as they come in. each time the buffer is updated, peak into the buffer if a full batch is available (e.g. the buffer would decide to exclude zero advantage problems from this, areal is invariant to this). if so, yield the batch

src/prime_rl/orchestrator/buffer.py
S1ro1: This was just copied from old Jackmin code, I think n=1 should work to be fair. I'll readapt

src/prime_rl/orchestrator/orchestrator.py
S1ro1: The issue with this, is we don't know how many rollouts to generate for the default scheduling loop, so after filtering we get a full batch. Or do you imagine it something like that the default loop would be something like:

```
while not buffer.is_ready():
    generate(n=batch_size - len(buffer))
    buffer.update()
```

PR #1131, opened 2025-10-20
update debug moe and add docs debug

samsja:
Refactor the debug config for moe, preparing for end to end rl integration tests with moe


configs/debug_moe/sft/train.toml
mikasenghaas, line 1: personally, i think the no underscore is cleaner

configs/debug_moe/infer.toml
```diff
@@ -0,0 +1,3 @@
+[model]
+enforce_eager = true
```
mikasenghaas: we can prob remove eager mode from here. also in the dense debug config. this is an artifact on vllm with cudagraph having slow init time, but this is not the case anymore since vLLM v0.10.0

docs/debug.md
```diff
@@ -0,0 +1,50 @@
+# Debugging
+
+this page reference to the debug configs in the `configs/debug` and `configs/debug_moe` directories. 
+
+This config are designed for debugging and reducing the iteration loop for development. Most of this configuration are also tested in our CI pipeline.
+
+
+All dense debug are using
```
mikasenghaas, line 8: ?

docs/debug.md
```diff
@@ -0,0 +1,50 @@
...
+uv run torchrun --local-ranks-filter 0 --nproc_per_node=2 src/prime_rl/trainer/rl/train.py @ configs/debug_moe/rl/train.toml
+```
+
+## Inference
+
+```bash
+uv run inference @ configs/debug/infer.toml
+```
+
+there is no inference server for moe model yet.
```
mikasenghaas, line 44: why not?

samsja: ain't working

PR #1132, opened 2025-10-21
[Easy] Remove duplicate 'torchtitan' dep

mgoin:
Just was rummaging around and noticed there was a duplicate entry for 'torchtitan' in pyproject.toml


mikasenghaas:
thanks!

PR #1136, opened 2025-10-21
Feature: Add prime env hub integration.

d42me:
Added prime env hub integration:

```
uv run eval \
  --client.base-url https://api.openai.com/v1 \
  --client.api-key-var OPENAI_API_KEY \
  --model.name gpt-4o-mini \
  --environment-ids gsm8k \
  --num-examples 2 \
  --rollouts-per-example 1 \
  --save.env-hub
```

Hub ID will be auto-resolved, and it also works well for multiple environments


src/prime_rl/orchestrator/config.py
```diff
@@ -159,9 +159,24 @@ class EvalSaveHFConfig(BaseConfig):
...
+
+    run_id: Annotated[
+        str | None,
+        Field(description="Link to existing training run ID. Either run_id or env_hub_id must be provided."),
+    ] = None
+
+    env_hub_id: Annotated[
+        str | None,
+        Field(description="Environment Hub ID. Either run_id or env_hub_id must be provided."),
+    ] = None
```
```suggestion
    run_id: Annotated[
        str | None,
        Field(description="Link to existing training run ID. Either run_id or env_hub_id must be provided."),
    ] = None

    env_hub_id: Annotated[
        str | None,
        Field(description="Environment Hub ID. Either run_id or env_hub_id must be provided."),
    ] = None
    
        @model_validator(mode="after")
    def id_check(self):
        if self.run_id is None and self.env_hub_id is None:
            raise ValueError(f"Either run_id or env_hub_id must be provided but got {self.run_id}, {self.env_hub_id}"
 
        return self
```

src/prime_rl/eval/utils.py
```diff
@@ -188,6 +188,48 @@ async def run_eval(
...
+            from prime_core import APIClient
+            from prime_evals import EvalsClient
+
+            results_list = []
+            for i in range(len(dataset)):
+                sample = dataset[i]
+                results_list.append(dict(sample))
+
+            # Push to Prime Hub
+            api_client = APIClient()
```
samsja: can we create only one client per task instead of each task ?

src/prime_rl/orchestrator/config.py
```diff
@@ -159,9 +159,31 @@ class EvalSaveHFConfig(BaseConfig):
...
```
mikasenghaas: i dont understand thees two values. what is the difference? why do i need to set them at all? i would expect it to generate a uuid on the fly and push to evals viewer of the environment that is being evaluated. might be missing smth here tho but seems odd

src/prime_rl/eval/utils.py
```diff
@@ -199,28 +246,30 @@ async def run_evals(
...
-            )
-            for env_id, num_examples, rollouts_per_example, max_concurrent in zip(
-                eval_config.environment_ids,
-                eval_config.num_examples,
-                eval_config.rollouts_per_example,
-                eval_config.max_concurrent,
-            )
-        ]
-    )
+    async with get_evals_client_if_needed(eval_config.save.hub) as evals_client:
```
mikasenghaas: not a huge fan of this being a context manager, and also dont think it needs to be conditional. imo fine to just always create the eval client and pass it through, like we do with our oai clients and admin clients. unless there is some heavy setup involved in creating the client? but that'd be surprising

src/prime_rl/eval/utils.py
```diff
@@ -188,6 +200,41 @@ async def run_eval(
                 f"Pushed {'private' if save_config.hf.private else 'public'} eval results for {env_id} to HF Hub (https://huggingface.co/datasets/{repo_name})"
             )
 
+        if save_config.hub is not None:
+            results_list = []
+            for i in range(len(dataset)):
+                sample = dataset[i]
+                results_list.append(dict(sample))
```
mikasenghaas: `dataset.to_list()` should do the same?

src/prime_rl/eval/utils.py
```diff
@@ -188,6 +200,41 @@ async def run_eval(
...
+
+            eval_name = f"{env_id}--{model_config.name.replace('/', '--')}"
+            metadata_dict = sanitize_metadata(results.metadata)
```
mikasenghaas: we already do this when saving to disk, imo fine to move up to where we also call `make_dataset`

src/prime_rl/eval/utils.py
```diff
@@ -188,6 +200,41 @@ async def run_eval(
...
+
+            # Determine environments or run_id
+            environments = [{"id": save_config.hub.env_hub_id}] if save_config.hub.env_hub_id else None
```
mikasenghaas: can you explain this? why is it a list of dicts?

src/prime_rl/eval/utils.py
```diff
@@ -188,6 +200,41 @@ async def run_eval(
...
+                framework="verifiers",
+                metadata=metadata_dict,
+                metrics=eval_metrics,
+            )
+
+            eval_id = create_response.get("evaluation_id") or create_response.get("id")
+
+            # Push samples if provided
+            if results_list:
+                await evals_client.push_samples(eval_id, results_list)
```
mikasenghaas: there will always be samples

src/prime_rl/eval/utils.py
```diff
@@ -188,6 +200,41 @@ async def run_eval(
...
+
+            # Finalize evaluation
+            await evals_client.finalize_evaluation(eval_id, metrics=eval_metrics)
```
mikasenghaas, line 212: what does this do? why is it not part of the main results push?

src/prime_rl/orchestrator/config.py
d42me: So, atm the eval needs to be connected either to a run or an environment, and this needs to be specified. But I see your point, let me make some changes here. I agree that this should be handled automatically.

src/prime_rl/eval/utils.py
d42me: prime-evals takes a list of dicts for environment, to optionally specify the version for the appropriate env

src/prime_rl/eval/utils.py
d42me: prime-evals works like:

1. Create an evaluation
2. Push samples
3. Finalize evaluation

The idea here was to be flexible, especially for RFT runs.

src/prime_rl/orchestrator/config.py
```diff
@@ -162,6 +162,12 @@ class EvalSaveHFConfig(BaseConfig):
 class EvalSaveConfig(BaseConfig):
     disk: EvalSaveDiskConfig | None = None
     hf: EvalSaveHFConfig | None = None
+    hub: Annotated[
```
mikasenghaas: hub is a little unspecific, ig one could figure out that its the env hub, not hf hub because there's already hf but making it more explicit with `--save.hf-hub` and `--save.env-hub` would be better here i think

src/prime_rl/orchestrator/config.py
```diff
@@ -162,6 +162,12 @@ class EvalSaveHFConfig(BaseConfig):
...
+        bool,
+        Field(
+            description="Whether to push eval results to Environment Hub. Automatically pushes all evaluated environments."
```
mikasenghaas: maybe mention that this only works for environments that one is authorized for?

mikasenghaas (commented):
nice! lgtm, one minor nit on config arg naming

mikasenghaas (approved):
code looks good now and happy to merge!

i was just testing this out tho and got an unauthorized error. ran a small eval suite from 

```bash
uv run eval --environment-ids math500 --model.name gpt-4.1-mini --num-examples 5 --rollouts-per-example 3 --client.base-url https://api.openai.com/v1 --save.env-hub
```

this was done with an api key from our prime team account with read/write access to the environment hub. 

Image

i also tried generating a new key on the fly with `prime login` but still got the unauthorized error. am i doing something wrong?

PR #1137, opened 2025-10-21
add support for local ckpt loading hf

samsja:
# Add support for local ckpt loading hf

previously we could not just point to a local folder with the hf model and load it to start the training and had to first push  to hub and re download which is expensive for large model and slow for local debugging.

This pr skip the snapshot download if the model.name is a a local directory

example

first ckpt
```
uv run torchrun --local-ranks-filter 0 --nproc_per_node=2 src/prime_rl/trainer/sft/train.py @ configs/debug/moe/sft/train.toml --model.load_using_meta --model.compile --weights.interval 2 --output_dir output_sft --weights.save_format safetensors
```

then try to load from it
```
uv run torchrun --local-ranks-filter 0 --nproc_per_node=2 src/prime_rl/trainer/sft/train.py @ configs/debug/moe/sft/train.toml --model.load_using_meta --model.compile --model.name output_sft/weights/step_2
```

### DIFF
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -1,4 +1,5 @@
 import logging
+from pathlib import Path
 import time
 from typing import cast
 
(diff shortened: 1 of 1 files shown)


mikasenghaas (approved):
looks good!

S1ro1 (approved):
LGTM

PR #1142, opened 2025-10-22
fix glm init

samsja:
maybe fix glm


mikasenghaas:
Closed in favor of  #1145

PR #1143, opened 2025-10-22
More conservative OAI client limits

mikasenghaas:
We decrease the (alive) connection limit to 1024 to avoid unnecessarily using system resources. Previously we used 28000 which was fine for a single client setup, but now that we have multiple clients it may lead to resource contention. Also, the vLLM server by default only processes 1024 sequences in parallel (which is already above the critical batch size for most inference workloads), so we are not losing performance here.


samsja (approved):
lgtm

PR #1144, opened 2025-10-22
Fix asyncio concurrency

mikasenghaas:
Because we now schedule "group rollout requests" the `max_concurrent` argument of `env.generate` does not limit the global concurrency anymore. Since [#504](https://github.com/PrimeIntellect-ai/verifiers/pull/504) of verifiers, we support explicitly passing a semaphore which prime-rl uses to limit the global asyncio concurrency. This PR makes the necessary code changes to support this in prime-rl as well.


samsja (approved):
lgtm

PR #1145, opened 2025-10-22
Fix MoE weight loading

mikasenghaas:
This PR fixes a weight loading issue that occurs when we try to load a custom (MoE) model (`--model.impl custom`) from meta device (`--model.load-using-meta`). Our custom model uses the TT format for MoE layers. When trying to load a HF MoE model (e.g. `zai-org/GLM-4.5-Air`) the FQNs do not match leading to wrongly loading the weights. Because we had `DefaultLoadPlanner(allow_partial_load=True)` this would error siltently and would manifest in high loss on the SFT trainer and high trainer-inference mismatch on the RL trainer.

Changes made:
- Use the default planner (`allow_partial_load=False`) to get explicit logs
- Automatically detect and convert if there is a mismatch between the snapshot weights and the model weights and do a best-effort attempt at conversion. This only happens on the master rank (all non-master ranks wait), and is "cached", i.e. we write the converted weights to disk for subsequent operations.

Note: The bug will still occur if one tries to load with `--model.impl custom` on CPU device.

### DIFF
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -29,6 +29,7 @@ from prime_rl.trainer.weights import (
     load_state_dict,
     save_state_dict,
 )
+from prime_rl.trainer.world import get_world
 from prime_rl.utils.logger import get_logger
 from prime_rl.utils.tensor_hashing import get_module_signature
 
@@ -169,9 +170,6 @@ def load_dcp_from_hf(model: nn.Module, config: ModelConfig):
         logger.warning("Randomly initializing model. Skipping loading weights from HF.")
         return
 
-    logger.info("Loading weights using HF DCP")
-    load_dcp_start_time = time.time()
-
     if not Path(config.name).exists():
         snapshot_path = Path(snapshot_download(repo_id=config.name, repo_type="model"))
     else:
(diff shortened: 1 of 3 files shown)


src/prime_rl/trainer/model.py
```diff
@@ -152,35 +163,68 @@ def setup_fsdp(model: nn.Module, config: ModelConfig, parallel_dims: ParallelDim
...
+
+    logger.info(f"Loading weights using HF DCP from {snapshot_path}")
+    load_dcp_start_time = time.time()
+    dcp_load(
         model.state_dict(),
-        storage_reader=HuggingFaceStorageReader(path=path_snapshot),
+        storage_reader=HuggingFaceStorageReader(path=snapshot_path.as_posix()),
         # Note: This allow is needed by weight tying but could cause silent issues
-        planner=DefaultLoadPlanner(allow_partial_load=True),
+        # planner=DefaultLoadPlanner(allow_partial_load=True),
```
samsja, line 224: why commented out, btw should we put allow_partial_load to false here

src/prime_rl/trainer/weights.py
```diff
@@ -99,6 +161,55 @@ def _convert_tt_moe_to_hf_(state_dict: dict[str, Tensor]):
         del state_dict[f"model.layers.{i}.mlp.experts.w3"]
 
 
+def load_state_dict(path: Path) -> dict[str, Tensor]:
```
samsja:
```suggestion
def load_state_dict_safetensor(path: Path) -> dict[str, Tensor]:
```

src/prime_rl/trainer/weights.py
```diff
@@ -99,6 +161,55 @@ def _convert_tt_moe_to_hf_(state_dict: dict[str, Tensor]):
...
+    safetensors_paths.sort(key=lambda x: int(x.stem.split("-")[1].split("of")[0]))
+    state_dict = {}
+    for safetensor_path in tqdm(safetensors_paths, desc="Loading weight shards", total=len(safetensors_paths)):
+        with safe_open(safetensor_path, framework="pt", device="cpu") as f:
+            for key in f.keys():
+                state_dict[key] = f.get_tensor(key)
+    return state_dict
+
+
+def save_state_dict(state_dict: dict[str, Tensor], path: Path):
```
samsja:
```suggestion
def save_state_dict_safetensor(state_dict: dict[str, Tensor], path: Path):
```

src/prime_rl/trainer/model.py
mikasenghaas: commented out bc i wanted to get your opinion here. given jack's comments it seems like this is needed for some models? but also, i really don't like allowing partial load because it fails siltently. disabling this showed me exactly which keys in the state dict were not matching which is a much nicer debugging experience.

PR #1147, opened 2025-10-22
refactor: remove args passing weight ckpt

samsja:
refactor: remove args passing weight ckpt


mikasenghaas (approved):
ah yea that was bad before, my bad (i think)

PR #1149, opened 2025-10-23
nccl broadcast

samsja:
image

two fsdp two tp weight update

image

image

image

test on single node:

* [x] 2 fsdp 2 inferece tp
* [x] 2 fsdp 1 inferece
* [x] 1 fsdp 1 inference
* [x] 2 fsdp 2 dp  
* [x] 8 fsdp 3 x 4 TP with IB


src/prime_rl/orchestrator/config.py
```diff
@@ -499,7 +506,7 @@ class OrchestratorConfig(BaseSettings):
             ge=0,
             description="Maximum number of async levels to use. If 0, will do synchronous RL. Else, it will allow to go `async_level` steps ahead of training.",
         ),
-    ] = 2
+    ] = 1
```
S1ro1, line 502: ~~Is this intended?~~ edit: i see why now, makes sense

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -156,7 +170,10 @@ async def orchestrate(config: OrchestratorConfig):
             ckpt_step = progress.step - config.async_level
             logger.info(f"Waiting for weight checkpoint {ckpt_step}")
             wait_for_weight_ckpt_start_time = time.time()
+
+            # if config.broadcast_backend == "filesystem":
```
S1ro1: nit: can remove this?

src/prime_rl/inference/vllm/worker/nccl.py
mikasenghaas, line 1: repo style nit: i dont like `worker` and `worker_nccl`. what about:
- `worker/filesystem.py` and `worker/nccl.py` _or_
- just put `FileSystemCheckpointWorker` and `NCCLCheckpointWorker` both in `worker.py`

src/prime_rl/inference/vllm/server.py
```diff
@@ -28,6 +28,12 @@
 logger = init_logger("vllm.entrypoints.openai.api_server")
 
 
+WORKER_EXTENSION_CLS = {
```
mikasenghaas, line 31: style: do we need this being a dict? never gonna be more than two types of workers no? could just if else imo

src/prime_rl/orchestrator/config.py
```diff
@@ -380,6 +380,13 @@ class AdvantageConfig(BaseConfig):
     neg_clipped: bool = False
 
 
+class NCCLBroadcastConfig(BaseConfig):
+    """Configures the NCCL broadcast."""
+
+    host: Annotated[str, Field(description="The host to use for the NCCL broadcast.")] = "localhost"
+    port: Annotated[int, Field(description="The port to use for the NCCL broadcast.")] = 29500
```
mikasenghaas: is 29500 dangerous if this is the rdvz port for torchrun or is it meant to be shared?

src/prime_rl/orchestrator/config.py
```diff
@@ -510,6 +517,19 @@ class OrchestratorConfig(BaseSettings):
 
     seed: Annotated[int | None, Field(description="Random seed for the orchestrator.")] = 42
 
+    nccl_broadcast: NCCLBroadcastConfig = NCCLBroadcastConfig()
+
+    broadcast_backend: Annotated[
```
mikasenghaas: these seem like they should be mutually exclusive but are not from the config, ie. its possible to configure nccl broadcast but have broadcast backend filesystem

src/prime_rl/trainer/rl/broadcast/nccl_broadcast.py
```diff
@@ -0,0 +1,95 @@
...
+                value = value.to(self.dtype)
+
+            if self.training_rank == 0:
+                value = value.to(self.dtype)
+                self.communicator.broadcast(value, src=0)
+
+        self.logger.info("Weights broadcasted to inference pool")
+
+
+class NCCLBroadcastInference:
```
mikasenghaas: seems weird that this is called inference. maybe sender and receiver? and then the receiver is used on inference?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -105,10 +113,16 @@ async def orchestrate(config: OrchestratorConfig):
...
-        ckpt_step = max(progress.step - config.async_level, 0)
-        await update_weights(admin_clients, get_step_path(get_weights_dir(config.output_dir), ckpt_step))
+        if config.broadcast_backend == "filesystem":
+            logger.info(f"Resuming training from checkpoint step `{config.ckpt.resume_step}`")
+            ckpt_manager.load(progress, buffer, step=config.ckpt.resume_step)
+            ckpt_step = max(progress.step - config.async_level, 0)
+            await update_weights(admin_clients, get_step_path(get_weights_dir(config.output_dir), ckpt_step))
+        elif config.broadcast_backend == "nccl":
+            logger.info("Resuming training form latest ckpt via NCCL broadcast")
+            await update_weights_nccl(admin_clients)
```
mikasenghaas: it seems like we may not be resuming correctly here? if broadcast backend is nccl, we are not calling `ckpt_manager.load`

src/prime_rl/trainer/rl/broadcast/nccl_broadcast.py
```diff
@@ -0,0 +1,95 @@
...
+            pg = StatelessProcessGroup.create(host=host, port=port, rank=rank, world_size=world_size)
+            self.communicator = PyNcclCommunicator(pg, device=device)
+
+            self.logger.info(f"NCCL broadcast initialized for rank {rank} and world size {world_size}")
+
+        self.device = device
+        self.dtype = dtype
+
+    def broadcast_state_dict(self, model: torch.nn.Module) -> None:
+        self.logger.info("Broadcasting weights to inference pool")
```
mikasenghaas: debug logs?

src/prime_rl/utils/client.py
```diff
@@ -127,3 +127,44 @@ async def _reload_weights(admin_client: AsyncClient) -> None:
...
+
+    await asyncio.gather(
+        *[
+            _init_nccl_broadcast(admin_client, host, port, client_num)
+            for client_num, admin_client in enumerate(admin_clients)
+        ]
+    )
+
+
+async def update_weights_nccl(admin_clients: list[AsyncClient]) -> None:
```
mikasenghaas: this is basically the same function as update_weights right with the difference that this one doesn't hve a weight_dir. could we remove this func and change the original one to have `weight_dir: Path | None` where `None` implies recv via NCCL communicator?

src/prime_rl/inference/vllm/worker_nccl.py
```diff
@@ -0,0 +1,56 @@
...
+                host=host,
+                port=port,
+                rank=rank + self.dp_rank,
+                world_size=(world_size - 1) * self.dp_world_size + 1,
+                device=self.device,
+                logger=logger,
+            )
+
+    def update_weights(self, weight_dir: str) -> None:
+        """Update weights from a specified path pointing to a .pt file."""
```
mikasenghaas: docstring wrong?

mikasenghaas (commented):
is it correct that we send from trainer master rank to each tp master rank on inference? im guessing this was built in mind for tp/dp parallelism types in vllm, right? do we know if this breaks with other parallelism types, like pp/ep etc.?

src/prime_rl/orchestrator/config.py
samsja: yup will change

src/prime_rl/orchestrator/config.py
samsja: wondering what is the easiest way to do tho

having some kind of broadcast_config which map to either weight or nccl, but also we might want to use nccl broadcast with weight ckpt

samsja: lets keep it like this and refactor later

src/prime_rl/inference/vllm/server.py
samsja: > style: do we need this being a dict? never gonna be more than two types of workers no? could just if else imo

defo gonna be more than two, nccl is easy to setup but not the best solution for our training, we gonna move to nixl soon

src/prime_rl/orchestrator/config.py
mikasenghaas: sure

src/prime_rl/inference/vllm/worker/nccl.py
```diff
@@ -0,0 +1,59 @@
...
+    using NCCL across multiple GPUs.
+    """
+
+    def init_broadcaster(self, host: str, port: int, server_rank: int, num_inference_server: int, timeout: int) -> None:
+        """Initialize the process group for NCCL broadcast."""
+        logger = init_logger("vllm.inference.vllm.worker_nccl")
+        self.tp_rank = get_tp_group().rank
+
+        tp_size = get_tp_group().world_size
+        tp_rank = get_tp_group().rank
```
S1ro1, line 30: nit: getting `tp_rank` twice

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -147,27 +158,40 @@ async def orchestrate(config: OrchestratorConfig):
...
-            )
 
-            # Wait for the checkpoint to be available
-            ckpt_step = progress.step - config.async_level
-            logger.info(f"Waiting for weight checkpoint {ckpt_step}")
-            wait_for_weight_ckpt_start_time = time.time()
-            await wait_for_path(get_step_path(get_weights_dir(config.output_dir), ckpt_step) / "STABLE")
-            wait_for_weight_ckpt_time = time.time() - wait_for_weight_ckpt_start_time
-            logger.debug(f"Waited {wait_for_weight_ckpt_time:.2f}s for weight checkpoint")
+        if config.broadcast_backend == "filesystem":
```
S1ro1: I'm wondering if we can abstract this a bit better. I don't like polluting the main orchestrator method with extra `if/else`. Preferably this should be something totally opaque as `xxx.update_weights` where the orchestrator doesn't have to worry about extra checks for which ckpt to load etc. This is just a food for thought though

PR #1165, opened 2025-10-27
Reward is the same as advantage

damoonsh:
As part of writing-zero bounty, I have created the [wo-brpo](https://app.primeintellect.ai/dashboard/environments/dmnsh001/w0-brpo) to run the BRPO Rl algorithm. The only issues is that current prime-rl stack runs advantage computation for all rewards. Whereas, the [BRPO algorithm](https://www.alphaxiv.org/abs/2506.00103) does not utilize any normalization just raw reward values:
Screenshot 2025-10-27 at 00 03 44

The changes are minimal where the normalization is skipped if no_norm is set to True. The code has been ran and tested within GPU instances and does not break the code.

### DIFF
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -373,7 +373,8 @@ DataBufferConfigType: TypeAlias = SimpleBufferConfig | DifficultyPoolBufferConfi
 
 
 class AdvantageConfig(BaseConfig):
-    std_norm: Literal["local", "global"] | None = None
+    global_std_norm: bool = False
+    local_std_norm: bool = False
     length_weighted_mean: bool = False
     leave_one_out: bool = False
     neg_clipped: bool = False


src/prime_rl/orchestrator/config.py
```diff
@@ -373,6 +373,7 @@ class OnlineDifficultyBufferConfig(BufferConfig):
 
 
 class AdvantageConfig(BaseConfig):
+    no_norm: bool = False
     global_std_norm: bool = False
     local_std_norm: bool = False
```
mikasenghaas: im starting to not like how we configure advantage computation. it's not immediately obvious which combinations of these boolean values are allowed, and if multiple are allowed which takes precedence.

this is alr the case for global vs. local std norm. it should prob be `std_norm: Literal["local", "global"] | None` to make clear that these are mutually exclusive. cc @faresobeid 

for the feature of this pr, i'd probably suggest the following: make the whole advantage config optional, ie. `advantage: AdvantageConfig | None` in the orchestrator config. here `None` would mean that we skip advantage estimation entirely and assign advantages = rewards.

damoonsh: I could change:
```python
# config.py
advantage: AdvantageConfig = AdvantageConfig() | None = None
```
But that would break the current logic since even if all the other values are set to False, the `base_line` (average of rewards) gets subtracted to get the advantages. 

the other change would be:
```python
# config.py
class AdvantageConfig(BaseConfig):
    std_norm: Literal["local", "global"] | None = None
    length_weighted_mean: bool = False
    leave_one_out: bool = False
    neg_clipped: bool = False

# advantage.py
.....
   if advantage_config.std_norm == 'local':
        advantages = advantages / global_std
   elif advantage_config.std_norm == 'global':
        advantages = advantages / (rewards.std() + 1e-8)
.....
```

Do let me know if I should apply these changes

mikasenghaas: well i'd simply abort the entire advantage estimation if advantage config is None, i.e. add
```python
if advantage_config is None:
     return rewards
```
that should work no?

damoonsh: pushed the change

src/prime_rl/orchestrator/config.py
```diff
@@ -402,7 +401,7 @@ class OrchestratorConfig(BaseSettings):
     buffer: Annotated[DataBufferConfigType, Field(discriminator="type")] = SimpleBufferConfig()
 
     # The advantage configuration
-    advantage: AdvantageConfig = AdvantageConfig()
+    advantage: AdvantageConfig | None = None
```
mikasenghaas: we cannot change the default configuration. default should be `AdvantageConfig()` here

src/prime_rl/orchestrator/config.py
```diff
@@ -373,8 +373,7 @@ class OnlineDifficultyBufferConfig(BufferConfig):
 
 
 class AdvantageConfig(BaseConfig):
-    global_std_norm: bool = False
-    local_std_norm: bool = False
+    std_norm: Literal["local", "global"] | None = None
```
mikasenghaas: also this may require changes to some config files, not super sure tho. you can run `uv run pytest -m "not gpu"` to see if all config files still pass

src/prime_rl/orchestrator/config.py
damoonsh: if we are to use the default config, then the original code is the way to go since advantage_config will not be None through .toml file. Am I missing something?

mikasenghaas: im not sure i undestand? the whole line should be `advantage: AdvantageConfig | None = AdvantageConfig()`. in this way, by default we compute the advantages as zero normalized. but one may configure advantage config to be None to skip any normalization

src/prime_rl/orchestrator/config.py
damoonsh: Screenshot 2025-10-27 at 14 11 23

src/prime_rl/orchestrator/config.py
damoonsh: I apologise for the confusion, I will fix and push

src/prime_rl/orchestrator/config.py
damoonsh: skipped one test so I will revert since it is somewhat out of the scope for me

mikasenghaas:
ci failure is fixed on main. should be good to go

PR #1167, opened 2025-10-27
Changes to the trainer to minimise potential mismatch with vllm

faresobeid:
Currently have matched rope implementations (seemingly vllm does rope in bf16 instead of fp32 which is weird)
Add residuals in fp32 instead of bf16
Stop using liger kernel for our RMS Norm

Planned:
Check MoE difference
Vllm fuses the residual and rms norm, not sure if we can do the equivalent
torch.topk apparently is unstable so will try to find an alternative
Test actual mismatch_kl difference and mfu difference (although mismatch_kl here might not be perfect as it can have an impact late in training)

### DIFF
--- a/src/prime_rl/trainer/models/layers/attn.py
+++ b/src/prime_rl/trainer/models/layers/attn.py
@@ -61,7 +61,7 @@ class FlashAttention(nn.Module):
         cos_sin_cache: torch.Tensor,
         cu_seqlens: torch.LongTensor | None = None,
         max_seqlen: int | None = None,
-        position_ids: Optional[torch.LongTensor] = None,
+        position_ids: torch.LongTensor | None = None,
     ) -> tuple[torch.Tensor, torch.Tensor | None]:
         input_shape = hidden_states.shape[:-1]
         hidden_shape = (*input_shape, -1, self.head_dim)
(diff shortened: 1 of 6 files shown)


mikasenghaas (commented):
gods work

src/prime_rl/trainer/models/layers/attn.py
```diff
@@ -58,9 +58,10 @@ def __init__(self, config: AttentionConfig):
     def forward(
         self,
         hidden_states: torch.Tensor,
-        position_embeddings: tuple[torch.Tensor, torch.Tensor],
+        cos_sin_cache: torch.Tensor,
         cu_seqlens: torch.LongTensor | None = None,
         max_seqlen: int | None = None,
+        position_ids: Optional[torch.LongTensor] = None,
```
samsja:
```suggestion
        position_ids: torch.LongTensor | None = None,
```

src/prime_rl/trainer/models/layers/attn.py
```diff
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-
+from typing import Optional
```
samsja: no Optional, we use x | None

src/prime_rl/trainer/models/layers/rms_norm.py
```diff
@@ -21,12 +20,19 @@ def __init__(self, config: RMSNormConfig) -> None:
         self.weight = nn.Parameter(torch.ones(config.hidden_size))
         self.variance_epsilon = config.eps
 
-    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+    def forward(self, hidden_states: torch.Tensor, residual: Optional[torch.Tensor] = None) -> torch.Tensor:
```
samsja:
```suggestion
    def forward(self, hidden_states: torch.Tensor, residual: torch.Tensor | None = None) -> torch.Tensor:
```

PR #1168, opened 2025-10-27
Use torch.compile for selective_log_softmax

faresobeid:
Much cleaner implementation of selective_log_softmax, and a bit of a higher mfu
image


mikasenghaas:
what am i looking at in the plot? rerun is from this pr? in that case it would be slightly lower mfu than yellow baseline with compile on, no?

faresobeid:
Compile false is the current way we do things and compile true is with the pr. I just reran each twice to be safe

PR #1176, opened 2025-10-27
Mask out entire sequences with dangerous tokens

faresobeid:
Have a threshold sequence_mask_ratio_low where if any token in a sequence falls under this, the entire sequence is masked.
Should handle dangerous sequences as mentioned in https://www.arxiv.org/abs/2510.02387
Screenshot 2025-10-27 at 17 59 23

Borrowed from the slime repo https://github.com/THUDM/slime/blob/83a3c964d68eb7f5cc8b2d05ef165d4ede4df813/examples/train_infer_mismatch_helper/mis.yaml#L24

### DIFF
--- a/src/prime_rl/trainer/rl/config.py
+++ b/src/prime_rl/trainer/rl/config.py
@@ -31,7 +31,7 @@ class LossConfig(BaseModel):
         Field(
             ge=0,
             description=(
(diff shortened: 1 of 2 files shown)


samsja (approved):
lgtm

PR #1189, opened 2025-10-29
Fix Pydantic serialization overhead

mikasenghaas:
At large batch sizes and context, constructing OAI's Pydantic model `ChatCompletion` with logprobs is causing heavy CPU overhead (~200ms per object at 32K context, which translates to >10min overhead at 4K batch size). This PR monkeypatches the OAI type and the post-processing utils in `verifiers` to make this fast.

---

**GitHub Issue**: #1191 
**Linear Issue**: Resolves PRIMERL-189

### DIFF
--- a/src/prime_rl/orchestrator/utils.py
+++ b/src/prime_rl/orchestrator/utils.py
@@ -23,7 +23,7 @@ def monkey_patch_chat_completion_logprobs():
 
     # Patch the logprob field which is causing CPU overhead
     openai.types.chat.chat_completion.Choice = ChoiceAny
-    openai.types.chat.ChatCompletion = ChatCompletionAny
+    openai.types.chat.chat_completion.ChatCompletion = ChatCompletionAny
 
     # Patch verifiers parse_chat_completion_logprobs
     def patched_parse_chat_completion_logprobs(chat_completion: ChatCompletionAny) -> list[float]:
--- a/src/prime_rl/orchestrator/utils.py
+++ b/src/prime_rl/orchestrator/utils.py
@@ -1,8 +1,9 @@
-from typing import Any
+from typing import Any, List, Optional
 
+import openai.types.chat
 import pandas as pd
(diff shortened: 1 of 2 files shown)


src/prime_rl/orchestrator/utils.py
```diff
@@ -13,6 +14,51 @@
 )
 
 
+def monkey_patch_chat_completion_logprobs():
+    class ChoiceAny(Choice):
+        logprobs: Optional[Any] = None
+
+    class ChatCompletionAny(ChatCompletion):
+        choices: List[ChoiceAny]  # type: ignore
```
samsja, line 34: can we add comment here why we do this, also please link to the og open client ChatCompletion base model so that people can understand what it does

mikasenghaas: added more docs, i think the link is clear enough with the monkey-patch lines and `go-to` IDE functionality

```py
openai.types.chat.chat_completion.Choice = ChoiceAny
openai.types.chat.chat_completion.ChatCompletion = ChatCompletionAny
```

mikasenghaas:
Screenshot 2025-10-29 at 11 00 57 PM
Screenshot 2025-10-29 at 11 01 13 PM

samsja (approved):
LFGTM

samuelcolvin:
Please use `model_validate_json` - see my reply to your twitter, also see https://gist.github.com/samuelcolvin/b432a00393849cadddde440697c23c73

samsja:
> Please use `model_validate_json` - see my reply to your twitter, also see https://gist.github.com/samuelcolvin/b432a00393849cadddde440697c23c73

thanks @samuelcolvin for the hint, we are going to implement it using `model_validate_json` as it defo seems like the best option

samuelcolvin:
LMK if you need more help, I can probably skip validation of logprops if you don't need them and need a bigger boost.

PR #1198, opened 2025-10-31
auto-infer inference dp if not set

willccbb:
Currently, users need to set both inference_gpu_ids and dp/tp values in their configs. If inference_gpu_ids is provided and dp is not (in RLConfig), this PR infers dp = len(inference_gpu_ids) / tp. 

Sample rl.toml config provided, along with test case tweak to allow rl.toml files.


willccbb:
@cursoragent fix pls

mikasenghaas (commented):
can we rebase this on current main? a little hard to review this way

willccbb:
yes will do before un-drafting

samsja (approved):
lgtm !

PR #1202, opened 2025-11-01
all-in-one uv sync for fa2

willccbb:
Uses tool.uv.extra_build_dependencies to match runtime for torch + flash-attn, enables single-command `uv sync` without `--all-extras`


samsja (commented):
this is great !

PR #1203, opened 2025-11-01
Allow simpler pattern-matching for lora module names

willccbb:
Adds "PEFT-style" name matching so that configs can set e.g. "q_proj" without needing regex.

Also, sets default modules_to_save = [] -- most users won't expect LoRA to keep embeddings + norms trainable by default; we should surface this in the example configs.


samsja (approved):
ah yeah this is way better !

mikasenghaas (approved):
yep i like it

PR #1205, opened 2025-11-02
log unfiltered_reward if difficulty filtering enabled

willccbb:
If online difficulty filtering is used, we currently log the reward only for the train batch, which can make reward curves harder to interpret. This PR adds additional logging for the reward before overly easy/hard samples are dropped, which gives a (nearly*) unbiased estimator of the current distribution-level reward.

(*slightly biased estimator when multiple rounds of sampling are needed to fill the batch if oversampling level is insufficient for single pass)

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -29,7 +29,7 @@ from prime_rl.utils.client import (
     update_weights,
 )
 from prime_rl.orchestrator.config import OrchestratorConfig
-from prime_rl.orchestrator.buffer import setup_buffer, make_rollouts, Rollout
+from prime_rl.orchestrator.buffer import setup_buffer, make_rollouts, Rollout, OnlineDifficultyBuffer
 from prime_rl.orchestrator.batch import prepare_batch
 from prime_rl.utils.logger import setup_logger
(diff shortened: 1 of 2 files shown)


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -204,6 +204,11 @@ async def orchestrate(config: OrchestratorConfig):
         problem_requests, completion_requests, calls_to_generate = 0, 0, 0
         problems_per_batch = config.batch_size // config.rollouts_per_example
         problems_to_sample = problems_per_batch
+
+        # Reset unfiltered reward tracking if using online difficulty filtering
+        if isinstance(buffer, OnlineDifficultyBuffer):
```
mikasenghaas, line 209: i dont think this should be a pure online difficulty buffer field, but a general buffer feature

mikasenghaas (commented):
this logic doesn't work for the pool buffer. here the reward trends down because we don't sample easy _problems_ multiple times so over time we see harder and harder problems. this would not be captured by the current impl which only filters on calls to `sample_rollouts`

i think there's no way around making a proper validation split that we run through the simple buffer to get a true estimates of how the reward trends without any fancy data filtering

mikasenghaas:
Closing in favor of #1211

PR #1208, opened 2025-11-02
Make NCCL beautiful

mikasenghaas:
A couple of improvements to bring NCCL into a more mergable state:
- Merge the redundant `broadcast_backend` and `nccl_broadcast` config into a single `weight_broadcast` config with a discriminated union, matching our general config system
- Simplify the checkpoint/ async resume logic for `filesystem` and `nccl` broadcast on the orchestrator. We can basically reuse the same logic, just have to enforce `async_level=1` for NCCL with pydantic
- Only set `--weight.interval 1` when filesystem is used on RL trainer, else use the default (None) which only saves at the checkpoint interval and at the end of training
- Validate from `rl.py` that the weight broadcast config aligns across all three components
- Add a guard that NCCL broadcast only works on at least 2 GPU (trainer and inference have to be on distinct devices)
- Minor file restructures
- Simplify checkpoint resume logic (do not save last async_level steps with either backend to simplify things) - needs some changes on trainer weight checkpointing and orchestrator resume

### DIFF
--- a/src/prime_rl/inference/config.py
+++ b/src/prime_rl/inference/config.py
@@ -118,11 +118,11 @@ class InferenceConfig(BaseSettings):
     ] = None
 
     broadcast_backend: Annotated[
-        Literal["nccl", "filesystem"], Field(description="The backend to use for broadcast.")
+        Literal["nccl", "filesystem"], Field(description="The backend to use for updating the weights.")
     ] = "filesystem"
 
     @model_validator(mode="after")
-    def nccl_and_dp(self) -> bool:
+    def nccl_and_dp(self):
         if self.broadcast_backend == "nccl" and self.parallel.dp != 1:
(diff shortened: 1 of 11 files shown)


mikasenghaas:
Tested working with `reverse-text` in default setting (using filesystem)

```bash
uv run rl \
  --inference @ examples/reverse_text/rl/infer.toml
```

[W&B](https://wandb.ai/primeintellect/mika?nw=jn9qpehzk1l)

Screenshot 2025-11-02 at 5 33 56 PM

```bash
uv run rl \
  --inference @ examples/reverse_text/rl/infer.toml \
  --weight-broadcast.type nccl
```

NCCL didn't work on RTX 6000 Blackwell, going to retry with another dev machine too

samsja (approved):
beautiful

PR #1209, opened 2025-11-02
w0 example

damoonsh:
This is the example workflow for writin-zero bounty that is completed!
This structure would work for any other GenRM model as well.


damoonsh:
@mikasenghaas please advise on the errors

damoonsh:
@mikasenghaas rebased, fixed the errors!

PR #1211, opened 2025-11-03
Implement validation split

mikasenghaas:
This is the easiest way I can think of supporting reporting validation results, e.g. reward scores independent of the buffer strategy used. It adds the `ValConfig` which specifies `--val.num_examples` and `--val.rollouts_per_example` to run as a validation batch at a specified `--val.interval`. If the env supports a train/eval split, it will use the eval samples for validation, i.e. samples that are not used for training. If not eval split is defined in the env, it will fallback to training samples (but still independent of the buffer used).

## Example

Here, we train `hendrycks-math` with the difficulty pool buffer (does not resample easy problems) and we can see that the train reward stays ~constant while the validation reward goes up.

Screenshot 2025-11-03 at 7 39 19 PM

## Multi-Env Train/Val

```bash
uv run rl --trainer @ configs/multi-reverse-text/rl/train.toml --orchestrator @ configs/multi-reverse-text/rl/orch.toml 
```

Screenshot 2025-11-03 at 10 04 05 PM

---

**GitHub Issue**: #1206 
**Linear Issue**: Resolves PRIMERL-192


configs/multi-env/hendrycks_math/rl/infer.toml
samsja, line 1: is this multi-env/hendrcysk catch by ci config ?

src/prime_rl/orchestrator/config.py
```diff
@@ -195,6 +195,24 @@ class EvalEnvConfig(EnvConfig):
     ] = None
 
 
+class ValidationConfig(BaseConfig):
```
```suggestion
class ValidationSplitConfig(BaseConfig):
```

Validation remind me more of pydantic validation than val set

src/prime_rl/orchestrator/config.py
```diff
@@ -219,7 +237,7 @@ class OnlineEvalConfig(EvalConfig):
     interval: Annotated[
         int,
         Field(
-            ge=0,
+            ge=1,
```
samsja, line 234: why 1 here ?

configs/multi-env/hendrycks_math/rl/infer.toml
mikasenghaas: should be, yea? but im gonna remove them before merging, just for my testing. or maybe i will propose changing the main hendrycsk config

src/prime_rl/orchestrator/config.py
mikasenghaas: ah ok, what abt `ValConfig` for brevity👀

src/prime_rl/orchestrator/config.py
mikasenghaas: can you cant do smth every 0th interval

samsja (approved):
lgtm

PR #1215, opened 2025-11-03
Do not log extras by default

mikasenghaas:
Logging samples and distributions, by definition, is an extra and should not be enabled by default. We have seen that, especially on the trainer, these args can create a lot of overhead, so this I think opt-in instead of of opt-out is better here.


samsja (approved):
lgtm

PR #1222, opened 2025-11-04
bump verifier

samsja:
https://github.com/PrimeIntellect-ai/verifiers/pull/526


mikasenghaas:
we alr have this commit in. we are always pinning to the latest commit on vf [prime-rl](https://github.com/PrimeIntellect-ai/verifiers/pull/483) branch which ideally includes everything on vf main and additional features prime-rl needs

PR #1225, opened 2025-11-04
Sami/fix nccl moe

samsja:
image

this pr fix the mismatch by respecting dtype more

### DIFF
--- a/src/prime_rl/trainer/rl/broadcast/nccl_broadcast.py
+++ b/src/prime_rl/trainer/rl/broadcast/nccl_broadcast.py
@@ -43,7 +43,7 @@ def send_state_dict(state_dict: dict[str, torch.Tensor], communicator: PyNcclCom
 
 
 def receive_state_dict(
-    communicator: PyNcclCommunicator | dist.ProcessGroup, dtype: torch.dtype
+    communicator: PyNcclCommunicator | dist.ProcessGroup,
 ) -> Generator[tuple[str, torch.Tensor], None, None]:
     """Stream tensors in a state dict broadcasted over NCCL."""
     size_tensor = torch.tensor([10], dtype=torch.long).to(communicator.device)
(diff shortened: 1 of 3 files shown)


mikasenghaas (approved):
nice, lgtm!

src/prime_rl/trainer/rl/broadcast/nccl_broadcast.py
```diff
@@ -146,8 +146,7 @@ def broadcast_state_dict(self, model: torch.nn.Module) -> None:
             for key, value in list(state_dict.items()):
                 if isinstance(value, DTensor):
                     value = value.to(self.dtype).full_tensor()
-                else:
-                    value = value.to(self.dtype)
```
S1ro1, line 150: Is this legit? I.e. if tensor is a DTensor, we cast to self.dtype, if tensor is vanilla tensor we don't?

S1ro1 (approved):
LGTM, left one comment which seems like a tiny mistake, so just address that and ready

src/prime_rl/trainer/rl/broadcast/nccl_broadcast.py
samsja: yes its not a typo, its actually the whole point of the PR otherwise we have a high mismatch. We do the same when savin the file to disk btw

S1ro1: but why would the behaviour need to be different per tensor/dtensor, that's what I'm confused about.

samsja: don't ask ! 

its not clear to me neither, we basically use the .to(bf16) as a way to reduce memory pressure for the all gather. The take was that all tensor are saved in hf format in b16 anyway but for some reason it breaks for moe. So there must be some layer sensitive to it, I don't know which

S1ro1: lol, sg (it doesn't)

PR #1227, opened 2025-11-04
more performant nccl broadcast

samsja:
make nccl broadcast more performant by bucketing tensor together
image


S1ro1 (approved):
scared as hell of the dtype shenanigans, but lgtm

Issue #980, opened 2025-09-22
PoC of areal continuous batching


samsja:
[https://github.com/ServiceNow/PipelineRL/blob/main/pipelinerl/vllm1.py#L97](https://github.com/ServiceNow/PipelineRL/blob/main/pipelinerl/vllm1.py#L97)

PR #1229, opened 2025-11-04
Continuous batching

mikasenghaas:
This PR adds AReal-type continuous batching to PRIME-RL. We define the `Scheduler` abstraction which is responsible for scheduling rollout and weight update requests. 
- Asynchronously schedules group rollout requests and re-schedules them as they complete (continuous batching)
-  Continuously polls the weight directory to check if a new policy is available, and updates if so. Halts, if inference is more than `max_async_level` steps away. With `strict_async_level`, we enforce that rollouts from a policy _exactly_ async level steps old. Also, we add a new parameter `max_off_policy_steps` which denotes the maximum number of policies that are allowed to generate a rollout before it is discarded. We currently set this value to `8` by default.

We ensure that a constant (high) number of group rollout request is in-flight to increase inference throughput. This is especially advantageous if the sequence length distribution of the RL rollouts has heavy tails and/or online rollout filtering techniques (e.g. online difficulty filtering) are used.

Known issues:
- There is some weird random fluctuation in the update weight and step time when using the ARealScheduler. However, since we are completely async this is probably fine for as long as we don't timeout completely.

### Reverse Text

First, we check the `reverse-text` debuggingg example to see that our default configuration is running and working as before. [W&B](https://wandb.ai/primeintellect/mika?nw=3r81itvl4j2).

```bash
uv run rl \
--inference @ examples/reverse_text/rl/infer.toml \
--wandb.project mika \
--wandb.name pr-1229-reverse-text
```

Screenshot 2025-11-04 at 12 49 32 PM

### Hendrycks Math

```bash
prime env install primeintellect/hendrycks-math
prime env install primeintellect/math500
prime env install primeintellect/aime2024
prime env install primeintellect/aime2025
```

```bash
uv run rl \
--trainer @ configs/hendrycks-math/rl/train.toml \
--orchestrator @ configs/hendrycks-math/rl/orch.toml \
--inference @ configs/hendrycks-math/rl/infer.toml \
--trainer-gpu-ids 0,1 \
--inference-gpu-ids 2,3,4,5,6,7 \
--wandb.project hendrycks-math \
--wandb.name pr-1229
```

Screenshot 2025-11-06 at 10 41 20 AM

## Acereason Math

Finally, we train on an adapted version of `acereason-math` with online difficulty filtering enabled. 

```bash
prime env install primeintellect/acereason-math
```

```bash
uv run rl \
--trainer @ configs/acereason_math/stage1/rl/train.toml \
--orchestrator @ configs/acereason_math/stage1/rl/orch.toml \
--inference @ configs/acereason_math/stage1/rl/infer.toml \
--trainer-gpu-ids 0,1 --inference-gpu-ids 2,3,4,5,6,7
--orchestrator.buffer.type online-difficulty \
--orchestrator.buffer.min-reward 0.01 \
--orchestrator.buffer.max-reward 0.99 \
--wandb.project mika \
--wandb.name pr-1229-acereason-math
```

---

**GitHub Issue**: #980 
**Linear Issue**: Resolves PRIMERL-127

### DIFF
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -265,6 +265,20 @@ class CheckpointConfig(BaseConfig):
         ),
     ] = None
 
+    skip_progress: Annotated[
+        bool,
+        Field(
+            description="Whether to skip loading the progress from checkpoint.",
+        ),
+    ] = False
+
+    skip_buffer: Annotated[
+        bool,
+        Field(
+            description="Whether to skip loading the buffer from checkpoint.",
+        ),
+    ] = False
+
 
 class BufferConfig(BaseModel):
     """Base config for all buffer types."""
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -444,6 +444,14 @@ class OrchestratorConfig(BaseSettings):
 
     batch_size: Annotated[int, Field(ge=1, description="Number of samples to train on per step.")] = 128
 
+    oversampling_factor: Annotated[
+        float,
+        Field(
+            ge=1,
+            description="Factor by which to oversample the batch. Will lead to more in-flight group rollout requests at the same time.",
+        ),
+    ] = 1.0
+
     rollouts_per_example: Annotated[
         int,
         Field(
@@ -493,18 +501,40 @@ class OrchestratorConfig(BaseSettings):
         ),
     ] = None
 
-    async_level: Annotated[
+    max_off_policy_steps: Annotated[
         int,
         Field(
             ge=0,
-            description="Maximum number of async levels to use. If 0, will do synchronous RL. Else, it will allow to go `async_level` steps ahead of training.",
+            description="Maximum number of policies that are allowed to generate a single rollout. Rollouts that are generated from more than `max_off_policy_steps` steps ahead of training will be discarded. Higher values yield better throughput, but lead to more off-policyness in training.",
+        ),
+    ] = 8
+
+    max_async_level: Annotated[
+        int,
+        Field(
+            ge=0,
+            description="Maximum number of steps the inference can be ahead of training. If 0, will degenerate to synchronous on-policy RL. If >=1, training and inference will be overlapped.",
         ),
     ] = 1
+
+    strict_async_level: Annotated[
+        bool,
+        Field(
+            description="Whether to strictly enforce the max async level. If True, will always ensure that the policy used for generating rollouts is exactly `max_async_level` steps ahead of training. If False, any policy that is at most `max_async_level` steps ahead of training is allowed, i.e. we always use the latest available policy.",
+        ),
+    ] = True
 
     bench: Annotated[
         bool,
         Field(
             description="Whether to run in benchmark mode. It will automatically set the maximum number of steps to run to 5, max async level to ~infinity and disable W&B.",
+        ),
+    ] = False
+
+    continuous_batching: Annotated[
+        bool,
+        Field(
+            description="Whether to use continuous batching. If True, will use continuous batching to generate rollouts.",
         ),
     ] = False
 
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -394,6 +394,18 @@ class NCCLWeightBroadcastConfig(BaseModel):
 WeightBroadcastConfigType: TypeAlias = FileSystemWeightBroadcastConfig | NCCLWeightBroadcastConfig
 
 
+class EnvMixConfig(BaseModel):
+    """Configures the mixing of environments."""
+
+    strategy: Literal["interleave", "concatenate"] = "interleave"
+    probabilities: Annotated[list[float] | None, Field(description="Probabilities to use for each environment.")] = None
+    stopping_strategy: Annotated[
+        Literal["first_exhausted", "all_exhausted"],
+        Field(description="Stopping strategy to use for interleaving environment datasets."),
+    ] = "all_exhausted"
+    seed: Annotated[int | None, Field(description="Random seed to use for the environment mixing.")] = None
+
+
 class OrchestratorConfig(BaseSettings):
     """Configures the orchestrator for RL training."""
 
@@ -405,6 +417,9 @@ class OrchestratorConfig(BaseSettings):
 
     # The sampling configuration
     sampling: SamplingConfig = SamplingConfig()
+
+    # The environment mixing configuration
+    env_mix: EnvMixConfig = EnvMixConfig()
 
     # The environment configuration
     env: list[EnvConfig] = [EnvConfig()]
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -552,13 +552,6 @@ class OrchestratorConfig(BaseSettings):
         ),
     ] = False
 
-    continuous_batching: Annotated[
-        bool,
-        Field(
-            description="Whether to use continuous batching. If True, will use continuous batching to generate rollouts.",
-        ),
-    ] = False
-
     seed: Annotated[int | None, Field(description="Random seed for the orchestrator.")] = 42
 
     @model_validator(mode="after")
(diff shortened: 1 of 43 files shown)


src/prime_rl/orchestrator/config.py
```diff
@@ -493,28 +493,50 @@ class OrchestratorConfig(BaseSettings):
...
             description="Whether to run in benchmark mode. It will automatically set the maximum number of steps to run to 5, max async level to ~infinity and disable W&B.",
         ),
     ] = False
 
+    continuous_batching: Annotated[
+        bool,
+        Field(
+            description="Whether to use continuous batching. If True, will use continuous batching to generate rollouts.",
+        ),
+    ] = False
```
samsja: would link to relevant paper and also say that it make inference faster so that ppl get what we are talking about

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -0,0 +1,243 @@
...
+        self.tokenizer = tokenizer
+        self.config = config
+        self.batch_size = config.batch_size
+        self.rollouts_per_example = config.rollouts_per_example
+        self.seq_len = config.seq_len
+        self.problems_per_batch = int(oversampling_factor * self.batch_size // self.rollouts_per_example)
+        self.max_async_level = max_async_level
+        self.max_off_policy_steps = max_off_policy_steps
+        self.strict_async_level = strict_async_level
+        self.inflight_group_rollouts: dict[asyncio.Task, tuple[int, AsyncOpenAI]] = {}
```
samsja:
```suggestion
        self.inflight_group_rollouts: dict[asyncio.Task, InflightRolloutInfo] = {}
```
```python
class InflightRolloutInfo(NamedTuple):
    """Metadata for an in-flight group rollout request."""
    off_policy_steps: int
    client: AsyncOpenAI
```


would be cleaner imo, I am not a fane of list or dict of tuple in lage codebase

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -0,0 +1,243 @@
...
+from prime_rl.orchestrator.config import OrchestratorConfig
+from prime_rl.orchestrator.utils import get_sampling_args, parse_is_truncated_completions
+from prime_rl.utils.client import update_weights
+from prime_rl.utils.logger import get_logger
+from prime_rl.utils.utils import get_latest_ckpt_step, get_step_path, get_weights_dir, sync_wait_for_path
+from prime_rl.utils.vf import generate_group, make_rollouts
+
+
+class Scheduler:
+    """Asynchronously schedules group rollout requests and re-schedules them as they complete (continuous batching). Updates policy in between group rollout requests."""
```
samsja: I am suggesting adding more detail docstring about the scheduler. I used ai to unroll the flow and edge case but it help me fully get what each function do and how to think about the Scheduler class in general

```suggestion


class Scheduler:
    """
    Asynchronously schedules group rollout requests and re-schedules them as they complete (continuous batching).
    Updates policy in between group rollout requests.

    ASYNC OFF-POLICY TRAINING FLOW:
    ================================

    The orchestrator runs two concurrent async tasks:
    1. Main training loop calls generate_batch(step=N) at each step
    2. update_policy_loop() runs continuously in background (started in orchestrator.py:163)

    Key Concepts:
    - max_async_level: Max steps inference can be ahead (async_level = step - ckpt_step)
    - max_off_policy_steps: Max policy updates a rollout survives before cancellation
    - inflight_group_rollouts: Tracks in-flight rollouts as {Task: (off_policy_steps, client)}

    Example Timeline (max_async_level=2, max_off_policy_steps=2, training≈inference time):

    Step 0: generate_batch(step=0) called
      - Schedules 25 rollouts (A-Z), all with off_policy_steps=0
      - Enters continuous batching loop waiting for completions
      - Takes ~10s to generate batch

    Step 0.5: update_policy_loop() runs (background, every ~1s)
      - update_policy() checks for checkpoint 0
      - Trainer still processing, no checkpoint yet

    Step 1: generate_batch(step=0) completes
      - Batch generated in ~10s
      - Main loop calls generate_batch(step=1)
      - Trainer still processing step 0 (takes ~10s)
      - Some slow rollouts still in-flight from step 0

    Step 1.5: update_policy_loop() runs - CHECKPOINT 0 AVAILABLE
      - Trainer finishes step 0, checkpoint 0 becomes available
      - update_policy() detects checkpoint 0
      - Updates inference weights via update_weights()
      - Increments all in-flight rollouts: off_policy_steps += 1
      - Sets ckpt_step = 0, async_level = 1 - 0 = 1
      - Edge Case: Rollout Y0 started at step 0, very slow (still generating)
      - Y0: off_policy_steps = 1 (survived 1 update)

    Step 2: generate_batch(step=1) completes
      - Batch generated, main loop calls generate_batch(step=2)
      - Trainer still processing step 1
      - Mix of rollouts from steps 0, 1, 2 in-flight

    Step 2.5: update_policy_loop() runs - CHECKPOINT 1 AVAILABLE + CANCELLATION
      - Trainer finishes step 1, checkpoint 1 becomes available
      - Updates to ckpt_step = 1
      - Increments rollouts: A0: 0→1→2, B0: 0→1→2, Y0: 1→2
      - Y0: off_policy_steps=2 (survived 2 updates, still valid)
      - async_level = 2 - 1 = 1 (below max)

    Step 3: generate_batch(step=2) completes
      - Main loop calls generate_batch(step=3)
      - Trainer still processing step 2
      - async_level = 3 - 1 = 2 (at max_async_level!)

    Step 3.5: update_policy_loop() runs - CHECKPOINT 2 AVAILABLE + CANCELLATION
      - Trainer finishes step 2, checkpoint 2 becomes available
      - Updates to ckpt_step = 2
      - Increments rollouts: Y0 now at off_policy_steps = 3
      - Y0: off_policy_steps=3 > max_off_policy_steps=2 → CANCEL
      - task.cancel() stops Y0 mid-generation
      - schedule_group_rollout() immediately re-schedules Y1 with fresh policy (off_policy_steps=0)
      - async_level = 3 - 2 = 1 (back below max)

    Edge Case: Async Barrier with strict_async_level=True
      - At step 4, async_level = 4 - 2 = 2 (at max)
      - update_policy() computes async_away_ckpt_step = max(4-2, 0) = 2
      - If checkpoint 2 already loaded, waits for checkpoint 3 before updating
      - Prevents inference from getting more than max_async_level steps ahead

    Key Behaviors:
    - With training≈inference time, checkpoints arrive regularly, async_level stays stable
    - Most rollouts complete before exceeding max_off_policy_steps
    - Cancellations rare, mainly for unusually slow rollouts
    - update_policy_loop() maintains policy freshness while generate_batch() keeps pool full

    The "x.5" timing represents async execution: update_policy_loop() runs independently every ~1s,
    not tied to training steps. Policy updates happen concurrently with batch generation.
    """

```

src/prime_rl/orchestrator/config.py
RPasquale: https://arxiv.org/pdf/2509.19128 ?

src/prime_rl/rl.py
```diff
@@ -337,18 +337,18 @@ def auto_setup_max_steps(self):
     @model_validator(mode="after")
     def auto_setup_async_level(self):
         # If specified, use the same async level for trainer and orchestrator
-        if self.async_level:
-            self.trainer.async_level = self.async_level
-            self.orchestrator.async_level = self.async_level
+        if self.max_async_level is not None:
+            self.trainer.max_async_level = self.max_async_level
+            self.orchestrator.max_async_level = self.max_async_level
```
samsja, line 342: wouldn't this break with nccl broadcast ? We have a check there that async level should be 1. Imo worth removing the check on nccl config side

mikasenghaas: yea if you specify >1 `max_async_level` on `rl.py` I think it should still trigger the model validator when we build the trainer config no? let me double check this tho

samsja: yes it will, my question is should we allow max_async_level > 1 with nccl

samsja (approved):
lgtm

PR #1236, opened 2025-11-05
Online difficulty pool buffer

faresobeid:
Prob a bad name but basically combines the online difficulty filtering and difficulty pools.
So we sample problems from the normal, easy, and hard pools
Generate rollouts and update a problems pool
Only release rollouts where the average reward is between certain thresholds

### DIFF
--- a/src/prime_rl/orchestrator/buffer.py
+++ b/src/prime_rl/orchestrator/buffer.py
@@ -412,29 +412,85 @@ class OnlineDifficultyBuffer(Buffer):
 
 class OnlineDifficultyPoolBuffer(Buffer):
     """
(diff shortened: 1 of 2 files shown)


mikasenghaas:
i have a pretty clear idea on how i wanna refactor this so that this becomes trivial/ for free: basically, pull apart the abstraction into a `ExampleBuffer` and `RolloutBuffer` which can be mixed and matched as desired. so your abstraction would then be an examle of `DifficultyPoolExampleBuffer` and `OnlineDifficultyRolloutBuffer` used together. 

not prio to impl this now tho so happy to merge this if we decide to run with it

PR #1237, opened 2025-11-05
remove stable file nccl

samsja:
remove stable file writing when using nccl as we don't use it anymore


mikasenghaas (commented):
i love this, but im not sure what change made us not need it anymore? will just test it out:)

mikasenghaas (approved):
ah yea this works, nice! less code smell

PR #1238, opened 2025-11-05
Activations offloading

S1ro1:
Adds activation offloading to cpu from torchtune. Enables us to scale context length even further for almost no cost. @samsja

### DIFF
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -27,13 +27,7 @@ class ActivationCheckpointConfig(BaseModel):
 class ActivationOffloadingConfig(BaseModel):
     """Configures the activation offloading."""
 
-    enabled: Annotated[bool, Field(description="Whether to enable activation offloading.")] = False
-
     pin_memory: Annotated[bool, Field(description="Whether to pin the offloaded activations to CPU memory.")] = True
-
-    use_cuda_streams: Annotated[
-        bool, Field(description="Whether to use CUDA streams for communication/computation overlap.")
-    ] = True
 
     max_inflight_activations: Annotated[
         int,
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -24,6 +24,26 @@ class ActivationCheckpointConfig(BaseModel):
     ] = 1
 
 
+class ActivationOffloadingConfig(BaseModel):
+    """Configures the activation offloading."""
+
+    enabled: Annotated[bool, Field(description="Whether to enable activation offloading.")] = False
+
+    pin_memory: Annotated[bool, Field(description="Whether to pin the offloaded activations to CPU memory.")] = True
+
+    use_cuda_streams: Annotated[
+        bool, Field(description="Whether to use CUDA streams for communication/computation overlap.")
+    ] = True
+
+    max_inflight_activations: Annotated[
+        int,
+        Field(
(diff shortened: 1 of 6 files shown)


src/prime_rl/trainer/rl/config.py
```diff
@@ -88,6 +89,9 @@ class RLTrainerConfig(BaseSettings):
     # The learning rate scheduler configuration
     scheduler: Annotated[SchedulerConfigType, Field(discriminator="type")] = ConstantSchedulerConfig()
 
+    # The activation offloading configuration
+    activation_offloading: ActivationOffloadingConfig = ActivationOffloadingConfig()
```
samsja: can we move this to the AcitvationCheckpointingConfig

src/prime_rl/trainer/config.py
```diff
@@ -24,6 +24,26 @@ class ActivationCheckpointConfig(BaseModel):
     ] = 1
 
 
+class ActivationOffloadingConfig(BaseModel):
+    """Configures the activation offloading."""
```
samsja, line 28: would add this to the other ac config

src/prime_rl/trainer/config.py
```diff
@@ -24,6 +24,26 @@ class ActivationCheckpointConfig(BaseModel):
...
+
+    enabled: Annotated[bool, Field(description="Whether to enable activation offloading.")] = False
```
samsja: lets jsut the | None pattern instead of enable

src/prime_rl/trainer/config.py
```diff
@@ -24,6 +24,26 @@ class ActivationCheckpointConfig(BaseModel):
...
+
+    pin_memory: Annotated[bool, Field(description="Whether to pin the offloaded activations to CPU memory.")] = True
+
+    use_cuda_streams: Annotated[
+        bool, Field(description="Whether to use CUDA streams for communication/computation overlap.")
+    ] = True
```
samsja: is there a reason we would disable both of them ? if not I would not even add them to the config and just put them to true

src/prime_rl/trainer/rl/train.py
```diff
@@ -232,8 +233,9 @@ def train(config: RLTrainerConfig):
             temperature = micro_batch["temperature"]
 
             # Forward pass
-            with maybe_record_function("forward"):
+            with maybe_record_function("forward"), maybe_activation_offloading(config.activation_offloading):
```
samsja: I am happy to finally enter the maybe_xxx meta

src/prime_rl/trainer/config.py
S1ro1: There is a possible one for pin memory (lower resource system) agree for CUDA streams.

src/prime_rl/trainer/sft/config.py
```diff
@@ -109,6 +110,9 @@ class SFTTrainerConfig(BaseSettings):
     # The learning rate scheduler configuration
     scheduler: Annotated[SchedulerConfigType, Field(discriminator="type")] = ConstantSchedulerConfig()
 
+    # The activation offloading configuration
+    activation_offloading: ActivationOffloadingConfig | None = None
```
samsja:
```suggestion
    ac_offloading: ActivationOffloadingConfig | None = None
```

samsja (approved):
lfgtm !

PR #1244, opened 2025-11-06
fix ckpt start

samsja:
fix nccl ckpt resume on areal.

restarting from step 230 would look at ckpt 229 on disk


mikasenghaas (commented):
should pr into areal branch right? but yea good catch

PR #1255, opened 2025-11-07
auto-install envs in user/env-id format in rl.py

willccbb:
Convenience feature: if train/eval envs are given in config as `user/env-id`, install via `prime env install ...` subprocess inside of `rl.py`

Minimizes commands needed to launch a run, and allows users to be more explicit about which environment source is intended to be used with a config.

(Also re-adding LoRA for wiki-search config for 2-GPU training without OOMing)

### DIFF
--- a/src/prime_rl/rl.py
+++ b/src/prime_rl/rl.py
@@ -458,7 +458,7 @@ def rl(config: RLConfig):
     # Install each environment
     for env_id in env_ids_to_install:
         logger.info(f"Installing environment: {env_id}")
-        install_cmd = ["uv", "run", "prime", "env", "install", env_id]
+        install_cmd = ["uv", "run", "--no-sync", "prime", "env", "install", env_id]
         result = subprocess.run(install_cmd, capture_output=True, text=True)
         if result.returncode != 0:
             logger.error(f"Failed to install environment {env_id}: {result.stderr}")
--- a/src/prime_rl/rl.py
+++ b/src/prime_rl/rl.py
@@ -458,7 +458,7 @@ def rl(config: RLConfig):
     # Install each environment
     for env_id in env_ids_to_install:
         logger.info(f"Installing environment: {env_id}")
-        install_cmd = ["prime", "env", "install", env_id]
+        install_cmd = ["uv", "run", "prime", "env", "install", env_id]
         result = subprocess.run(install_cmd, capture_output=True, text=True)
         if result.returncode != 0:
             logger.error(f"Failed to install environment {env_id}: {result.stderr}")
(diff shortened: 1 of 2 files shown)


src/prime_rl/rl.py
```diff
@@ -441,6 +441,30 @@ def rl(config: RLConfig):
...
+    # Install each environment
+    for env_id in env_ids_to_install:
+        logger.info(f"Installing environment: {env_id}")
+        install_cmd = ["prime", "env", "install", env_id]
+        result = subprocess.run(install_cmd, capture_output=True, text=True)
+        if result.returncode != 0:
+            logger.error(f"Failed to install environment {env_id}: {result.stderr}")
+            raise RuntimeError(f"Failed to install environment {env_id}")
+        logger.info(f"Successfully installed environment: {env_id}")
+
```
samsja, line 467: not sure if it should belong here or in the orchestrator. I can see argument for both

src/prime_rl/rl.py
```diff
@@ -441,6 +441,30 @@ def rl(config: RLConfig):
...
+    # Collect evaluation environment IDs
+    if config.orchestrator.eval:
+        for eval_env_config in config.orchestrator.eval.env:
+            if "/" in eval_env_config.id:
+                env_ids_to_install.add(eval_env_config.id)
+
+    # Install each environment
+    for env_id in env_ids_to_install:
+        logger.info(f"Installing environment: {env_id}")
+        install_cmd = ["prime", "env", "install", env_id]
```
samsja: hmm not sure how it works with different venv and path

willccbb: guarded with `uv run`

willccbb: rl.py is a bit less intrusive for now; if we decide we want the pattern to be more load-bearing, i think orchestrator def could make sense

src/prime_rl/rl.py
```diff
@@ -458,7 +458,7 @@ def rl(config: RLConfig):
     # Install each environment
     for env_id in env_ids_to_install:
         logger.info(f"Installing environment: {env_id}")
-        install_cmd = ["prime", "env", "install", env_id]
+        install_cmd = ["uv", "run", "prime", "env", "install", env_id]
```
samsja:
```suggestion
        install_cmd = ["uv", "run", "--no-sync", "prime", "env", "install", env_id]
```

samsja (approved):
lgtm, would merge once its tested on couple of env at the same time to be sure uv doesn't uninstall them

willccbb:
yep multi-env working smoothly

PR #1260, opened 2025-11-08
fix(orchestrator): prevent extra padded micro batches

cwwarren:
Hey folks, I noticed what appears to be a minor bug that could lead to extra padding batches wasting compute. Currently `num_padding_batch` is calculated as `num_train_workers - len(micro_batches) % num_train_workers` which creates `num_train_workers` extra padding batches if the micro batch count is evenly divisible by the number of train workers since `len(micro_batches) % num_train_workers` would be 0 leaving `num_train_workers - 0 = num_train_workers` padding batches.

This change modifies the calculation to `-len(micro_batches) % num_train_workers` which will find the required count of padding batches for the micro batch count to be divisible by the number of train workers.

Scenarios:

```python
# 4 mb, 2 train workers - fixed
2 - 4 % 2 # = 2
-4 % 2    # = 0

# 5 mb, 2 train workers - no change
2 - 5 % 2 # = 1
-5 % 2    # = 1

# 11 mb, 4 train workers - no change
4 - 11 % 4 # = 1
-11 % 4    # = 1
```

Testing: `uv run pytest -v`

### DIFF
--- a/src/prime_rl/orchestrator/batch.py
+++ b/src/prime_rl/orchestrator/batch.py
@@ -150,7 +150,7 @@ def prepare_batch(
         prepare_micro_batch_packing(micro_batch, max_seq_len, temperature) for micro_batch in micro_batches_list
     ]
 
-    num_padding_batch = num_train_workers - len(micro_batches) % num_train_workers
+    num_padding_batch = -len(micro_batches) % num_train_workers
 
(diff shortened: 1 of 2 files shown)


mikasenghaas (approved):
ah nice, good catch! i havent looked at this line of code in months lol but this looks correct to me

PR #1263, opened 2025-11-10
Refactor buffers

faresobeid:
Refactor Buffers into one class. 
Allows now for adding new datasets between checkpoints and changes the logic for sampling and updating problems.
Allows for the use of a replay buffer which can fill in sampled rollouts if the rollout buffer doesn't have enough.
Allows take_all_rollouts which at sample_problems takes the entire rollout buffer


src/prime_rl/orchestrator/config.py
```diff
@@ -281,14 +281,14 @@ class CheckpointConfig(BaseConfig):
 
 
 class BufferConfig(BaseModel):
-    """Base config for all buffer types."""
+    """Configures the buffer for the orchestrator."""
 
-    from_scratch: Annotated[
-        bool,
+    dataset_path: Annotated[
```
mikasenghaas: im pro removing the `from_scratch` flag bc it should be ckpt manager concern but im con making dataset path part of the buffer config. imo should be part of ckpt config

src/prime_rl/orchestrator/config.py
```diff
@@ -436,7 +408,7 @@ class OrchestratorConfig(BaseSettings):
     eval: OnlineEvalConfig | None = None
 
     # Data buffer configuration
-    buffer: Annotated[DataBufferConfigType, Field(discriminator="type")] = SimpleBufferConfig()
+    buffer: BufferConfig = Field(default_factory=BufferConfig)
```
mikasenghaas: dont need a factory here?

mikasenghaas (commented):
should be based off areal branch no?

src/prime_rl/orchestrator/config.py
faresobeid: hmm good point will fix

Issue #1251, opened 2025-11-06
Checkpoint edge case with `--ckpt.keep 1`


mikasenghaas:
It can happen that you end up with no full checkponits because orch checkpoints next step, deletes prev one, but trainer does not yet have next one

PR #1266, opened 2025-11-10
Fix/checkpoint cleanup incomplete

dhyaneesh:
Fixes edge case where checkpoint cleanup with `--ckpt.keep 1` could delete all complete checkpoints before the trainer finished saving, leaving no valid checkpoints.

**Changes:**
- Added `_is_complete_checkpoint()` method to check if both orchestrator and trainer checkpoints exist
- Updated `maybe_clean()` to only delete old checkpoints when the newest kept checkpoint is complete
- Added `get_step_path()` to trainer CheckpointManager for consistency
- Fixed tests to handle existing directories

**GitHub Issue**: #1251 

### DIFF
--- a/src/prime_rl/trainer/ckpt.py
+++ b/src/prime_rl/trainer/ckpt.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Any
 
 import torch
+import torch.distributed as dist
 from torch import nn
 from torch.distributed.checkpoint.state_dict import get_state_dict, set_state_dict
 from torch.distributed.checkpoint.state_dict_loader import load as dcp_load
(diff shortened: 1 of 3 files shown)


dhyaneesh:
I only added exist_ok=True for the test cases, not sure why it is considering it as a whole file change.

mikasenghaas:
im wondering if the more elegant solution would be to just make the trainer ckpt manager responsible of cleanup as it is always lags behind the orchestrator

dhyaneesh:
Yea actually that makes, sense I will update this PR, I got thrown off by this comment 
`    """Test that _is_complete_checkpoint returns True when both orchestrator and trainer checkpoints exist."""
`

dhyaneesh:
> im wondering if the more elegant solution would be to just make the trainer ckpt manager responsible of cleanup as it is always lags behind the orchestrator

What about orchestrator-only workloads like inference? Should I include a cleanup for just the orchestrator in case it's only running without a trainer ?

mikasenghaas:
orchestrator should never really run outside an RL so don't really have to consider this case imo, unless you have a good usecase in mind?

dhyaneesh:
I thought benchmarking was not going to call the trainer that's why I asked

dhyaneesh:
Fix distributed checkpoint cleanup and prevent deletion of incomplete checkpoints

Fixes edge case where checkpoint cleanup with --ckpt.keep 1 could delete all 
complete checkpoints before the trainer finished saving, leaving no valid 
checkpoints.

Changes:

- Implemented distributed checkpoint cleanup: each rank deletes its own 
  dataloader files in parallel, master deletes shared DCP files and step 
  directories

- Made trainer checkpoint manager responsible for cleanup (lags behind 
  orchestrator), ensuring safe deletion of entire step directories

- Added _is_complete_checkpoint() to verify both orchestrator and trainer 
  checkpoints exist

- Updated maybe_clean() to only delete when newest kept checkpoint is complete, 
  preventing deletion of partial checkpoints

- Added get_step_path() method for consistency with orchestrator API

- Fixed dist.broadcast_object_list() usage (handles in-place mutation correctly)

- Fixed assert statement (removed unnecessary type annotation)

This ensures at least one complete checkpoint always exists and improves 
cleanup performance through distributed I/O.

GitHub Issue: #1251

mikasenghaas:
hey! sorry this was already fixed as part of a general cleanup refactor in #1299. thank you for the effort still!

PR #1267, opened 2025-11-11
Fix redundant weight checkpoint saving on resume

mikasenghaas:
Previously, we were saving the weight checkpoint redundantly on the RL trainer when resuming from weight checkpoint. Now, we skip if we `is_first_step`. We also use the same pattern as for the full checkpoint of writing a final checkpoint outside the main loop no matter the interval configuration so that one always has the final weight checkpoint available


samsja (approved):
lgtm

samsja (changes requested):
I don't think we should merge this pr, the first restart from ckpt save is basically a broadcast I don't think we should remove it or we won't be able to start from ckpt unless we also save the weight step

mikasenghaas:
will be solved in #1299

PR #1284, opened 2025-11-13
Replay buffer and take_all_rollout feature

faresobeid:


src/prime_rl/orchestrator/config.py
```diff
@@ -343,6 +343,20 @@ class BufferConfig(BaseModel):
...
     ] = None
 
+    use_replay_buffer: Annotated[
+        bool,
+        Field(
+            description="Whether to use a replay buffer to sample rollouts from. If True, will sample rollouts from the replay buffer instead of the rollout buffer if not full",
+        ),
+    ] = False
+
+    take_all_rollouts: Annotated[
```
mikasenghaas, line 353: i dont get this one, why would i do that?

mikasenghaas (commented):
code looks good! can you explain why i'd use `take_all_rollouts`?

PR #1286, opened 2025-11-13
add cpu offloading fsdp

samsja:
add fsdp optimizer and grad offloading to reduce memory usage with adam


src/prime_rl/trainer/utils.py
```diff
@@ -22,11 +22,16 @@
...
     torch.cuda.set_device(get_world().local_rank)
+    # Use Gloo backend when CPU offloading is enabled (Gloo supports both CPU and GPU)
+    # Otherwise use NCCL for better GPU performance
+    backend = "gloo" if enable_gloo else "nccl"
+    if enable_gloo:
+        get_logger().info("Using Gloo backend for FSDP CPU offloading support")
     dist.init_process_group(
-        backend="nccl",
-        device_id=torch.device("cuda", torch.cuda.current_device()),
+        backend=backend,
```
S1ro1: we can use "cuda:nccl,cpu:gloo" to specify both

PR #1289, opened 2025-11-15
add filtered rollout logging

faresobeid:
Log the ratio of rollouts that finished within a step that got filtered because they were above the easy threshold (too easy) or below the hard threshold (too hard)

Screenshot 2025-11-15 at 7 57 18 PM

### DIFF
--- a/src/prime_rl/orchestrator/buffer.py
+++ b/src/prime_rl/orchestrator/buffer.py
@@ -15,7 +15,7 @@ class Buffer:
     """A buffer for storing rollouts and metadata."""
 
     POOLS = ["easy", "normal", "hard"]
-    FILTER_THRESHOLDS = ["easy", "normal", "hard"]
+    DIFFICULTY_LEVELS = ["easy", "normal", "hard"]
     def __init__(self, dataset: Dataset, buffer_config: BufferConfig, buffer_path: Path | None = None):
         self.config = buffer_config
         if self.config.seed is not None:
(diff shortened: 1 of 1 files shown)


src/prime_rl/orchestrator/buffer.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, dataset: Dataset, buffer_config: BufferConfig, buffer_path: P
         # The number of problems/ rollouts sampled from each pool at the current step (will reset with every call to get_metrics)
         self.num_sampled_problems_per_pool = defaultdict(int)  # Will reseet every step
```
mikasenghaas: 😭

src/prime_rl/orchestrator/buffer.py
```diff
@@ -15,7 +15,7 @@ class Buffer:
     """A buffer for storing rollouts and metadata."""
 
     POOLS = ["easy", "normal", "hard"]
-
+    FILTER_THRESHOLDS = ["easy", "normal", "hard"]
```
mikasenghaas: hmm not sure about filter threshold normal? like its a bit ill defined no?

mikasenghaas: like by definition its not filtered if its in normal

src/prime_rl/orchestrator/buffer.py
```diff
@@ -140,6 +144,12 @@ def get_metrics(self) -> dict[str, float]:
         prefix = "buffer/sampled_rollouts"
         metrics.update({f"{prefix}/{pool}": value for pool, value in zip(self.POOLS, rollout_pool_ratio)})
 
+        # Add ratio of rollouts filtered by each threshold this step
```
mikasenghaas: wondering if we should not do a ratio here, but show absolute count of easy and hard filtered rollouts?

faresobeid: hmm not sure tbh

src/prime_rl/orchestrator/buffer.py
```diff
@@ -111,12 +112,15 @@ def update(self, rollouts: list[Rollout]):
...
-            if (self.config.filter_min_threshold is not None and avg_reward <= self.config.filter_min_threshold) or (
-                self.config.filter_max_threshold is not None and avg_reward >= self.config.filter_max_threshold
-            ):
+            if self.config.filter_min_threshold is not None and avg_reward <= self.config.filter_min_threshold:
+                self.num_sampled_rollouts_per_difficulty["easy"] += 1
                 continue
-
-            self.rollout_buffer.extend(example_rollouts)
+            elif self.config.filter_max_threshold is not None and avg_reward >= self.config.filter_max_threshold:
+                self.num_sampled_rollouts_per_difficulty["hard"] += 1
```
mikasenghaas: HMMM but its not sampled no haah?

faresobeid: AHHHH it was technically rolled out and sampled, just not put in the buffer?

PR #1292, opened 2025-11-15
Fix breaking configs

mikasenghaas:
This PR fixes a couple of misc issues around testing our configs:
- Some config classes (e.g. `BufferConfig`) were not using Pydantic's strict mode so our CI didn't catch that some  TOML config files were setting values which were already deprecated. I switched all config classes that I could to strict mode to prevent this from happening in the future and fixed the TOML files. I also renamed the `BaseConfig` class to `StrictModel` to make its functionality clearer.
- Remove the `validate_device` model validator on `RLConfig` because it's a runtime argument and config CI shouldn't fail if run on less (or no) GPUs than specified--this restores local CI functionality

Note: There's still a somewhat nasty edge case around the usage discriminated union configs. If the default model has attributes which another type in the union does not have, these values carry over because we allow partial updates to the models and cause validation errors that should not happen. Those models are hence not in strict model which may lead to other, but less concerning, footgun. After a little investigation, there doesn't  seem to be a great way out of this. We can choose to:
- Stop settings defaults in the Pydantic models (this is very annoying because each run needs to set the whole config then)
- Stop using discriminated unions (it's a possibility but quite a breaking change so not doing for now)
- Do not validate strictly (our current, and probably, best option for now)

### DIFF
--- a/configs/debug/moe/sft/train.toml
+++ b/configs/debug/moe/sft/train.toml
@@ -13,4 +13,3 @@ load_using_meta = true
 type = "fake"
 batch_size = 2
 seq_len = 1024
-fake = "fixed"
--- a/configs/skywork-acereason-math/rl.toml
+++ b/configs/skywork-acereason-math/rl.toml
@@ -39,11 +39,10 @@ num_examples = 32
 rollouts_per_example = 16
 
 [orchestrator.buffer]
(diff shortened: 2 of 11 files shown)


samsja (approved):
lgtm

PR #1305, opened 2025-11-17
Feat ring flash attn

S1ro1:
## RL

Hendrycks math 1.5B deepseek distill with CP=2

image

image

## SFT

Reverse text on 0.6B Qwen3 - 1024 Seq Len just as an example
- removes the old context parallelism, replaces it with ring-flash-attn as this one supports cat dataset
- grad norm is weird, but had the same issue in hf as well, loss is better signal here

image

---

> [!NOTE]
> Adds ring-flash-attn and full CP support across RL/SFT (attention, training loops, data/packing, loss), with DP-aware loaders and optional micro-batch padding.
> 
> - **Core/Attention**:
>   - Integrate `ring-flash-attn` and add `substitute_prime_rl_flash_attn` to override FA2/FA3 for CP.
>   - Minor fixes in `attn.py` comments.
> - **Context Parallelism (CP) Support**:
>   - New `utils/cp.py` for sharding, CU seqlens setup, and padding-logit exchange; used in RL/SFT training.
>   - RL/SFT trainers: CP-aware input/position/targets, cross-rank logprob/entropy gathering, and FA substitutions.
>   - SFT config validations for CP: `pack_function=cat`, seq len divisible by CP, micro batch size = 1.
> - **Data & Packing**:
>   - `prepare_batch`: optional `pad_to_multiple_of` and micro-batch padding helper.
>   - RL `DataLoader`/`Packer`: DP-rank–based receivers, plumb `pad_to_multiple_of`; Fake loader can generate packed samples and is DP-consistent.
> - **Training/Loss/Perf**:
>   - `shift_logits` accepts optional left-pad logit (for CP boundaries).
>   - RL tokens/sec computation based on DP size; auto trainer workers divide by `cp*tp`.
> - **Misc/Deps**:
>   - Add `ring-flash-attn` dependency; ignore `outputs/` in `.gitignore`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a1e5d4d6e4c07ed7bd52ada323e2dd6b44210b7b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/trainer/models/layers/attn.py
```diff
@@ -178,7 +170,73 @@ def forward(
...
+            )
+            out = out.contiguous()
+            attn_output = out.view(1, out.shape[0], -1)
+            attn_weights = None
+
+            # attn_output = attn_output.reshape(*input_shape, -1).contiguous()
+            attn_output = self.o_proj(attn_output)
+            return attn_output, attn_weights
+
+    ATTN_IMPL2CLASS["flash_attention_2"].forward = RingFlashAttention.forward
```
samsja: not the biggest fan of this pattern, smell like transformer magic.

what about just having flash_attenion_2_cp in attn.impl ? and doing parralel.cp 2 would automatically change the attention to it

src/prime_rl/trainer/rl/train.py
```diff
@@ -207,34 +235,62 @@ def train(config: RLTrainerConfig):
...
+            )
+            advantages = maybe_shard_for_cp(
+                micro_batch["advantages"].to("cuda"), cp_rank=cp_rank, cp_world_size=config.model.cp
+            )
+            loss_mask = maybe_shard_for_cp(
+                micro_batch["loss_mask"].to("cuda"), cp_rank=cp_rank, cp_world_size=config.model.cp
+            )
+            inference_logprobs = maybe_shard_for_cp(
+                micro_batch["inference_logprobs"].to("cuda"), cp_rank=cp_rank, cp_world_size=config.model.cp
+            )
```
samsja: I would rather do
```suggestion
            if config.model.cp > 1:
                 shard_for_cp(...)
```


not fan of the maybe pattern

src/prime_rl/trainer/models/layers/attn.py
S1ro1: The issue with that is we need the process group already for this, which we do not have at init time. You can see that the process group is captured in this call.

src/prime_rl/trainer/rl/train.py
S1ro1: yep, not a fan of this either, not a strong preference here.

src/prime_rl/trainer/rl/data.py
```diff
@@ -23,16 +23,54 @@ class MicroBatch(TypedDict):
...
+        inference_logprobs = torch.randn(input_ids.shape[0])
+
+        return {
+            "input_ids": input_ids.unsqueeze(0),
+            "position_ids": position_ids.unsqueeze(0),
+            "advantages": advantages.unsqueeze(0),
+            "inference_logprobs": inference_logprobs.unsqueeze(0),
+            "temperature": 1.0,
+            "loss_mask": loss_mask.unsqueeze(0),
+        }
```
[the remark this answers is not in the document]
S1ro1: This is not something to care about.

Jackmin801 (commented):
Nice! Looks rlly good. Some of the batch creation functions change a bit with the multi-dir merge. But otherwise I think this is great! Oh and as discussed in weekly, we can delete the stacking dataset logic and only do catting.

src/prime_rl/trainer/rl/config.py
```diff
@@ -43,7 +43,9 @@ class FakeDataLoaderConfig(BaseConfig):
 
     batch_size: Annotated[int, Field(ge=1)] = 2
     seq_len: Annotated[int, Field(ge=1)] = 128
-    generate_documents: Annotated[bool, Field(description="Whether to generate documents instead of micro batches.")] = False
+    generate_documents: Annotated[
+        bool, Field(description="Whether to generate documents instead of micro batches.")
+    ] = False
```
samsja: would need to find a new better name imo. also lets use sample instead of documents

src/prime_rl/trainer/rl/data.py
```diff
@@ -39,18 +39,70 @@ def micro_batch_to_tensor(micro_batch: MicroBatch) -> TensorMicroBatch:
...
+            get_micro_batch_fn = self._get_micro_batch
+        else:
+            get_micro_batch_fn = self._get_document_micro_batch
+
+        # This is a pretty ugly hack to ensure that all CP ranks in a data parallel group receive the same micro batch.
+        micro_batches = []
+        for micro_batch_idx in range(self.num_micro_batches):
+            seed = self.dp_rank * 1000000 + self.batch_counter * 1000 + micro_batch_idx
+            generator = torch.Generator().manual_seed(seed)
+            micro_batches.append(get_micro_batch_fn(generator))
```
samsja, line 68: seems very ugly indeed haha

src/prime_rl/trainer/rl/data.py
```diff
@@ -39,18 +39,70 @@ def micro_batch_to_tensor(micro_batch: MicroBatch) -> TensorMicroBatch:
...
+
+        self.batch_counter += 1
+        return micro_batches
+
+    def _get_document_micro_batch(self, generator: torch.Generator) -> TensorMicroBatch:
```
samsja:
```suggestion
    def _get_sample_micro_batch(self, generator: torch.Generator) -> TensorMicroBatch:
```

samsja: need docstring here

src/prime_rl/trainer/rl/data.py
S1ro1: it's just fake data loader so it's a price i'm willing to pay

Jackmin801 (approved):
LFGTM!

src/prime_rl/trainer/rl/data.py
```diff
@@ -39,18 +39,70 @@ def micro_batch_to_tensor(micro_batch: MicroBatch) -> TensorMicroBatch:
...
 
     def wait_for_batch(self) -> None:
         return
 
     def get_batch(self) -> list[TensorMicroBatch]:
-        return [self._get_micro_batch() for _ in range(self.num_micro_batches)]
+        if not self.generate_samples:
+            get_micro_batch_fn = self._get_micro_batch
+        else:
+            get_micro_batch_fn = self._get_document_micro_batch
```
Jackmin801: with the rename now this is wrong?
```suggestion
            get_micro_batch_fn = self._get_sample_micro_batch
```

src/prime_rl/trainer/rl/data.py
samsja: oh okay but then why do we even need it here ? for fake data we don't really care they have the same data right ?

samsja (approved):

PR #1309, opened 2025-11-18
feat fa3 - v2

S1ro1:
We make flash-attn-3 really optional by just adding `uv pip install ...` in the readme, we can't have it as a dependency due to this:
https://github.com/astral-sh/uv/issues/7299

TLDR: if it's an extra dependency, it has no static dependency metadata so uv tries to build it to get its dependencies, even if it's extra - then it fails. 

---

> [!NOTE]
> Adds optional FlashAttention 3 across models with runtime availability checks, updates docs for manual install, and switches CI/Docker/install to lock-only uv sync (dropping extras).
> 
> - **Models/Trainer**
>   - Add `flash_attention_3` to `AttnImplementation` and handle in `glm.py`, `llama.py`, `qwen.py` sequence length prep.
>   - Introduce FA versioned attention: `FlashAttention` now selects v2/v3 (`flash_attn` vs `flash_attn_interface`) and normalizes return type; update `ATTN_IMPL2CLASS` mapping.
>   - Guard FA3 usage via `is_flash_attn_3_available()` with clear install error in `setup_model`.
> - **Docs**
>   - README: add optional FA3 install instructions and caveats; environment validation notes.
> - **Infra**
>   - CI workflow, Dockerfile, and `scripts/install.sh`: use `uv sync --locked` only; remove `--all-extras` and extra sync steps.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 309b3799e1b9e92e4e5461bd24c1a3f4a6aaacec. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ uv sync
 3.1. Optional: Install Flash Attention 3 (on Hopper GPUs only, for flash_attention_3 attention backend)
 
 > *NOTE*: This step will take a while, as it builds the Flash Attention 3 extension from source, as it has no wheels prebuilt.
-> *NOTE*: After this step, you can't run `uv sync` as it will uninstall the package, you can avoid it by running `uv sync --inexact`
+> *NOTE*: After this step, you can't run `uv sync` or `uv run` as it will uninstall the package, you can avoid it by running `uv sync --inexact` or `uv run --no-sync`
 
 ```bash
 uv pip install "flash-attn-3 @ git+https://github.com/Dao-AILab/flash-attention.git@main#subdirectory=hopper" --no-build-isolation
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -79,6 +79,15 @@ source $HOME/.local/bin/env
 uv sync
 ```
 
+3.1. Optional: Install Flash Attention 3 (on Hopper GPUs only, for flash_attention_3 attention backend)
+
+> *NOTE*: This step will take a while, as it builds the Flash Attention 3 extension from source, as it has no wheels prebuilt.
+> *NOTE*: After this step, you can't run `uv sync` as it will uninstall the package, you can avoid it by running `uv sync --inexact`
```
samsja: would also add uv run --no-sync here

samsja (approved):

PR #1316, opened 2025-11-19
Fp8 inference 

faresobeid:
First have to convert your model to fp8 by running:
```
python scripts/convert_to_fp8.py <model_name_or_path> <output_path>
```
Then to run inference:
```
uv run inference --model.name <output_path> --quantization
```
Can optionally use 
```
uv run inference --model.name <output_path> --quantization --use_deep_gemm
```
for potentially faster inference.
For training, you would set the orchestrator and inference model names to the fp8 model, and keep the trainer on the original bf16 model

### DIFF
--- a/src/prime_rl/trainer/fp8_triton.py
+++ b/src/prime_rl/trainer/fp8_triton.py
@@ -1,0 +1,66 @@
+from typing import Tuple
+
+import torch
+import triton
+import triton.language as tl
+
+fp8_dtype = torch.float8_e4m3fn
+fp8_max = torch.finfo(fp8_dtype).max
+fp8_min = -fp8_max
+
+
+def ceil_div(x: int, y: int) -> int:
+    return (x + y - 1) // y
+
+
+@triton.jit
+def _blockwise_cast_to_fp8_triton(
+    X,
+    Y,
+    S,
+    stride_xm,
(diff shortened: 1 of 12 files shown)


src/prime_rl/trainer/fp8_triton.py
```diff
@@ -0,0 +1,66 @@
+from typing import Tuple
```
samsja, line 1: lets up the link to the original slime code

src/prime_rl/utils/validation.py
```diff
@@ -34,7 +34,7 @@ def validate_shared_model_name(
     orchestrator: OrchestratorConfig,
     inference: Optional[InferenceConfig] = None,
 ) -> None:
-    if trainer.model.name.startswith("Jackmin108/"):  # The TT MoE models will have a different name on the orchestrator
+    if trainer.model.name.startswith("Jackmin108/") or orchestrator.model.name.endswith("-fp8"):  # The TT MoE models will have a different name on the orchestrator
```
samsja, line 37: arf not a fan of this tbh but this is prior your work

faresobeid: ya i guess the other option is to just always allow the orch + inference model name to be different to the trainer

PR #1317, opened 2025-11-19
decouple environment execution from orchestrator eventloop

JannikSt:
- PR is still work in progress 

Core goal is that with merging this, the user does not have to change anything. I've introduced an env_manager that simply boots up the environments when running a single orchestrator. This still gives us the benefit of executing an environment in a different process.

You can test running a remote environment easily with e.g. reverse text: 
First, start the environment: 
```bash
uv run python -m prime_rl.environment.zmq_server \
  --bind tcp://0.0.0.0:5555 \
  --env-id reverse-text \
  --instance-name reverse-text-env
```

Adjust the config on your orchestrator: 
config
```bash
[[orchestrator.env]]
id = "math"
# Replace 192.168.x.x with your remote machine's actual IP address
zmq_endpoint = "tcp://192.168.1.50:5555"
```

Start the reverse text example: 
```bash
uv run rl \
  --inference @ examples/reverse_text/rl/infer.toml --log.level debug
```

Things that are still WIP: 
- [ ] Explicit Mixing Ratios missing: The following configuration and logic for explicit environment mixing (via verifiers.EnvGroup) was removed in favor of the ZMQ implementation. This functionality provided weighted dataset interleaving and custom stopping strategies and needs to be readded
- [ ] Thinking about the load dataset calls - this kind of currently defeats the purpose as we are sending the whole dataset in that call. Might just go back to sending ids only


mikasenghaas (commented):
nice, i like the user facing api! wondering if it would make sense to make this a first class entrypoint, like rl and sft, ie. instead of 

```
uv run python -m prime_rl.environment.zmq_server \
  --bind tcp://0.0.0.0:5555 \
  --env-id reverse-text \
  --instance-name reverse-text-env
```

have smth like

```
uv run env \
  --bind tcp://0.0.0.0:5555 \
  --id reverse-text \
  --name reverse-text-env
```

here id and name would one to one match how envs are configured on the orchestrator

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -87,23 +89,82 @@ async def orchestrate(config: OrchestratorConfig):
...
+            # Auto-start worker
+            logger.info(f"Auto-starting {env_name} worker on {endpoint}")
+            manager = EnvironmentManager(
+                env_id=env_config.id,
+                instance_name=instance_name,
+                endpoint=endpoint,
+                env_args=env_config.args,
+                output_dir=config.output_dir,
+                log_level=config.log.level,
+            )
```
samsja, line 126: not fan of a env manager starting a uv run, can't we do sub process ?

PR #1356, opened 2025-11-29
refactor models to dir and add model specific conversion function

Jackmin801:
> [!NOTE]
> Adds PreTrainedModelPrimeRL with model-specific HF↔PrimeRL state dict conversion, refactors GLM4/Qwen3/Llama into dedicated modules, updates checkpoint/loading/broadcast to use it, removes legacy conversion scripts, and adds unit tests.
> 
> - **Models**:
>   - Introduce `PreTrainedModelPrimeRL` with `is_hf_state_dict`/`is_prime_state_dict` and `convert_to_(hf|prime)` (+ per-layer variants).
>   - Refactor into modules: `glm4_moe/`, `qwen3_moe/`, `llama/` (configs, modeling, conversion helpers).
>   - Implement model-specific MoE conversions for `glm4_moe` and `qwen3_moe`; `llama` uses HF format passthrough.
> - **Checkpoint/Loading/Broadcast**:
>   - `ckpt.py`/`model.py`: auto-convert weights HF↔PrimeRL using model APIs; call `init_buffers_post_meta` for PrimeRL models.
>   - RL broadcast (filesystem + NCCL): convert to HF (whole or per-layer) via model APIs before sending/saving.
> - **Weights/Utils**:
>   - Remove global MoE conversion helpers; keep shard/save/load utilities.
>   - Add `utils.default_dtype` context manager.
>   - MoE layer: disable grouped mm on non-SM90 GPUs.
> - **Tests**:
>   - Add unit tests for `glm4_moe`, `qwen3_moe`, and `llama` verifying HF parity and conversion round-trips.
> - **Scripts**:
>   - Remove legacy `scripts/convert_moe_to_hf.py` and `scripts/extract_hf_from_ckpt.py`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b5d73e99218fba098ca3ff611473bbc3ccd37331. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


scripts/convert_moe_to_hf.py
mikasenghaas, line 1: do we still need the conversion scripts even?

mikasenghaas (commented):
nice, i really like it. can we add tests for this somehow? maybe one a dummy model on cpu with one layer?

scripts/convert_moe_to_hf.py
Jackmin801: `convert_moe_to_hf` can be scrapped. `extracting_hf_from_ckpt` might still be useful if we want to save storage by not saving weights and just saving ckpt. the script would let us extract the weights from the ckpt if needed

Jackmin801: I believe the MoE layer doesnt work on cpu because the histogram counting for tokens_per_expert used to report max_vio doesnt have a cpu impl. Writing a gpu test that tests that fwd for HF model and TT model yield the same logits. But it seems like rn they dont :think:

Jackmin801: ugh, it was because huggingface puts attention_mask false on position_id 0 and youre supposed to start at 1

src/prime_rl/trainer/rl/broadcast/nccl.py
```diff
@@ -122,7 +122,7 @@ def broadcast_weights(self, model: nn.Module, step: int) -> None:
                 state_dict[key] = value
 
             if has_tt_moe_layers(state_dict):
```
samsja: I would also make this a model function, like basically any model should be able to define a custom mapping.

src/prime_rl/trainer/model.py
```diff
@@ -23,8 +23,6 @@
 from prime_rl.trainer.models import AutoModelForCausalLMPrimeRL
 from prime_rl.trainer.parallel_dims import ParallelDims
 from prime_rl.trainer.weights import (
-    convert_hf_to_tt_moe,
-    convert_tt_to_hf_moe,
     has_hf_moe_layers,
     has_tt_moe_layers,
     load_state_dict,
```
samsja, line 26: so ideally we disantagle the checkpointing converstion part from the moe part

samsja (commented):
this looks good, I would go one step further tho and disentangle moe logic from ckpt conversion.

any model should allow to define checkpoint conversion. 

we should expoe something like

* model.need_conversion(state_dict)
* model.convert(snapshot_state_dict)
* model.init_buffer_post_meta()

...

src/prime_rl/trainer/models/glm4_moe/converting_glm4_moe.py
```diff
@@ -0,0 +1,131 @@
...
+        del state_dict[f"model.layers.{i}.mlp.experts.w1"]
+        del state_dict[f"model.layers.{i}.mlp.experts.w2"]
+        del state_dict[f"model.layers.{i}.mlp.experts.w3"]
+
+
+def convert_tt_to_hf_moe(state_dict: dict[str, Tensor]):
+    """Convert MoE weights from TT to HF format in-place."""
+    num_layers = get_max_layer_num(state_dict)
+    for i in range(1, num_layers + 1):
+        convert_tt_layer_to_hf(state_dict, i)
```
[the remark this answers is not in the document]
Jackmin801: iirc this is because there is an extra layer of unused model weights in glm for the multi token prediction. So if you load the model in single token prediction, you are missing one layer to be converted. it should only be in HF -> TT and not TT -> HF though so this might be a catch

samsja (approved):
lfgtm !

PR #1368, opened 2025-12-01
Integrate weave

mikasenghaas:
Not ready to merge yet. Have to explore a bit more, and should make sure we don't throttle performance by blocking event loop or some bs


mikasenghaas:
im scared of their [monkey-patches](https://github.com/wandb/weave/blob/master/weave/integrations/verifiers/verifiers.py#L191)

PR #1369, opened 2025-12-01
Do not double load weights on resume

faresobeid:
> [!NOTE]
> When resuming from a checkpoint, models are initialized on meta and `to_empty` is used to avoid loading base HF weights; trainers pass a flag based on `ckpt.resume_step`.
> 
> - **Model initialization (`trainer/model.py`)**:
>   - `setup_model` now accepts `skip_load_weights` and uses meta device when `config.load_using_meta` or resuming from checkpoint.
>   - On resume (`skip_load_weights=True`): allocate with `model.to_empty("cuda")`, sync, and init buffers (`init_buffers_post_meta` or `fix_model_post_empty`) instead of loading HF weights; otherwise load via DCP.
>   - Falls back to CPU load if meta-path unsupported (`can_load_dcp_from_hf`).
> - **Trainers**:
>   - **RL (`trainer/rl/train.py`)** and **SFT (`trainer/sft/train.py`)** pass `skip_load_weights = config.ckpt and config.ckpt.resume_step` to `setup_model` to prevent double-loading during checkpoint resume.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d9203d775a93d19c84b7d0aa5592f39a1277ccf5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/trainer/model.py
```diff
@@ -300,22 +302,26 @@ def apply_compile(model: nn.Module, compile_config: CompileConfig):
...
         config,
-        device=torch.device("meta" if config.load_using_meta else "cpu"),
+        device=torch.device("meta" if use_meta else "cpu"),
         dtype=DTYPE_MAP[config.optimization_dtype],
     )
 
-    # Reload the model to CPU if we cannot load from
-    if config.load_using_meta and not can_load_dcp_from_hf(model):
+    # Reload the model to CPU if we cannot load from meta device (only when not skipping weights)
+    if use_meta and not can_load_dcp_from_hf(model) and not skip_load_weights:
```
Jackmin801: I think the `can_load_dcp_from_hf` check here is to make sure we know how to re-init the non-persistent buffers and you dont get an incorrect model. If we dont fall back to cpu on `skip_loading_weights` caused by ckpt, the ckpt model might be loaded incorrectly (e.g. have wrong rotary buffers)

src/prime_rl/trainer/model.py
```diff
@@ -331,8 +337,16 @@ def setup_model(config: ModelConfig, parallel_dims: ParallelDims) -> nn.Module:
 
     setup_fsdp(model, config, parallel_dims)
 
-    if config.load_using_meta and can_load_dcp_from_hf(model):
-        load_dcp_from_hf(model, config)
+    if use_meta:
+        if skip_load_weights:
```
Jackmin801, line 337: I think this also needs to check `can_load_dcp_from_hf`. Or maybe rename the function to `can_reinit_empty_buffers` or `can_load_meta` because thats mainly what it is checking.

Jackmin801 (approved):
nice! lgtm! should save us quite some time on resume.
added a nit comment around the edge case where we dont know how to reinit the buffers post empty. im not aware of any model on HF that would trigger the fallback to cpu atm so its just in case one gets used some day

Jackmin801 (approved):
lgtm!

PR #1372, opened 2025-12-02
Fix/orchestrator starvation

pmahdavi:
### Problem Description
When running the **GSM8K RL training example** (`configs/gsm8k/rl.toml`), we observed a severe performance regression. When the orchestrator hits the `max_async_level` barrier, it correctly pauses to let the trainer catch up. However, once the barrier clears, the previous logic only scheduled a **single** new group rollout (e.g., 16 requests).

This caused the scheduling loop to enter a "one-out, one-in" state, maintaining a concurrency of only 1 task (16 requests) instead of refilling to the target batch capacity (e.g., 32 tasks / 512 requests).

**Symptoms observed on GSM8K run:**
- **Throughput Collapse:** Dropped from ~15k tokens/s to ~500 tokens/s after the first barrier was hit (Step 2/3).
- **Step Latency:** Exploded from ~90s to >15 minutes.
- **Inference Starvation:** Logs showed `Running: 16 reqs` with `Waiting: 0` (server idle).

### Solution
Modified `src/prime_rl/orchestrator/scheduler.py` to aggressively refill the `inflight_group_rollouts` queue up to `problems_per_batch` immediately after the barrier lifts. This restores full concurrency and throughput instantly.

### Verification
Verified on a 2-GPU setup running the `gsm8k` configuration.

#### 1. Performance Metrics Comparison
The chart below clearly shows the impact.
- **After Fix:** Throughput and MFU remain stable throughout the run.
- **Before Fix:** Performance collapses immediately after Step 2 (when the barrier is first hit) and never recovers.
perf

#### 2. Trainer Logs Comparison

**Before Fix:**
Steps take **>15 minutes (~950s)** once the async barrier is hit (see Step 3 onwards).

before-trainer

**After Fix:**
Steps remain consistent at **~1.5 minutes (~100s)** even after hitting barriers.

after-trainer

---

> [!NOTE]
> Refills `inflight_group_rollouts` up to `problems_per_batch` once the async barrier clears, restoring full rollout concurrency.
> 
> - **Scheduler** (`src/prime_rl/orchestrator/scheduler.py`):
>   - **Rollout scheduling after async barrier**:
>     - In `generate_batch`, when `inflight_group_rollouts` is empty and the async barrier has cleared, repeatedly schedule group rollouts until `problems_per_batch` is reached (unless the barrier re-triggers), instead of scheduling only one.
>   - Preserves existing barrier wait/log behavior and concurrent completion handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 31baac8f7a14bed781289c29454c0facbcadde2f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/gpu_tests.yaml
+++ b/.github/workflows/gpu_tests.yaml
@@ -72,6 +72,7 @@ jobs:
           # This is to allow running tests on forks without WANDB_API_KEY
           WANDB_MODE: ${{ secrets.WANDB_API_KEY && 'online' || 'offline' }}
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           PRIME_API_KEY: ${{ secrets.PRIME_API_KEY }}
           GITHUB_REF_NAME: ${{ github.ref_name }}
           GITHUB_HEAD_REF: ${{ github.head_ref }}
--- a/configs/debug/eval/multi_env.toml
+++ b/configs/debug/eval/multi_env.toml
@@ -1,0 +1,13 @@
+# Example of generating data in two single-turn environments (gsm8k and hendrycks)
+num_examples = 16
+rollouts_per_example = 4
+
+[[env]]
+id = "primeintellect/single-turn-math"
+name = "gsm8k"
+args = { dataset_name = "openai/gsm8k", dataset_subset = "main" }
+
+[[env]]
+id = "primeintellect/single-turn-math"
+name = "hendrycks"
+args = { dataset_name = "PrimeIntellect/Hendrycks-Math", dataset_subset = "default" }
--- a/configs/debug/eval/single_env.toml
+++ b/configs/debug/eval/single_env.toml
@@ -1,0 +1,7 @@
+# Example of generating data in a single single-turn environment (gsm8k)
+num_examples = 16
+rollouts_per_example = 4
+
+[[env]]
+id = "primeintellect/single-turn-math"
+args = { dataset_name = "openai/gsm8k", dataset_subset = "main" }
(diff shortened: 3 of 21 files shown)


mikasenghaas:
ah yea nice, we are aware of this issue but didnt fix yet because in of our runs we setup trainer and inference compute such that we are never bottlenecked by the trainer and consequently dont hit the async barrier and this situation. i wasnt quite aware how severe the issue is tho, so happy to merge this asap

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -193,7 +193,10 @@ async def generate_batch(self, step: int) -> list[vf.State]:
                     pbar.set_description("Generating rollouts (train)")
                     was_waiting = False
 
-                await self.schedule_group_rollout()
+                while len(self.inflight_group_rollouts) < self.problems_per_batch:
```
mikasenghaas, line 196: hm i think there is a more elegant solution implementation wise. the async barrier is a little messy atm in general but fine to merge for now to  improv perf

mikasenghaas:
out of curiosity, what hardware did you run this on? on my rtx pro 6000 trainer step ~ inference step time for gsm8k so i was blind to this haha

mikasenghaas:
either way, really nice catch. thanks a lot!

src/prime_rl/orchestrator/scheduler.py
pmahdavi: Thanks for checking this, yeah I agree it’s probably not a clean one, but it helps with performance.

pmahdavi:
> out of curiosity, what hardware did you run this on? on my rtx pro 6000 trainer step ~ inference step time for gsm8k so i was blind to this haha

I was using [2 H100s](https://gist.github.com/pmahdavi/32525d8a6e555cbc2795a3321a82928a) to run that. The trainer step and inference are actually balanced for me as well after the first few steps. If I’m not missing something, the issue should show up as soon as the inference gets far ahead and needs to pause for the trainer to catch up.

pmahdavi:
> either way, really nice catch. thanks a lot!

Sure, thanks for reviewing this, I appreciate it.

PR #1374, opened 2025-12-02
robust Eval Entrypoint

rasdani:
Refactors eval entrypoint after new syn data gen endpoint. rollout instead of group based.

- adds retries on API errors
- adds smart retry on max_token error
- adds streaming results saving
- adds resume feature
- deprecates saving eval results to HF
- displays running average

starting to match official `gpt-oss-120b` high reasoning numbers on GPQA with `Avg@2=0.7904`

you can run with
```bash
uv run --no-sync eval @ configs/intellect_3/evals/models/gpt-oss-120b-vertex.toml @ configs/intellect_3/evals/envs/gpqa.toml --num-examples 5 --save.stream  # --resume-uuid 08db6279
```
---

**GitHub Issue**: [Issue ID]

---

> [!NOTE]
> Refactors evals to support per‑rollout scheduling with streaming saves, resumable runs, and robust retry logic, plus new configs for INTELLECT‑3 and GPT‑OSS‑120B Vertex.
> 
> - **Eval pipeline**:
>   - Add per‑rollout scheduling with live progress and running-average display.
>   - Implement streaming result writes to `results.jsonl` and resumable evals (filtering completed rollouts/examples).
>   - Add retry/backoff via `tenacity`, including smart `max_tokens` reduction on context errors.
>   - Support custom `reasoning_field` and propagate to result construction.
> - **Config & APIs**:
>   - Extend `EvalSaveConfig` with `stream`; add `RetryConfig`; extend `EvalConfig` with `retry`, `reasoning_field`, `per_rollout`.
>   - `OfflineEvalConfig` adds `reasoning_field`, `resume_path` with validators to enforce streaming and single-env/checkpoint use.
>   - `run_evals/run_eval` updated to pass `reasoning_field`, `retry`, `per_rollout`, and `resume_path`.
>   - New helpers in `prime_rl.utils.vf`: `generate_rollout`.
> - **Configs**:
>   - Add `configs/intellect_3/evals/api/pinference.toml` for `PrimeIntellect/INTELLECT-3`.
>   - Add `configs/.../models/gpt-oss-120b-vertex.toml` (OpenRouter Vertex; high reasoning; large `max_tokens`).
>   - Update `configs/.../envs/gpqa.toml`: `rollouts_per_example=8`, `args = { verifier = "regex" }`.
> - **Dependencies**:
>   - Add `tenacity` runtime dependency.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ec990bac2ab25251ebdee021b4a4ad50d3c4e178. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


configs/intellect_3/evals/api/openrouter/fireworks.toml
```diff
@@ -0,0 +1,5 @@
+max_concurrent = 64
+
+[client]
+base_url = ["https://openrouter.ai/api/v1"]
+api_key_var = "OPENROUTER_API_KEY"
```
samsja: is the idea that we can overide this key via cli ?

configs/intellect_3/evals/api/openrouter/fireworks.toml
```diff
@@ -0,0 +1,5 @@
+max_concurrent = 64
+
+[client]
+base_url = ["https://openrouter.ai/api/v1"]
```
samsja: its not clear from the api here that its using fireworks necessary

src/prime_rl/orchestrator/config.py
```diff
@@ -148,45 +148,20 @@ class EvalSaveDiskConfig(BaseConfig):
...
         Field(
             description="Whether to push eval results to Prime Environment Hub. Automatically pushes all evaluated environments. Requires PRIME_API_KEY and authorization for the environments."
         ),
     ] = False
+    stream: Annotated[
+        bool,
+        Field(
+            description="Whether to save results incrementally as rollouts complete.",
+        ),
+    ] = False
```
samsja, line 195: its not clear to me how we can do this with the trainer as we need to balance all the data

src/prime_rl/eval/utils.py
```diff
@@ -95,29 +224,95 @@ async def run_eval(
     dataset = env.get_eval_dataset(n=num_examples)
     sampling_args = prepare_sampling_args(sampling_config)
 
+    # Determine streaming save path
+    stream_save = save_config.stream
+    if resume_uuid is not None:
+        if not stream_save:
+            logger.warning("Resume requested but stream=False. Enabling streaming saves for resume to work.")
+            stream_save = True
```
mikasenghaas: move this to config as pydantic model validator. main reason is that config is logged correctly, e.g. to wandb (no runtime config changes)

src/prime_rl/utils/vf.py
```diff
@@ -1,15 +1,51 @@
...
+        return context_length - prompt_tokens
+    return None
+
+
+@retry(
+    stop=stop_after_attempt(10),
+    wait=wait_exponential(multiplier=1, min=1, max=60),
+    before_sleep=_log_retry_attempt,
+    reraise=True,
+)
```
samsja: this value seems fine for eval but not sure about training

src/prime_rl/eval/utils.py
```diff
@@ -95,29 +224,95 @@ async def run_eval(
...
+        base_path = get_results_path(env_name_or_id, model_config.name, base_path=output_dir)
+        path_to_save = base_path.parent / resume_uuid / "results.jsonl"
+    else:
+        path_to_save = get_results_path(env_name_or_id, model_config.name, base_path=output_dir) / "results.jsonl"
```
mikasenghaas: is this duplicated?

src/prime_rl/eval/utils.py
```diff
@@ -95,29 +224,95 @@ async def run_eval(
     dataset = env.get_eval_dataset(n=num_examples)
     sampling_args = prepare_sampling_args(sampling_config)
 
+    # Determine streaming save path
+    stream_save = save_config.stream
+    if resume_uuid is not None:
+        if not stream_save:
+            logger.warning("Resume requested but stream=False. Enabling streaming saves for resume to work.")
+            stream_save = True
+        base_path = get_results_path(env_name_or_id, model_config.name, base_path=output_dir)
```
samsja: lets not duplicate this line in both branch of the if statement

samsja (changes requested):
not too sure about the stream usecase during training, otherwise looks good

src/prime_rl/eval/utils.py
```diff
@@ -181,17 +376,6 @@ async def run_eval(
             save_to_disk(dataset, metadata_dict, save_path)
             logger.info(f"Saved eval results for {env_name_or_id} to disk ({save_path})")
 
-        if save_config.hf is not None:
```
mikasenghaas, line 184: oh we dont want to support this anymore?

mikasenghaas: i think we should still give the option to do this. once we are 100% happy with env hub behavior we can deprecate this but not confident enough in it yet that i want to remove from repo entirely

src/prime_rl/eval/utils.py
```diff
@@ -69,20 +103,115 @@ def prepare_sampling_args(sampling_config: EvalSamplingConfig) -> dict[str, Any]
     return sampling_args
 
 
+# TODO: This is a hotfix for as long as verifiers doesn't support reasoning content parsing
```
mikasenghaas: should move funcs into shared utils if exact copy. for now fine to import from syn data gen entrypoint

src/prime_rl/utils/vf.py
```diff
@@ -1,15 +1,51 @@
...
+    prompt_match = re.search(r"(\d+) of text input", error_message)
+
+    if context_match and prompt_match:
+        context_length = int(context_match.group(1))
+        prompt_tokens = int(prompt_match.group(1))
+        return context_length - prompt_tokens
+    return None
+
+
+@retry(
```
mikasenghaas: i would prefer if we
a) decorate with specific values per entrypoint (e.g. evals might want different settings here than rl) 
b) have these configurable
also, oai alr has a built-in retry mechanism (we configure this in `setup_clients` iirc), should double check that there is no weird interaction effects

src/prime_rl/utils/vf.py
```diff
@@ -1,15 +1,51 @@
...
+    """Log retry attempts at WARNING level using the global logger."""
+    logger = get_logger()
+    exception = retry_state.outcome.exception()
+    wait_time = retry_state.next_action.sleep
+    logger.warning(
+        f"Retrying {retry_state.fn.__name__} in {wait_time:.1f} seconds as it raised {exception.__class__.__name__}: {exception}"
+    )
+
+
+def _parse_and_calculate_max_tokens(error_message: str) -> int | None:
```
mikasenghaas: this is super nice, should acc be upstreamed to verifiers. this should be default behavior. esp important in multi-turn envs where we hit this case quite often if
- max tokens is set (this is a per turn limit bc its passed as sampling args to each chat completions call)
- the llm goes to max ctx

mikasenghaas (commented):
banger!:) couple nits, o/w let's get this in asap

src/prime_rl/eval/utils.py
rasdani: uuid gets replaced. but made it clearer now.

src/prime_rl/utils/vf.py
rasdani: OAI will retry itself until it throws an exception only then our retry will kick in.
it uses [headers and response code](https://github.com/openai/openai-python/blob/43e324ed7a8a8b6f0738254c59d732be2807a06f/src/openai/_base_client.py#L750) to determine a retry.

our retry mostly catches JSON decode errors of the request body itself being malformed (happens with openrouter). nothing in OAI handles that.

configs/intellect_3/evals/api/openrouter/fireworks.toml
rasdani: cli args always take precedence afaik

src/prime_rl/orchestrator/config.py
rasdani: streaming is necessary for resuming, which we need with flaky API providers.
it's disabled by default and we can keep it disabled for training.

rasdani:
> not too sure about the stream usecase during training, otherwise looks good

streaming save is necessary for resuming, which we need with flaky API providers.
it's disabled by default and we can keep it disabled for training.

src/prime_rl/utils/vf.py
rasdani: implemented you suggestions 👍

src/prime_rl/utils/vf.py
rasdani: pulled out into eval entry point and made it configurable

src/prime_rl/eval/utils.py
```diff
@@ -95,29 +271,92 @@ async def run_eval(
     dataset = env.get_eval_dataset(n=num_examples)
     sampling_args = prepare_sampling_args(sampling_config)
 
+    # Determine streaming save path
+    base_path = get_results_path(env_name_or_id, model_config.name, base_path=output_dir)
+    if resume_uuid is not None:
+        path_to_save = base_path.parent / resume_uuid / "results.jsonl"  # replace the uuid in the path
```
mikasenghaas: oh acc one thing i didnt catch earlier. what if i had output folder `a` with uuid `1234`, now i want to resume with output folder `b`. this is current not supported. should we use `--resume-path` arg instead to be agnostic to output dir changes?

src/prime_rl/orchestrator/config.py
mikasenghaas: @samsja the stream arg is only part of eval save config, so rl training is not affected. only potentially for online evals but it doent really matter so all good imo. @rasdani i think we should make this true by default in eval entrypoint tho

src/prime_rl/orchestrator/config.py
```diff
@@ -245,10 +294,20 @@ class EvalConfig(BaseConfig):
         default_factory=EvalSaveConfig,
         description="Configures how to save the eval results.",
     )
+    retry: RetryConfig = Field(
```
mikasenghaas, line 297: hm should we make this per-env or global? leaning towards the latter

src/prime_rl/eval/utils.py
```diff
@@ -69,20 +108,157 @@ def prepare_sampling_args(sampling_config: EvalSamplingConfig) -> dict[str, Any]
...
+            # Check if this is a context length error and retry with adjusted max_tokens
+            error_message = str(e)
+            new_max_tokens = parse_and_calculate_max_tokens(error_message)
+
+            if new_max_tokens is not None:
+                logger.warning(f"Context length error: reducing max_tokens to {new_max_tokens}.")
+                retry_sampling_args = deepcopy(sampling_args)
+                retry_sampling_args["max_tokens"] = new_max_tokens
+                return await generate_rollout(client, env, model_name, example, retry_sampling_args)
+            raise
```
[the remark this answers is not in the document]
mikasenghaas: @rasdani chat is this true?

src/prime_rl/utils/vf.py
```diff
@@ -31,6 +31,22 @@ async def generate_group(
...
+    model_name: str,
+    example: dict,
+    sampling_args: dict,
+) -> vf.State:
+    """Asynchronously generate and score a single rollout."""
+    semaphore = await get_semaphore()
+    rollout_input = vf.RolloutInput(**example)
+
+    state = await env.run_rollout(semaphore, rollout_input, client, model_name, sampling_args)
+    await env.rubric.score_rollout(state, score_sem=semaphore)
```
willccbb, line 46: We should wait to score until we have full groups, as reward functions are allowed to be group-level

willccbb (dismissed):

src/prime_rl/orchestrator/config.py
rasdani: fine by we since we have this assertion in vf `score_rollout`
https://github.com/PrimeIntellect-ai/verifiers/blob/eec990c3d5dd69e5395d37f91ead9ff657bd22db/verifiers/rubrics/rubric.py#L214

src/prime_rl/eval/config.py
```diff
@@ -81,3 +95,40 @@ def validate_eval_base(self):
...
+    def validate_resume_stream(self):
+        """Enforce per_rollout when resume_path or save.stream is enabled, as both require per-rollout scheduling."""
+        if self.resume_path is not None:
+            if not self.save.stream:
+                raise ValueError(
+                    "resume_path requires save.stream. Streaming saves are required for resume functionality."
+                )
+        return self
+
+    @model_validator(mode="after")
```
mikasenghaas, line 109: oh, we should enable this functionality sooner rather than later but fine to table for now

src/prime_rl/orchestrator/config.py
```diff
@@ -245,10 +294,26 @@ class EvalConfig(BaseConfig):
...
     rollouts_per_example: Annotated[
         int, Field(ge=1, description="Number of samples to generate per example for each environment.")
     ] = 1
+    reasoning_field: Annotated[
+        str,
+        Field(
+            description="The field in the raw model response that contains the reasoning content. Defaults to 'reasoning_content', which is the default for vLLM when serving a model with a reasoning parser. Other APIs (e.g. DeepSeek, GLM, etc.) may use different field names.",
+        ),
+    ] = "reasoning_content"
+    per_rollout: Annotated[
```
mikasenghaas, line 311: should be false by default because its the safer behavior. opt-in for more speed/ more frequent saving, _if_ user knows that no per-group information is needed for scoring

src/prime_rl/eval/config.py
rasdani: will do in follow-up pr

samsja (approved):

willccbb (approved):

PR #1393, opened 2025-12-06
[Fix] patch vllm statslogger to be ok with lora dp and helm chart to support multi-client for inference replicas

Jackmin801:
> [!NOTE]
> Builds INFERENCE_URL as a comma-separated list of all inference replicas and applies a vLLM Prometheus logger monkeypatch to enable LoRA in DP mode.
> 
> - **Kubernetes/Helm**:
>   - `INFERENCE_URL` for `orchestrator` and `trainer` now expands to all inference replicas (comma-separated), instead of a single pod URL, using values from `inference.service.port` and `inference.replicas` in `templates/deployment.yaml`.
>   - Example `reverse-text.yaml` updated to pass `--client.base-url $INFERENCE_URL` (no JSON quoting); `values.yaml` docs updated accordingly.
> - **Inference Server**:
>   - New `src/prime_rl/inference/patches.py` monkeypatches `vllm.v1.metrics.loggers.PrometheusStatLogger` to bypass LoRA DP mode check and reinitialize LoRA metrics.
>   - `src/prime_rl/inference/vllm/server.py` imports and applies the monkeypatch at startup.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 642bed2a0fb444f9f4cb0f4b190dd9c70ebb2e7c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/k8s/prime-rl/examples/reverse-text.yaml
+++ b/k8s/prime-rl/examples/reverse-text.yaml
@@ -11,7 +11,7 @@ image:
 orchestrator:
   enabled: true
   autoStart: true
-  command: "uv run orchestrator @ /app/examples/reverse_text/rl/orch.toml --output-dir /data/outputs --client.base-url '[\"'$INFERENCE_URL'\"]' ; sleep infinity"
+  command: "uv run orchestrator @ /app/examples/reverse_text/rl/orch.toml --output-dir /data/outputs --client.base-url $INFERENCE_URL ; sleep infinity"
   resources:
     requests:
       memory: "1Gi"
--- a/src/prime_rl/rl.py
+++ b/src/prime_rl/rl.py
@@ -131,6 +131,10 @@ class RLConfig(BaseSettings):
     inference_gpu_ids: Annotated[list[int], Field(description="The GPU IDs to use for inference.")] = [0]
     trainer_gpu_ids: Annotated[list[int], Field(description="The GPU IDs to use for trainer.")] = [1]
 
+    # Computed at runtime - number of inference servers to spawn
+    _num_inference_servers: int = 1
+    _inference_ports: list[int] = []
+
     ### Shared configurations
 
     output_dir: Annotated[
@@ -193,11 +197,18 @@ class RLConfig(BaseSettings):
 
     @model_validator(mode="after")
     def auto_setup_dp(self):
-        if self.inference and len(self.inference_gpu_ids) != self.inference.parallel.dp * self.inference.parallel.tp:
+        """
+        Instead of using VLLM's internal DP, we spawn multiple independent inference servers.
+        Each server has DP=1, and we calculate how many servers to spawn based on the GPU count.
+        """
+        if self.inference:
             assert len(self.inference_gpu_ids) % self.inference.parallel.tp == 0, (
                 "Number of inference GPUs must be divisible by the tensor parallel size"
             )
-            self.inference.parallel.dp = len(self.inference_gpu_ids) // self.inference.parallel.tp
+            # Calculate number of servers to spawn (this is our "external DP")
+            self._num_inference_servers = len(self.inference_gpu_ids) // self.inference.parallel.tp
+            # Force VLLM DP to 1 - we handle parallelism by spawning multiple servers
+            self.inference.parallel.dp = 1
         return self
 
     @model_validator(mode="after")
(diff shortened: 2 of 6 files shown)


src/prime_rl/rl.py
```diff
@@ -463,35 +487,55 @@ def rl(config: RLConfig):
...
-            monitor_thread.start()
-            monitor_threads.append(monitor_thread)
+                logger.debug(f"Inference start command: {' '.join(inference_cmd)}")
+
+                # If we don't log stdout, the server hangs
+                log_suffix = f".{server_idx}" if num_servers > 1 else ""
+                with open(log_dir / f"inference{log_suffix}.stdout", "w") as log_file:
+                    inference_process = Popen(
+                        inference_cmd,
+                        env={**os.environ, "CUDA_VISIBLE_DEVICES": ",".join(map(str, server_gpu_ids))},
```
samsja: do the inferences log to the same file ?

src/prime_rl/rl.py
```diff
@@ -428,6 +440,18 @@ def rl(config: RLConfig):
     logger.info("Starting RL run")
     logger.debug(f"RL start command: {' '.join(start_command)}")
 
+    # Pre-allocate ports for all inference servers and configure orchestrator client
+    if config.inference:
+        base_port = config.inference.server.port
+        config._inference_ports = [base_port + i for i in range(config._num_inference_servers)]
```
eexwhyzee: this will prob also require some changes to the k8s helm chart since only the single inference api port (8000) is exposed by the k8s svc for the inference deployment: https://github.com/PrimeIntellect-ai/prime-rl/blob/main/k8s/prime-rl/templates/service.yaml#L66

src/prime_rl/rl.py
```diff
@@ -131,6 +131,10 @@ class RLConfig(BaseSettings):
     inference_gpu_ids: Annotated[list[int], Field(description="The GPU IDs to use for inference.")] = [0]
     trainer_gpu_ids: Annotated[list[int], Field(description="The GPU IDs to use for trainer.")] = [1]
 
+    # Computed at runtime - number of inference servers to spawn
+    _num_inference_servers: int = 1
```
mikasenghaas: i dont think we should compute runtime args in config?

src/prime_rl/rl.py
Jackmin801: eh each inference server logs to inference.x.stdout where x is the rank

src/prime_rl/rl.py
Jackmin801: good catch! fixed the helm chart to build an $INFERENCE_URL that includes all inference pods in the statefulset

Jackmin801:
Orchestrator pod now correctly picks up each inference replica
image

k8s/prime-rl/examples/reverse-text.yaml
```diff
@@ -11,7 +11,7 @@ image:
 orchestrator:
   enabled: true
   autoStart: true
-  command: "uv run orchestrator @ /app/examples/reverse_text/rl/orch.toml --output-dir /data/outputs --client.base-url '[\"'$INFERENCE_URL'\"]' ; sleep infinity"
+  command: "uv run orchestrator @ /app/examples/reverse_text/rl/orch.toml --output-dir /data/outputs --client.base-url $INFERENCE_URL ; sleep infinity"
```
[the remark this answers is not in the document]
Jackmin801: it expects comma separated strings so its actually correct to not json it

src/prime_rl/rl.py
mikasenghaas: also not a fan of "private" attr in configs, we can just compute this outside of pydantic or is there any argument against?

mikasenghaas (commented):
im wondering if we should block dp from our inference config all together then if started manually? also, to stresstest this, let's make both wordle and wiki-search example full node again so they run faster in CI and we test this feature in the nightly tests as well

src/prime_rl/rl.py
eexwhyzee: nice nice

also i didn't realize initially that this port change was specifically in the `rl` entrypoint (single node usage?)...but just so i understand the usage correctly:
1. single node usage (via `rl` entrypoint): spawn child processes to run multiple vllm servers using different ports
2. multi node usage (via k8s): increase replica count of the inference statefulset, running multiple vllm servers at the pod level

is that right?

src/prime_rl/rl.py
Jackmin801: we're gonna just patch the not implemented error

Jackmin801:
Tested that monkeypatch works and allows dp

```toml
inference_gpu_ids = [0,1]
trainer_gpu_ids = [2]
max_steps = 20

[model]
name = "PrimeIntellect/Qwen3-0.6B-Reverse-Text-SFT"

[trainer.model.experimental.lora]
rank = 8

[orchestrator]
batch_size = 128
rollouts_per_example = 16
seq_len = 2048

[orchestrator.sampling]
max_tokens = 128

[[orchestrator.env]]
id = "reverse-text"

[trainer.optim]
lr = 1e-4

[inference]

[inference.parallel]
dp = 2

[log]
level = "debug"
```

samsja (approved):
LGTM

PR #1408, opened 2025-12-10
round up max_lora_rank to valid vLLM values

kalomaze:
vLLM only accepts specific values for max_lora_rank: (8, 16, 32, 64, 128, 256, 320, 512).
When using LoRA ranks like 4 that aren't in this list, vLLM would reject them

This adds a validator that automatically rounds up to the minimum valid value that can serve adapters of the requested rank (e.g., rank 4 -> max_lora_rank 8). A warning is emitted when rounding occurs.

---

> [!NOTE]
> Automatically rounds configured `max_lora_rank` up to the nearest supported vLLM rank and errors if it exceeds the maximum.
> 
> - **Config (`src/prime_rl/inference/config.py`)**
>   - **LoRA rank handling**:
>     - Add `VALID_VLLM_LORA_RANKS` constant defining supported ranks.
>     - Add `InferenceConfig.round_up_max_lora_rank` validator to round `max_lora_rank` up to the nearest valid vLLM value; raises if above max.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9fe47b5bb2a1be11e0dae9a0af4a487ece4a73d0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/inference/config.py
+++ b/src/prime_rl/inference/config.py
@@ -1,4 +1,5 @@
 import os
+import warnings
 from argparse import Namespace
 from typing import Annotated, Literal
 
@@ -107,6 +108,10 @@ class WeightBroadcastConfig(BaseSettings):
     )
 
 
+# Valid vLLM max_lora_rank values (from vllm/config/lora.py)
+VALID_VLLM_LORA_RANKS = (8, 16, 32, 64, 128, 256, 320, 512)
+
+
 class InferenceConfig(BaseSettings):
     """Configures inference."""
 
@@ -161,6 +166,30 @@ class InferenceConfig(BaseSettings):
     def auto_setup_dynamic_lora_updating(self):
         if self.enable_lora:
             os.environ["VLLM_ALLOW_RUNTIME_LORA_UPDATING"] = "True"
+        return self
+
+    @model_validator(mode="after")
+    def round_up_max_lora_rank(self):
+        """Round up max_lora_rank to the nearest valid vLLM value.
+
+        vLLM only accepts specific values for max_lora_rank: (1, 8, 16, 32, 64, 128, 256, 320, 512).
+        This validator ensures that any configured rank is rounded up to the minimum valid value
+        that can serve adapters of the requested rank.
+        """
+        if self.max_lora_rank is not None:
+            original_rank = self.max_lora_rank
+            for valid_rank in VALID_VLLM_LORA_RANKS:
+                if valid_rank >= self.max_lora_rank:
+                    self.max_lora_rank = valid_rank
+                    break
+            else:
+                raise ValueError(
+                    f"max_lora_rank={original_rank} exceeds vLLM maximum of {VALID_VLLM_LORA_RANKS[-1]}"
+                )
+            if self.max_lora_rank != original_rank:
+                warnings.warn(
+                    f"vLLM max_lora_rank {original_rank} -> {self.max_lora_rank} (serving allocation only, adapter rank unchanged)"
+                )
         return self
 
     def to_vllm(self) -> Namespace:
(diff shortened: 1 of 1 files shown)


src/prime_rl/inference/config.py
```diff
@@ -107,6 +108,10 @@ class WeightBroadcastConfig(BaseSettings):
     )
 
 
+# Valid vLLM max_lora_rank values
+VALID_VLLM_LORA_RANKS = (1, 8, 16, 32, 64, 128, 256, 320, 512)
```
mikasenghaas: can we get this from vllm internals somehow? just so we dont risk diverging bc of hardcode

src/prime_rl/inference/config.py
```diff
@@ -163,6 +168,30 @@ def auto_setup_dynamic_lora_updating(self):
...
+                    break
+            else:
+                raise ValueError(
+                    f"max_lora_rank={original_rank} exceeds vLLM maximum of {VALID_VLLM_LORA_RANKS[-1]}"
+                )
+            if self.max_lora_rank != original_rank:
+                warnings.warn(
+                    f"vLLM max_lora_rank {original_rank} -> {self.max_lora_rank} (serving allocation only, adapter rank unchanged)"
+                )
+        return self
```
mikasenghaas: not sure we need this warning

src/prime_rl/inference/config.py
```diff
@@ -163,6 +168,30 @@ def auto_setup_dynamic_lora_updating(self):
...
+        that can serve adapters of the requested rank.
+        """
+        if self.max_lora_rank is not None:
+            original_rank = self.max_lora_rank
+            for valid_rank in VALID_VLLM_LORA_RANKS:
+                if valid_rank >= self.max_lora_rank:
+                    self.max_lora_rank = valid_rank
+                    break
+            else:
+                raise ValueError(
```
mikasenghaas, line 186: is this a for..else??

src/prime_rl/inference/config.py
kalomaze: was just added in case someone saw the max serving size in vllm logged somewhere and was confused bc it's higher than what they specified but can remove if frivolous

src/prime_rl/inference/config.py
kalomaze: it's a local variable that sets the allowed ranks, not sure how we would be able to borrow it in a clean way :/

src/prime_rl/inference/config.py
kalomaze: we could instead do a next() variant ie `next(r for r in VALID if r >= original)` ?

src/prime_rl/inference/config.py
kalomaze: so i checked upstream, it looks like there is a cleaner way we could do this if we were on a newer vllm but we're not on the latest ver, and migrating to that is its own beast lololo

mikasenghaas: ah, all good then. maybe put a comment that once we migrate to new vllm, we can check to make this cleaner

src/prime_rl/inference/config.py
mikasenghaas: yea lets remove, i dont think we need to warn users here

src/prime_rl/inference/config.py
mikasenghaas: ah just looked this up, didn't know the for..else behavior. looks good then

src/prime_rl/inference/config.py
kalomaze: done

src/prime_rl/inference/config.py
kalomaze: removed

mikasenghaas (approved):

PR #1419, opened 2025-12-11
Move packing to trainer and support run directories

Jackmin801:
> [!NOTE]
> Shift batching/packing from orchestrator to trainer via a new transport layer and run_* directories, updating configs, weight broadcast, and tests accordingly.
> 
> - **Core architecture**:
>   - Introduce `prime_rl.transport` (filesystem sender/receiver) and msgspec types (`TrainingSample`, `TrainingBatch`, `MicroBatch`).
>   - Move batching/packing to trainer: new `trainer/rl/packer.py` and `trainer/batch.py`; orchestrator now sends `TrainingBatch` instead of writing micro-batches; remove `orchestrator/batch.py` and `orchestrator/types.py`.
>   - Add `trainer/runs.py` to manage concurrent `run_*` directories and per-run progress; new `utils/pathing.py` for common paths/waits.
> - **Orchestrator**:
>   - Save orch config to `outputs/run_*/configs/orch.toml` and send training batches via `setup_training_batch_sender`.
>   - Update trajectories to use `TrainingSample` with boolean masks.
> - **Trainer**:
>   - Data pipeline reads `TrainingBatch` via transport, packs into micro-batches, and distributes per DP rank.
>   - Weight broadcast (filesystem/NCCL) updated to handle multi-runs and STABLE notifications; optional cleanup per-run.
>   - Move `Progress` to `trainer/runs`; integrate `seq_len` from `trainer.model`.
> - **Configs & validation**:
>   - Add `rollout_transport` to orchestrator/trainer; default `orchestrator.output_dir` to `outputs/run_default`; RL `output_dir` maps to parent with per-run subdir; add `trainer.model.seq_len`.
>   - Validation updated to compare trainer output dir with orchestrator parent; RL auto-setup adjusted; inference adds `max_loras` and vLLM arg mapping.
> - **CI/Docs/Tests**:
>   - CI: run RL LoRA test separately for stability.
>   - Docs: change RL benchmarking flag to `--model.seq-len`.
>   - Tests updated for new types/paths and add `tests/unit/train/test_runs.py`.
>   - Gitignore: ignore `run_*/` and tidy `.idea`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f5c4b9d704e1ace4fbe8d9bd12a78da7a9032e73. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/transport/__init__.py
```diff
@@ -0,0 +1,52 @@
+from pathlib import Path
+
```
samsja, line 2: can we name the folder data_transport

src/prime_rl/orchestrator/config.py
```diff
@@ -441,6 +442,8 @@ class OrchestratorConfig(BaseSettings):
         FileSystemWeightBroadcastConfig()
     )
 
+    transport: Annotated[TransportConfigType, Field(discriminator="type")] = FileSystemTransportConfig()
```
```suggestion
    data_transport: Annotated[TransportConfigType, Field(discriminator="type")] = FileSystemTransportConfig()
```

src/prime_rl/transport/__init__.py
Jackmin801: I think not making the folder specific to rollout / training batch data but transports in general might be nicer

src/prime_rl/orchestrator/config.py
Jackmin801: maybe rollout transport? data seems a bit general

src/prime_rl/trainer/rl/broadcast/nccl.py
```diff
@@ -155,6 +158,29 @@ def broadcast_weights(self, model: nn.Module, step: int, adapter_only: bool = Fa
         self.logger.debug("Starting broadcasting weights to inference engine via NCCL")
         start_time = time.perf_counter()
         if self.world.is_master:
-            self.notify_orchestrator(step)
+            self.notify_orchestrator()
         self.nccl_broadcast_sender.broadcast_weights(model, step)
```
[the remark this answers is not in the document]
Jackmin801: its just the way it works I think

src/prime_rl/transport/types.py
```diff
@@ -0,0 +1,35 @@
+import msgspec
+
+
+# Orchestrator -> Packer
+class TrainingExample(msgspec.Struct, array_like=True, gc=False, omit_defaults=True):
```
samsja:
```suggestion
class TrainingSample(msgspec.Struct, array_like=True, gc=False, omit_defaults=True):
```

I feel example is misleading here. I rpopose to rename to sample.

Jackmin801: yea I think that makes sense

src/prime_rl/orchestrator/trajectories.py
```diff
@@ -19,11 +19,11 @@ def interleave_rollout(state: vf.State) -> list[TrainingExample]:
     # Initialize the rollout with prompt and completion from first trajectory step
     trajectory = state["trajectory"]
     first_step = trajectory[0]
-    interleaved_rollout = TrainingExample(
+    interleaved_rollout = TrainingSample(
         prompt_ids=deepcopy(first_step["tokens"]["prompt_ids"]),
-        prompt_mask=deepcopy(first_step["tokens"]["prompt_mask"]),
+        prompt_mask=[i != 0 for i in first_step["tokens"]["prompt_mask"]],
```
mikasenghaas: lol is this to have bool type? if so, can we just call `bool` on it, seems a bit nicer. or we just use the `int` type on the orch side and only turn into bool on trainer when packing into tensor? (then orch type would match vf which i dont dislike as well)

src/prime_rl/trainer/rl/broadcast/nccl.py
```diff
@@ -155,6 +158,29 @@ def broadcast_weights(self, model: nn.Module, step: int, adapter_only: bool = Fa
...
+                if not self.runs.ready_to_update[idx]:
+                    continue
+
+                try:
+                    save_dir = get_step_path(
+                        get_broadcast_dir(self.runs.get_run_dir(idx)), self.runs.progress[idx].step
+                    )
+                    save_dir.mkdir(parents=True, exist_ok=True)
+
+                    # Notify the orchestrator at the end of step to signal that it is safe to load weights from shared filesystem
```
mikasenghaas: adjust comment

src/prime_rl/trainer/rl/broadcast/base.py
```diff
@@ -5,23 +5,14 @@
...
         self.logger = get_logger()
         self.output_dir = output_dir
-        self.broadcast_dir = get_broadcast_dir(output_dir)
         self.lora_config = lora_config
 
     @abstractmethod
     def broadcast_weights(self, model: nn.Module, step: int):
         pass
-
-    def notify_orchestrator(self, step: int):
```
mikasenghaas, line 22: why does the this method move into the fs and nccl broadcast classes? is it diff for the diff broadcast types now?

src/prime_rl/transport/__init__.py
mikasenghaas, line 1: wondering abt the position of this module. rn we have 1 module for each of the components of rl run + then shared utils. so by that logic ig should live in `utils.transport`? but also we have been violating this principle alr with e.g. the weight broadcast stuff which is shared logic between trainer and inference which lives on rl trainer rn.. maybe fine to leave it here and cleanup when we rearrange the modules

src/prime_rl/orchestrator/config.py
```diff
@@ -453,7 +456,7 @@ class OrchestratorConfig(BaseSettings):
         Field(
             description="Directory to write outputs to. Will be populated with checkpoints, weights, rollouts and logs as subdirectories. Should be set to a persistent directory with enough disk space. This value should be distinct across experiments running on a single node. See the README for more details."
         ),
-    ] = Path("outputs")
+    ] = Path("outputs/run_default")
```
mikasenghaas, line 459: does this mean that trainer writes into `outputs/trainer.log` and orch writes into `outputs/run_default/orchestrator.log`? if this is the case do we also need to update tmux.sh or am i missing smth?

mikasenghaas (commented):
super well done, love the code structure. we should defo get a multi-run example into our tests. do you think it would make sense to do 2 reverse text or smth? btw,could we run this in regular ci or you think better in nightly tests?

oh also, doing multi-run requires lora right? do we assert this somewhere/ fail loudly if we detect that a config violates this

src/prime_rl/orchestrator/config.py
Jackmin801: Ah yup. good catch!

src/prime_rl/transport/__init__.py
Jackmin801: yea imo utils should be lightweight stuff else every component is util which I guess could be what we decide to do. yea maybe separate PR that maybe also moves broadcast

src/prime_rl/orchestrator/trajectories.py
Jackmin801: I dont think we should b passing around an int for a value that really only accepts bool. bool(i) I guess is ok

src/prime_rl/trainer/rl/broadcast/base.py
Jackmin801: its not meant to be used outside of the class and fs and nccl seem to mean different things with it so I felt like it shouldnt belong in a base class that is defining the general contract of how the Broadcasters should be used

src/prime_rl/orchestrator/config.py
Jackmin801: Ah actually tmux.sh is tailing orchestrator.stdout which is still in outputs/logs so its fine

src/prime_rl/trainer/rl/packer.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def pack(self):
+        training_batches: dict[int, TrainingBatch] = self.get_batch()
+        start_time = time.time()
+        while not self.has_enough_tokens(training_batches):
+            if time.time() - start_time > TIMEOUT_SECONDS and training_batches:
+                self.logger.warning("Timeout waiting for enough tokens to pack")
+                break
+            time.sleep(1)
+            training_batches = self.get_batch()
```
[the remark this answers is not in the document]
Jackmin801: this is intended

src/prime_rl/trainer/rl/train.py
```diff
@@ -153,9 +165,13 @@ def train(config: RLTrainerConfig):
             # Clean up old broadcast directories (unless at ckpt interval if using filesystem weight broadcast)
             ckpt_interval = config.ckpt and config.ckpt.interval
             interval_to_keep = ckpt_interval if config.weight_broadcast.type == "filesystem" else None
-            maybe_clean(weight_broadcast.broadcast_dir, progress.step, config.max_async_level, interval_to_keep)
+            if config.weight_broadcast.type == "filesystem":
+                weight_broadcast.maybe_clean(config.max_async_level, interval_to_keep)
```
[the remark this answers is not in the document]
Jackmin801: should be fine

.github/workflows/gpu_tests.yaml
```diff
@@ -89,4 +89,6 @@ jobs:
           GITHUB_REF_NAME: ${{ github.ref_name }}
           GITHUB_HEAD_REF: ${{ github.head_ref }}
           PYTEST_OUTPUT_DIR: /tmp/outputs
-        run: uv run pytest tests/integration -m gpu
+        run: |
```
mikasenghaas, line 92: why are we separating those out again? each test should have its own vllm server now no?

src/prime_rl/trainer/rl/broadcast/base.py
mikasenghaas: well i think they mean the same but one notifies that its about to send, while the other notifies that its finished "sending" (writing to disk), but yea prolly fine to keep it in child classes

src/prime_rl/trainer/rl/data.py
```diff
@@ -55,18 +71,33 @@ def _get_micro_batch(self) -> MicroBatch:
...
+                tokenizer=tokenizer,
+                config=config,
+                start_step=start_step,
+            )
+
+        self.receiver: MicroBatchReceiver = setup_micro_batch_receiver(output_dir, self.world.rank, start_step, config)
 
     def wait_for_batch(self) -> None:
-        sync_wait_for_path(self.get_rollout_path())
+        if self.world.is_master:
```
mikasenghaas, line 97: we pack and then wait?

mikasenghaas: would be nice in general to maybe have a descriptive docstring on the dataloader that explains the io that happens here and the interaction with the packer

.github/workflows/gpu_tests.yaml
Jackmin801: the test seems to fail in ci without separating. its not clear to me why. figured better to fix later as its fine to separate them

src/prime_rl/trainer/rl/data.py
Jackmin801: yep. rank 0 isnt special to the packer so rank 0 also waits for file. we could make packer return 0s batch but i dont imagine it speeds it up that much

Jackmin801: yea docstring sounds good. will do

.github/workflows/gpu_tests.yaml
mikasenghaas: ah ok good to know

samsja (approved):
LFGTM SIR

PR #1420, opened 2025-12-12
Include system prompt in logged rollouts

snimu:
Add "prompt" column back to wandb logging. This now includes the system prompt even if it comes from a chat template instead of an environment, which is important for debugging.

---

> [!NOTE]
> Reworks W&B sample logging to decode concatenated prompt+completion token IDs, logging the full message text (including system prompt) and raw input ID lists.
> 
> - **W&B Monitoring**:
>   - **Sample logging**:
>     - Derives `messages` by concatenating `tokens.prompt_ids` and `tokens.completion_ids` and decoding via tokenizer.
>     - Logs `input_ids` as the raw concatenated ID list (`full_ids`).
>     - Removes chat-template-based construction (`apply_chat_template`) of messages and input IDs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 699bac6432b2d177c249ee930c3e34b8c13a2c55. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/utils/monitor.py
+++ b/src/prime_rl/utils/monitor.py
@@ -54,7 +54,7 @@ class WandbMonitor:
         if config is not None and isinstance(config, WandbWithExtrasConfig) and config.log_extras:
             if config.log_extras.samples:
                 self.last_log_samples_step = -1
-                self.samples_cols = ["step", "example_id", "messages", "input_ids", "reward"]
+                self.samples_cols = ["step", "example_id", "prompt", "messages", "input_ids", "reward"]
                 self.samples_table = wandb.Table(
                     columns=self.samples_cols,
                     log_mode="INCREMENTAL",
@@ -97,9 +97,16 @@ class WandbMonitor:
         start_time = time.perf_counter()
         for rollout in rollouts:
             messages = rollout["trajectory"][-1]["prompt"] + rollout["trajectory"][-1]["completion"]
+            tokens = rollout["trajectory"][-1]["tokens"]
+            if tokens is not None and "prompt_ids" in tokens:
+                prompt_text = self.tokenizer.decode(tokens["prompt_ids"])
+            else:
+                prompt = rollout["trajectory"][-1]["prompt"]
+                prompt_text = self.tokenizer.apply_chat_template(prompt, tokenize=False)
             sample = {
                 "step": step,
                 "example_id": rollout["example_id"],
+                "prompt": prompt_text,
                 "messages": self.tokenizer.apply_chat_template(messages, tokenize=False),
                 "input_ids": str(self.tokenizer.apply_chat_template(messages)),
                 "reward": rollout["reward"],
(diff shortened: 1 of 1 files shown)


src/prime_rl/utils/monitor.py
```diff
@@ -97,9 +97,16 @@ def log_samples(self, rollouts: list[vf.State], step: int) -> None:
         start_time = time.perf_counter()
         for rollout in rollouts:
             messages = rollout["trajectory"][-1]["prompt"] + rollout["trajectory"][-1]["completion"]
+            tokens = rollout["trajectory"][-1]["tokens"]
+            if tokens is not None and "prompt_ids" in tokens:
+                prompt_text = self.tokenizer.decode(tokens["prompt_ids"])
```
mikasenghaas: the last trajectory step's prompt tokens is all the prompt+completions up to the last turn. if we want to bring back differentiating between prompt/ completion we would need to get prompt tokens from the first trajectory step

personally, would just keep messages flat, so:
- get messages from `rollout["trajectory"][-1]["prompt"] + rollout["trajectory"][-1]["completion"]`, as before
- get tokens from `rollout["trajectory"][-1]["tokens"]["prompt_ids"] + rollout["trajectory"][-1]["tokens"]["completion_ids"]`
both of these are guaranteed to be there. then just call decode on all all tokens (this should include the sys prompt)

src/prime_rl/utils/monitor.py
```diff
@@ -97,9 +97,16 @@ def log_samples(self, rollouts: list[vf.State], step: int) -> None:
...
+            else:
+                prompt = rollout["trajectory"][-1]["prompt"]
+                prompt_text = self.tokenizer.apply_chat_template(prompt, tokenize=False)
```
mikasenghaas: if we do what i suggest above we dont need this fallback of applying chat template

snimu: I removed the "prompt" column and instead create the "messages" column directly from the token ids. This now successfully includes the system prompt.

I also removed the fallback, to simplify the code.

src/prime_rl/utils/monitor.py
```diff
@@ -96,12 +96,14 @@ def log_samples(self, rollouts: list[vf.State], step: int) -> None:
         self.logger.info(f"Logging samples to W&B table at step {step}")
         start_time = time.perf_counter()
         for rollout in rollouts:
-            messages = rollout["trajectory"][-1]["prompt"] + rollout["trajectory"][-1]["completion"]
+            tokens = rollout["trajectory"][-1]["tokens"]
+            full_ids = tokens["prompt_ids"] + tokens["completion_ids"]
```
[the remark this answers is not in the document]
mikasenghaas: yes this is guaranteed, rl training wont work if tokens is not present
the fallbacks in the other functions bugbot mentions are also used by eval and synthesize entrypoint which may hit apis, so then they're not present, hence the fallbacks

mikasenghaas (approved):

PR #1425, opened 2025-12-12
Working multi dir

Jackmin801:
This is just for jannik to build stabler working images off of


src/prime_rl/trainer/config.py
```diff
@@ -135,6 +135,8 @@ class ModelConfig(BaseConfig):
         ),
     ] = "Qwen/Qwen3-0.6B"
 
+    seq_len: Annotated[int, Field(description="The sequence length to use for the model.")] = 2048
```
samsja, line 138: should we move this to data ?

S1ro1: I like this here tbf, will fix the ugly autosetup/check for pad_to_multiple_of with cp.

PR #1430, opened 2025-12-16
add multiple off

samsja:
# Checkpoint

add an option to keep last ckpt as a multiple of.

Example

```
uv run torchrun --local-ranks-filter 0 --nproc_per_node=2 src/prime_rl/trainer/sft/train.py @ configs/debug/sft/train.toml  --max-steps 5  --ckpt.interval 1  --ckpt.keep 1 --ckpt.keep_multiple_of 2
```

will keep
```
❯ ls outputs/checkpoints
step_2  step_4  step_5
```

### DIFF
--- a/src/prime_rl/trainer/ckpt.py
+++ b/src/prime_rl/trainer/ckpt.py
@@ -210,28 +210,25 @@ class CheckpointManager:
         if self.config.keep is None:
             return
 
-        # Get all the checkpoint steps to delete
+        # Compute which steps to keep: recent ones and multiples of keep_interval
         assert list(self.ckpt_steps) == sorted(self.ckpt_steps)
-        ckpt_steps_to_delete = self.ckpt_steps[: -self.config.keep]
-
-        # Filter out checkpoints that are multiples of keep_multiple_of
-        if self.config.keep_interval is not None:
-            ckpt_steps_to_delete = [step for step in ckpt_steps_to_delete if step % self.config.keep_interval != 0]
-
-        for ckpt_step in ckpt_steps_to_delete:
-            trainer_ckpt_path = self.get_ckpt_path(ckpt_step)
-            ckpt_path = trainer_ckpt_path.parent
-            if ckpt_path.exists():
-                self.logger.debug(f"Removing past checkpoint for step {ckpt_step} ({ckpt_path})")
-                shutil.rmtree(ckpt_path)
-
-        # Update checkpoint steps - keep recent ones and multiples
         kept_steps = self.ckpt_steps[-self.config.keep :]
         if self.config.keep_interval is not None:
             kept_steps.extend(
                 [step for step in self.ckpt_steps[: -self.config.keep] if step % self.config.keep_interval == 0]
             )
             kept_steps = sorted(set(kept_steps))
+
+        # Delete steps that are not kept
+        kept_set = set(kept_steps)
+        for ckpt_step in self.ckpt_steps:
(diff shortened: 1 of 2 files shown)


Jackmin801 (approved):
yee lgtm

src/prime_rl/trainer/ckpt.py
```diff
@@ -213,15 +213,26 @@ def maybe_clean(self) -> None:
...
         for ckpt_step in ckpt_steps_to_delete:
             trainer_ckpt_path = self.get_ckpt_path(ckpt_step)
             ckpt_path = trainer_ckpt_path.parent
             if ckpt_path.exists():
                 self.logger.debug(f"Removing past checkpoint for step {ckpt_step} ({ckpt_path})")
                 shutil.rmtree(ckpt_path)
 
-        # Update checkpoint steps
-        self.ckpt_steps = self.ckpt_steps[-self.config.keep :]
+        # Update checkpoint steps - keep recent ones and multiples
```
mikasenghaas: codex ah comment

src/prime_rl/trainer/config.py
```diff
@@ -425,6 +425,14 @@ class CheckpointConfig(BaseConfig):
         ),
     ] = None
 
+    keep_multiple_of: Annotated[
```
mikasenghaas: would prefer naming this keep interval, as this is what we use in the rest of repo

src/prime_rl/trainer/ckpt.py
```diff
@@ -328,14 +341,23 @@ def maybe_clean(self) -> None:
         # Get all the checkpoint steps to delete
         assert list(self.ckpt_steps) == sorted(self.ckpt_steps)
         ckpt_steps_to_delete = self.ckpt_steps[: -self.keep]
+
+        # Filter out checkpoints that are multiples of keep_multiple_of
```
mikasenghaas: i dont like that we have interval to keep logic in two places. right now we first filter those steps before deleting. and then we select those to keep. these are inverse operations so one time should be enough?

src/prime_rl/trainer/ckpt.py
```diff
@@ -210,18 +210,26 @@ def maybe_clean(self) -> None:
...
+        # Compute which steps to keep: recent ones and multiples of keep_interval
         assert list(self.ckpt_steps) == sorted(self.ckpt_steps)
-        ckpt_steps_to_delete = self.ckpt_steps[: -self.config.keep]
-        for ckpt_step in ckpt_steps_to_delete:
-            trainer_ckpt_path = self.get_ckpt_path(ckpt_step)
-            ckpt_path = trainer_ckpt_path.parent
-            if ckpt_path.exists():
-                self.logger.debug(f"Removing past checkpoint for step {ckpt_step} ({ckpt_path})")
-                shutil.rmtree(ckpt_path)
+        kept_steps = self.ckpt_steps[-self.config.keep :]
```
mikasenghaas, line 215: oh so we keep the last `keep` checkpoints _and_ all multiples of `keep_interval`? this would mean we violate the assertion to only keep `keep` checkpoints, e.g. we may have infinite ckpts on disks. im not sure this is desired behavior given that this arg is primarily used to not run out of disk space?

samsja: yes this is the goal, I was inclide of renaming keep to keep_last. 

Imo more important that we don't delete old checkpoint by mistake

PR #1438, opened 2025-12-17
Feat: remove config.load_using_meta and default to it now + remove weight double loading

S1ro1:
Give feedback on this, can cleanup if agreed on

Proposes to remove config.load_using_meta, defaulting to True and only loading to cpu if not possible to use meta device (some buffers cannot be reinitialized)

Qwen3-30B-A3B and Intellect 3 (30 layers as I can't fit more on 1 node)

image
image

image

---

> [!NOTE]
> Default model init to meta with CPU fallback, remove `model.load_using_meta`, and make trainers init checkpoints earlier to optionally defer weight loading.
> 
> - **Model loading**:
>   - Default to loading models on `meta` with CPU fallback if `can_reinit_empty_buffers` fails; remove `ModelConfig.load_using_meta` and its validator.
>   - Rename/replace `can_load_dcp_from_hf` with `can_reinit_empty_buffers`; use `fix_model_post_empty`/`init_buffers_post_meta` after `to_empty`.
>   - `setup_model(config, parallel_dims, loading_from_checkpoint_later)` now supports deferring weight load when resuming from checkpoints and otherwise loads via HF DCP.
> - **Trainers (RL/SFT)**:
>   - Initialize checkpoint managers before model construction; resolve `checkpoint_step` and pass `loading_from_checkpoint_later` to `setup_model`.
>   - Remove duplicated checkpoint-step resolution and post-optimizer scheduler fixes are handled accordingly.
> - **Config & debug**:
>   - Drop `load_using_meta` option from `ModelConfig` and the random-init/meta constraint validator.
>   - Minor config file tidy in `configs/debug/moe/sft/train.toml` (sequence length retained).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fc657f4d30fde64bb9b71c5409a43f665eb8ec85. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -330,15 +330,14 @@ def setup_model(
         )
     logger = get_logger()
 
+    # 1. We load to meta device by default
+    model = get_model(config, device=torch.device("meta"), dtype=DTYPE_MAP[config.optimization_dtype])
     possible_to_load_to_meta = can_reinit_empty_buffers(model)
 
     # 1a. We load to CPU if we cannot reinit empty buffers
     if not possible_to_load_to_meta:
         logger.warning("Cannot load model to meta device only, loading to CPU instead.")
         model = get_model(config, device=torch.device("cpu"), dtype=DTYPE_MAP[config.optimization_dtype])
-    # 1b. if we can reinit empty buffers, we load to meta device
-    else:
-        model = get_model(config, device=torch.device("meta"), dtype=DTYPE_MAP[config.optimization_dtype])
 
     # Apply LoRA before FSDP setup
     if config.experimental.lora is not None:
@@ -363,7 +362,7 @@ def setup_model(
             model.init_buffers_post_meta()
         else:
             fix_model_post_empty(model)
-    # 3. We fallback to loading the checkpoint from HF if and only if:
+    # 3. We fallback to loading the checkpoint from HF if either of the following is true:
     # - we cannot load to meta device (we cannot reinitialize some buffers)
     # - we are not loading from checkpoint
     else:
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -265,7 +265,7 @@ def load_dcp_from_hf(model: nn.Module, config: ModelConfig, parallel_dims: Paral
     logger.debug(f"Loaded weights using HF DCP in {time.perf_counter() - load_dcp_start_time:.2f} seconds")
 
 
-def can_load_dcp_from_hf(model: nn.Module):
+def can_reinit_empty_buffers(model: nn.Module):
     """Whether the model will be loaded correctly by load_dcp_from_hf.
 
     The main issue is with anything that is not in the checkpoint.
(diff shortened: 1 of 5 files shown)


src/prime_rl/trainer/rl/train.py
```diff
@@ -91,10 +84,15 @@ def train(config: RLTrainerConfig):
...
     parallel_dims = get_parallel_dims(config.model)
+    if config.model.cp > 1:
+        raise ValueError(
+            "CP is not supported for RL. No reason it shouldn't, we just didn't test it. If you need it, please open an issue."
+        )
 
     # Initialize the model and tokenizer
     logger.info(f"Initializing model ({config.model})")
-    model = setup_model(config.model, parallel_dims)
+    loading_from_ckpt_later = config.ckpt and config.ckpt.resume_step
```
Jackmin801: would this trigger for -1? I think with sami's PR that allows -1 to start from scratch this might cause it to never load the ckpt

S1ro1: true, will adapt, forgot about that PR

src/prime_rl/trainer/sft/train.py
```diff
@@ -87,14 +85,10 @@ def train(config: SFTTrainerConfig):
...
 
-    if parallel_dims.cp_enabled:
-        assert config.data.seq_len % parallel_dims.cp == 0, "Sequence length must be divisible by CP degree"
-        substitute_hf_flash_attn(parallel_dims.world_mesh["cp"].get_group(), heads_k_stride=1)
-        substitute_prime_rl_flash_attn(parallel_dims.world_mesh["cp"].get_group(), heads_k_stride=1)
-
     # Initialize the model and tokenizer
     logger.info(f"Initializing model ({config.model})")
-    model = setup_model(config.model, parallel_dims)
+    loading_from_ckpt_later = config.ckpt and config.ckpt.resume_step
```
Jackmin801: same here

src/prime_rl/trainer/model.py
```diff
@@ -352,7 +351,21 @@ def setup_model(config: ModelConfig, parallel_dims: ParallelDims) -> nn.Module:
...
+        torch.distributed.barrier()
+        if isinstance(model, PreTrainedModelPrimeRL):
+            model.init_buffers_post_meta()
+        else:
+            fix_model_post_empty(model)
+    # 3. We fallback to loading the checkpoint from HF if either of the following is true:
+    # - we cannot load to meta device (we cannot reinitialize some buffers)
+    # - we are not loading from checkpoint
+    else:
         load_dcp_from_hf(model, config, parallel_dims)
```
Jackmin801: I think this needs to `elif possible_to_load_to_meta`. We only do load_dcp_from_hf if it was loaded in meta. If we fellback to cpu load, the weights would already be correct at this point and we dont need to load again.

src/prime_rl/trainer/rl/train.py
S1ro1: not true viz https://github.com/PrimeIntellect-ai/prime-rl/pull/1438#discussion_r2625178748

src/prime_rl/trainer/model.py
```diff
@@ -352,8 +356,22 @@ def setup_model(config: ModelConfig, parallel_dims: ParallelDims) -> nn.Module:
...
+        if loading_from_checkpoint_later:
+            logger.warning(
+                "Skipping loading weights. Initializing an empty model on device, loading from checkpoint later."
+            )
+            model.to_empty(device="cuda")
+            torch.distributed.barrier()
+            if isinstance(model, PreTrainedModelPrimeRL):
+                model.init_buffers_post_meta()
+            else:
+                fix_model_post_empty(model)
```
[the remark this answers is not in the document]
S1ro1: @Jackmin801 this is a real issue, right?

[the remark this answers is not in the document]
Jackmin801: it does happen but I dont think you'd ever do this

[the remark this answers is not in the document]
Jackmin801: Like either you want to do lora from scratch, in which you would point to the base model and not resume from a ckpt. Or you want to resume a lora training in which the ckpt will have the lora weights.

I dont think we should support resuming from full weight ckpt but want to train loras off of it

Jackmin801 (approved):
nice! lgtm. I think the merging got a bit weird though and its adding some things from sami's PR and removing some. the core logic for loading from meta looks great though

Jackmin801:
oh u just moved the ckpt manager init up. eh all good then

PR #1446, opened 2025-12-17
PyZmq - rollout transfer using zmq

S1ro1:
8x B200 4 trainer + 4 inference 1.5B qwen distill (as the CI) hendrycks math

image

## Benchmarks

image

---

**GitHub Issue**: [Issue ID]

---

> [!NOTE]
> Introduces a ZMQ transport alongside the filesystem transport for batch exchange between orchestrator and trainers.
> 
> - Adds `ZMQTrainingBatchSender/Receiver` using PUSH/PULL; receiver buffers per-run pending batches and returns the oldest step first per call
> - Adds `ZMQMicroBatchSender/Receiver` using PUB/SUB with a READY barrier (PULL/PUSH) to avoid slow-joiner drops; per-rank topics `data_rank|<n>|`
> - Respects `ZMQTransportConfig` (`host`, `port`, `hwm`) and integrates via `setup_*` helpers to select `transport.type == "zmq"`
> - Exposes new classes from `prime_rl.transport` and logs endpoints; maintains non-blocking polling with `zmq.Poller`
> - Adds dependency `pyzmq>=27.1.0`; updates `CHANGELOG.md` to document `{orchestrator,trainer}.transport.zmq`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 57d4301b2fe8d2b20413422b0e9b44eb8a2958c7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/transport/zmq.py
```diff
@@ -0,0 +1,285 @@
...
+            step = int(batch.step)
+            per_idx = self._pending.setdefault(idx, {})
+            assert step not in per_idx, f"Step {step} already in pending for idx, this should not happen: {per_idx}"
+            per_idx[step] = batch
+
+    def receive(self) -> list[TrainingBatch]:
+        """Return at most one (oldest) pending batch per run idx; buffer newer ones for later calls."""
+        batches: list[TrainingBatch] = []
+
+        current_ids = [self.runs.get_run_dir(idx).stem for idx in self.runs.used_idxs]
```
Jackmin801: this is current_run_dirs not really current_ids?

src/prime_rl/transport/zmq.py
```diff
@@ -0,0 +1,285 @@
...
+                break
+
+            try:
+                sender_id, payload = self.socket.recv_multipart(flags=zmq.NOBLOCK)
+            except zmq.Again:
+                break
+
+            idx = self.get_idx_from_sender_id(sender_id)
+            if idx is None:
+                self.logger.warning(f"Dropping message for unknown sender_id={sender_id!r}")
```
Jackmin801: I think this would make it that oversubscription wouldnt be handled properly.

The Runs object has a max_runs attribute it uses to cap how many active runs the trainer should keep track of at any one time. If the max_runs is exceeded (there are more runs than there are max_runs slots) it ignores the new runs. When runs get freed by the RFT API by deleting the run directory for the freed run, space becomes available and the runs that were ignored might get slots. They would now be out of sync with trainer and deadlock because the orchestrator thought it had sent its batches but the trainer threw them away.

So we should have per-id queues rather than per-idx.

I think it doesnt matter if it isnt fixed in this PR since this already helps with the single run case and we can fix the multi-run ocersubscribe case later. Just pointing it out

src/prime_rl/transport/zmq.py
```diff
@@ -0,0 +1,285 @@
...
+        try:
+            self.socket.close(linger=0)
+            self.ready_socket.close(linger=0)
+        finally:
+            self.logger.info("ZMQ micro batch sender closed")
+
+
+class ZMQMicroBatchReceiver(MicroBatchReceiver):
+    def __init__(self, output_dir: Path, data_rank: int, current_step: int, transport: ZMQTransportConfig):
+        """ZMQ micro batch receiver that receives micro batches from the through ZMQ transport."""
```
Jackmin801: typo? sami? joschu?

src/prime_rl/transport/zmq.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def can_receive(self) -> bool:
+        events = dict(self.poller.poll(timeout=0))
+        return self.socket in events
+
+    def receive(self) -> list[MicroBatch]:
+        """Receive a micro batch from the trainer."""
+        _, payload = self.socket.recv_multipart()
+        micro_batches: list[MicroBatch] = self.decoder.decode(payload)
+        self.current_step += 1
```
Jackmin801: You can keep this but I dont think its actually used by anything. Filesystem needed current step to know where to look next. Here youd just keep receiving and not care too much about which step its for

Jackmin801 (approved):
Nice! LFGTM! batches go vrm vrm. Some small nits and notes

PR #1451, opened 2025-12-18
Refactor buffer env mixing + checkpointing

mikasenghaas:
This PR refactors the buffer such that environment mixing (e.g. training w.r.t a certain environment distribution) is controlled via random sampling inside the buffer instead of explicitly changing the dataset in `vf.EnvGroup`. The main goal is to robustify resuming from a buffer state even when resuming with different env mixes (e.g. add/ remove env, rename env, change underlying dataset per-env, etc.).

Initialization logic:
- Initialize `example_buffer` as a nested dict-of-dicts with `env_name -> (example_id -> example)` and `env_probs` as a probability distribution over environments based on `--config.env-ratio` (fallback to ratio of examples per-env)
- Initialize `easy_pool` and `hard_pool` as flat lists because we do not have to sample from it
- Initialize `rollout_buffer` as flat list (FIFO queue)

Core methods:
- On `sample_example` first sample the env w.r.t `env_probs`, then randomly sample from the examples
- On `update`, move example from `example_buffer` to `easy_pool` or `hard_pool` according to configurable thresholds, and maybe add to rollout buffer (if passing online difficulty filter)
- On `sample_rollouts`, return the most recent rollouts

Resume logic:
- On `save`, save the easy/ hard pool and rollout buffer as `.jsonl` files
- On `load`, load the `.jsonl` files. Then, do a _best effort load_: Match examples in the easy and hard pool to the current example buffer based on configurable hash keys. For all hits, move them to the pools. Also, extend the rollout buffer with all rollouts for active envs. If configured, move back a fraction of the easy/ hard samples to example buffer (reset pool)

Misc. changes:
- Bump vf commit to latest commit which is **fully up-to-date** with vf main! (e.g. the last diverging commit was env-mixing in env groups which was never in vf main, but now got reverted)
- Moved terminology from `problem` -> `example` to match `verifiers` and the rest of the codebase (e.g. it is an `example_buffer` not `problem_buffer`
- Restricted logging to "global across envs" metrics, i.e. we log the pool distribution, ratio of moved samples/ filtered rollouts across all envs, but not per-env yet. I want to find a good _general_ solution for per-env logging in a separate PR

## Examples

```bash
# stage 1
uv run  rl @ examples/alphabet_sort/rl.toml --log.level debug --max-steps 20 --orchestrator.buffer.easy-threshold 0.8 --orchestrator.buffer.hard-threshold 0.2 --orchestrator.buffer.hard-fraction 1.0 --orchestrator.buffer.online-difficulty-filtering --ckpt

# stage 2
uv run  rl @ examples/alphabet_sort/rl.toml --log.level debug --max-steps 30 --orchestrator.buffer.easy-threshold 0.8 --orchestrator.buffer.hard-threshold 0.2 --orchestrator.buffer.hard-fraction 1.0 --orchestrator.buffer.online-difficulty-filtering --ckpt.resume-step 20
```

As we can see, the buffer works correctly:
- Initially all examples are in the normal pool and get moved into easy/ hard pools over time
- Over time more too easy samples are filtered out (bc of online difficulty filtering being enabled)
- Upon resume, the easy pool remains unchanged and the hard pool is reset fully

Screenshot 2025-12-18 at 3 48 59 PM

### Resume

**No pools.** We save empty `.jsonl` files for the easy/ hard pool and rollout buffer. On resume, we are being notified that nothing to load was found in checkpoint.

```bash
# start with
uv run rl @ configs/multi_reverse_text/rl.toml --log.level debug --max-steps 5 --ckpt

# resume with
uv run rl @ configs/multi_reverse_text/rl.toml --log.level debug --max-steps 10 --ckpt.resume-step -1
```

**With pools.** We move easy/ hard examples from the checkpoint into the buffer. All examples have cache hits and get correctly moved.

```bash
# start with
uv run rl @ configs/multi_reverse_text/rl.toml --log.level debug --max-steps 5 --ckpt --orchestrator.buffer.easy-threshold 0.8 --orchestrator.buffer.hard-threshold 0.2

# resume with
uv run rl @ configs/multi_reverse_text/rl.toml --log.level debug --max-steps 10 --ckpt.resume-step -1 --orchestrator.buffer.easy-threshold 0.8 --orchestrator.buffer.hard-threshold 0.2
```

**With pools + resume w/ different envs.** If we only resume with 1 env, we only get cache hits for the `reverse-text` env, not `reverse-text-2` (removed), and so we load a partial checkpoint (best-effort checkpoint loading)

```bash
# like above, but resume with one env less
uv run rl @ configs/multi_reverse_text/rl.toml --log.level debug --max-steps 15 --ckpt.resume-step -1 --orchestrator.buffer.easy-threshold 0.8 --orchestrator.buffer.hard-threshold 0.2 --orchestrator.env '[{"id": "reverse-text"}]'
```

**With pools + reset pools.** We can specify `easy_fraction` or `hard_fraction` to move a fraction of the easy/hard pools back into normal pool (so that it can be sampled again)

```bash
# like above, but resume with one env less
 uv run rl @ configs/multi_reverse_text/rl.toml --log.level debug --max-steps 25 --ckpt.resume-step -1 --orchestrator.buffer.easy-threshold 0.8 --orchestrator.buffer.hard-threshold 0.2 --orchestrator.env '[{"id": "reverse-text"}]' --orchestrator.buffer.easy-fraction 1.0 --orchestrator.buffer.hard-fraction 1.0
```

---

**GitHub Issue**: [Issue ID]

---

> [!NOTE]
> Moves env mixing into the buffer (via `env_ratios`) and adds hash-keyed checkpoint resume; updates orchestrator/scheduler flow, metrics, configs, and tests.
> 
> - **Buffer (`src/prime_rl/orchestrator/buffer.py`)**:
>   - Rewritten to source data from `vf.EnvGroup` and sample by env using `buffer.env_ratios`.
>   - Introduces `sample_examples`, easy/hard pools as lists, and per-step pool/rollout metrics; adds `mean` helper use.
>   - Checkpointing now saves/loads JSONL (`easy_examples.jsonl`, `hard_examples.jsonl`, `rollout_buffer.jsonl`) and resumes via `buffer.hash_keys`; supports converting a fraction of easy/hard back to normal on load.
> - **Config (`src/prime_rl/orchestrator/config.py`)**:
>   - Removes `orchestrator.env_mix`; adds `buffer.env_ratios`, `buffer.hash_keys`; redefines `easy_fraction`/`hard_fraction` semantics; adds validators (thresholds/ratios/length checks).
> - **Orchestrator/Scheduler**:
>   - `EnvGroup` creation no longer passes env-mix args; buffer instantiated with `EnvGroup` for train/val.
>   - Replace `sample_problems` with `sample_examples`; log per-env batch ratios instead of counts.
>   - Guard checkpoint load when `ckpt_manager` is present.
> - **Utils/Trainer**:
>   - Add `mean`, tweak `mean_normalize` return type; fix a log string in filesystem weight broadcast.
> - **Configs/Tests/Deps**:
>   - Update Hendrycks Math config buffer block; overhaul unit tests for new buffer API/behavior.
>   - Bump `verifiers` revision in `pyproject.toml`/`uv.lock`.
>   - Update `CHANGELOG.md` with deprecations and new fields.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 991ee0d6ef78f84b172d5a94c36b923af940aac3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/orchestrator/config.py
```diff
@@ -421,13 +406,45 @@ class BufferConfig(BaseConfig):
...
+
     online_difficulty_filtering: Annotated[
         bool,
         Field(
-            description="Whether to filter rollouts based on their average reward. If True, rollouts with average reward == 0.0 will be marked as hard and rollouts with average reward == 1.0 will be marked as easy.",
+            description="Whether to filter rollouts based on difficulty. If True, rollouts with average reward 0.0 or 1.0 are not added to the buffer.",
         ),
     ] = False
 
+    hash_keys: Annotated[
```
faresobeid, line 434: is this necessary? Do we have a good example for another way to use it than the prompt?

src/prime_rl/utils/utils.py
```diff
@@ -259,6 +259,11 @@ def get_latest_ckpt_step(weights_dir: Path) -> int | None:
     return None
 
 
+def mean(values: list[float] | list[int]) -> float:
+    """Compute the mean of a list of values."""
+    return sum(values) / len(values) if values else 0
```
snimu: both `mean` and `mean_normalize` should set their default values to `0.0`, not `0`, in order to be consistent with their type annotations.

src/prime_rl/orchestrator/config.py
mikasenghaas: two envs that have the same prompt, should prob hash based on (task, prompt)

src/prime_rl/orchestrator/config.py
```diff
@@ -421,13 +406,51 @@ class BufferConfig(BaseConfig):
...
+        list[str],
+        Field(
+            min_length=1,
+            description="Keys to use for computing example hashes. Will be used to match examples from buffer checkpoints and determine buffer resume behavior.",
+        ),
+    ] = ["prompt"]
```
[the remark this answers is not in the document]
mikasenghaas: yea the reason to not put task in by-default is to have resumes be robust to env name changes (e.g. resumed with the same env, but renamed)

src/prime_rl/orchestrator/config.py
faresobeid: ah ya good point

src/prime_rl/orchestrator/config.py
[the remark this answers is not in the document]
faresobeid: thinking though if we just choose to always hash with task, prompt and assume the env names stay the same. Like seems something that shouldnt be configurable, and the assumption seems like it should hold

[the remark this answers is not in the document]
mikasenghaas: fine by me. would keep the config tho that _if_ you rename mid-run you can still recover the buffer ckpt

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -187,7 +183,7 @@ async def orchestrate(config: OrchestratorConfig):
         else:
             checkpoint_step = config.ckpt.resume_step
 
-    if checkpoint_step is not None:
+    if checkpoint_step is not None and ckpt_manager is not None:
```
samsja, line 186: wut I am pretty sure I added this before in my pr

samsja (approved):
lgtm

PR #1477, opened 2025-12-23
threaded OpenAI client

rasdani:
introduce similar threading pattern from threaded sandbox client in verifiers for OpenAI client to relieve event loop

---

**GitHub Issue**: [Issue ID]

---

> [!NOTE]
> Introduces a threaded OpenAI client and wires it into RL orchestration to avoid event loop flooding.
> 
> - Adds `ThreadedAsyncOpenAIClient` (thread pool per base URL; per-thread `AsyncOpenAI` with isolated event loops) and `teardown()` in `utils/threaded_client.py`
> - New `ThreadedClientConfig` with `max_workers_per_client` and `max_connections`; `OrchestratorConfig` now uses it and auto-sets `max_concurrent` and derives `max_workers_per_client`
> - Orchestrator switches to `setup_threaded_clients(...)` and calls `teardown_clients(...)`; logs include `max_workers_per_client`
> - Client helpers: add `setup_threaded_clients(...)` and `teardown_clients(...)` in `utils/client.py` (legacy `setup_clients(...)` retained)
> - Minor: adjust import in `synthesize/utils.py`; increase RL integration test timeouts to 15 minutes
> - Update `CHANGELOG.md` to document new/auto-configured fields
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3d10ea3fb1feec2276ad20ad235d92d6a9101a09. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -6,6 +6,7 @@ from pydantic import BaseModel, Field, model_validator
 
 from prime_rl.transport.config import FileSystemTransportConfig, TransportConfigType
 from prime_rl.utils.config import (
+    ClientConfig,  # noqa
     HeartbeatConfig,
     LogConfig,
     ModelConfig,
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -6,7 +6,6 @@ from pydantic import BaseModel, Field, model_validator
 
 from prime_rl.transport.config import FileSystemTransportConfig, TransportConfigType
 from prime_rl.utils.config import (
-    ClientConfig,
     HeartbeatConfig,
     LogConfig,
     ModelConfig,
@@ -486,11 +485,8 @@ WeightBroadcastConfigType: TypeAlias = FileSystemWeightBroadcastConfig | NCCLWei
 class OrchestratorConfig(BaseSettings):
     """Configures the orchestrator for RL training."""
 
-    # The OAI client configuration
-    client: ClientConfig = ClientConfig()
-
-    # The threaded client configuration
-    threaded_client: ThreadedClientConfig = ThreadedClientConfig()
+    # The threaded OpenAI client configuration
+    client: ThreadedClientConfig = ThreadedClientConfig()
 
     # The model configuration
     model: ModelConfig = ModelConfig()
@@ -676,8 +672,8 @@ class OrchestratorConfig(BaseSettings):
 
     @model_validator(mode="after")
     def auto_setup_max_workers_per_client(self):
-        if self.threaded_client.max_workers_per_client is None:
+        if self.client.max_workers_per_client is None:
             num_clients = len(self.client.base_url)
             if self.max_concurrent is not None:
-                self.threaded_client.max_workers_per_client = math.ceil(self.max_concurrent / num_clients)
+                self.client.max_workers_per_client = math.ceil(self.max_concurrent / num_clients)
         return self
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -674,6 +674,6 @@ class OrchestratorConfig(BaseSettings):
     def auto_setup_max_workers_per_client(self):
         if self.client.max_workers_per_client is None:
             num_clients = len(self.client.base_url)
-            if self.max_concurrent is None:
+            if self.max_concurrent is not None:
                 self.client.max_workers_per_client = math.ceil(self.max_concurrent / num_clients)
         return self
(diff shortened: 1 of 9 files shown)


src/prime_rl/utils/config.py
```diff
@@ -52,6 +52,20 @@ class ClientConfig(BaseConfig):
...
+            description="Use threaded clients to offload API calls from the main event loop. Recommended for high-throughput scenarios.",
+        ),
+    ] = True
+
+    max_workers_per_client: Annotated[
+        int,
+        Field(
+            description="Number of worker threads per client (only used when threaded=True).",
+        ),
+    ] = 64
```
samsja: I woiuld make this its own config

src/prime_rl/utils/threaded_client.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @property
+    def models(self) -> "_Models":
+        return _Models(self)
+
+    def teardown(self, wait: bool = True) -> None:
+        self.executor.shutdown(wait=wait)
+
+
+class _Chat:
```
mikasenghaas: any way we can avoid those?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -93,9 +95,12 @@ async def orchestrate(config: OrchestratorConfig):
 
     # Setup client
     logger.info(
-        f"Initializing OpenAI client (base_url={', '.join(config.client.base_url)}, api_key_var={config.client.api_key_var}, headers={config.client.headers})"
+        f"Initializing OpenAI client (base_url={', '.join(config.client.base_url)}, api_key_var={config.client.api_key_var}, headers={config.client.headers}, threaded={config.client.threaded})"
     )
-    clients = setup_clients(config.client)
+    if config.client.threaded:
```
mikasenghaas: imo fine to not make optional? wanna avoid branching logic as much as possible and just have good default behaviour. would have option max workers on client, which is set to some default. and can then auto configure in validator based on max concurrency and num clients

src/prime_rl/orchestrator/config.py
```diff
@@ -668,3 +669,11 @@ def auto_setup_bench(self):
                 self.prime_monitor.log_extras = None
 
         return self
+
+    @model_validator(mode="after")
+    def auto_setup_max_workers_per_client(self):
+        if self.client.max_workers_per_client is None:
+            num_clients = len(self.client.base_url)
+            if self.max_concurrent is not None:
```
mikasenghaas: if max concurrent is None, we should infer from `batch size * oversampling factor`

src/prime_rl/orchestrator/config.py
```diff
@@ -668,3 +669,11 @@ def auto_setup_bench(self):
...
+                self.client.max_workers_per_client = math.ceil(self.max_concurrent / num_clients)
```
mikasenghaas: also this is too high no? if i do 1k bs and only have 1 client (which is a pretty standard config) i have 1k workers in thread. should take into consideration the connections per thread client as well. in general we should aim to have

num_clients * max_workers_per_client * max_alive_collections ~ max concurrent (or if None, batch_size * oversampling_factor)

num_clients is user-configured. max alive collections i would put leave at oai default (100, maybe 128 to have power of 2 lol), and then i would set `max_workers_per_client` as `max_concurrent // (num_clients * max_alive_collections)`

src/prime_rl/utils/config.py
```diff
@@ -53,6 +53,17 @@ class ClientConfig(BaseConfig):
     ] = {}
 
 
+class ThreadedClientConfig(ClientConfig):
```
mikasenghaas, line 56: this is to keep default behavior in other entrypoints, right?

src/prime_rl/utils/threaded_client.py
```diff
@@ -0,0 +1,84 @@
...
+        max_workers: int,
+        timeout: int = 1200,
+        max_retries: int = 10,
+        headers: dict[str, str] | None = None,
+    ):
+        self.executor = ThreadPoolExecutor(
+            max_workers=max_workers,
+            thread_name_prefix="oai-client",
+        )
+        self._base_url = base_url
```
mikasenghaas: fine to make those public?

src/prime_rl/utils/threaded_client.py
```diff
@@ -0,0 +1,84 @@
...
+        self._api_key = api_key
+        self._timeout = timeout
+        self._max_retries = max_retries
+        self._headers = headers or {}
+        self._tls_key = f"oai_client_{id(self)}"
```
mikasenghaas: is this needed?

rasdani: i guess so. need a key to index into thread local storage.

PR #1484, opened 2025-12-24
add multi env worker

samsja:
## Isolate Environment Execution to Subprocess Workers

This PR refactors the orchestrator to run each environment in a dedicated subprocess worker. The goal is to prevent blocking operations in one environment from affecting other environments or the main orchestrator event loop.
Previously, all environments ran in the main orchestrator process. Any synchronous/blocking code in an environment's reward function (e.g., CPU-bound computation, synchronous I/O) would block the entire event loop, causing lag spikes that affected all rollouts.

With this change:

* Each environment runs in its own EnvWorker subprocess
* The orchestrator communicates with workers via IPC queues
* Workers instantiate their own OpenAI clients (with round-robin load balancing)

```
┌─────────────────────────────────────────────────────────────┐
│                      Orchestrator                           │
│  - Manages training loop                                    │
│  - Calls scheduler.generate_batch()                         │
│  - Logs metrics to wandb (including worker lag)             │
└──────────────────────┬──────────────────────────────────────┘
                       │
                       ▼
┌─────────────────────────────────────────────────────────────┐
│                       Scheduler                             │
│  - Maintains pool of EnvWorker instances                    │
│  - Round-robin dispatches requests to workers               │
│  - Collects responses and resolves futures                  │
│  - Handles policy updates and off-policy cancellation       │
└──────────────────────┬──────────────────────────────────────┘
                       │ (IPC Queues)
        ┌──────────────┼──────────────┐
        ▼              ▼              ▼
┌──────────────┐ ┌──────────────┐ ┌──────────────┐
│  EnvWorker   │ │  EnvWorker   │ │  EnvWorker   │
│  (Process 1) │ │  (Process 2) │ │  (Process 3) │
│              │ │              │ │              │
│ - Own env    │ │ - Own env    │ │ - Own env    │
│ - Own clients│ │ - Own clients│ │ - Own clients│
│ - Own lag    │ │ - Own lag    │ │ - Own lag    │
│   monitor    │ │   monitor    │ │   monitor    │
└──────────────┘ └──────────────┘ └──────────────┘
```

this pr add one the ability to have multiple process per environment

image

---

> [!NOTE]
> Introduces subprocess workers for environment execution and rewires the training loop to dispatch rollouts via IPC, reducing event-loop contention and enabling multiple workers per environment.
> 
> - **New** `EnvWorker` (subprocess) with `RolloutRequest/Response`, per-worker event-loop lag monitoring, round-robin client usage, and least-pending routing
> - **Scheduler rewrite**: routes samples to workers, tracks in-flight futures with off-policy cancellation, updates worker model names on checkpoint change, and returns simplified rollout dicts
> - **Orchestrator changes**: constructs `Buffer` from `Dataset` and `env_names`, initializes/starts/stops `Scheduler` workers, adjusts LoRA handling, and simplifies `error` field logging
> - **Config**: adds `workers_per_env` to control subprocess count per environment
> - **Logging/metrics**: W&B samples include `task`; scheduler exposes per-worker lag and pending counts
> - **Buffer refactor**: no longer depends on `EnvGroup`; assertions/logging updated; unchanged pool/rollout semantics
> - **Tests**: updated to use new `Buffer` constructor and behaviors
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ecab443742a0da8ea95036e95a73cb95ec918d8d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -135,27 +135,30 @@ async def orchestrate(config: OrchestratorConfig):
 
     # Setup buffer
     logger.info(f"Setting up buffer ({config.buffer})")
-    buffer = Buffer(env, config.buffer)
+    train_dataset = env.get_dataset(seed=config.buffer.seed)
+    buffer = Buffer(train_dataset, env.env_names, config.buffer)
     if config.val is not None:
         val_buffer_config = BufferConfig(env_ratios=config.buffer.env_ratios)
-        val_buffer = Buffer(env, val_buffer_config, dataset_type="val")
+        # Preserve historical behavior: validation dataset seeding follows val buffer config (defaults to None).
```
mikasenghaas: remove comment

src/prime_rl/orchestrator/buffer.py
mikasenghaas, line 1: i dont quite get why the buffer needs to be initted with the dataset instead of the env here. afaict, on orch we still load the full env group so we might as well init with env. maybe im missing smth tho?

to give some ctx on the historical evolution of this: initially, the buffer wrapped a dataset because it seemed like the more "general" abstraction. however, as the buffer evolved, it used more and more features of the "vf-specific dataset", e.g. it relies on specific cols like `example_id` and `task` so it seemed more and more silly to assume generality with the constructor args so i changed it pretty recently to be initted by a env, making clear that the buffer is an abstraction that allows sampling problems and rollouts from a verifiers environment not a general dataset

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+
+import verifiers as vf
+from openai import AsyncOpenAI
+
+from prime_rl.utils.client import setup_clients
+from prime_rl.utils.config import ClientConfig
+
+
+@dataclass
+class RolloutRequest:
```
mikasenghaas, line 21: from first principles, this should only need `id` and `rollouts_per_example`, right? all other is redunant:
- example can be inferred from id
- samling args and model name is the same across all reqs to an env?

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+    model_name: str
+
+
+@dataclass
+class RolloutResponse:
+    """Response containing rollout results."""
+
+    id: str
+    results: list[dict]  # Simplified state dicts
+    error: str | None = None
```
mikasenghaas: why is err not part of results? its usually part of vf.State

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+    lag_metrics: dict | None = None  # Event loop lag metrics from worker
```
mikasenghaas, line 35: arfh, i really hate the way do metrics in general rn. always exporting them in weird ways from classes like buffer, scheduler and now across process boundaries. wondering what would be a good holistic pattern here

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+        "prompt": state.get("prompt"),
+        "completion": state.get("completion"),
+        "trajectory": trajectory,
+    }
+
+
+async def process_request(
+    request: RolloutRequest,
+    env: vf.Environment,
+    clients: list[AsyncOpenAI],
```
mikasenghaas: `clients` unused?

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+    max_concurrent: int,
+    env_id: str,
+):
+    """Main async loop for processing requests."""
+    from prime_rl.orchestrator.event_loop_lag import EventLoopLagMonitor
+
+    client_cycle = cycle(clients)
+    semaphore = asyncio.Semaphore(max_concurrent) if max_concurrent > 0 else asyncio.Semaphore(10000)
```
mikasenghaas, line 115: concurrency is per env worker now? if true, should document that in configs. 

let's think abt whether max concurrent arg should be global. advantage is that  you wouldnt need to tune the param as you scale workers/env

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+
+            # Wait for at least one task to complete
+            done, _ = await asyncio.wait(pending_tasks.keys(), timeout=0.1, return_when=asyncio.FIRST_COMPLETED)
+
+            for task in done:
+                pending_tasks.pop(task)
+                try:
+                    response = task.result()
+                    # Attach lag metrics to response
+                    response.lag_metrics = lag_monitor.get_metrics()
```
mikasenghaas: hm we pass the list of all lag metrics with each finished req

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+                    response_queue.put(response)
+                except Exception as e:
+                    # Should not happen since process_request catches exceptions
+                    response_queue.put(RolloutResponse(id="unknown", results=[], error=str(e)))
```
mikasenghaas: lets not put dead exceptions, very opus behavior haha and dont like the 'empty' rollout response side effect

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+):
+    """Main entry point for worker process."""
+    # Load environment
+    env = vf.load_environment(env_id, **env_args)
+    env.set_max_seq_len(seq_len)
+    env.set_interleaved_rollouts(interleaved_rollouts)
+
+    # Create clients
+    client_config = ClientConfig(**client_config_dict)
+    clients = setup_clients(client_config)
```
mikasenghaas, line 186: quite a lot of clients: `num_envs * workers_per_env * num_inference_nodes`. will have another multiplier if we do threaded client. if we do threaded + multi-proc, should prob reinvestigate if we need the multi-client for inference scaling. plausible that vllm fixed their dp issues

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+            # Non-blocking check for responses
+            try:
+                while not self.response_queue.empty():
+                    response: RolloutResponse = self.response_queue.get_nowait()
+                    # Store latest lag metrics from worker
+                    if response.lag_metrics:
+                        self.latest_lag_metrics = response.lag_metrics
+                    if response.id in self.pending_futures:
+                        future = self.pending_futures.pop(response.id)
+                        if response.error:
```
mikasenghaas: which case is handled by this? is this for vf errors?

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+                            future.set_exception(RuntimeError(response.error))
+                        else:
+                            future.set_result(response.results)
+            except Exception:
+                pass
+            await asyncio.sleep(0.01)
+
+    def update_model_name(self, model_name: str):
```
mikasenghaas, line 307: do we need this?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -68,32 +71,93 @@ def __init__(
...
+    async def stop(self):
+        """Stop all workers and collectors."""
+        for task in self._response_collectors:
+            task.cancel()
+        for workers in self.workers.values():
+            for worker in workers:
+                worker.stop()
+
+    def _get_next_worker(self) -> EnvWorker:
+        """Round-robin through envs, least-pending within each env."""
```
mikasenghaas: huh how can we round-robin through envs if we wanna respect arbitrary env ratios?

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -0,0 +1,307 @@
...
+class RolloutResponse:
+    """Response containing rollout results."""
+
+    id: str
+    results: list[dict]  # Simplified state dicts
+    error: str | None = None
+    lag_metrics: dict | None = None  # Event loop lag metrics from worker
+
+
+def extract_result(state: vf.State) -> dict:
```
mikasenghaas, line 38: im not a huge fan of this: it means that any new field that is added to vf state that orch might want, will have to be manually added here.

i understand that we cannot just pass state as-is through ipc, bc state might not be serializable. but im wondering if we can maybe guarantee that all explicitly typed out fields in vf state are serializable and hence we automatically have access to all fields vf defines. we would still miss env-specific fields in this way but i think a) we cannot prevent this and b) its kinda fine bc those are typically runtime args that are only used during rollout + scoring and are not needed afterwards

mikasenghaas (commented):
nice one, super crucial for perf! as mentioned alr i think we wanna think carefully abt the case workers_per_env>1 and im wondering if we should even support it in the way it currently is. gut feeling is that this should be pretty co-designed with vf and not just "wrapped" on top like we currently do.

on a more general note: my general feeling is that we start to have pretty convoluted code on the orch. this is not all due to this pr ofc, started with e.g. buffer + scheduler not being super clean abstractions, metrics handling being a bit awkward etc. but it all gets exacerbated by going across process boundaries. stuff like manually passing back lag metrics, selecting specific result fields to be passed back to orch, all seem very handcrafted to the current shape and form of the framework to "just make it work" but im fearing are not very susintable. im fine with merging smth a bit more hacky for perf reasons but pretty soon i think we will wanna consolidate abstractions and cleanup  otherwise im fearing it will get very unpleasant to maintain

src/prime_rl/orchestrator/buffer.py
samsja: okay good to know.

Yes as of now this change is not needed. It just that we should probably avoid very soon having to initiated the environment inside the main event loop just to get the datasaet this seems silly to me.

THe right pattern imo is to have a load_dataset from an environment. There is really no need to have both environment runtime and dataset being couple in anyway

src/prime_rl/orchestrator/env_worker.py
samsja: yeah I agree here, quickly hacked this because vf state was not serializable

src/prime_rl/orchestrator/scheduler.py
samsja: we only round robin throw subprocess of the same env afaiu

src/prime_rl/orchestrator/env_worker.py
samsja: yeah hence my point that I don't think we need threaded client anymore with this

src/prime_rl/orchestrator/env_worker.py
samsja: its the error coming from the env worker not from vf

src/prime_rl/orchestrator/env_worker.py
samsja: yes v important

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -144,6 +144,18 @@ async def orchestrate(config: OrchestratorConfig):
     else:
         val_buffer = None
 
+    # Get checkpoint manager
+    logger.info(f"Initializing checkpoint manager ({config.ckpt})")
+    ckpt_manager = setup_ckpt_manager(config.output_dir, config.ckpt)
+
+    # Determine checkpoint step early to set correct model name before starting workers
```
samsja: remove comment @cursoragent

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -130,109 +194,143 @@ async def update_policy(self):
...
             "time/wait_for_ckpt": self.wait_for_ckpt_time,
             "time/update_weights": self.update_weights_time,
             "batch/async_level": self.async_level,
             "batch/off_policy_level/max": self.max_off_policy_level,
             "batch/off_policy_level/mean": self.mean_off_policy_level,
             "batch/off_policy_level/min": self.min_off_policy_level,
         }
+
+        # Add per-worker lag metrics and pending counts
+        for env_name, workers in self.workers.items():
```
Jackmin801: this env_name isnt used I think
```suggestion
        for workers in self.workers.values():
```

Jackmin801 (approved):
Nice! lgtm

PR #1488, opened 2025-12-27
[feat] auto impl selection

Jackmin801:
A pretty common footgun in the codebase rn is to run the MoE models with HF impl accidentally which causes the code to hang at either model loading or first backward.

This PR makes the default model.impl auto. With auto, we always use the custom impl if we support it, otherwise, we fallback to hf.

---

> [!NOTE]
> Makes model implementation selection automatic to avoid accidentally using incompatible implementations.
> 
> - **Config**: `model.impl` adds `"auto"` and becomes the default; validator updated to allow `trust_remote_code` with `"hf"` or `"auto"` (`config.py`)
> - **Model loading**: Auto-detects support via `supports_custom_impl` and selects `custom` or `hf` accordingly; logging added; switch uses `impl_to_use` (`model.py`)
> - **Models registry**: Adds `supports_custom_impl(model_config: PretrainedConfig)` to check if custom implementation is available; exports updated (`models/__init__.py`)
> - **Changelog**: Documents the new default and behavior for `model.impl`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 51d95061ce1c43b17f0e099f92312cbb3e7ab73f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


Jackmin801:
@cursoragent Add a line to changelog about the changes to model.impl here

samsja (approved):
lgtm

PR #1553, opened 2026-01-07
Re-raise `vf.SandboxError` in eval to trigger rollout rescheduling

rasdani:
So online-evals may never crash again.

  - Check state["error"] after rollout completion in _generate_rollout and _generate_group
  - Re-raise vf.SandboxError to trigger existing @retry decorator
  - Enables automatic rescheduling of rollouts with sandbox infrastructure failures

---

**GitHub Issue**: [Issue ID]

---

> [!NOTE]
> Ensures infra failures reschedule rollouts by surfacing errors from returned states to the retry mechanism.
> 
> - In `_generate_rollout` and `_generate_group`, check `state["error"]` (or each in `states`) and re-raise if it's a `vf.InfraError` to trigger `@retry`
> - Preserve context-length handling: on `BadRequestError`, parse max tokens, retry with reduced `max_tokens`, and still re-raise `vf.InfraError` if present
> - Applies to both per-rollout and group generation paths without altering saving/progress behavior (`results.jsonl`, progress bar)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b8d8cbe19c48d1bd1e0cdab9de203d29488f0036. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/eval/utils.py
+++ b/src/prime_rl/eval/utils.py
@@ -231,7 +231,11 @@ async def generate_and_save_rollout(
         """Asynchronously generate and score a single rollout."""
         logger = get_logger()
         try:
-            return await generate_rollout(client, env, model_name, example, sampling_args)
+            state = await generate_rollout(client, env, model_name, example, sampling_args)
+            # Re-raise infrastructure errors to trigger retry
+            if state.get("error") and isinstance(state["error"], vf.SandboxError):
+                raise state["error"]
+            return state
         except BadRequestError as e:
             # Check if this is a context length error and retry with adjusted max_tokens
             error_message = str(e)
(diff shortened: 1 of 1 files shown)


src/prime_rl/eval/utils.py
```diff
@@ -240,7 +244,10 @@ async def _generate_rollout(
             if new_max_tokens is not None:
                 logger.warning(f"Context length error: reducing max_tokens to {new_max_tokens}.")
                 sampling_args["max_tokens"] = new_max_tokens
-                return await generate_rollout(client, env, model_name, example, sampling_args)
+                state = await generate_rollout(client, env, model_name, example, sampling_args)
+                if state.get("error") and isinstance(state["error"], vf.SandboxError):
```
mikasenghaas: this seems ultra-specific, wondering what would be a good general pattern to discriminate between errors that should trigger retries and those that should not

rasdani: we can make this catch `vf.Error` or `vf.InfraError` instead.

general pattern:
- eval  -> must always retry to have best-effort accuracy
- training -> can mask

rn `vf` will never raise itself but gracefully hand on `state["error"]`. it think the way we treat this differently in `prime-rl` is already a good pattern.

i very much like the outer caller having the responsibility here.

samsja (approved):
lgtm

PR #1554, opened 2026-01-07
QoL: Checkpoint warnings + Upload CLI

rasdani:
Checkpoint Upload CLI (upload-ckpt):
  - Upload checkpoints to HuggingFace Hub with configurable org/repo prefix
  - Supports three checkpoint types: weights, trainer, orchestrator, or all
  - Auto-discovers available steps or accepts explicit --steps list
  - Uses upload_large_folder with configurable worker count for performance
  - Includes --dry-run mode and --public flag for repo visibility
  - Temporarily disables HF offline mode environment variables during upload
  - Can be turned into `maybe_save()` down the line

  Checkpoint Consistency Warnings:
  - New warn_if_ckpts_inconsistent() called on orchestrator resume
  - Warns if resume_step=-1 but latest steps differ across checkpoint directories
  - Warns if explicit resume_step doesn't exist in all checkpoint directories
  - Skips warning for multi-tenant setups (multiple run_* directories)

To be tested in prod.

---

**GitHub Issue**: [Issue ID]

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ dependencies = [
     "prime>=0.5.0",
     "tenacity>=8.2.0",
     "pyzmq>=27.1.0",
+    "huggingface_hub>=0.32.0",
 ]
 
 [project.scripts]
@@ -51,6 +52,7 @@ inference = "prime_rl.inference.server:main"
 sft = "prime_rl.trainer.sft.train:main"
 eval = "prime_rl.eval.eval:main"
 synthesize = "prime_rl.synthesize.synthesize:main"
+upload-ckpt = "prime_rl.utils.upload:main"
 
 [project.optional-dependencies]
 flash-attn = [
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -55,6 +55,7 @@ from prime_rl.utils.utils import (
     install_env,
     resolve_latest_ckpt_step,
     to_col_format,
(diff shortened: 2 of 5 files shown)


mikasenghaas (commented):
hmm this seems like a lot of boilerplate code around `hf upload-large-folder` with not that _gain_ in utility. like the main thing is that it abstracts the location of the checkpoint paths, right? im wondering if we should rather aim to have a simpler checkpoint structure. people are alr used to hf upload cli, this would be one more thing to learn.

agree tho, that we defo want ckpt upload on-the-fly during training--this seems like the bigger unlock

rasdani:
> hmm this seems like a lot of boilerplate code around `hf upload-large-folder` with not that _gain_ in utility. like the main thing is that it abstracts the location of the checkpoint paths, right? im wondering if we should rather aim to have a simpler checkpoint structure. people are alr used to hf upload cli, this would be one more thing to learn.
> 
> agree tho, that we defo want ckpt upload on-the-fly during training--this seems like the bigger unlock

partially (dis-)agree. the utility here is from having an easy way to push *all common steps present in all three checkpoint directories*. otherwise it's a pain in the ass to upload them manually in three panes varying in upload time a lot, being deleted while you're uploading (haven't cherry-picked the file lock yet), or ending up with incomplete upload bc of one piece missing.

and yes my goal it  to have this part of checkpoint manager as e.g. `maybe_upload()` for sure, but didn't want to put this into the orchestrator for prod training as the blocking eval already caused us trouble.

PR #1529, opened 2026-01-01
fix wandb

samsja:
wip fix step wandb

---

> [!NOTE]
> Standardizes W&B step handling by making `monitor.log` accept an explicit `step` and propagating it to `wandb.log`.
> 
> - Changes `WandbMonitor.log(metrics, step)` to forward the provided `step` to W&B instead of inferring from `metrics`
> - Updates all call sites to pass `step=progress.step` in RL/SFT trainers and `step=None` in evals (while still including `step` in the metrics payload)
> - No functional training logic changes; minor line-wrap/formatting adjustments
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 28e542c9f55c49665341da8615bdbb858d04298c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1556, opened 2026-01-07
Fix out-of-sync W&B step

mikasenghaas:
#1529 and #1538 made `step` an explicit method arg in the monitors (as opposed to just being popped by the W&B monitor), but didn't include the arg when logging the main orchestrator metrics. This would lead to the following out-of-sync behavior:
1. Log metrics at step t (but no step provided), so auto-increments to t+1
2. Log samples at step t (step provided) now conflicts because W&B counter is already at t+1

This PR passes the correct step, and commits on the final log call of the step. This required adding the `commit` flag to the logging calls.

---

> [!NOTE]
> Ensures logging step alignment and proper step commits, especially for W&B.
> 
> - Orchestrator: pass `step=progress.step` to `monitor.log`; call `monitor.log_samples(..., commit=True)` as the final log per step
> - Monitor interfaces: add optional `commit` to `log` and `log_samples`; update `NoOpMonitor`, `MultiMonitor`, `WandbMonitor`, and `PrimeMonitor` to accept/forward it
> - W&B: forward `commit` to `wandb.log` for metrics and samples
> - No changes to distribution logging behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cc6d3ee58da594725cdb1f8f7be9af3ea5d2388d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -509,11 +509,13 @@ async def orchestrate(config: OrchestratorConfig):
                 to_log.update({f"val_batch/{env}": ratio for env, ratio in per_env_ratio.items()})
 
         # Log metrics to monitor(s)
-        monitor.log(to_log)
+        monitor.log(to_log, step=progress.step)
 
         # Log samples to monitor(s) if enabled
         subset_train_rollouts = random.sample(train_rollouts, min(8, len(train_rollouts)))
-        monitor.log_samples(subset_train_rollouts, step=progress.step)
+        monitor.log_samples(
+            subset_train_rollouts, step=progress.step, commit=True
+        )  # this is the last place we log to w&b from orch, so commit
 
         # Log distributions (rewards, advantages) if enabled
         monitor.log_distributions(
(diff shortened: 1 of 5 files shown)


rasdani (approved):

Jackmin801 (approved):
nice! eh two nit suggestions
1. should we put commit=True in the evals?
2. should we make commit=True default? I think its better to flush too often than to not flush. is there a reason we shouldnt just always commit and tank the warning?

mikasenghaas:
> should we make commit=True default? I think its better to flush too often than to not flush. is there a reason we shouldnt just always commit and tank the warning?

i think its not just a warning, but the metrics do not get pushed if the steps are out-of-sync (at least this is what happened with the samples log), so i dont think we can do this?

mikasenghaas:
> should we put commit=True in the evals?

ohhh, i just saw this in the evals

```py
    monitor.log(eval_metrics, step=None)
```

this means our step count is just wrong when we do online evals.. should defo fix yea

PR #1568, opened 2026-01-11
Unique rollout request ids to avoid collisions

snimu:
avoid orphaning future from example by overwriting the future with newly submitted rollout from that same example. Fixed with specific `request_id`.

---

> [!NOTE]
> Ensures rollout tracking uses unique identifiers to prevent future collisions when multiple rollouts originate from the same example.
> 
> - Introduces `request_id` (UUID) in `RolloutRequest`/`RolloutResponse`; separates `example_id`
> - `EnvWorker`: keys `pending_futures` by `request_id`; `submit_request` returns `(future, request_id)`; worker loop maps tasks to `request_id`
> - `Scheduler`: captures `request_id` directly from `EnvWorker.submit_request` and stores it in `InflightRolloutInfo`
> - Improves response matching and avoids orphaned/overwritten futures during concurrent rollouts
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f72421457f381d6353e2ac5bc35ee502ee80ec6d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/env_worker.py
+++ b/src/prime_rl/orchestrator/env_worker.py
@@ -135,7 +135,7 @@ async def worker_loop(
             task = asyncio.create_task(
                 process_request(request, env, client_cycle, semaphore, example_lookup, sampling_args)
             )
-            pending_tasks[task] = request.id
+            pending_tasks[task] = request.request_id
         return True
 
     try:
--- a/src/prime_rl/orchestrator/env_worker.py
+++ b/src/prime_rl/orchestrator/env_worker.py
@@ -6,6 +6,7 @@ Runs environment rollouts in a separate process to isolate event loop lag.
 
 import asyncio
 import queue
+import uuid
 from dataclasses import dataclass
 from itertools import cycle
 from multiprocessing import Process, Queue
@@ -21,7 +22,8 @@ from prime_rl.utils.config import ClientConfig
 class RolloutRequest:
     """Request to generate rollouts for an example."""
 
-    id: int  # example_id
+    request_id: str
+    example_id: int
     rollouts_per_example: int
     model_name: str  # Model name to use for this request (may change for LoRA)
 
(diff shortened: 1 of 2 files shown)


src/prime_rl/orchestrator/env_worker.py
```diff
@@ -269,20 +271,22 @@ async def submit_request(
...
         )
 
         loop = asyncio.get_event_loop()
         future = loop.create_future()
-        self.pending_futures[str(example_id)] = future
+        self.pending_futures[request_id] = future
 
         self.request_queue.put(request)
-        return future
+        return future, request_id
```
mikasenghaas, line 289: no need to change the signature here, right?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -143,14 +143,11 @@ async def schedule_group_rollout(self):
         workers = self.workers[task]
         worker = min(workers, key=lambda w: w.pending_count)
 
-        future = await worker.submit_request(
+        future, request_id = await worker.submit_request(
```
mikasenghaas, line 146: then, would not need to change return type here

mikasenghaas (commented):
yep, this looks very similar to what i did. minor comment, but else lgtm!

src/prime_rl/orchestrator/env_worker.py
snimu: without this, access to the `request_id` in the scheduler becomes a pain. Either, we have to add an attribute to the future directly (I'm not comfortable with that), or we need to loop over all the pending futures like this: `request_id = [k for k, v in worker.pending_futures.items() if v is future][0]` which is kinda ugly (unless you have some other access mechanism in mind, I could just overlook sth).

mikasenghaas: `request_id = [k for k, v in worker.pending_futures.items() if v is future][0]` this is ~what we do right now, no? so would be fine for me to keep like this given that major refactor is coming either way

but honestly, also fine as-is, can just merge

mikasenghaas (approved):

PR #1596, opened 2026-01-15
Fix trainer crash when all rollouts in a batch fail

JannikSt:
## Summary
- Fixes crash when all rollouts fail and tensor is empty
- Adds guard to skip stats computation for empty tensors

## Context
Discovered while testing diplomacy environment - multi-turn conversations can exceed context limits, causing all rollouts to fail occasionally.

---

> [!NOTE]
> Prevents a crash during distributed stats aggregation when batches produce no valid rollouts.
> 
> - In `Tensors.compute_stats`, after `flexible_all_gather`, detect empty 1D tensors and record `NaN` for `mean/median/std/min/max`, skipping computation
> - Leaves non-empty path unchanged; still gathers tensors and appends aggregated values back to the accumulator
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 13ea8f9df1c84c1ed31a08937f9cc46c6a80d646. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/utils.py
+++ b/src/prime_rl/trainer/utils.py
@@ -215,6 +215,10 @@ class Tensors(defaultdict):
             tensors = flexible_all_gather(tensors)
             assert tensors.ndim == 1, "Can only aggregate 1D tensors"
 
+            # Skip empty tensors (can happen when all rollouts in a batch fail)
+            if tensors.numel() == 0:
+                continue
+
             # Compute relevant tensor statistics
             metrics[f"{key}/mean"] = tensors.mean().item()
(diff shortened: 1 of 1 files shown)


Jackmin801:
The orchestrator shouldnt be able to create a sample with all failed rollouts can it?

JannikSt:
> The orchestrator shouldnt be able to create a sample with all failed rollouts can it?

I'll get back to you on this one

Jackmin801 (approved):

PR #1617, opened 2026-01-20
Add elastic inference pool with DNS-based service discovery

JannikSt:
## Summary
- Adds `ElasticInferencePool` for dynamic pod discovery via K8s headless service DNS
- Tracks which pods have the correct LoRA adapter loaded
- Only exposes "ready" pods to workers, preventing model-not-found errors
- Backwards compatible with static `base_url` config

## Usage
```toml
[client.elastic]
headless_service = "inference-headless.ns.svc.cluster.local"
port = 8000
sync_interval = 5.0
```

## Changes
- `elastic.py` - new ElasticInferencePool class
- `config.py` - add ElasticConfig
- `env_worker.py` - handle URL updates from main process
- `scheduler.py` - integrate elastic pool, send ready URLs to workers
- `orchestrator.py` - create/manage elastic pool lifecycle
- `test_elastic.py` - 31 unit tests

---

> [!NOTE]
> Enables dynamic, DNS-based scaling of inference servers with automatic LoRA adapter synchronization and safe routing to only "ready" servers.
> 
> - Adds `ElasticInferencePool` and `ServerDiscovery` (`utils/elastic.py`) for DNS discovery, health/model checks, adapter load/sync, and pool metrics
> - Extends `ClientConfig` with `elastic` and `is_elastic`; introduces `InferencePool` protocol, `StaticInferencePool`, and `setup_inference_pool` in `utils/client.py`
> - Orchestrator now initializes/manages a unified `inference_pool`, waits for readiness, updates weights via the pool, and stops it on shutdown
> - Scheduler uses `InferencePool` for weight updates and metrics; switches workers to LoRA model names after sync
> - Env workers support elastic mode: dynamic client discovery/refresh, queuing until clients available, and model-name switching per request
> - Updates `CHANGELOG.md`; adds focused unit tests for elastic pool and worker model-switching behavior
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1bb3eec0d6c6579f286b3feabcf0c6edf902d375. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tests/unit/utils/test_elastic.py
```diff
@@ -0,0 +1,601 @@
...
+        client_config=client_config,
+        base_model="Qwen/Qwen2-0.5B",
+        port=8000,
+        sync_interval=5.0,
+    )
+
+
+class TestElasticInferencePoolInit:
+    def test_init_sets_attributes(self, pool):
+        assert pool.hostname == "inference-headless.ns.svc.cluster.local"
```
samsja: can we keep pytest style tests here

src/prime_rl/utils/config.py
```diff
@@ -21,8 +21,43 @@ class ModelConfig(BaseConfig):
...
+    ]
+
+    port: Annotated[
+        int,
+        Field(
+            description="Port that inference servers listen on.",
+        ),
+    ] = 8000
+
+    sync_interval: Annotated[
```
samsja, line 44: can we precise what is the unit for "often" second ? or step ?

src/prime_rl/utils/elastic.py
```diff
@@ -0,0 +1,472 @@
...
+    """Discover server IPs via DNS lookup.
+
+    Args:
+        hostname: DNS hostname that resolves to one or more IP addresses
+
+    Returns:
+        List of server IP addresses (empty if lookup fails or no records)
+    """
+    try:
+        _, _, ips = socket.gethostbyname_ex(hostname)
```
samsja, line 31: so the whole logics is base on the dns returning many ips ?

tests/unit/utils/test_elastic.py
JannikSt: yes

src/prime_rl/utils/elastic.py
```diff
@@ -0,0 +1,472 @@
...
+
+            server.status = "unhealthy"
+            server.sync_failures += 1
+            return False
+
+        except Exception as e:
+            server.status = "unhealthy"
+            server.sync_failures += 1
+            self.logger.error(f"Failed to sync server {ip}: {e}")
+            return False
```
samsja: not sure about this large try expect, can we rather add try expect on the line we expect it might fail ?

src/prime_rl/utils/config.py
JannikSt: fixing thx

src/prime_rl/utils/elastic.py
```diff
@@ -0,0 +1,472 @@
...
+
+        except Exception as e:
+            self.logger.debug(f"Failed to add server {ip}: {e}")
+            self._servers.pop(ip, None)
+            if ip in self._admin_clients:
+                try:
+                    await self._admin_clients.pop(ip).aclose()
+                except Exception:
+                    pass
+            return False
```
samsja, line 397: can we put try expect where we expect it to fail

samsja, line 397: I think we already handle the failing case in subfunction

src/prime_rl/utils/elastic.py
```diff
@@ -0,0 +1,472 @@
...
+
+    async def _remove_server(self, ip: str) -> None:
+        """Remove a server from the pool."""
+        self.logger.debug(f"Inference server removed: {ip}")
+        self._servers.pop(ip, None)
+        if ip in self._admin_clients:
+            try:
+                await self._admin_clients.pop(ip).aclose()
+            except Exception:
+                pass
```
samsja: no need to try expect here

src/prime_rl/utils/elastic.py
JannikSt: yes

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -131,15 +152,70 @@ async def worker_loop(
...
+        nonlocal current_clients, client_cycle
+
+        logger = get_logger()
+
+        # Close old clients to avoid connection leaks
+        for client in current_clients:
+            try:
+                client.close()
+            except Exception:
+                pass
```
samsja: remove try edxpect

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -131,15 +152,70 @@ async def worker_loop(
     # Track in-flight tasks
     pending_tasks: dict[asyncio.Task, str] = {}
 
+    def handle_url_update(urls: list[str]) -> None:
+        """Handle URL update by recreating clients with new ready URLs."""
+        nonlocal current_clients, client_cycle
```
samsja: why non local here ?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -230,11 +272,13 @@ async def orchestrate(config: OrchestratorConfig):
...
-            get_weight_dir(config.output_dir, scheduler.ckpt_step),
-            lora_name=config.model.lora.name if config.model.lora else None,
-        )
+        weights_path = get_weight_dir(config.output_dir, scheduler.ckpt_step)
+        lora_name = config.model.lora.name if config.model.lora else None
+
+        if elastic_pool is not None:
+            await elastic_pool.update_weights(weights_path, lora_name=lora_name, step=scheduler.ckpt_step)
+        else:
+            await update_weights(admin_clients, weights_path, lora_name=lora_name)
```
samsja: I don't like that we have two path here, why not just having pool return a list of clients that we fed tu update weight ?
```suggestion
        if elastic_pool is not None:
            current_admin_clients = await elastic_pool.get_admin_client        
        else:
            current_admin_clients = admin_clients
             
        await update_weights(admin_clients, weights_path, lora_name=lora_name)
```

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -60,6 +63,7 @@ def __init__(
         strict_async_level: bool,
         lora_name: str | None = None,
         output_dir: Path | None = None,
+        elastic_pool: "ElasticInferencePool | None" = None,
```
samsja: I don't think we need `"` here as we import for type check

samsja: I don't think we need `"` here as we import for type check

src/prime_rl/orchestrator/orchestrator.py
JannikSt: yeah the function name here is actually misleading - it is rather an adapter sync that also makes sure the inference pool has loaded all the right adapters

samsja (commented):
looks good,

I would try to reduce the change of code on the orctrestator side I especially don't like if we need to branch behavior between elastic and normal for things like weight update logic. There should only be one place that define this logic.

I think we should also add more doc either a .md or docstring on how the elastic is working


also need to add in the config that this is not compatible with nccl boadcast

src/prime_rl/utils/config.py
```diff
@@ -21,8 +21,43 @@ class ModelConfig(BaseConfig):
...
+    ]
+
+    port: Annotated[
+        int,
+        Field(
+            description="Port that inference servers listen on.",
+        ),
+    ] = 8000
+
+    sync_interval_s: Annotated[
```
samsja: ah lol would have kept the same name I just wanted the docstring to change to add seconds

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -90,12 +92,32 @@ async def orchestrate(config: OrchestratorConfig):
...
+        await elastic_pool.start()
+        clients = elastic_pool.get_inference_clients()
+        admin_clients = elastic_pool.admin_clients
+        logger.info(f"Elastic pool started with {elastic_pool.num_servers} server(s) ({elastic_pool.num_ready_servers} ready)")
+    else:
+        logger.info(
+            f"Initializing OpenAI client (base_url={', '.join(config.client.base_url)}, api_key_var={config.client.api_key_var}, headers={config.client.headers})"
+        )
+        clients = setup_clients(config.client)
+        admin_clients = setup_admin_clients(config.client)
```
samsja: wondering if the env worker shouldn't be agnostic to the concept of elastic client, we should just imo have a way to update the client of a worker via an api, then we have some kind of elastic check that update the client list and then change the client on each worker

samsja (commented):
proposal: what about instead of having a "elastic client" we keep the current logic that handle multi client, but we have some logic at the beginning of every step that potentially refresh the list of client and update the env worker accordingly

src/prime_rl/orchestrator/orchestrator.py
JannikSt: technically its already agnostic of the concept of an elastic client - it just has callback URLs to update the URLs

Jackmin801 (commented):
hrmm I think it would be easier to understand and more resilient of the env workers had their own elastic pools instead. as discussed, the env workers dont actually need to be aware of what step is on the inference server, only that the inference server has the model that was requested. this is because the env workers are a proxy of the requests initiated by the orchestrator process, which is in charge of blocking further requests on max_async. so as long as admin clients on orchestrator process are aware of what step the inference servers are at and block until all the inference servers it can see are updated, we shouldnt be in the case where the env worker will call the wrong inference server.

the orchestrators model update loop needs to make sure it broadcasts out the update to all discovered servers before unblocking and allowing new requests.

the env workers only make sure that the model_name in the request is on the discovered inference servers

JannikSt:
> hrmm I think it would be easier to understand and more resilient of the env workers had their own elastic pools instead. as discussed, the env workers dont actually need to be aware of what step is on the inference server, only that the inference server has the model that was requested. this is because the env workers are a proxy of the requests initiated by the orchestrator process, which is in charge of blocking further requests on max_async. so as long as admin clients on orchestrator process are aware of what step the inference servers are at and block until all the inference servers it can see are updated, we shouldnt be in the case where the env worker will call the wrong inference server.
> 
> the orchestrators model update loop needs to make sure it broadcasts out the update to all discovered servers before unblocking and allowing new requests.
> 
> the env workers only make sure that the model_name in the request is on the discovered inference servers

bro we discussed this in person

Jackmin801:
> bro we discussed this in person

if its not in the LLM training data it didnt happen

src/prime_rl/utils/elastic.py
```diff
@@ -0,0 +1,578 @@
...
+        self._clients = setup_clients(
+            ClientConfig(
+                timeout=self._client_config.timeout,
+                base_url=urls,
+                api_key_var=self._client_config.api_key_var,
+                headers=self._client_config.headers,
+            )
+        )
+        self._client_index = 0  # Reset round-robin on refresh
+        return True
```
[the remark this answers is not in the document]
JannikSt: this is on purpose to ensure old rollouts can complete

src/prime_rl/utils/client.py
```diff
@@ -11,6 +14,99 @@
...
+        self._clients = clients
+        self._admin_clients = admin_clients
+        self._elastic_pool = elastic_pool
+
+    @property
+    def clients(self) -> list[AsyncOpenAI]:
+        """Get inference clients (refreshed from elastic pool if applicable)."""
+        if self._elastic_pool is not None:
+            return self._elastic_pool.get_inference_clients()
+        return self._clients
```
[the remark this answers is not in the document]
JannikSt: low prio given its mainly an admin client at work here but will fix anyways

JannikSt:
@codex review

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -82,75 +83,85 @@ def extract_result(state: vf.State) -> dict:
...
-    client_cycle = cycle(clients)
     semaphore = asyncio.Semaphore(max_concurrent) if max_concurrent > 0 else asyncio.Semaphore(10000)
 
     # Start event loop lag monitor for this worker
     lag_monitor = EventLoopLagMonitor(interval=0.1)  # More frequent sampling for workers
     lag_monitor_task = asyncio.create_task(lag_monitor.run())
 
+    # Setup client discovery (elastic mode) or static client cycle
+    discovery = WorkerServerDiscovery(client_config, model_name) if client_config.is_elastic else None
+    static_cycle = cycle(clients) if clients else None
```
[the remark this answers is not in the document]
JannikSt: that's a temporary race condition that we are accepting

src/prime_rl/utils/elastic.py
```diff
@@ -0,0 +1,589 @@
...
+    """Lightweight server discovery for env workers.
+
+    Unlike ElasticInferencePool, this doesn't manage adapters or admin clients.
+    It just discovers and provides inference clients for rollouts, with
+    round-robin client selection.
+    """
+
+    def __init__(self, client_config: ClientConfig, model_name: str):
+        self._client_config = client_config
+        self._model_name = model_name
```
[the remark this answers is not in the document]
JannikSt: yes

JannikSt:
@codex review

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -169,7 +167,6 @@ async def orchestrate(config: OrchestratorConfig):
         else:
             checkpoint_step = config.ckpt.resume_step
 
-    # Setup scheduler (uses subprocess workers for env execution)
     scheduler = Scheduler(
         admin_clients=admin_clients,
         client_config=config.client,
```
Jackmin801: do we still need to pass these into Scheduler constructor if they can be gotten from inference_pool?

src/prime_rl/utils/elastic.py
```diff
@@ -0,0 +1,589 @@
...
+    sync_failures: int = 0
+
+
+@dataclass
+class DesiredAdapterState:
+    """Desired adapter state for all pods."""
+
+    lora_name: str | None = None
+    lora_path: Path | None = None
+    step: int = 0
```
Jackmin801: Do you think it makes sense to have one class `AdapterState`? `LoadedAdapter` and `DesiredAdapterState` are basically the same

src/prime_rl/utils/client.py
```diff
@@ -11,6 +14,99 @@
 from prime_rl.utils.config import ClientConfig
 from prime_rl.utils.logger import get_logger
 
+if TYPE_CHECKING:
+    from prime_rl.utils.elastic import ElasticInferencePool
+
+
+class InferencePool:
```
Jackmin801: discussed in person

src/prime_rl/utils/elastic.py
```diff
@@ -0,0 +1,589 @@
...
+    except Exception as e:
+        logger.debug(f"Failed to check server {url}: {e}")
+        return False, False
+
+
+async def discover_ready_servers(hostname: str, port: int, model_name: str) -> list[str]:
+    """Discover servers via DNS with majority vote logic.
+
+    - If NO servers have the model: return all healthy servers (base model mode)
+    - If ANY server has the model: return only those with it (adapter mode)
```
Jackmin801: I dont think `If NO servers have the model: return all healthy servers (base model mode)` is correct. discovery._model_name is updated based on request model names so it will start wth base model and then become the lora name when it sees a broadcast or after max_async steps. In both cases, we shouldnt be using that inference server if it doesnt have the model the request is looking for.

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -82,75 +83,89 @@ def extract_result(state: vf.State) -> dict:
...
+                    request = request_queue.get_nowait()
+                except queue.Empty:
+                    break
+                if request is None:  # Shutdown signal
+                    return
+
+                # Update discovery model name if it changed (e.g., switched to LoRA)
+                if discovery and request.model_name != discovery._model_name:
+                    discovery._model_name = request.model_name
+
```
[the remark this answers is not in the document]
Jackmin801: yea I think this makes sense

Jackmin801 (approved):
wowow lgtm

PR #1622, opened 2026-01-20
Add retry with exponential backoff for empty training batches

JannikSt:
## Summary
- Replaces hard crash assertion with retry logic when batch has 0 samples
- Retries up to 5 times with exponential backoff (30s -> 300s cap)
- Handles transient inference unavailability (DNS failures, connection errors)

---

> [!NOTE]
> Introduces resilient handling when a training step yields no samples to avoid hard crashes and transient inference outages.
> 
> - Replaces assert with retry logic when `TrainingBatch.examples` is empty, tracking `empty_batch_retries` (max 5)
> - Exponential backoff between retries: 30s, 60s, 120s, 240s, capped at 300s; raises `RuntimeError` after max attempts
> - Cancels in-flight `val_task` during retry to prevent background task buildup; resets retry counter on success
> - No other functional changes outside `src/prime_rl/orchestrator/orchestrator.py`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 45ae66eb28ab97dad6bb59b692eff99bce2c3908. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -249,6 +249,10 @@ async def orchestrate(config: OrchestratorConfig):
     # Start update policy loop
     update_policy_task = asyncio.create_task(scheduler.update_policy_loop())
 
+    # Track consecutive empty batches for retry logic
+    empty_batch_retries = 0
+    max_empty_batch_retries = 5
+
     while True:
         # Check if update_policy_task has failed and propagate the exception
         if update_policy_task.done():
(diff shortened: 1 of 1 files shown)


Jackmin801 (approved):
hrmm this should cover the case when all rollouts in the batch fail but I think ideally we retry whenever any rollouts fails. that way we fill full batches and dont sometimes have steps that have a lesser number of samples in them.

that said I think rn in hosted rl users are more often seeing all their samples fail so this is good

Jackmin801 (approved):
nice! yea lgtm. im wondering if this logic can be encapsulated such that its not so many lines in the orchestrator though. but good to merge and get it out first to see if it improves UX

PR #1624, opened 2026-01-20
add Temp Scheduler

hallerite:
> [!NOTE]
> Introduces configurable temperature scheduling and propagates per-token temperatures end-to-end through rollout, transport, packing, and training.
> 
> - Adds `sampling.temp_scheduler` (linear/cosine) with validation; defaults to `temperature=1.0` if neither is set. New `compute_temperature()` drives per-step temps; `get_sampling_args()` now requires an explicit `temperature` and strips scheduler details.
> - Orchestrator/Scheduler compute stepwise temperature and pass per-request `sampling_args`; env workers embed turn temperature into trajectories; metrics log `sampling/temperature`.
> - Transport types now carry `completion_temperatures`; trajectory builders populate it per token.
> - Trainer pipeline updated for per-token temps: `MicroBatch.temperatures`, packing/prepare logic, CP sharding, and loss. Model `forward()` and LM heads accept `temperature: Tensor`; fused LM head computes logprobs/entropy with per-token temps; vanilla path applies scaling externally.
> - Packer refactored to be temperature-agnostic (can pack mixed-temperature samples) and to group by run without mixing adapters.
> - Extensive unit tests added/updated for scheduler/worker, trajectories, batching/packing, and fused head behavior. CHANGELOG updated.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 376fda2747cb74a417d5ac9bd055f55fac843bf0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/rl/packer.py
+++ b/src/prime_rl/trainer/rl/packer.py
@@ -8,6 +8,7 @@ from transformers.tokenization_utils import PreTrainedTokenizer
 from prime_rl.trainer.batch import prepare_batch
 from prime_rl.trainer.runs import get_multi_run_manager
 from prime_rl.transport import (
+    MicroBatch,
     MicroBatchSender,
     TrainingSample,
     TransportConfigType,
@@ -267,12 +268,14 @@ class MultiPacker(BasePacker):
         selected_samples = self._select_samples_round_robin(token_budget)
         assert selected_samples, "No samples selected"
 
-        samples: list[TrainingSample] = []
-        idxs: list[int] = []
+        # Group samples by run_idx - each microbatch must contain samples from only ONE run
+        # because MultiLoRAGroupedExperts (MoE) only supports one adapter per microbatch
+        samples_by_run: dict[int, list[TrainingSample]] = {}
         per_run_stats: dict[int, tuple[int, int]] = {}
         for run_idx, sample, step in selected_samples:
-            samples.append(sample)
-            idxs.append(run_idx)
+            if run_idx not in samples_by_run:
+                samples_by_run[run_idx] = []
+            samples_by_run[run_idx].append(sample)
 
             num_tokens = len(sample.prompt_ids) + len(sample.completion_ids)
             if run_idx in per_run_stats:
@@ -284,16 +287,23 @@ class MultiPacker(BasePacker):
         for run_idx, (num_samples, num_tokens) in per_run_stats.items():
             self._update_run_progress(run_idx, num_samples, num_tokens)
 
-        micro_batch_grid = prepare_batch(
-            rollouts=samples,
-            seq_len=self.seq_len,
-            pad_to_multiple_of=self.pad_to_multiple_of,
-            num_train_workers=self.dp_world_size,
-            idxs=idxs,
-            num_loras=self.multi_run_manager.max_runs,
-        )
+        # Pack each run separately to ensure no mixing of runs in microbatches
+        all_micro_batches: list[list[MicroBatch]] = [[] for _ in range(self.dp_world_size)]
+        for run_idx in sorted(samples_by_run.keys()):
+            run_samples = samples_by_run[run_idx]
+            run_micro_batch_grid = prepare_batch(
+                rollouts=run_samples,
+                seq_len=self.seq_len,
+                pad_to_multiple_of=self.pad_to_multiple_of,
+                num_train_workers=self.dp_world_size,
+                idxs=[run_idx] * len(run_samples),
+                num_loras=self.multi_run_manager.max_runs,
+            )
+            # Merge into combined grid
+            for worker_idx, worker_batches in enumerate(run_micro_batch_grid):
+                all_micro_batches[worker_idx].extend(worker_batches)
 
-        self.sender.send(micro_batch_grid)
+        self.sender.send(all_micro_batches)
 
 
 def setup_packer(


Jackmin801:
hey @hallerite! nice PR, excited to be supporting temperature scheduling in the orchestrator. some quick comments:

1. You're probably going to have resolve a lot of merge conflicts when https://github.com/PrimeIntellect-ai/prime-rl/pull/1619 merges soon. Sorry about that. Its mostly renaming stuff so I would think the agents are capable enough to figure out the merge conflict resolution but yea sorry for the trouble.

2. Im not sure if grouping by temperature in the packer is necessary atm. We already group by run and implicitly by step (we never pack across step boundary for a run in a step) and so we will never be in a case where the prepare_batch loop sees any unique run having different temps in the same batch.

Though with async I guess we can have multiple sampling temperatures within a batch for a run with `max unique temps <= max async`. But in that case, I would think the better solution is to make the loss support passing a tensor of temperatures rather than packing such that similar temperatures are together.

src/prime_rl/orchestrator/config.py
```diff
@@ -78,6 +113,13 @@ class SamplingConfig(BaseConfig):
         ),
     ] = 1.0
 
+    temperature_schedule: Annotated[
```
```suggestion
    temp_scheduler: Annotated[
```

src/prime_rl/orchestrator/env_worker.py
```diff
@@ -379,7 +378,9 @@ async def collect_responses(self):
                     # Track successful responses; reset restart count after stable operation
                     self._responses_since_restart += 1
                     if self._responses_since_restart >= 10 and self._restart_count > 0:
-                        logger.debug(f"Worker '{self.worker_name}' stable after {self._responses_since_restart} responses, resetting restart count")
+                        logger.debug(
```
samsja, line 410: not sure what is up with your ruff , did you used the pre-commit hook ?

hallerite: I did now

src/prime_rl/orchestrator/config.py
```diff
@@ -67,16 +67,58 @@ class ModelConfig(BaseModelConfig):
     ] = None
 
 
+class TemperatureScheduleConfig(BaseConfig):
+    """Configures temperature scheduling over training steps. Use this OR sampling.temperature, not both."""
```
samsja, line 71: do we have a config failure for this case ?

samsja: i saw we have it after all good

src/prime_rl/trainer/models/layers/lm_head.py
```diff
@@ -40,15 +40,20 @@ def forward(
...
     ) -> PrimeLmOutput:
         assert labels is not None, "FusedOutputLinear requires labels for chunked logprob computation"
 
-        inv_t = 1.0 / float(temperature)
         b, s, h = hidden_states.shape
         hidden_states = hidden_states.reshape(b * s, h).contiguous()
         labels = labels.reshape(b * s).contiguous()
 
+        # Handle both scalar and per-token temperatures
+        if isinstance(temperature, Tensor):
```
samsja: I would just handle per token temp to be honest

src/prime_rl/trainer/rl/packer.py
```diff
@@ -263,38 +262,31 @@ def pack(self):
...
-                temperature=temperature,
-                seq_len=self.seq_len,
-                pad_to_multiple_of=self.pad_to_multiple_of,
-                num_train_workers=self.dp_world_size,
-                idxs=[run_idx] * num_samples,
-                num_loras=self.multi_run_manager.max_runs,
-            )
-
-            for i, micro_batch in enumerate(_micro_batch_grid):
-                micro_batch_grid[i].extend(micro_batch)
```
samsja, line 297: why did we need to change the order here ?

hallerite: We don't need to, but I think the old order was like this because we did not have access to the temp per-token and hence had to call `prepare_batch` per-run with the temp float

src/prime_rl/trainer/models/qwen3_moe/modeling_qwen3_moe.py
```diff
@@ -360,16 +360,16 @@ def forward(
         output_router_logits: Optional[bool] = None,
         cache_position: Optional[torch.LongTensor] = None,
         logits_to_keep: Union[int, torch.Tensor] = 0,
-        temperature: Optional[float] = 1.0,
+        temperature: Optional[torch.Tensor] = None,
```
samsja:
```suggestion
        temperature: torch.Tensor | None = None,
```

samsja:
move Optional[] -> move to " | "

src/prime_rl/trainer/rl/packer.py
```diff
@@ -263,38 +267,31 @@ def pack(self):
...
+        micro_batch_grid = prepare_batch(
+            rollouts=samples,
+            seq_len=self.seq_len,
+            pad_to_multiple_of=self.pad_to_multiple_of,
+            num_train_workers=self.dp_world_size,
+            idxs=idxs,
+            num_loras=self.multi_run_manager.max_runs,
+        )
```
Jackmin801: We still need to group by run because MultiLoRAMoE requires same run_idx in microbatch across all ranks. I would think you actually dont need to change much from what the code was before but make it that temperature doesnt get passed to prepare_batch now that it is in samples

hallerite: Are you sure? My understanding was that the old code actually grouped because of the temperature

Packing already sorts by `lora_idx` in batch.py:66-67 and the only cross-rank constraint seems to be same number of microbatches for FSDP sync, not same run composition (batch.py:147-153).

Running
```python
from prime_rl.trainer.batch import packed_samples_into_micro_bs, MicroBatch

# Create samples from different runs (simulating mixed-run input)
# run_idx, MicroBatch
samples = [
    (2, MicroBatch(input_ids=[1,2,3], loss_mask=[True]*3, advantages=[0.0]*3,
                  inference_logprobs=[0.0]*3, temperatures=[1.0]*3, position_ids=[0,1,2],
                  lora_num_tokens=[], teacher_logprobs=None)),
    (0, MicroBatch(input_ids=[4,5], loss_mask=[True]*2, advantages=[0.0]*2,
                  inference_logprobs=[0.0]*2, temperatures=[1.0]*2, position_ids=[0,1],
                  lora_num_tokens=[], teacher_logprobs=None)),
    (1, MicroBatch(input_ids=[6,7,8,9], loss_mask=[True]*4, advantages=[0.0]*4,
                  inference_logprobs=[0.0]*4, temperatures=[1.0]*4, position_ids=[0,1,2,3],
                  lora_num_tokens=[], teacher_logprobs=None)),
    (0, MicroBatch(input_ids=[10,11], loss_mask=[True]*2, advantages=[0.0]*2,
                  inference_logprobs=[0.0]*2, temperatures=[1.0]*2, position_ids=[0,1],
                  lora_num_tokens=[], teacher_logprobs=None)),
]

# Pack with max_seq_len=20, num_loras=3
result = packed_samples_into_micro_bs(samples, max_seq_len=20, num_loras=3)

for i, mb in enumerate(result):
    print(f"Microbatch {i}:")
    print(f"  input_ids: {mb.input_ids}")
    print(f"  lora_num_tokens: {mb.lora_num_tokens}")
```
yields

```
Input: samples from runs [2, 0, 1, 0] (out of order, mixed)

Output: 1 microbatch(es)

Microbatch 0:
  input_ids: [4, 5, 10, 11, 6, 7, 8, 9, 1, 2, 3]
  lora_num_tokens: [4, 4, 3]
  Adapter 0: tokens 0 to 3 (4 tokens)
  Adapter 1: tokens 4 to 7 (4 tokens)
  Adapter 2: tokens 8 to 10 (3 tokens)
```

which seems correct.

UPDATE:

Yeah, you are right. This would not support Multi-Lora for MoEs.

src/prime_rl/trainer/models/qwen3_moe/modeling_qwen3_moe.py
```diff
@@ -360,7 +360,7 @@ def forward(
         output_router_logits: Optional[bool] = None,
         cache_position: Optional[torch.LongTensor] = None,
         logits_to_keep: Union[int, torch.Tensor] = 0,
-        temperature: Optional[torch.Tensor] = None,
+        temperature: Union[torch.Tensor, None] = None,
```
samsja, line 363: lets use | instead of Union

Jackmin801:
just a quick clarification. the temperature used in the TrainingBatch isnt exactly correct right? it uses what the scheduler currently has but the rollouts could have been scheduled from before and thus have actually used slightly different temp. guessing it doesnt matter though, just wanted to clarify if it was true

hallerite:
> just a quick clarification. the temperature used in the TrainingBatch isnt exactly correct right? it uses what the scheduler currently has but the rollouts could have been scheduled from before and thus have actually used slightly different temp.

The temperature stored in each trajectory step comes from request.sampling_args["temperature"], i.e. the temperature that was in the sampling_args at the time the request was submitted, not the current scheduler temperature, so I think it's correct. @Jackmin801

PR #1629, opened 2026-01-21
Assert when only trainer or orchestrator wandb is configured

kalomaze:
Previously, users could specify [trainer.wandb] or [orchestrator.wandb] independently and it would silently log only one of the two wandb run pages. Now, we assert in the case that the user is missing one or the other.

---

> [!NOTE]
> **W&B config validation tightened**
> 
> - In `validate_shared_wandb_config`, now raises if `trainer.wandb` is set without `orchestrator.wandb`, or vice versa, with guidance to configure both or use shared `[wandb]`.
> - Retains check that both `wandb.project` values match when both are present.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 398069d61ccb3365039fe6743dabc866d8b4da1d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/utils/validation.py
+++ b/src/prime_rl/utils/validation.py
@@ -61,6 +61,16 @@ def validate_shared_wandb_config(
     trainer: RLTrainerConfig,
     orchestrator: OrchestratorConfig,
 ) -> None:
+    if trainer.wandb and not orchestrator.wandb:
+        raise ValueError(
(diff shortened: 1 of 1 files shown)


samsja (approved):

Jackmin801 (approved):
lgtm!

PR #1653, opened 2026-01-24
Add missing [ckpt] section to reverse_text rl.toml

minpeter:
## Summary
- Add `[ckpt]` section to `examples/reverse_text/rl.toml` to enable weight checkpoint saving
- Without this section, `outputs/weights/step_20` is never created despite README.md documenting its existence

## Problem
The README.md states that after running RL training, a checkpoint will be saved at `outputs/weights/step_20`. However, the `rl.toml` config was missing the `[ckpt]` section.

When `[ckpt]` is absent, `setup_ckpt_managers()` in `src/prime_rl/trainer/ckpt.py` returns `(None, None)`, causing no checkpoints to be saved.

## Test plan
- [x] Run `uv run rl @ examples/reverse_text/rl.toml` and verify `outputs/weights/step_20` is created

### DIFF
--- a/examples/reverse_text/rl/train.toml
+++ b/examples/reverse_text/rl/train.toml
@@ -5,3 +5,5 @@ name = "PrimeIntellect/Qwen3-0.6B-Reverse-Text-SFT"
 
 [optim]
 lr = 3e-6
+
+[ckpt] # Checkpoint at the end of training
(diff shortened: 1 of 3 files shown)


minpeter:
I created this PR using Claude Code, but it’s a minor issue I noticed while following the example code. I’d appreciate it if you could keep that in mind during your review. (I made sure to avoid producing any AI slop. 😉)

PR #1654, opened 2026-01-24
Fix tokenizer name not updated when using shared model config

minpeter:
## Summary
- Fix bug where `trainer.tokenizer.name` was not updated when using `--model.name` CLI option
- This occurred due to Pydantic's nested model validation order: `RLTrainerConfig.auto_setup_tokenizer` runs before `RLConfig.auto_setup_model`

## Problem
When running:
```bash
uv run rl @ examples/reverse_text/rl.toml --model.name minpeter/Qwen3-0.6B-Reverse-Text-SFT
```

The tokenizer was being loaded from the default model (`Qwen/Qwen3-0.6B`) instead of the specified model, causing the saved checkpoint to have incorrect tokenizer/chat_template.

## Solution
Update `trainer.tokenizer.name` in `auto_setup_model` if it was auto-configured from the old model name.

## Test plan
- [ ] Run RL training with `--model.name` and verify tokenizer is loaded from the correct model

### DIFF
--- a/tests/unit/test_configs.py
+++ b/tests/unit/test_configs.py
@@ -53,10 +53,10 @@ def test_rl_config_updates_tokenizer_name_when_not_explicit():
     config = RLConfig(
         trainer=trainer,
         orchestrator=orchestrator,
-        model={"name": "minpeter/Qwen3-0.6B-Reverse-Text-SFT"},
+        model={"name": "PrimeIntellect/Qwen3-0.6B-Reverse-Text-SFT"},
     )
-    assert config.trainer.model.name == "minpeter/Qwen3-0.6B-Reverse-Text-SFT"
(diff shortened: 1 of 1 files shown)


minpeter:
@samsja @mikasenghaas, Could you please review this PR?

minpeter:
Ci failed 👀  (I'll check)

samsja:
hey @minpeter is the goal to be using a model with a different tokenizer that its model name ?

minpeter:
> hey @minpeter is the goal to be using a model with a different tokenizer that its model name ?

No, there's currently a bug where specifying only the model name forces the tokenizer to default to Qwen/Qwen3-0.6B. This fixes this, but you can still choose a different tokenizer.

PR #1657, opened 2026-01-25
fix checkpoint resume with NCCL broadcast mode

samsja:
The `.exists()` check introduced in d7b0be7e broke checkpoint resume with NCCL broadcast mode. In NCCL mode, weights are broadcasted directly from trainer to inference servers, not stored on disk. When resuming from checkpoint, `get_weight_dir()` would raise `FileNotFoundError` because the broadcast directory doesn't exist yet.

Changes:
- Add `signal_nccl_ready()` function to decouple NCCL readiness signaling from `update_weights()` - this separates the coordination concern from the weight loading concern
- Remove NCCL_READY marker creation from `update_weights()` since it's now handled explicitly where needed
- Add `check_exists` parameter to `get_weight_dir()` - when False, returns the broadcast directory path without checking existence (for NCCL mode)
- Update orchestrator checkpoint resume logic to skip existence check and only signal NCCL readiness in NCCL mode (no weight loading needed)
- Update scheduler to explicitly call `signal_nccl_ready()` before `update_weights()` in NCCL mode during normal training

The NCCL_READY marker path remains unchanged:
`output_dir/broadcasts/step_{step}/NCCL_READY`

---

> [!NOTE]
> Ensures checkpoint resume works in NCCL broadcast mode by separating readiness signaling from weight loading and avoiding premature filesystem checks.
> 
> - Add `signal_nccl_ready()` to create `NCCL_READY` marker independently of weight updates
> - Remove NCCL marker creation from `update_weights()`; call `signal_nccl_ready()` explicitly where needed
> - Extend `get_weight_dir(output_dir, step, check_exists)` to allow bypassing existence checks in NCCL mode
> - Orchestrator: on resume, skip weight dir existence check; in NCCL mode only signal readiness (no disk load); otherwise load from filesystem
> - Scheduler: before `update_weights()` in NCCL mode, signal readiness for the current step
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 32f2bd781e066eadc27c8d17a033777e8ae4048e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -43,6 +43,7 @@ from prime_rl.utils.client import (
     setup_admin_clients,
     setup_clients,
     setup_inference_pool,
+    signal_nccl_ready,
 )
 from prime_rl.utils.heartbeat import Heartbeat
 from prime_rl.utils.logger import intercept_verifiers_logging, setup_logger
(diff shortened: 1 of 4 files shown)


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -232,9 +233,17 @@ async def orchestrate(config: OrchestratorConfig):
...
+        check_exists = config.weight_broadcast.type != "nccl"
+        weights_path = get_weight_dir(config.output_dir, scheduler.ckpt_step, check_exists=check_exists)
+        if config.weight_broadcast.type == "nccl":
+            # Signal readiness for NCCL broadcast - trainer will send weights when it starts
+            signal_nccl_ready(weights_path)
+            logger.info(f"NCCL mode: signaled readiness at {weights_path}, waiting for trainer to broadcast weights")
+        else:
+            # Filesystem mode: load weights from disk
+            lora_name = config.model.lora.name if config.model.lora else None
+            await inference_pool.update_weights(weights_path, lora_name=lora_name, step=scheduler.ckpt_step)
```
Jackmin801, line 246: can we encapsulate this into the nccl broadcast class?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -207,6 +207,10 @@ async def update_policy(self):
             # Update weights on inference servers
             update_weights_start_time = time.perf_counter()
             weights_path = get_step_path(get_broadcast_dir(self.config.output_dir), next_ckpt_step)
+
+            if self.config.weight_broadcast.type == "nccl":
```
Jackmin801, line 211: same here. I would think the if elses arent needed other than the one that picks which broadcast class to create.

PR #1670, opened 2026-01-27
[fix] zombie guard on checkpoint

Jackmin801:
Fixes two issues:
1. If a run gets deleted while we are checkpointing, the trainer will crash because it will try to write stable file into non-existent directory. We need to include the stable writing in the try except that ignores this and also try to cleanup the run directory if we created it while it was being deleted.
2. If the trainer crashes when a step has orchestrator ckpt but not trainer ckpt, the respawned trainer will try to resume from this step and fail because it doesnt have trainer ckpt. We thus need to check trainer ckpt stable file before resuming in the trainer.

> [!NOTE]
> **Medium Risk**
> Touches distributed checkpoint save/resume ordering and run lifecycle cleanup; race conditions or incorrect barriers could cause missed checkpoints or accidental directory deletion.
> 
> **Overview**
> Fixes multi-run checkpoint stability and cleanup behavior. Multi-run resume/progress discovery now only considers *stable* checkpoints (those with a `STABLE` marker) via new `get_stable_ckpt_steps`, avoiding resuming from partially-written steps.
> 
> During `MultiCheckpointManager.save`, the `STABLE` marker is now written only after all ranks synchronize, and master additionally deletes a run directory if the run was removed mid-checkpoint to prevent recreating “zombie” runs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a1501d04e99b658cdc78faae209677d530d60146. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/multi_ckpt.py
+++ b/src/prime_rl/trainer/multi_ckpt.py
@@ -212,9 +212,6 @@ class MultiCheckpointManager:
             state_dict = torch.load(ckpt_path / f"rank_{self.world.rank}.pt")
             run_state.load_state_dict(state_dict)
 
-            # Set progress.step to match the loaded checkpoint
-            self.multi_run_manager.progress[idx].step = step
-
             self.logger.info(f"Resumed run {self.multi_run_manager.idx_2_id[idx]} from step {step}")
             return True
         except Exception as e:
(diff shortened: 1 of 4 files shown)


samsja (approved):
lgtm

PR #1674, opened 2026-01-28
best-effort interleaving

hallerite:
Deprecating the explicit duality of interleave vs branching in favor of "best effort interleaving", i.e. interleaving where possible and branching, otherwise. If two consecutive trajectory steps in a rollout fulfill the extension property – meaning that the earlier's prompt + completion tokens is a true prefix of the latter's prompt tokens – they are merged into one training sample (following the earlier interleaving strategy). Otherwise, we create 2 samples (following the earlier branching strategy). This way we get O(T) where we can and only get O(T^2) where there is a prefix mismatch.

---

> [!NOTE]
> **Medium Risk**
> Changes core rollout-to-training-sample conversion logic and removes a configuration switch, which can affect training data construction and throughput across multi-turn and multimodal workloads. The surface area is well-covered by new unit tests, but regressions could impact correctness if prefix matching or image attachment behavior differs from expectations.
> 
> **Overview**
> Implements **best-effort interleaving** for multi-turn rollouts: `interleave_rollout` now merges trajectory steps when the prompt tokens extend any prior step’s prompt+completion prefix, and automatically starts a new `TrainingSample` when the prefix match breaks (including multi-prefix tracking for interleaved agents). This replaces the explicit `branch_rollout` path and adds VLM support by updating each merged sample with cumulative `VLMImageCache` images up to its last merged step.
> 
> Configuration and runtime behavior are simplified by removing `orchestrator.trajectory_strategy` and forcing `env.set_interleaved_rollouts(True)` in the orchestrator, scheduler, and env workers; multimodal docs/configs are updated accordingly, and unit tests are rewritten/expanded to cover extension breaks, interleaved agents, error masking, and VLM image accumulation semantics.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ecc12f4c13ed3e77b0d649e3663ad6a89aa99c62. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/orchestrator/config.py
```diff
@@ -707,7 +707,7 @@ class OrchestratorConfig(BaseSettings):
     trajectory_strategy: Annotated[
```
mikasenghaas: we can just deprecate, and denote this in the CHANGELOG.md

src/prime_rl/orchestrator/trajectories.py
```diff
@@ -23,92 +25,74 @@ def interleave_rollout(state: vf.State) -> list[TrainingSample] | None:
...
-    rollouts = []
-    trajectory = state["trajectory"]
-    if len(trajectory) == 0:
-        logger.warning(f"No trajectory steps for example {state['example_id']}. Skipping rollout.")
-        return None
-
-    has_error = state["error"] is not None
-    for step in state["trajectory"]:
-        assert "tokens" in step
+    def make_sample(step: dict) -> TrainingSample:
```
mikasenghaas: we can actually type this as `step: TrajectoryStep` right?

mikasenghaas (approved):
two mini comment, but looks great --preapprove

src/prime_rl/orchestrator/trajectories.py
hallerite: we add temperatures per token in prime-rl, but I will create a type for it for better readability

snimu (approved):

snimu (approved):

PR #1684, opened 2026-01-29
[QoL] Multiple small sanity checks

kalomaze:
- validate orchestrator client port matches inference server port
- error if existing GPU processes are on target GPUs before training
- fix trainer to respect `resume_step` config (prevents deadlock on restart)

---

> [!NOTE]
> **Medium Risk**
> Touches RL process startup and checkpoint resume/progress logic; failures could block launches or resume from the wrong step if assumptions about NVML availability or config values are incorrect.
> 
> **Overview**
> Adds preflight sanity checks to `rl.py`: the launcher now errors early if any target GPUs already have compute processes (via NVML), and it validates that `orchestrator.client.base_url`’s port matches `inference.server.port` when using a non-elastic client.
> 
> Updates `MultiRunManager._create_run_data` in `trainer/runs.py` to set initial run progress from `config.ckpt.resume_step` (including `-1` meaning “latest stable checkpoint”), aligning trainer behavior with the orchestrator and avoiding incorrect restart behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e17c28c67f9e43942917cffc3e76d2e336660e1e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/rl.py
```diff
@@ -546,6 +573,34 @@ def monitor_process(process: Popen, stop_event: Event, error_queue: list, proces
         stop_event.set()
 
 
+def check_gpus_available(gpu_ids: list[int]) -> None:
+    """Raise error if there are existing processes on the specified GPUs."""
+    try:
```
samsja: lets not to try expect on importat, either the lib is here or we add it to pyproecrt toml

src/prime_rl/rl.py
```diff
@@ -546,6 +573,34 @@ def monitor_process(process: Popen, stop_event: Event, error_queue: list, proces
...
+        import pynvml
+
+        pynvml.nvmlInit()
+    except Exception:
+        return  # pynvml not available, skip check
+
+    occupied = []
+    for gpu_id in gpu_ids:
+        try:
```
samsja: not super fan of the try expeft here would rather fail

samsja (approved):

src/prime_rl/rl.py
```diff
@@ -513,6 +514,28 @@ def validate_teacher_model(self):
             )
         return self
 
+    @model_validator(mode="after")
+    def validate_client_port_matches_inference(self):
```
samsja: lets do this check outside of the configs. config check should be purely decouple from runtime

samsja (changes requested):
see comment

samsja (approved):

src/prime_rl/rl.py
```diff
@@ -564,6 +585,25 @@ def rl(config: RLConfig):
...
+
+    # Validate client port matches inference server port
+    if config.inference is not None and not config.orchestrator.client.is_elastic:
+        from urllib.parse import urlparse
+
+        base_url = config.orchestrator.client.base_url[0]
+        parsed = urlparse(base_url)
+        client_port = parsed.port
+        expected_port = config.inference.server.port
+        if client_port != expected_port:
```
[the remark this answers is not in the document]
kalomaze: base urls are expected to have these

src/prime_rl/rl.py
```diff
@@ -564,6 +585,25 @@ def rl(config: RLConfig):
     logger.info("Starting RL run")
     logger.debug(f"RL start command: {' '.join(start_command)}")
 
+    # Check for existing processes on GPUs
+    all_gpu_ids = list(set(config.inference_gpu_ids + config.trainer_gpu_ids + (config.teacher_gpu_ids or [])))
+    check_gpus_available(all_gpu_ids)
```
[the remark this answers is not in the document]
kalomaze: user should set inference_gpu_ids = [] for this case OR we should make it so a run cant start with any inference gpu IDs set while also having inference set to None (which would be an out of scope change)

PR #1692, opened 2026-01-30
[fix] remove cc check for grouped mm

Jackmin801:
grouped mm support CC >= 8 which is basically all the cards we care about. ppl who are still running v100 probably have a lot more troubles than just this grouped mm being unhappy with them

---

> [!NOTE]
> **Medium Risk**
> Changes the runtime selection of the grouped GEMM code path in core MoE/LoRA compute; on unsupported GPU/PyTorch combinations this could surface as performance regressions or runtime errors rather than falling back to the for-loop.
> 
> **Overview**
> Removes CUDA compute capability checks that previously disabled `use_grouped_mm` on non-Hopper GPUs for MoE and multi-LoRA layers, relying instead on the existing 8-byte alignment constraints to decide whether to use grouped GEMM.
> 
> `MultiLoRAGroupedExperts` now accepts a `use_grouped_mm` constructor flag (defaulting to `True`) rather than hard-forcing a per-GPU override, and `MoE` no longer mutates `moe_args.use_grouped_mm` based on detected device capability.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e980ff323e3d56666256a0b7197253ca610b8b08. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/models/layers/lora/multi_linear.py
+++ b/src/prime_rl/trainer/models/layers/lora/multi_linear.py
@@ -53,13 +53,6 @@ class MultiLoRALinear(MultiLoRAModule):
         if rank <= 0 or n_adapters <= 0:
             raise ValueError("rank and n_adapters must be > 0")
 
-        # Set use_grouped_mm to False if CUDA compute capability < 9.0
-        if torch.cuda.is_available():
-            cc_major, _ = torch.cuda.get_device_capability()
-            if cc_major != 9:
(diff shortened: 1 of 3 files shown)


samsja (approved):
LGTM FASTER 3090 training is all I need

S1ro1 (approved):
lgtm

PR #1694, opened 2026-01-30
Add optimizer state CPU offload

samsja:
## Summary

Adds `CPUOffloadOptimizer` - keeps optimizer states on CPU, moves to GPU only during `optimizer.step()`.

Unlike FSDP's `CPUOffload`, this keeps weights on GPU (no H2D all-gather on forward pass).

## Key insight

With activation checkpointing, activations and optimizer states are never on GPU simultaneously:
- Forward/backward: activations on GPU, optimizer states on CPU
- Optimizer step: optimizer states on GPU, activations freed

Peak memory becomes `max(activations, optimizer_states)` instead of `activations + optimizer_states`.

## Results

**4B model, 4 GPUs, AC enabled:**
- 2 GB saved per GPU (~8% reduction)
- ~2% slower

Best for memory-constrained scenarios where fitting larger model/batch matters more than throughput.

## Usage

```toml
[model]
optim_cpu_offload = true
```

ckpt  resume tests
image

---

> [!NOTE]
> **Medium Risk**
> Touches optimizer stepping, scheduler binding, and checkpoint save/load paths; mistakes could break resume correctness or cause device-transfer/perf issues during training.
> 
> **Overview**
> Adds **optimizer-state CPU offloading** via a new `CPUOffloadOptimizer` wrapper that keeps optimizer states on CPU and transfers them to CUDA only during `step()`.
> 
> Introduces `model.optim_cpu_offload` config (mutually exclusive with `fsdp_cpu_offload`) and wires it into both RL and SFT trainers, with RL additionally disallowing use when `max_concurrent_runs > 1`. Checkpointing and scheduler setup are updated to unwrap/wrap base optimizers so optimizer state save/load works correctly when offloading is enabled.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b6a64f1c47f7c0c6c66f3e19775cfb70126b7783. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -167,6 +167,13 @@ class ModelConfig(BaseConfig):
         ),
     ] = False
 
+    optim_cpu_offload: Annotated[
+        bool,
+        Field(
+            description="Whether to enable optimizer state CPU offloading. Unlike fsdp_cpu_offload, this only moves optimizer states (momentum, variance) to CPU, keeping weights on GPU. This avoids the H2D all-gather overhead while still saving GPU memory.",
+        ),
+    ] = False
+
     reshard_after_forward: Annotated[
         bool, Field(description="Whether to reshard the model after each forward pass.")
     ] = True
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -301,6 +301,12 @@ class ModelConfig(BaseConfig):
         return self
 
     @model_validator(mode="after")
+    def cpu_offload_mutual_exclusion(self):
+        if self.fsdp_cpu_offload and self.optim_cpu_offload:
+            raise ValueError("Cannot enable both fsdp_cpu_offload and optim_cpu_offload. Use one or the other.")
+        return self
+
+    @model_validator(mode="after")
     def fused_lm_head_chunk_size_is_valid(self):
         if isinstance(self.fused_lm_head_chunk_size, int):
             low = 512
--- a/src/prime_rl/trainer/optim.py
+++ b/src/prime_rl/trainer/optim.py
@@ -4,6 +4,7 @@ from typing import Callable
 import torch
 from dion import Muon
 from torch import nn
+from torch.distributed.tensor import DTensor
 from torch.optim import SGD, AdamW, Optimizer
 
 from prime_rl.trainer.config import OptimizerConfigType
@@ -27,8 +28,6 @@ class CPUOffloadOptimizer:
         self._initialized = False
 
     def _move_states(self, device: str):
-        from torch.distributed.tensor import DTensor
-
         for p in self.optimizer.state:
             state = self.optimizer.state[p]
             for k, v in state.items():
--- a/src/prime_rl/trainer/optim.py
+++ b/src/prime_rl/trainer/optim.py
@@ -29,6 +29,7 @@ class CPUOffloadOptimizer:
         self._initialized = False
 
     def _move_states(self, device: str):
+        """Move optimizer states to CPU or back to GPU (matching each parameter's device)."""
         for p in self.optimizer.state:
             state = self.optimizer.state[p]
             for k, v in state.items():
--- a/src/prime_rl/trainer/optim.py
+++ b/src/prime_rl/trainer/optim.py
@@ -41,10 +41,7 @@ class CPUOffloadOptimizer:
                         if self.pin_memory and not new_local.is_pinned():
                             new_local = new_local.pin_memory()
                     else:
-                        target_device = torch.device("cuda")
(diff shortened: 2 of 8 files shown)


src/prime_rl/trainer/optim.py
```diff
@@ -12,12 +13,104 @@
...
+    at the same time: peak memory becomes max(activations, opt_states) instead of sum.
+    """
+
+    def __init__(self, optimizer: Optimizer, pin_memory: bool = True):
+        self.optimizer = optimizer
+        self.pin_memory = pin_memory
+        self._initialized = False
+
+    def _move_states(self, device: str):
+        from torch.distributed.tensor import DTensor
```
Jackmin801: move this import up?

mikasenghaas (commented):
do we wanna deprecate the fsdp offload then?

samsja:
> do we wanna deprecate the fsdp offload then?

I would still keep fsdp offload, its very minor right now and can be usefull for very low gpu ressources

src/prime_rl/trainer/optim.py
samsja: @cursoragent can u move the import up ?

S1ro1 (approved):
Do you have some plots proving convergence? Ideally w checkpoint restart. In general looks good but skeptical without proof

src/prime_rl/trainer/optim.py
```diff
@@ -12,12 +15,110 @@
...
+                    local_tensor = v._local_tensor
+                    if device == "cpu":
+                        non_blocking = not self.pin_memory
+                        new_local = local_tensor.to("cpu", non_blocking=non_blocking)
+                        if self.pin_memory and not new_local.is_pinned():
+                            new_local = new_local.pin_memory()
+                    else:
+                        target_device = torch.device("cuda")
+                        if isinstance(p, DTensor):
+                            target_device = p._local_tensor.device
```
S1ro1: Why are we overriding the target device here? Shouldn't we send to a device as specified by the function argument?

src/prime_rl/trainer/optim.py
```diff
@@ -12,12 +15,110 @@
...
+                    state[k] = new_dtensor
+                elif isinstance(v, torch.Tensor):
+                    if device == "cpu":
+                        non_blocking = not self.pin_memory
+                        cpu_tensor = v.to("cpu", non_blocking=non_blocking)
+                        if self.pin_memory and not cpu_tensor.is_pinned():
+                            cpu_tensor = cpu_tensor.pin_memory()
+                        state[k] = cpu_tensor
+                    else:
+                        target_device = p.device if hasattr(p, "device") else "cuda"
```
S1ro1: Likewise here

samsja:
Re: target_device override question - The `device` arg is a direction flag, not a specific target. When moving back to GPU, we need to match each state tensor to its parameter's device (cuda:0, cuda:1, etc.) for multi-GPU FSDP. Using just `"cuda"` would default to the current device, which could be wrong.

src/prime_rl/trainer/scheduler.py
```diff
@@ -1,14 +1,19 @@
...
 from prime_rl.utils.logger import get_logger
 
-if TYPE_CHECKING:
-    from prime_rl.trainer.optim import MultiLoRAOptimizer
+
+def _get_base_optimizer(optimizer: Optimizer | CPUOffloadOptimizer) -> Optimizer:
+    """Extract the base optimizer from a potentially wrapped optimizer."""
+    if isinstance(optimizer, CPUOffloadOptimizer):
+        return optimizer.optimizer
+    return optimizer
```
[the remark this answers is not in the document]
S1ro1: What is being suggested here is good, we should use the `._get_base_optimizer` or just match them

S1ro1 (approved):
lgtm

src/prime_rl/trainer/config.py
```diff
@@ -167,6 +167,13 @@ class ModelConfig(BaseConfig):
         ),
     ] = False
 
+    optim_cpu_offload: Annotated[
+        bool,
+        Field(
+            description="Whether to enable optimizer state CPU offloading. Unlike fsdp_cpu_offload, this only moves optimizer states (momentum, variance) to CPU, keeping weights on GPU. This avoids the H2D all-gather overhead while still saving GPU memory.",
+        ),
+    ] = False
```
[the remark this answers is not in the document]
S1ro1: Also this lol???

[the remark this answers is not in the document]
samsja: NAH WHAT ?????????????????????

PR #1712, opened 2026-02-03
fix: use stable checkpoints for orchestrator resume

JannikSt:
## Summary
- Only consider checkpoints with STABLE file when resuming orchestrator
- Fixes crash when orchestrator restarts before trainer finishes saving weights
- Uses existing `get_stable_ckpt_steps` instead of `get_all_ckpt_steps` in `resolve_latest_ckpt_step`

---

> [!NOTE]
> **Medium Risk**
> Touches checkpoint/weight-loading on resume and adds blocking wait logic, which could impact startup behavior or hang if misconfigured, but is scoped to resume paths.
> 
> **Overview**
> Improves orchestrator resume behavior by selecting weight directories only once a `STABLE` marker is present, avoiding crashes when the orchestrator restarts while the trainer is still writing a checkpoint.
> 
> Adds `ckpt.wait_for_weights_timeout` to optionally wait for the stable weights directory to appear on resume, and threads this through `orchestrator.py` into `get_weight_dir`, which now prefers stable checkpoint/broadcast paths and otherwise errors (or falls back to broadcast path when existence checks are disabled).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 55b24625e5906b7f3ba983f2370b66397f3f5f20. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/utils.py
+++ b/src/prime_rl/orchestrator/utils.py
@@ -186,9 +186,9 @@ def get_weight_dir(output_dir: Path, step: int, check_exists: bool = True, wait_
     broadcast_weight_dir = get_step_path(get_broadcast_dir(output_dir), step)
 
     def check_dirs() -> Path | None:
-        if ckpt_weight_dir.exists():
+        if (ckpt_weight_dir / "STABLE").exists():
             return ckpt_weight_dir
-        if broadcast_weight_dir.exists():
+        if (broadcast_weight_dir / "STABLE").exists():
             return broadcast_weight_dir
         return None
 
--- a/src/prime_rl/orchestrator/utils.py
+++ b/src/prime_rl/orchestrator/utils.py
@@ -188,11 +188,11 @@ def get_weight_dir(output_dir: Path, step: int, check_exists: bool = True, wait_
     def find_stable_dir() -> Path | None:
         # For checkpoint weights, check STABLE file in parent directory (checkpoints/step_{step}/STABLE)
         ckpt_step_dir = get_step_path(get_ckpt_dir(output_dir), step)
-        if (ckpt_step_dir / "STABLE").exists():
+        if (ckpt_step_dir / "STABLE").exists() and ckpt_weight_dir.exists():
             return ckpt_weight_dir
 
         # For broadcast weights, check STABLE file in the broadcast directory itself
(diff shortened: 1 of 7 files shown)


Jackmin801:
Hrmm I think this breaks the single tenant `--resume-step -1` case. the test passes because i believe it passes `--resume-step 20` but if it has `--resume-step -1` it would now start from scratch

src/prime_rl/orchestrator/utils.py
```diff
@@ -178,13 +180,36 @@ def get_weight_dir(output_dir: Path, step: int, check_exists: bool = True) -> Pa
...
-    if ckpt_weight_dir.exists():
-        return ckpt_weight_dir
-    if broadcast_weight_dir.exists():
-        return broadcast_weight_dir
+
+    def check_dirs() -> Path | None:
+        if ckpt_weight_dir.exists():
+            return ckpt_weight_dir
+        if broadcast_weight_dir.exists():
+            return broadcast_weight_dir
```
Jackmin801: I think for both of these you need to check that their stable file exists and then return that dir

src/prime_rl/orchestrator/utils.py
```diff
@@ -178,13 +180,36 @@ def get_weight_dir(output_dir: Path, step: int, check_exists: bool = True) -> Pa
...
     broadcast_weight_dir = get_step_path(get_broadcast_dir(output_dir), step)
-    if ckpt_weight_dir.exists():
-        return ckpt_weight_dir
-    if broadcast_weight_dir.exists():
-        return broadcast_weight_dir
+
+    def check_dirs() -> Path | None:
+        if ckpt_weight_dir.exists():
+            return ckpt_weight_dir
+        if broadcast_weight_dir.exists():
```
Jackmin801:
```suggestion
        if (ckpt_weight_dir / "STABLE").exists():
            return ckpt_weight_dir
        if (broadcast_weight_dir / "STABLE").exists():
```

Jackmin801 (approved):
fixing the issue with checking stable file should make it work

Jackmin801 (approved):

PR #1715, opened 2026-02-03
Feature: bring your own algorithms

samsja:
## Summary

Adds support for custom algorithms ("bring your own"), allowing researchers to experiment with different RL objectives and techniques.

### 1. Custom Loss Functions

- Loss is computed **per-sequence** - you provide a function for a single sample
- `LossInputs` / `LossOutputs` dataclasses define the interface
- `CustomLossConfig` with `byo_function` and `kwargs`

### 2. Custom Advantage Functions

- Advantages are computed **per-problem** (grouped by samples_per_problem)
- `AdvantageInputs` / `AdvantageOutputs` dataclasses define the interface
- `CustomAdvantageConfig` with `byo_function` and `kwargs`

### Example configs

```toml
# Custom loss
[loss]
byo_function = "my_module.ppo_clip_loss"
kwargs = { clip_eps = 0.2 }

# Custom advantage
[advantage]
byo_function = "my_module.gae_advantage"
kwargs = { gamma = 0.99, lam = 0.95 }
```

See `docs/bring-your-own-algorithms.md` for full documentation.

## Test plan

- [x] Unit tests for custom loss
- [x] Unit tests for custom advantage
- [x] Existing tests still pass

---

> [!NOTE]
> **Medium Risk**
> Touches core training math and configuration parsing; while defaults are preserved, dynamic imports and changed loss/metric aggregation could affect training stability and logging if misconfigured.
> 
> **Overview**
> Enables *bring-your-own* RL components by allowing users to configure and dynamically import **custom loss** and **custom advantage** functions via `type="custom"` + `import_path` + `kwargs`.
> 
> The trainer refactors loss computation to be **per-sequence** using new `LossInputs`/`LossOutputs` dataclasses, a `setup_loss_fn()` factory, and generic metric aggregation; training logging now tolerates missing default metrics like `mismatch_kl` when using custom losses. The orchestrator similarly introduces `AdvantageInputs`/`AdvantageOutputs`, `setup_advantage_fn()`, and a discriminated union config type to support external advantage implementations, plus adds a shared `import_object()` helper, tests, and new documentation (`docs/bring-your-own-algorithms.md`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 559f7ab99baadd7eb3bfc6c51a5dfa2e35426a7a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


docs/bring-your-own-algorithms.md
```diff
@@ -0,0 +1,138 @@
...
+[loss]
+byo_function ="my_module.ppo_clip_loss"
+kwargs = { clip_eps = 0.2 }
+```
+
+---
+
+## 2. Custom Advantage Functions
+
+Advantages are computed **per-problem** (grouped by `samples_per_problem`). You provide a function that computes advantages for a batch of problems.
```
mikasenghaas:
```suggestion
Advantages are computed **per-example** (grouped by `rollouts_per_example`). You provide a function that computes advantages for a batch of problems.
```

consistent notation. in vf, `problem`=`example` and `group_size`=`rollouts_per_example`

docs/bring-your-own-algorithms.md
```diff
@@ -0,0 +1,138 @@
...
+    ...
+```
+
+#### AdvantageInputs
+
+```python
+@dataclass
+class AdvantageInputs:
+    rewards: Float[Tensor, "num_problems samples_per_problem"]
+    completion_lengths: Int[Tensor, "num_problems samples_per_problem"]
```
mikasenghaas: let's double check all common grpo-like advantages are computable using these so we don't have to change this too much in the future, cc @faresobeid

docs/bring-your-own-algorithms.md
```diff
@@ -0,0 +1,138 @@
...
+
+#### AdvantageOutputs
+
+```python
+@dataclass
+class AdvantageOutputs:
+    advantages: Float[Tensor, "num_problems samples_per_problem"]
+```
+
+### Example: GAE-style Advantage
```
mikasenghaas: this is just GRPO, no? why GAE?

docs/bring-your-own-algorithms.md
```diff
@@ -0,0 +1,138 @@
...
+kwargs = { eps = 1e-8 }
+```
+
+---
+
+## Default Implementations
+
+If no custom function is specified:
+
+- **Loss**: Uses `prime_rl_loss` (GRPO-style with importance ratio clipping and masking)
```
mikasenghaas: our loss is actually not very GRPO-style because we don't do PPO-clipping and also dont have a KL term to a frozen ref. would prob say CISPO-style, clipped/masked importance sampling with optional KL term to a moving reference. @faresobeid ideas for wording here? imo should not be too precise so we don't have to update docs with any minor change to the default loss function

docs/bring-your-own-algorithms.md
```diff
@@ -0,0 +1,138 @@
...
+- **Advantage**: Uses `grpo_advantage` (reward minus per-problem baseline)
```
mikasenghaas: this is dr.grpo right which omits the divsion by std

src/prime_rl/orchestrator/advantage.py
```diff
@@ -1,13 +1,79 @@
...
+AdvantageFn = Callable[..., AdvantageOutputs]
+"""Type for an advantage function.
+
+Expected signature:
+    def my_advantage(inputs: AdvantageInputs, **kwargs) -> AdvantageOutputs:
+        ...
+"""
+
+
+def _import_object(path: str) -> Any:
```
mikasenghaas: imo fine to make public, not a huge fan of private funcs outside of classes

mikasenghaas: also can prob become a more general util so that loss can also use it?

src/prime_rl/orchestrator/advantage.py
```diff
@@ -1,13 +1,79 @@
...
+    """Import an object from a dotted path."""
+    module_path, _, name = path.rpartition(".")
+    module = importlib.import_module(module_path)
+    return getattr(module, name)
 
-from prime_rl.orchestrator.config import AdvantageConfig
+
+def grpo_advantage(inputs: AdvantageInputs, length_weighted_mean: bool = False) -> AdvantageOutputs:
```
mikasenghaas: would prob name this `default_advantage` -- that way it's clear this is always the default used by prime-rl, indepedent of what algorithm we implement

src/prime_rl/orchestrator/config.py
```diff
@@ -603,10 +603,26 @@ def validate_skip_verification(self):
         return self
 
 
-class AdvantageConfig(BaseConfig):
+class AdvantageConfig(BaseModel):
+    """Config for the default GRPO advantage computation."""
```
mikasenghaas:
```suggestion
    """Config for the default advantage."""
```

src/prime_rl/orchestrator/config.py
```diff
@@ -603,10 +603,26 @@ def validate_skip_verification(self):
...
+
+    byo_function: Annotated[
+        str, Field(description="Import path to the advantage function (e.g., 'my_module.my_advantage')")
+    ]
+    kwargs: Annotated[
+        dict[str, Any], Field(default_factory=dict, description="Kwargs to pass to the advantage function")
+    ]
+
+
+AdvantageConfigType: TypeAlias = AdvantageConfig | CustomAdvantageConfig
```
mikasenghaas: should prob make this a discriminator type, that way we get stronger type checking and don't have to `isinstance` on the config object

mikasenghaas: also in general wondering if we should make the default function an "instance" of the custom function (e.g. the specified path is just the path to where the default function is defined?)

src/prime_rl/orchestrator/config.py
```diff
@@ -603,10 +603,26 @@ def validate_skip_verification(self):
...
+class AdvantageConfig(BaseModel):
+    """Config for the default GRPO advantage computation."""
+
     length_weighted_mean: bool = False
 
 
+class CustomAdvantageConfig(BaseModel):
+    """Config for a custom external advantage function."""
+
+    byo_function: Annotated[
```
mikasenghaas: not a huge fan of `byo_function` as a name, i feel like there should be a standard name for the `module.submodule.func` syntax. dont remember rn tho

src/prime_rl/trainer/rl/loss.py
```diff
@@ -1,11 +1,48 @@
...
 
-from prime_rl.trainer.rl.config import LossConfig
+from prime_rl.trainer.rl.config import CustomLossConfig, LossConfig, LossConfigType
+
+
+def _import_object(path: str) -> Any:
+    """Import an object from a dotted path."""
+    module_path, _, name = path.rpartition(".")
+    module = importlib.import_module(module_path)
+    return getattr(module, name)
```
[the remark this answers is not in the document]
mikasenghaas: agree

mikasenghaas (commented):
yea, really nice, i like it! apart from little details, the main architectural q in my mind is:

should we make the default impls an instance of the custom functions. this would prob be less lines of code bc we avoid conditional logic and it would give users a concrete example of how to specific a loss function

samsja:
> yea, really nice, i like it! apart from little details, the main architectural q in my mind is:
> 
> should we make the default impls an instance of the custom functions. this would prob be less lines of code bc we avoid conditional logic and it would give users a concrete example of how to specific a loss function

we would lose the config parsing which I think is too important

docs/bring-your-own-algorithms.md
samsja: let me ask claude why it did this

docs/bring-your-own-algorithms.md
samsja: yeah fair

src/prime_rl/orchestrator/advantage.py
samsja: sure

docs/bring-your-own-algorithms.md
faresobeid: hmm ya i mean its Masked importance sampling with KL against the inference policy, and optional masking strategies

docs/bring-your-own-algorithms.md
faresobeid: ya lol GAE is not used here

docs/bring-your-own-algorithms.md
faresobeid: hmm ya i mean completion lengths is mainly for OPO. I think for most this is enough tho, cant think of any of the top of my head that need more

src/prime_rl/orchestrator/config.py
samsja: I don't think so, this would nmake it hard to validate the loss config which I care about for the default loss

src/prime_rl/orchestrator/config.py
```diff
@@ -612,9 +612,36 @@ def validate_skip_verification(self):
...
+
+def _advantage_config_discriminator(v: Any) -> str:
+    if isinstance(v, dict):
+        return v.get("type", "default")
+    return getattr(v, "type", "default")
+
+
+AdvantageConfigType: TypeAlias = Annotated[
+    Annotated[AdvantageConfig, Tag("default")] | Annotated[CustomAdvantageConfig, Tag("custom")],
+    Discriminator(_advantage_config_discriminator),
```
mikasenghaas, line 641: hm i think we can do this simpler? e.g. for the lr scheduler we have a simple union type and then in the config 

```py
scheduler: Annotated[SchedulerConfigType, Field(discriminator="type")] = ConstantSchedulerConfig()
```

src/prime_rl/orchestrator/config.py
```diff
@@ -683,7 +710,7 @@ class OrchestratorConfig(BaseSettings):
     buffer: BufferConfig = BufferConfig()
 
     # The advantage configuration
-    advantage: AdvantageConfig | None = AdvantageConfig()
+    advantage: AdvantageConfigType | None = AdvantageConfig()
```
mikasenghaas, line 713: wait how can it be `None`? (i know not related to this PR but i dont thins makes sense, no?)

src/prime_rl/trainer/rl/config.py
```diff
@@ -72,6 +73,26 @@ def validate_mask_bounds(self):
...
+def _loss_config_discriminator(v: Any) -> str:
+    if isinstance(v, dict):
+        return v.get("type", "default")
+    return getattr(v, "type", "default")
+
+
+LossConfigType: TypeAlias = Annotated[
+    Annotated[LossConfig, Tag("default")] | Annotated[CustomLossConfig, Tag("custom")],
+    Discriminator(_loss_config_discriminator),
+]
```
mikasenghaas, line 93: same here

src/prime_rl/trainer/rl/loss.py
```diff
@@ -74,120 +104,149 @@ def _safe_mean(values: Tensor, mask: Tensor) -> Tensor:
     return values[mask].sum() / denom
 
 
+def prime_rl_loss(inputs: LossInputs, loss_config: LossConfig) -> LossOutputs:
```
mikasenghaas:
```suggestion
def default_loss_fn(inputs: LossInputs, loss_config: LossConfig) -> LossOutputs:
```

src/prime_rl/orchestrator/advantage.py
```diff
@@ -1,13 +1,72 @@
...
+"""Type for an advantage function.
+
+Expected signature:
+    def my_advantage(inputs: AdvantageInputs, **kwargs) -> AdvantageOutputs:
+        ...
+"""
+
 
-from prime_rl.orchestrator.config import AdvantageConfig
+def default_advantage(inputs: AdvantageInputs, length_weighted_mean: bool = False) -> AdvantageOutputs:
```
mikasenghaas:
```suggestion
def default_advantage_fn(inputs: AdvantageInputs, length_weighted_mean: bool = False) -> AdvantageOutputs:
```

src/prime_rl/trainer/rl/train.py
```diff
@@ -139,9 +140,15 @@ def train(config: RLTrainerConfig):
     logger.info(f"Initializing tokenizer ({config.tokenizer})")
     tokenizer = setup_tokenizer(config.tokenizer)
 
+    # Set up the loss function
+    if isinstance(config.loss, LossConfig):
+        logger.info(f"Using `{config.loss.ratio_type}` importance ratio ({config.loss})")
+    else:
+        logger.info(f"Using custom loss function ({config.loss})")
```
mikasenghaas: imo, could just be 

```py
logger.info(f"Setting up loss function ({config.loss})")
```

this will also print the type which indicates whether default or custom

mikasenghaas (commented):
nice!

src/prime_rl/orchestrator/config.py
samsja: agree

docs/bring-your-own-algorithms.md
samsja: Done in latest push.

docs/bring-your-own-algorithms.md
samsja: Done in latest push.

docs/bring-your-own-algorithms.md
samsja: Done in latest push.

docs/bring-your-own-algorithms.md
samsja: Done in latest push.

src/prime_rl/orchestrator/advantage.py
samsja: Done in latest push.

samsja: Done in latest push.

src/prime_rl/orchestrator/advantage.py
samsja: Done in latest push.

src/prime_rl/orchestrator/config.py
samsja: Done in latest push.

src/prime_rl/orchestrator/config.py
samsja: Done in latest push.

src/prime_rl/orchestrator/config.py
samsja: Done in latest push.

src/prime_rl/trainer/rl/loss.py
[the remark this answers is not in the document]
samsja: Done in latest push.

src/prime_rl/orchestrator/config.py
samsja: Discussed offline — deferring for now to keep typed config parsing for the default. The runtime path in `setup_loss_fn` is already fairly uniform though.

src/prime_rl/orchestrator/config.py
samsja: The simple `Field(discriminator="type")` approach requires the `type` field to be present in the input data. If someone has `[advantage]\nlength_weighted_mean = true` in their TOML without `type`, pydantic fails with `union_tag_not_found`. The custom `Discriminator` callable defaults to `"default"` when `type` is missing, preserving backward compatibility. Happy to switch to the simpler form if we're okay requiring `type` in all configs.

src/prime_rl/orchestrator/config.py
samsja: Pre-existing — `None` means skip advantage computation entirely (just use raw rewards). See `compute_advantages`: `if not advantage_config: return rewards`. Agree it's a bit odd, could be a follow-up to clean up.

src/prime_rl/trainer/rl/config.py
samsja: Same reason as above — custom `Discriminator` callable for backward compat so existing configs without `type = "default"` still parse correctly.

src/prime_rl/trainer/rl/loss.py
samsja: Good call for consistency with `default_advantage`. Will rename.

PR #1720, opened 2026-02-04
handle empty completion_temperatures in prepare_sample

nph4rd:
4fc312b0 introduced per-token temperatures but didn't handle empty completions. if a model only outputs tool calls (no text), completion_ids is empty, so completion_temperatures is [] and training crashes.

defaulting to 1.0 since that's the previous default and prompt tokens are masked anyway.

---

> [!NOTE]
> **Low Risk**
> Small defensive change in batch construction that only affects an edge case (empty completions) and preserves the prior default temperature behavior.
> 
> **Overview**
> Fixes a crash in `prepare_sample` when `completion_temperatures` is empty (e.g., rollouts with tool-only outputs and no completion tokens).
> 
> The prompt-side temperature now falls back to `1.0` instead of indexing `[0]`, keeping per-token `temperatures` construction valid while leaving masked prompt tokens unaffected.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0581c1a4c70ea1d5ff93de08af882d3dab15aa67. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/batch.py
+++ b/src/prime_rl/trainer/batch.py
@@ -15,7 +15,8 @@ def prepare_sample(training_example: TrainingSample, seq_len: int) -> MicroBatch
     position_ids = list(range(len(input_ids)))
 
     # Per-token temperatures: prompt tokens use first completion temp (masked out anyway)
-    prompt_temp = training_example.completion_temperatures[0]
+    # Default to 1.0 if completion is empty (e.g., model generated only tool calls with no text)
+    prompt_temp = training_example.completion_temperatures[0] if training_example.completion_temperatures else 1.0
     temperatures = [prompt_temp] * len(training_example.prompt_ids) + training_example.completion_temperatures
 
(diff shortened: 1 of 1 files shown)


nph4rd:
@samsja  fwiw - i think it's actually something else. i thought it was tool-only completions, but it's actually caused by me hitting the condition of prompt_len > max_seq_lenin in verifiers ~ here: https://github.com/PrimeIntellect-ai/verifiers/blob/c31ef81bb191659f3d407365332a2947374c52ca/verifiers/utils/response_utils.py#L72

the completion ids are set to an empty list ---> temperatures are empty too and that's why it crashes

i think the patch still applies, though it made me wonder why these empty completions would reach that point in the first place

samsja:
> an

it should just fail imo if you hit `prompt_len > max_seq_len`

PR #1724, opened 2026-02-05
Strip @version from env IDs before module loading

JannikSt:
## Summary
- Platform sends versioned env IDs like `jack/meow@0.1.5` for installation
- `vf.load_environment()` internally calls `importlib.import_module()` which fails on `@version`
- Added `strip_env_version()` utility to strip the version suffix at all 4 `load_environment()` call sites
- `install_env()` still receives the full versioned ID as expected


Jackmin801 (approved):
nice! yea lgtm

Jackmin801:
Will the install work with @ though?

PR #1726, opened 2026-02-05
Flash-attention-4

S1ro1:
image

image

Flash-attention-2 vs Flash-attention-4 on 4xB200, llama3-8B, 32k seq len, no ac.

image

Hendrycks - fa4 diverges which is rather weird, but could also be just model, never ran this config tbf, can see early signs of divergence in fa2 as well

---

> [!NOTE]
> **Medium Risk**
> Touches core attention selection/execution and adds a new CUDA kernel dependency with special-case calling conventions and import/namespace validation, which could impact training stability and startup failures.
> 
> **Overview**
> Adds **FlashAttention 4** as a new attention backend for the custom trainer path, wiring a new internal `fa4` option through model config validation, attention layer dispatch, and several custom model implementations.
> 
> Introduces a new `flash-attn-cute` optional dependency (plus lockfile updates) and runtime safeguards: config aliasing from user-facing `flash_attention_4` to internal `fa4`, enforcement that FA4 only runs with `model.impl="custom"`, and a startup check + helper script to detect/fix the `flash_attn.cute` namespace collision when FA2 and FA4 extras are both installed.
> 
> Updates benchmarking scripts and the changelog to recognize/report the new FA4 attention choice.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c5bee9a1f04482f6aabd8953f30ef476e5ea0fc3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/config.py
+++ b/src/prime_rl/trainer/config.py
@@ -143,7 +143,12 @@ class ModelConfig(BaseConfig):
 
     seq_len: Annotated[int, Field(description="The sequence length to use for the model.")] = 2048
 
-    attn: Annotated[AttnImplementation, Field(description="The attention implementation to use.")] = "flash_attention_2"
+    attn: Annotated[
+        AttnImplementation,
+        Field(
+            description="The attention implementation to use. When CP is enabled, ring attention uses the matching kernel family (FA2 for flash_attention_2, FA3 for flash_attention_3).",
+        ),
+    ] = "flash_attention_2"
 
     compile: Annotated[
         CompileConfig | None,
@@ -305,6 +310,11 @@ class ModelConfig(BaseConfig):
     def cp_only_with_flash_attn(self):
         if self.cp > 1 and self.attn not in ["flash_attention_2", "flash_attention_3"]:
             raise ValueError("CP is only supported with flash attention 2 or flash attention 3")
+        if self.cp > 1 and self.attn == "flash_attention_3" and self.impl != "custom":
+            raise ValueError(
+                "CP with flash_attention_3 requires model.impl='custom' "
+                "(the FA3 ring-attention kernel is only implemented for the custom model path)"
+            )
         return self
 
     @model_validator(mode="after")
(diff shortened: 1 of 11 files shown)


src/prime_rl/trainer/config.py
```diff
@@ -5,7 +5,7 @@
 
 from prime_rl.utils.pydantic_config import BaseConfig
 
-AttnImplementation: TypeAlias = Literal["sdpa", "flash_attention_2", "flash_attention_3"]
+AttnImplementation: TypeAlias = Literal["sdpa", "flash_attention_2", "flash_attention_3", "flash_attention_4", "fa4"]
```
samsja: what is the difference between fa4 and flash attention 4 ?

src/prime_rl/trainer/config.py
```diff
@@ -316,6 +316,50 @@ def fused_lm_head_chunk_size_is_valid(self):
...
+                    raise ValueError(
+                        "flash-attn-cute has probably been overwritten by flash-attn, run `scripts/fix-flash-attn-cute.sh` to fix this behaviour."
+                    )
+
+            except ImportError:
+                raise ValueError(
+                    "flash-attn-cute is not installed. Please run `uv sync --extra flash-attn-cute` to install it."
+                )
+
+        return self
```
samsja: lets move this outside of config, config shouldn't be runtime related

src/prime_rl/trainer/config.py
```diff
@@ -316,6 +316,50 @@ def fused_lm_head_chunk_size_is_valid(self):
...
+            # if we keep the name that starts with `flash_attention_*`, it will trigger a hit in transformers and try to install kernel from hub lol, so we internall override to `fa4`
+            self.attn = "fa4"
+            from transformers import AttentionInterface
+
+            def empty(*args, **kwargs) -> None:
+                pass
+
+            # we need to register a dummy function so AutoConfig will think `fa4` is a legitimate attention implementation
+            # this never gets called bc fa4 is only supported with our custom implementation - checked in `flash_attention_4_only_with_custom_impl`
+            AttentionInterface.register("fa4", empty)
```
samsja: config shouldn't impact other library, lets move this outside of config check

scripts/fix-flash-attn-cute.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# Reinstalls flash-attn-cute to ensure it overwrites the older cute module from flash-attn
+# Run this after `uv sync` if you have both flash-attn and flash-attn-cute extras enabled
```
samsja: can u add a comment about here about why we nbeed this ?

PR #1729, opened 2026-02-06
Fix broadcast alpha rank

Jackmin801:
> [!NOTE]
> **Medium Risk**
> Touches LoRA configuration propagation and adapter checkpoint/broadcast metadata, which can affect multi-run scaling behavior and checkpoint compatibility if misconfigured.
> 
> **Overview**
> Aligns per-run LoRA `rank`/`alpha` handling between trainer and orchestrator so multi-run broadcasts and adapter metadata use the correct values.
> 
> `orchestrator.model.lora.alpha` is now optional (inherits trainer by default), `RLConfig.auto_setup_lora` validates that orchestrator LoRA settings don’t conflict with the trainer, propagates missing `rank`/`alpha` from the trainer, and generates adapter names from the effective (possibly inherited) values.
> 
> LoRA adapter config saving is refactored to pass explicit `rank`/`alpha`/`dropout` values, and filesystem broadcast now writes `adapter_config.json` using each run’s orchestrator LoRA `rank`/`alpha` (while `runs.py` also defaults missing alpha before computing scaling factors).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b2a37e1b76e23f45e6dc8a2e1a05affb711f0ed3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/examples/alphabet_sort/rl.toml
+++ b/examples/alphabet_sort/rl.toml
@@ -19,6 +19,9 @@ freq = 1
 
 [trainer.model.lora]
 rank = 32
+
+[orchestrator.model.lora]
+rank = 32
 alpha = 64
 
 [trainer.optim]
--- a/examples/alphabet_sort/rl.toml
+++ b/examples/alphabet_sort/rl.toml
@@ -19,9 +19,6 @@ freq = 1
 
 [trainer.model.lora]
 rank = 32
-
-[orchestrator.model.lora]
-rank = 32
 alpha = 64
 
 [trainer.optim]
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -48,12 +48,12 @@ class LoRAConfig(BaseConfig):
     ] = None
 
     alpha: Annotated[
-        float,
+        float | None,
         Field(
             ge=0,
-            description="LoRA alpha for this run.",
+            description="LoRA alpha for this run. If None, uses trainer's alpha.",
         ),
-    ] = 32.0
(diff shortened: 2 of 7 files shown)


examples/alphabet_sort/rl.toml
```diff
@@ -19,6 +19,9 @@ freq = 1
 
 [trainer.model.lora]
 rank = 32
+
+[orchestrator.model.lora]
+rank = 32
```
mikasenghaas: hmm i dont quite get why the orch even has to know abt any lora config?

mikasenghaas: like now the user has to remember to set the same lora params on both, otherwise it will still mismatch no? there's prob a good reason that this is a duplicated arg (like some args like model name, etc. are unavoidable to  duplicate). if there's no way around, maybe good to add this to the shared rl.py configs so that users only have to configure one place and we handle setting it correctly on trainer and orch

Jackmin801: > hmm i dont quite get why the orch even has to know abt any lora config?

For hosted RL, we want to allow runs to set lora hparams and thus it needs to be set by orchestrator

Jackmin801: made it that if you set the alphas differently for orch and trainer in `rl.py` it will error. and if you only set on trainer side then it defaults to the trainer's one.

samsja (approved):

PR #1746, opened 2026-02-09
add gibberish & repetition filtering

hallerite:
This PR adds a new `RolloutFilter` class that allows one to define rules for filtering out rollouts, for example when they contain gibberish or repetitions.

---

> [!NOTE]
> **Medium Risk**
> Touches the training data path by mutating rollout masks/stop conditions before advantage computation, which can change what samples reach the trainer. Logic is localized and covered by unit tests, but misconfiguration could silently drop learning signal.
> 
> **Overview**
> Adds **orchestrator-side rollout filtering** to detect degenerate generations and optionally prevent them from contributing to training.
> 
> Introduces configurable `orchestrator.filters` with `gibberish` (rare tokens at high entropy) and `repetition` (long high-confidence streaks) detectors, wires them into the orchestrator loop to zero `completion_mask`/set `stop_condition` when `enforce=True`, and logs aggregate filter metrics. Also removes the previous Chinese-character counting metrics/utilities and adds unit tests covering filter detection, enforcement behavior, and setup.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b1dd7e796a384c14073d71e7aced51771956ffcc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/config.py
+++ b/src/prime_rl/orchestrator/config.py
@@ -376,6 +376,13 @@ class EvalConfig(BaseConfig):
         ),
     ] = True
 
+    cancel_inflight_rollouts_on_eval: Annotated[
+        bool,
+        Field(
+            description="Whether to cancel in-flight training rollouts before starting online evals. This is useful to avoid congestion (e.g. do not have training + eval rollouts happening at the same time) but leads to slower training steps as rollouts get cancelled and the pipeline has to fill up after each eval",
+        ),
+    ] = False
+
 
 class CheckpointConfig(BaseConfig):
     """Configures checkpointing the orchestrator."""
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,3 +59,5 @@ Documenting changes which affect configuration usage patterns (added/moved/remov
 - **`dump_config`**: Added `--dump-config <path>` flag to the `rl` command. When set, writes the resolved subconfigs (trainer, orchestrator, inference, teacher_inference) to the given directory and exits without starting any processes (2026-02-12)
 - **`client.api_key_var`**: Changed default from "OPENAI_API_KEY" to "VLLM_API_KEY" (2026-02-12)
 - **`orchestrator.filters`**: Added orchestrator-side rollout filters for detecting degenerate generations. Supports `[[filters]] type = "gibberish"` (rare tokens at high entropy) and `[[filters]] type = "repetition"` (high-confidence token streaks). Detected rollouts get reward zeroed and completion mask cleared (2026-02-13)
+- **`orchestrator.eval.cancel_inflight_rollouts_on_eval`**: Added flag to optionally cancel in-flight training rollouts before starting online evals. When enabled, avoids congestion by preventing training and eval rollouts from running simultaneously, but slows training as the rollout pipeline must refill after each eval (default: False) (2026-02-16)
+
--- a/k8s/prime-rl/templates/deployment.yaml
+++ b/k8s/prime-rl/templates/deployment.yaml
@@ -27,7 +27,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       containers:
-      - name: prime-rl
+      - name: prime-rl-orchestrator
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         {{- if .Values.orchestrator.autoStart }}
@@ -133,7 +133,7 @@ spec:
       runtimeClassName: {{ .Values.inference.runtimeClassName }}
       {{- end }}
       containers:
-      - name: prime-rl
+      - name: prime-rl-inference
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         {{- if .Values.inference.autoStart }}
@@ -262,7 +262,7 @@ spec:
       runtimeClassName: {{ .Values.trainer.runtimeClassName }}
       {{- end }}
       containers:
-      - name: prime-rl
+      - name: prime-rl-trainer
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         {{- if .Values.trainer.autoStart }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ sft = "prime_rl.trainer.sft.train:main"
 
 [project.optional-dependencies]
 flash-attn = [
-    "flash-attn @ https://github.com/mjun0812/flash-attention-prebuild-wheels/releases/download/v0.7.16/flash_attn-2.6.3%2Bcu128torch2.10-cp312-cp312-linux_x86_64.whl",
+    "flash-attn @ https://github.com/mjun0812/flash-attention-prebuild-wheels/releases/download/v0.7.16/flash_attn-2.8.3%2Bcu128torch2.10-cp312-cp312-linux_x86_64.whl",
 ]
 flash-attn-3 = [
     "flash_attn_3 @ https://github.com/samsja/flash-attn-builds/releases/download/v0.1/flash_attn_3-3.0.0b1-cp39-abi3-linux_x86_64.whl",
--- a/scripts/mini_moe.py
+++ b/scripts/mini_moe.py
@@ -15,12 +15,14 @@ import argparse
 from pathlib import Path
 
 import torch
-from transformers import AutoConfig, AutoTokenizer
+from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
 from transformers import Glm4MoeForCausalLM as HFGlm4MoeForCausalLM
 
 from prime_rl.trainer.models.glm4_moe import Glm4MoeConfig
 from prime_rl.trainer.models.glm4_moe import Glm4MoeForCausalLM as PrimeRLGlm4MoeForCausalLM
 from prime_rl.trainer.models.layers.lm_head import inject_prime_lm_head
+from prime_rl.trainer.models.minimax_m2 import MiniMaxM2Config
+from prime_rl.trainer.models.minimax_m2 import MiniMaxM2ForCausalLM as PrimeRLMiniMaxM2ForCausalLM
 from prime_rl.utils.logger import setup_logger
 from prime_rl.utils.utils import default_dtype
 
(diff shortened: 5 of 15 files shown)


src/prime_rl/orchestrator/filters.py
```diff
@@ -0,0 +1,161 @@
...
+
+    for rollout in rollouts:
+        for filt in filters:
+            result = filt.check(rollout)
+            if result.detected:
+                rollout["reward"] = 0.0
+                for step in rollout["trajectory"]:
+                    tokens = step["tokens"]
+                    if tokens is not None:
+                        tokens["completion_mask"] = [0] * len(tokens["completion_mask"])
```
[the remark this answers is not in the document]
hallerite: I believe this is fine, as this is also how we handle truncation.

src/prime_rl/orchestrator/config.py
```diff
@@ -561,6 +561,54 @@ def _advantage_config_discriminator(v: Any) -> str:
...
+def _filter_config_discriminator(v: Any) -> str:
+    if isinstance(v, dict):
+        return v.get("type", "gibberish")
+    return getattr(v, "type", "gibberish")
+
+
+FilterConfigType: TypeAlias = Annotated[
+    Annotated[GibberishFilterConfig, Tag("gibberish")] | Annotated[RepetitionFilterConfig, Tag("repetition")],
+    Discriminator(_filter_config_discriminator),
+]
```
samsja: we don't need tag here, can just do normal distriminator and let the user always precise with type

src/prime_rl/orchestrator/config.py
```diff
@@ -561,6 +561,54 @@ def _advantage_config_discriminator(v: Any) -> str:
...
+
+
+class RepetitionFilterConfig(BaseConfig):
+    """Flags pathological repetition loops (Section 3.2, https://arxiv.org/abs/2506.13585)."""
+
+    type: Literal["repetition"] = "repetition"
+    window: Annotated[
+        int,
+        Field(ge=1, description="Number of consecutive high-probability steps before flagging."),
+    ] = 3_000
```
samsja, line 621: 3000 seems to high as a default ?

src/prime_rl/orchestrator/filters.py
```diff
@@ -0,0 +1,162 @@
...
+@dataclass
+class FilterResult:
+    detected: bool
+    detection_index: int | None = None
+
+
+class RolloutFilter(Protocol):
+    name: str
+
+    def check(self, rollout: vf.RolloutOutput) -> FilterResult: ...
```
samsja, line 30: FINALY SOMEONE USING PROTOCOL

src/prime_rl/orchestrator/config.py
faresobeid: It's what minimax m1 has as a default image

tests/unit/orchestrator/test_filters.py
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+
+# --- GibberishFilter tests ---
+
+
+def test_gibberish_detects_rare_low_prob_token():
+    vocab_size = 128_000
+    logprob_threshold = -math.log(vocab_size) - 2.0
+    f = GibberishFilter(name="gibberish", token_id_threshold=100_000, logprob_threshold=logprob_threshold)
```
samsja: I don;t like variable name withj one letter even in tests, word are cheap in the ai era

faresobeid:
Nice work! Maybe we can have an option to track always but have the filtering optional so we can judge whether we should turn it on or not if we didn't already.

src/prime_rl/orchestrator/filters.py
```diff
@@ -0,0 +1,162 @@
...
+        return {}
+
+    counts: dict[str, int] = {f.name: 0 for f in filters}
+    total_filtered = 0
+
+    for rollout in rollouts:
+        for filt in filters:
+            result = filt.check(rollout)
+            if result.detected:
+                rollout["reward"] = 0.0
```
samsja: what if reward is between -1 and 1 ?

hallerite: we already mask out the gradient, so I guess we could simply not override the reward at all to get an undistorted group mean

src/prime_rl/orchestrator/config.py
```diff
@@ -561,6 +561,64 @@ def _advantage_config_discriminator(v: Any) -> str:
 ]
 
 
+class GibberishFilterConfig(BaseModel):
+    """Flags rare tokens generated at high entropy (Section 5.2, https://arxiv.org/abs/2510.02387)."""
+
+    model_config = ConfigDict(extra="forbid")
```
mikasenghaas, line 585: this bc we cannot use base settings on discriminated unions right?

src/prime_rl/orchestrator/config.py
```diff
@@ -561,6 +561,64 @@ def _advantage_config_discriminator(v: Any) -> str:
...
+
+    type: Literal["gibberish"] = "gibberish"
+    enforce: Annotated[
+        bool,
+        Field(
+            description="If True, zero reward and mask for detected rollouts. If False, only track detection metrics."
```
mikasenghaas: zero reward and mask? if we mask reward doesn't matter right?

src/prime_rl/orchestrator/config.py
```diff
@@ -561,6 +561,64 @@ def _advantage_config_discriminator(v: Any) -> str:
...
+        ),
+    ] = False
+    token_id_threshold: Annotated[
```
mikasenghaas, line 594: not sure how much modern vocab sizes but could it make sense to make this a percentile? not too strong opinons here tho

src/prime_rl/orchestrator/config.py
```diff
@@ -561,6 +561,64 @@ def _advantage_config_discriminator(v: Any) -> str:
...
+        int,
+        Field(description="Token IDs above this are candidates for gibberish. BPE tokens are sorted by merge order."),
+    ] = 100_000
+    logprob_offset: Annotated[
+        float,
+        Field(description="Offset from uniform distribution logprob. Threshold = -log(vocab_size) - logprob_offset."),
+    ] = 2.0
+    vocab_size: Annotated[
```
mikasenghaas: why would i ever change this value?

src/prime_rl/orchestrator/config.py
```diff
@@ -561,6 +561,64 @@ def _advantage_config_discriminator(v: Any) -> str:
...
+    window: Annotated[
+        int,
+        Field(ge=1, description="Number of consecutive high-probability steps before flagging."),
+    ] = 3_000
+    prob_threshold: Annotated[
+        float,
+        Field(
+            gt=0,
+            le=1,
+            description="Token probability threshold. Steps where sampled prob exceeds this count toward the window.",
```
mikasenghaas: im not sure i get this description, can we make it clearer?

src/prime_rl/orchestrator/filters.py
```diff
@@ -0,0 +1,180 @@
...
+                    consecutive = 0
+                if consecutive >= self.window:
+                    return FilterResult(detected=True, detection_index=global_idx)
+                global_idx += 1
+        return FilterResult(detected=False)
+
+
+def setup_filter(config: FilterConfigType, vocab_size: int) -> RolloutFilter:
+    """Create a RolloutFilter from a filter config."""
+    if isinstance(config, GibberishFilterConfig):
```
mikasenghaas: can use `config.type == ...` here and get strong typing after

src/prime_rl/orchestrator/filters.py
```diff
@@ -0,0 +1,180 @@
...
+        return {}
+
+    counts: dict[str, int] = {f.name: 0 for f in filters}
+    total_detected = 0
+    total_enforced = 0
+
+    for rollout in rollouts:
+        if rollout.get("metrics") is None:
+            rollout["metrics"] = {}
+        for filt in filters:
```
mikasenghaas, line 146: lol any reason to not name it `filter` here?

src/prime_rl/orchestrator/filters.py
```diff
@@ -0,0 +1,180 @@
...
+                            tokens["completion_mask"] = [0] * len(tokens["completion_mask"])
+                    rollout["stop_condition"] = filt.name
+                    total_enforced += 1
+
+                break
+
+    n = len(rollouts)
+    metrics: dict[str, float] = {}
+    for f in filters:
+        metrics[f"filter/{f.name}_count"] = float(counts[f.name])
```
mikasenghaas, line 169: more in line with metric logging convention would be `filter/<name>/count` and `filter/<name>/mean` but idk have not been too happy with our consistency in logging for some time so could also fix later haha

src/prime_rl/orchestrator/filters.py
```diff
@@ -0,0 +1,162 @@
...
+                rollout["metrics"][f"filter/{filt.name}"] = 1.0
+                counts[filt.name] += 1
+                total_filtered += 1
+                break
+
+    n = len(rollouts)
+    metrics: dict[str, float] = {}
+    for f in filters:
+        metrics[f"filter/{f.name}_count"] = float(counts[f.name])
+        metrics[f"filter/{f.name}_rate"] = counts[f.name] / n if n > 0 else 0.0
```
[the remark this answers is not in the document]
mikasenghaas: ah yea should maybe cover this edge case and ensure that the filter config types are unique which should ensure unique names as well right?

mikasenghaas (commented):
really nice!

- we should probably aim to turn online difficulty filtering (currently part of buffer) into smth like `ZeroAdvantageFilter`. can do part of later pr tho
- can we do a hendrycks/acereason math run to double check the defaults are sensible (with enforce on)
- i think we change the reward in-place when we `apply_filter`. should think clearly about what behavior we want, e.g. should a filtered rollout contribute nothing to baseline or 0 or its original? same for simple logging of e.g. training reward should we include or exclude filtered rollouts?

mikasenghaas:
ah thinking about the zero adv filter again: ig its qualitatively a bit different bc the filter is based on a group not an individual rollout. i think we might still want to consider if we can consolidate both abstractions. i really like the filter one so if it were able to handle this one as well would be great

src/prime_rl/orchestrator/config.py
hallerite: exactly

src/prime_rl/orchestrator/config.py
hallerite: true, should make description more parsimonious

src/prime_rl/orchestrator/config.py
hallerite: I just followed the defaults of the paper, but we also make it a percentile

src/prime_rl/orchestrator/config.py
hallerite: you are right, this is an artifact from an earlier version that didn't derive `vocab_size` automatically

src/prime_rl/orchestrator/filters.py
hallerite: I think we should think about this more generally, but probably better to make a separate PR to consolidate

src/prime_rl/orchestrator/filters.py
hallerite: yeah, `filter` is a Python builtin, so it's convention not too shadow it (even if harmless here).

src/prime_rl/orchestrator/filters.py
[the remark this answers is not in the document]
hallerite: yes

samsja (approved):

PR #1753, opened 2026-02-10
refactor code and token-level batches

faresobeid:
- Make batches be defined at the token level (completion tokens). This follows https://arxiv.org/abs/2510.18855 and https://www.arxiv.org/abs/2510.02387

- Have online difficulty filtering always on but orch logs (reward, ...) arent affected and log based on all rollouts

- Have it so the orchestrator completes a group of rollouts and directly sends to the trainer, and the trainer accumulates these until they reach the batch size and takes a step

- Currently the code keeps unused rollouts from a step and uses them on the next step (becomes one more step off policy but prob should be fine), can consider clearing the queued up rollouts if we dont want this behaviour

---

> [!NOTE]
> **High Risk**
> Large refactor of the core RL dataflow (orchestrator↔trainer contract, batching/stepping, checkpoint/eval timing), so regressions could affect training stability, throughput, and resume behavior despite test updates.
> 
> **Overview**
> Switches RL training from fixed `batch_size` steps to a **streaming, token-budgeted** pipeline: the orchestrator now continuously emits completed group rollouts, and the trainer accumulates samples until `token_batch_size` is reached before stepping.
> 
> The orchestrator/scheduler are refactored around `max_inflight_rollouts` concurrency (removing `batch_size`, `max_async_level`, and validation), with eval/checkpointing triggered on `ckpt_step` transitions via a shared `_should_run_eval` helper. Buffer behavior is simplified to track easy/normal/hard pool metrics and persist only pool assignments (no rollout buffering), while filesystem transport/broadcast add reset/cleanup controls (`set_start_step`, `keep_last`) to match the new streaming semantics. Configs/examples/tests are updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit db3ab7864c0c704ee0c907be377da49ef3831640. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/orchestrator/config.py
```diff
@@ -680,15 +652,13 @@ class OrchestratorConfig(BaseSettings):
         ),
     ] = None
 
-    batch_size: Annotated[int, Field(ge=1, description="Number of samples to train on per step.")] = 128
-
-    oversampling_factor: Annotated[
-        float,
+    max_inflight_groups: Annotated[
```
mikasenghaas: lets change semantics here and have `max_inflight_rollouts` for it to be invariant of the group size

src/prime_rl/orchestrator/config.py
```diff
@@ -705,12 +675,6 @@ class OrchestratorConfig(BaseSettings):
         ),
     ] = 2048
 
-    # TODO(Mika): This should be automatic from the number of ZMQ connections
-    num_train_workers: Annotated[
```
mikasenghaas, line 720: we dont need this?

src/prime_rl/orchestrator/config.py
```diff
@@ -726,25 +690,10 @@ class OrchestratorConfig(BaseSettings):
         ),
     ] = 8
 
-    max_async_level: Annotated[
```
mikasenghaas, line 740: it was a good run;(

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -70,6 +67,34 @@
 from prime_rl.utils.vlm import is_vlm_model
 
 
+class OrchestratorContext:
```
mikasenghaas: hmmmm, we might as well make the orch a class at this point haha (i dont hate it)

mikasenghaas: but this abstraction feels a bit half baked

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -297,7 +308,9 @@ async def orchestrate(config: OrchestratorConfig):
     if checkpoint_step is not None and ckpt_manager is not None:
         ckpt_manager.load(progress, buffer, step=checkpoint_step)
         logger.info(f"Resuming training from checkpoint step {checkpoint_step}")
-        scheduler.ckpt_step = progress.step  # Always resume from the latest checkpoint
+        scheduler.ckpt_step = checkpoint_step
```
mikasenghaas: are these equivalent?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -320,76 +333,39 @@ async def orchestrate(config: OrchestratorConfig):
         else:
             logger.info("Training from scratch. Skipping base weight reload")
 
-    # Iterate over dataset in batches
-    max_steps = config.max_steps or int(1e9)
-    logger.info(f"Starting orchestrator loop (max_steps={max_steps or 'infinite'})")
-    is_first_step = True
+    # Streaming loop: continuously get group rollouts and send them to the trainer.
+    # The trainer decides when to step based on accumulated completion tokens.
+    logger.info(f"Starting orchestrator streaming loop (max_steps={config.max_steps or 'infinite'})")
```
mikasenghaas: i dont like the term "streaming loop", just loop is fine imo

mikasenghaas (commented):
ouf, i think i like a lot of stuff in here directionally but its stuffing a lot of orthogonal changes/ deprecations into one pr which makes it hard to review. can we break this up into a little more digestible pieces plss

src/prime_rl/orchestrator/config.py
faresobeid: they were never used, in prepare_batch we just have it equal to dp_world_size

PR #1770, opened 2026-02-12
[refactor] Separate IS and rejection in loss

LeonEricsson:
Off-policy correction is an important topic in RL with strong influence on large-scale and long-horizon training. The handling of this in `default_loss_fn` can be made more explicit, and at the same time we can expose general helpers that align with the BYOA vision.

## what this pr does

The loss function currently mixes two orthogonal concerns: importance sampling (soft reweighting) and rejection sampling (hard masking). This PR separates them into standalone helpers that `default_loss_fn` calls internally, and that BYOA authors can import and compose in custom loss functions.

**Importance sampling**: `compute_importance_weights(log_ratio, mask, ...)` returns an `ImportanceWeights` dataclass with token, sequence, and geo-mean weights, each with optional clipping.

**Rejection masking**: a family of `reject_by_*` functions, each taking `log_ratio` (and `mask` where applicable) and returning boolean masks:
- `reject_by_token`: per-token ratio bounds
- `reject_by_sequence_minmax`: mask sequence if any token breaches bounds
- `reject_by_sequence_sum`: mask sequence by product ratio (new, not employed in `default_loss_fn`)
- `reject_by_geo_mean`: mask sequence by geometric mean ratio
- `reject_by_geo_mean_k3`: mask sequence by mean k3 KL divergence (new, not employed in `default_loss_fn`)

`default_loss_fn` is rewritten in terms of these helpers. No behaviour change, with two caveats:

1. A numerical guard was added. We route all `exp()` calls through `_safe_exp_log_ratio`, which clamps `log_ratio` to `[-30, 30]` before exponentiating.
2. A `.detach()` was removed. The old code had `seq_log_importance_ratio = torch.clamp(log_importance_ratio[loss_mask].sum().detach(), max=10.0)`, detaching the sequence log-ratio before it flowed into `seq_importance_ratio` and then into `coeff`. This is redundant because `coeff.detach()` downstream already cuts the graph before importance weights reach the loss.

## motivator, again

As said, the previous `default_loss_fn` interleaved IS weighting and rejection masking in a single block with shared intermediates, making it hard to tell what served which purpose. Now the separation is explicit: `compute_importance_weights` handles soft reweighting, `reject_by_*` functions handle hard masking. Either can be used independently or composed together, as shown in `default_loss_fn`.

These helpers are useful for BYOA. Off-policy correction is ubiquitous across RL algos, often combining some importance weight strategies with a rejection strategy. Happy to put together some BYOA examples using these in a follow-up PR.

Rejection functions take `log_ratio` directly rather than pre-computed weights, so they are self-contained. BYOA authors don't need to know about `compute_importance_weights` to use rejection.

## testing

Added minimal unit tests covering the new helpers. Wasn't sure about the testing guidelines for this area so kept it light. Open to discuss what level of coverage we want here.

---

> [!NOTE]
> **Medium Risk**
> Changes core RL loss weighting/masking logic and adds clamped exponentiation; behavior is intended to match but could subtly affect training dynamics, gradients, or masking edge cases.
> 
> **Overview**
> Refactors `default_loss_fn` to explicitly separate **importance sampling** (soft reweighting) from **rejection masking** (hard token/sequence filtering), exposing reusable helpers for custom losses.
> 
> Adds `compute_importance_weights` (token/sequence/geometric-mean weights with optional clipping) plus `reject_by_*` utilities (`token`, `sequence_minmax`, `sequence_sum`, `geo_mean`, `geo_mean_k3`), and routes exponentiation through `_safe_exp_log_ratio` with `LOG_RATIO_EXP_BOUND` clamping for numerical stability.
> 
> Updates loss metrics to use the new masks/weights and adds unit tests covering the new helper behaviors (including empty-mask handling and k1-vs-k3 mismatch detection).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 64ee658318d3ab198ae243fee9ae9e9d7e1ec52e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/unit/train/rl/test_loss.py
+++ b/tests/unit/train/rl/test_loss.py
@@ -11,6 +11,7 @@ from prime_rl.trainer.rl.loss import (
     reject_by_geo_mean,
     reject_by_geo_mean_k3,
     reject_by_sequence_minmax,
+    reject_by_sequence_sum,
     reject_by_token,
     setup_loss_fn,
 )
@@ -139,6 +140,34 @@ def test_reject_by_sequence_minmax_ignores_unmasked():
 
     assert high_mask.item() is False
 
+
+def test_reject_by_sequence_sum():
+    # product ratio = exp(1.5) ≈ 4.48
+    log_ratio = torch.tensor([0.5, 0.5, 0.5], device="cuda")
+    mask = torch.ones(3, dtype=torch.bool, device="cuda")
+    
+    low, high = reject_by_sequence_sum(log_ratio, mask, low=0.1, high=10.0)
+    assert low.item() is False
+    assert high.item() is False
+
+    _, high = reject_by_sequence_sum(log_ratio, mask, high=4.0)
+    assert high.item() is True
+
+    # negative product ratio = exp(-1.5) ≈ 0.22
+    low, _ = reject_by_sequence_sum(-log_ratio, mask, low=0.5)
+    assert low.item() is True
+    
+    # big log_ratio at index 1, but it's unmasked → product only over indices 0, 2
+    log_ratio = torch.tensor([0.1, 10.0, 0.1], device="cuda")
+    mask = torch.tensor([True, False, True], device="cuda")
+
+    _, high = reject_by_sequence_sum(log_ratio, mask, high=5.0)
+    assert high.item() is False
+
+     # None bounds
+    low, high = reject_by_sequence_sum(log_ratio, mask)
+    assert low.item() is False
+    assert high.item() is False
 
 def test_k1_cancellation_hides_mismatch_k3_catches_it():
     """k1 geo mean allows cancellation (e^2 and e^-2 average to 1); k3 KL does not."""


faresobeid:
Thank you for the PR! We currently want to focus on providing our view of the SOTA in terms of RL algos for our setup. Can be more flexible with the loss personally tho

LeonEricsson:
> Thank you for the PR! We currently want to focus on providing our view of the SOTA in terms of RL algos for our setup. Can be more flexible with the loss personally tho

Thanks for the reply. I'd push back a bit and argue that this PR is in service of that goal, not a detour from it.

This change isn't introducing new algorithmic direction, we're restructuring existing off-policy correction logic to make two concerns explicit: importance sampling (soft reweighting) and rejection (hard masking). The goal is to make default_loss_fn easier to reason about, safer to modify, and easier to test.

As a side benefit, IS weighting and rejection are common across many RL algorithms. Exposing these helpers lets BYOA reuse vetted primitives rather than rolling their own.

That said, if keeping scope tight is a concern, we can drop the helpers that aren't tied to the current default-loss path.

PR #1774, opened 2026-02-13
Add SLURM entrypoint with Jinja2 template

samsja:
## Summary

Adds a SLURM deployment workflow for multi-node RL training. Instead of manually writing and maintaining SLURM batch scripts, you now define a single TOML config and the entrypoint handles everything: config resolution, sub-config dumping, script rendering, and job submission.

### New entrypoint: `rl_slurm`

```bash
uv run python -m prime_rl.slurm.rl @ examples/slurm/reverse_text.toml
```

This will:
1. Resolve the full RL config (trainer, orchestrator, inference) with all shared config propagation (model, wandb, ckpt, etc.)
2. Dump sub-configs as TOML files to `{output_dir}/configs/`
3. Render a SLURM batch script from a Jinja2 template to `{output_dir}/rl.sh`
4. Submit the job via `sbatch`

Use `--dry-run` to only generate the script without submitting.

### SLURM-specific config fields

| Field | Description |
|---|---|
| `job_name` | SLURM job name |
| `output_dir` | Directory for outputs, sub-configs, and logs |
| `base_dir` | Path to the project root on the cluster (defaults to cwd) |
| `num_train_nodes` | Number of nodes for training |
| `num_infer_nodes` | Number of nodes for inference |
| `slurm_template` | Path to a custom Jinja2 template (optional) |
| `dry_run` | Only generate the script without submitting (default: false) |

### Custom templates

The default template (`src/prime_rl/slurm/rl_slurm.sh.j2`) handles NCCL weight broadcast, InfiniBand detection, and `srun`-based process dispatch. For advanced use cases (custom partitions, account settings, module loads, different networking), provide your own template via `--slurm-template`.

### Tmux launcher

A tmux helper for monitoring running jobs:

```bash
bash scripts/slurm_tmux.sh my-rl-job /shared/outputs/my-rl-job
```

Creates a session with a terminal window and a logs window with three panes tailing trainer, orchestrator, and inference logs (using glob patterns for multi-node).

### Docs

Added `docs/slurm.md` with full configuration reference, usage examples, custom template guide, and monitoring instructions. Updated `docs/deployment.md` to link to it (replacing the old TBD sections).

---

> [!NOTE]
> **Medium Risk**
> Introduces a new job-submission path (`sbatch`) and refactors core RL config validation into a shared base class, which could subtly change config resolution behavior across existing RL runs.
> 
> **Overview**
> Adds a new `prime_rl.slurm.rl` entrypoint that takes a single TOML config, resolves the full RL config, writes resolved sub-config TOMLs to `{output_dir}/configs`, renders a batch script from `rl_slurm.sh.j2`, and optionally submits it via `sbatch` (or `--dry-run` for generation only).
> 
> Refactors shared RL config propagation/validation into a new `BaseRLConfig` (migrated out of `RLConfig`) so the same resolution logic can be reused by the SLURM workflow. Also adds a tmux helper (`scripts/slurm_tmux.sh`) for tailing SLURM logs, new SLURM docs and example configs, updates deployment docs to link to them, and updates dependencies/lockfile (including `math-env`) and test config coverage to include `RLSLURMConfig`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c9fa4e153df22c8debb8b19bbfbf3c1ba375aee0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/examples/slurm/hendrycks_math.toml
+++ b/examples/slurm/hendrycks_math.toml
@@ -1,4 +1,5 @@
 job_name = "hendrycks-math"
+output_dir = "outputs/hendrycks-math"
 
 num_train_nodes = 1
 num_infer_nodes = 1
--- a/examples/slurm/hendrycks_math_moe.toml
+++ b/examples/slurm/hendrycks_math_moe.toml
@@ -1,4 +1,5 @@
 job_name = "hendrycks-math-moe"
+output_dir = "outputs/hendrycks-math-moe"
 
 num_train_nodes = 1
 num_infer_nodes = 1


docs/slurm.md
```diff
@@ -0,0 +1,154 @@
...
+
+| Field | Description |
+|---|---|
+| `job_name` | SLURM job name |
+| `output_dir` | Directory for outputs, sub-configs, and logs |
+| `project_dir` | Path to the project root on the cluster (defaults to cwd) |
+| `num_train_nodes` | Number of nodes for training |
+| `num_infer_nodes` | Number of nodes for inference |
+| `gpus_per_node` | Number of GPUs per node (default: 8) |
+| `nodes_per_fsdp_group` | Number of train nodes per FSDP island (optional, auto-sets `trainer.dp_replicate`) |
```
mikasenghaas, line 33: isnt this redundant with the trainer config? why would i set this and not dp replicate?

examples/slurm/hendrycks_math.toml
mikasenghaas, line 1: should we not put the slurm template into each run dir, e.g. `examples/hendrycks_math/slurm.toml` or something like this

scripts/slurm_tmux.sh
mikasenghaas, line 1: nice should we also add add a pane for env logs? should prob be in new window

src/prime_rl/slurm/rl_slurm.sh.j2
```diff
@@ -0,0 +1,142 @@
...
+        --node-rank=$TRAIN_NODE_RANK \
+        --rdzv-endpoint=$MASTER_ADDR:$MASTER_PORT \
+        --rdzv-id=job_$SLURM_JOB_ID \
+        --log-dir=$OUTPUT_DIR/torchrun \
+        --tee=3 \
+        --redirects=3 \
+        --local-ranks-filter=$(seq -s, 0 $((GPUS_PER_NODE - 1))) \
+        -m prime_rl.trainer.rl.train \
+        @ $CONFIG_DIR/trainer.toml \
+        --weight_broadcast.host 0.0.0.0 \
```
mikasenghaas, line 135: this is prob default, no? in this case no need to set

src/prime_rl/slurm/rl_slurm.sh.j2
```diff
@@ -0,0 +1,142 @@
...
+    uv run torchrun \
+        --nnodes=$NUM_TRAIN_NODES \
+        --nproc-per-node=$GPUS_PER_NODE \
+        --node-rank=$TRAIN_NODE_RANK \
+        --rdzv-endpoint=$MASTER_ADDR:$MASTER_PORT \
+        --rdzv-id=job_$SLURM_JOB_ID \
+        --log-dir=$OUTPUT_DIR/torchrun \
+        --tee=3 \
+        --redirects=3 \
+        --local-ranks-filter=$(seq -s, 0 $((GPUS_PER_NODE - 1))) \
```
mikasenghaas, line 132: this means we get all rank logs right to stdout, right? isn't it nicer to only get rank 0 to stdout? afaik, we still write logs for all ranks to files via torchrun either way

src/prime_rl/slurm/rl_slurm.sh.j2
```diff
@@ -0,0 +1,142 @@
...
+        -m prime_rl.trainer.rl.train \
+        @ $CONFIG_DIR/trainer.toml \
+        --weight_broadcast.host 0.0.0.0 \
+        --weight_broadcast.port $BROADCAST_PORT \
+        --weight_broadcast.inference_world_size $((GPUS_PER_NODE * NUM_INFER_NODES)) \
+        --weight_broadcast.timeout $BROADCAST_TIMEOUT \
+        --dist_timeout_seconds $NCCL_COMM_TIMEOUT \
```
mikasenghaas, line 139: also let's set this via configs not via slurm template?

src/prime_rl/slurm/rl_slurm.sh.j2
mikasenghaas, line 1: imo we should try to make the slurm template as bare-bone as possible. e.g nccl of fs weight broadcast should be configurable in the rl or slurm rl configs not be hardcoded into the slurm template

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "prime>=0.5.37",
     "pyzmq>=27.1.0",
     "aiolimiter>=1.2.1",
+    "math-env",
```
mikasenghaas, line 28: hmm do we really want to add this?

docs/slurm.md
samsja: dp_replicate is dependent on the number of gpus. Wheras the nodes_per_fsdp_group is not. This is the same --dp_replicate num_nodes //8 param we had before essentialy

examples/slurm/hendrycks_math.toml
samsja: no the slurm template is general and should be available from the site packages,  user can extend the template for full flexibility but most of the time they shouldn't

src/prime_rl/slurm/rl_slurm.sh.j2
samsja: 0.0.0.0 is not the default

src/prime_rl/slurm/rl_slurm.sh.j2
samsja: its easier so far to have all the gpus logged into it so that we can get it from the tmux logs

scripts/slurm_tmux.sh
samsja: issue is that the file is based on the env name no ?

src/prime_rl/slurm/rl_slurm.sh.j2
samsja: yeah good point sir

src/prime_rl/slurm/rl_slurm.sh.j2
samsja: moving nccl to the config indeed

pyproject.toml
samsja: yes because the auto install is not working on slurm so otherwise the math example are just not working oob

PR #1795, opened 2026-02-15
feat(inference): auto-detect tool_call_parser from model name

samsja:
## Summary

- Auto-detect `tool_call_parser` from the model name when not explicitly configured, instead of silently defaulting to `"hermes"` for all models
- Add `MODEL_TOOL_CALL_PARSER_MAP` — a hardcoded, ordered mapping from model name patterns to vLLM tool call parser names
- Raise a clear error when the model is unknown and no parser is set, telling the user exactly what to do

### Supported model → parser mappings

| Model pattern | Parser |
|---|---|
| GLM-4.7 | `glm47` |
| GLM-4.5 | `glm45` |
| INTELLECT-3 | `qwen3_coder` |
| Qwen3 | `hermes` |
| Qwen2.5 | `hermes` |
| Llama-4 | `llama4_json` |
| Llama-3 | `llama3_json` |
| OLMo-3 | `olmo3` |
| Trinity | `hermes` |
| SmolLM3 | `hermes` |
| Nemotron | `hermes` |
| DeepSeek-V3/R1 | `deepseek_v3` |
| Mistral | `mistral` |

Explicit `tool_call_parser` in config still takes priority — auto-detection only kicks in when the field is left unset.

## Test plan

- [x] Unit tests for all model patterns (`tests/unit/test_tool_call_parser.py`)
- [x] Existing config loading tests pass (`tests/unit/test_configs.py` — 42 tests)
- [ ] Manual: run inference server with a Qwen3 model without setting `tool_call_parser` and verify it auto-selects `hermes`

---

> [!NOTE]
> **Medium Risk**
> Introduces a new model backend and modifies core attention normalization behavior, which can affect correctness/perf across models; also changes an inference default (`tool_call_parser`) that may alter runtime behavior for existing deployments relying on the implicit parser.
> 
> **Overview**
> Adds a new PrimeRL `minimax_m2` model implementation (config + modeling + HF/PrimeRL MoE weight conversion) and registers it with `AutoConfig`/`AutoModelForCausalLMPrimeRL` for discovery and loading.
> 
> Extends attention to support `qk_norm_type` (`per_head` vs `per_layer`) and updates the mini-checkpoint script (`scripts/mini_moe.py`) to generate/verify both `glm4_moe` and `minimax_m2`, including `trust_remote_code` AutoModel loading when no explicit HF class exists.
> 
> Changes inference `ModelConfig.tool_call_parser` default from `"hermes"` to `None` (disabled), updating the config schema/documentation accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0f3559dfcc4d45297b1cb96d7b1cbbded4ac740f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/inference/config.py
+++ b/src/prime_rl/inference/config.py
@@ -84,11 +84,11 @@ class ModelConfig(BaseConfig):
     ] = False
 
     tool_call_parser: Annotated[
-        str,
+        str | None,
         Field(
-            description="The tool call parser to use. Passed to vLLM as `--tool-call-parser`",
+            description="The tool call parser to use. Passed to vLLM as `--tool-call-parser`. Defaults to None (disabled).",
         ),
-    ] = "hermes"
+    ] = None
 
     reasoning_parser: Annotated[
         str | None,


src/prime_rl/inference/config.py
```diff
@@ -6,6 +7,42 @@
 from prime_rl.utils.pydantic_config import BaseConfig, BaseSettings, get_all_fields
 from prime_rl.utils.utils import rgetattr, rsetattr
 
+logger = logging.getLogger(__name__)
```
mikasenghaas: why std logger? we can acc now get a default prime-rl logger via `get_logger` in config files since #1804

src/prime_rl/inference/config.py
```diff
@@ -6,6 +7,42 @@
...
+
+# Mapping from lowercase model name patterns to vLLM tool call parser names.
+# Order matters: first match wins. Patterns are matched against the lowercased model name.
+MODEL_TOOL_CALL_PARSER_MAP: list[tuple[str, str]] = [
```
mikasenghaas: hm not sure the benefit of having model aliases is worth the overhead of needing to be careful on where to place models for it to resolve correctly

src/prime_rl/inference/config.py
```diff
@@ -104,6 +142,24 @@ class ModelConfig(BaseConfig):
...
+            if pattern in model_name_lower:
+                logger.info(f"Auto-detected tool_call_parser='{parser}' for model '{self.name}'")
+                self.tool_call_parser = parser
+                return self
+
+        raise ValueError(
+            f"Could not auto-detect tool_call_parser for model '{self.name}'. "
+            f"Please set `model.tool_call_parser` explicitly in your config. "
+            f"Supported model patterns: {[p for p, _ in MODEL_TOOL_CALL_PARSER_MAP]}"
+        )
```
mikasenghaas: not sure we should raise here? i feel like if i train without tool calls then i shouldn't need to configure it

samsja: good point

src/prime_rl/inference/config.py
samsja: so we should harcode the model name ?

mikasenghaas: why not?

PR #1805, opened 2026-02-16
fix resume from latest step for single-run

mikasenghaas:
## Summary

Fixes a resume bug where the trainer's resume logic (`resume_step: -1`) would fail to detect orchestrator checkpoints because STABLE marker files were not being created.

## Problem

When resuming a run with `resume_step: -1` (resume from latest checkpoint), the trainer looks for STABLE marker files to determine which step to resume from:
- In `src/prime_rl/trainer/runs.py:267`: `get_stable_ckpt_steps(run_dir / "checkpoints")`  
- In `src/prime_rl/trainer/multi_ckpt.py:194-198`: Similar logic for per-run checkpoints

Previously, only the trainer checkpoint manager created STABLE files. Orchestrator checkpoints existed but had no STABLE markers, causing the trainer to incorrectly default to step 0 even when orchestrator checkpoints existed at later steps.

## Solution

Do not require the run checkpoints directories to have a `STABLE` file to resume from it with `-1` for single-tenant runs.

---

> [!NOTE]
> **Medium Risk**
> Touches checkpoint save/resume logic and `STABLE`-marker semantics; mistakes could cause resuming from the wrong step or reading partially-written checkpoints, though the change is small and well-scoped.
> 
> **Overview**
> Fixes resume-from-latest (`resume_step: -1`) for *single-run* setups by selecting the latest checkpoint step from all `step_*` dirs (via `get_all_ckpt_steps`) instead of requiring `STABLE` marker files, which are only present in multi-run trainer-written checkpoints.
> 
> Refactors checkpoint utilities: orchestrator checkpoint helpers are made public (`save_to_path`/`load_from_path`), and `WeightCheckpointManager` centralizes `STABLE` creation in a new `mark_stable()` and ensures weight checkpoint writes only occur on the master rank before marking the step stable.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 238276a98c9c22ae766b8b58ef59569e80bc745d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/ckpt.py
+++ b/src/prime_rl/orchestrator/ckpt.py
@@ -29,6 +29,11 @@ class CheckpointManager:
     def get_ckpt_path(self, step: int) -> Path:
         return get_step_path(self.ckpt_dir, step) / "orchestrator"
 
+    def mark_stable(self, step: int) -> None:
+        """Write STABLE file to indicate checkpoint is complete."""
+        step_path = get_step_path(self.ckpt_dir, step)
+        (step_path / "STABLE").touch()
+
     def _save_to_path(
         self,
         ckpt_path: Path,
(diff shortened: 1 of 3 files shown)


samsja (approved):
lgtm

PR #1813, opened 2026-02-17
Make get_module_signature calls lazy

rotemtzaban:
The get_module_signature are only used for debug logging, and can be very slow for large models(since they hash all the model's parameters).
Specifically for qwen 30b moe was very slow

---

> [!NOTE]
> **Low Risk**
> Logging-only change; behavior is unchanged except that signature computation is deferred and may be skipped when debug logging is disabled.
> 
> **Overview**
> Switches expensive debug logging of `get_module_signature`/`get_optimizer_signature` to *lazy evaluation* so signatures are only computed if the debug log line is actually emitted.
> 
> This updates checkpoint save/load logging in `trainer/ckpt.py` and model setup logging in `trainer/model.py` to use `logger.opt(lazy=True).debug(...)` with lambda-bound fields, avoiding large-model parameter hashing overhead during normal runs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3710de37269c5fa5da145d1a21fc1561c6fdec24. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/ckpt.py
+++ b/src/prime_rl/trainer/ckpt.py
@@ -214,8 +214,12 @@ class CheckpointManager:
         if not ckpt_path.exists():
             raise FileNotFoundError(f"Checkpoint not found at {ckpt_path}")
         self.load_from_path(ckpt_path, model, optimizers, scheduler, progress, dataloader)
-        self.logger.debug(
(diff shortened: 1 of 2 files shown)


samsja:
hey you made a good point, actually we never want this logs I created a pr to remove them completely https://github.com/PrimeIntellect-ai/prime-rl/pull/1830

PR #1847, opened 2026-02-21
make tito client opt-in

mikasenghaas:
makes the tito client opt-in via `--use-token-client`. if set, uses tito client across all training environments. by default, we use the standard oai client which branches on retokenization discrepancies but interleaves all other messages (due to best-effort-interleave)

---

> [!NOTE]
> **Medium Risk**
> Changes the default inference client type used for training, which can affect request/retokenization behavior and rollout generation; scope is limited to client wiring and config.
> 
> **Overview**
> Training can now explicitly opt into the token-in-token-out (TITO) inference client via `orchestrator.use_token_client`, with a runtime warning when enabled.
> 
> Inference pool setup is updated to accept and propagate a `client_type` through both static and elastic pools, switching the default from the token-based client to the standard `openai_chat_completions` client unless the flag is set. Changelog is updated to document the new config option.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a92c0f9debcd59f8858b0b84462a92fd7e0da9fd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


willccbb:
We probably want a config path to auto-set the flags that the TITO client manually sets (logprobs=True, return_token_ids=True)

mikasenghaas:
> We probably want a config path to auto-set the flags that the TITO client manually sets (logprobs=True, return_token_ids=True)

these are auto-set in `get_sampling_config`. we wouldn't be getting any tokens back if we don't set those so these need to be set whether we are using tito client or not, right?

willccbb (approved):

PR #1857, opened 2026-02-23
Fix weight update race condition between trainer cleanup and orchestrator 

cdreetz:
current bug in prime-rl  when having excessive event loop lag, it will cause the cleanup of old broadcast/step_n that wont happen immediately, then inference will check step_n and see its there, then the delete takes place and the following load fails because step_n is missing by the time it tries to load

The trainer's maybe_clean deletes broadcast checkpoint directories on a timer, but the orchestrator may not have loaded them yet (its event loop can be blocked for 70+ seconds during rollout generation). This caused FileNotFoundError on the inference server and crashed the run.

Fix:
- Orchestrator writes a LOADED_STEP marker after each successful weight update so the trainer knows which checkpoints have been consumed.
- Trainer only cleans broadcast directories the orchestrator has moved past (candidate_step < loaded_step).
- Orchestrator catches update_weights failures and retries on the next poll as defense-in-depth.
- Fix DefaultModelLoader import for vllm >=0.16 (module renamed from loader to default_loader).

---

> [!NOTE]
> **Medium Risk**
> Touches training/orchestration coordination and filesystem deletion behavior, so bugs could lead to missed cleanup or premature deletion; lockfile downgrades may also affect runtime compatibility/performance.
> 
> **Overview**
> Fixes a race between trainer checkpoint cleanup and orchestrator weight loading by having the orchestrator write an atomic `LOADED_STEP` marker after each successful `update_weights`, and by updating trainer `maybe_clean` to only delete `step_*` broadcast dirs that are both old enough *and* strictly below the last loaded step (with a sweep to eventually remove skipped steps while respecting `interval_to_keep`).
> 
> Also updates `uv.lock` to pin older versions of `torch`/`torchaudio`/`torchvision`, `triton`, and `nvidia-nvshmem-cu12`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6b06cd41904fcd636c63dab394c589fa345e508b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/orchestrator/scheduler.py
```diff
@@ -164,13 +164,26 @@ async def update_policy(self):
...
             weights_path = get_step_path(get_broadcast_dir(self.config.output_dir), next_ckpt_step)
-            await self.inference_pool.update_weights(weights_path, lora_name=self.lora_name, step=next_ckpt_step)
+            try:
+                await self.inference_pool.update_weights(weights_path, lora_name=self.lora_name, step=next_ckpt_step)
+            except Exception:
+                self.logger.warning(
+                    f"Failed to load checkpoint {next_ckpt_step} (likely cleaned by trainer). "
+                    f"Will retry with a newer checkpoint on next poll."
+                )
+                return
```
samsja: I don't think we need the try expect here

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -164,13 +164,19 @@ async def update_policy(self):
                 f"Got new policy with step {next_ckpt_step}. Updating weights and cancelling old rollout requests."
             )
 
-            # Update weights on inference servers
+            # Update weights on inference servers.
```
mikasenghaas, line 258: 😭

src/prime_rl/trainer/utils.py
```diff
@@ -323,10 +323,30 @@ def step(self):
...
+    if not loaded_step_file.exists():
+        return
+
+    loaded_step = int(loaded_step_file.read_text().strip())
+    max_step_to_delete = min(candidate_step, loaded_step - 1)
+    if max_step_to_delete < 0:
+        return
+
+    # Sweep all eligible historical steps so skipped candidates are eventually cleaned.
+    for step_dir in path.glob("step_*"):
```
mikasenghaas, line 341: this will break if we ever rename `get_step_path` but ig we wont haha so ig we're good?

mikasenghaas (commented):
i think it should be fine? im a little lost with our ckpt logic (ie which ckpt goes where) since multi tenant but overall looks sensible

PR #1861, opened 2026-02-23
Fix/vlm prompt tokenization

cdreetz:
Fix to get webvoyager env training qwen3 vl


hallerite:
this was fixed in #1877

PR #1864, opened 2026-02-24
Rename use_token_client to trajectory

samsja:
## Summary

- Replaces the boolean `use_token_client` with a `trajectory_mode` field (`"interleave"` | `"branching"`)
- **Interleave** (default): uses the token-level client (`openai_chat_completions_token`). Requires linear history and a chat template with the extension property.
- **Branching**: uses the message-level client (`openai_chat_completions`), re-tokenizing each turn independently. The interleave algorithm still does best-effort compression of branches into a single trajectory when the extension property holds.

## Breaking change

`orchestrator.use_token_client = true/false` → `orchestrator.trajectory_mode = "interleave"/"branching"`

---

> [!NOTE]
> **Medium Risk**
> Breaking config change and it alters which inference client is used during training, which can affect rollout tokenization/throughput and compatibility with non-linear histories.
> 
> **Overview**
> Replaces `orchestrator.use_token_client` with `orchestrator.trajectory_strategy` ("interleaved" | "branching"), where the setting now explicitly controls how multi-turn trajectories are tokenized and which OpenAI client type is used.
> 
> Updates orchestrator startup to pick `openai_chat_completions_token` only for "interleaved" (and emit an updated warning), otherwise defaulting to message-level `openai_chat_completions`. The changelog is updated to document the breaking config rename and new semantics.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6aa87f4c7205d5b83c13ccd42121c50c7e912f05. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


willccbb:
@cursoragent  can we do "interleaved" not "interleave" ?

willccbb:
@cursoragent  Also "trajectory_strategy" for config name pls

mikasenghaas (commented):
not sure i like this renaming. seems less explicit to me. we should warn about the extension property assumption in the tito client in the config description + probably also a warning log or smth

PR #1865, opened 2026-02-24
individual rollouts

faresobeid:
Re-write the scheduler to do rollouts at the rollout level instead of group level. This way any long tails within groups are handled, so when a rollout within a group finishes it can move on to another group.
Currently if any env needs group scoring, we go back to previous behaviour of group rollouts although ideally we make it so verifiers is closer here and we can do scoring within prime-rl

---

> [!NOTE]
> **Medium Risk**
> Rewrites the training rollout scheduler and changes verification/scoring configuration, which can affect throughput, reward computation, and cancellation behavior during training.
> 
> **Overview**
> **Refactors rollout scheduling from group-level to individual rollout tasks** to reduce long-tail stalls: the scheduler now tracks per-example “groups” internally while keeping `max_inflight_rollouts` as the primary concurrency limit, and updates batch inflight/off-policy metrics accordingly.
> 
> Adds *deferred group scoring* support for tasks whose rubrics require group-level reward functions: training envs run with `score_rollouts=False` and the scheduler calls `rubric.score_group()` once all rollouts for an example complete.
> 
> Replaces `orchestrator.buffer.skip_verification` with a new top-level `orchestrator.verification.enabled` switch (with validation preventing reward-dependent buffer features when disabled), updates docs/changelog, and adds a unit test covering off-policy update behavior during interleaved group cancellation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 72d5b878ff395bfaa6a2f69b157a931cb5041add. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/orchestrator/scheduler.py
```diff
@@ -82,14 +79,23 @@ def __init__(
         # Inference pool - used for admin operations (adapter sync) and metrics
         self.inference_pool = inference_pool
 
+        self.use_group_scheduling = any(
```
mikasenghaas: can we not distinguish this per-env?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -134,27 +142,98 @@ async def _select_least_loaded_client(self) -> vf.ClientConfig:
...
                 model_name=self.model_name,
-                rollouts_per_example=self.config.rollouts_per_example,
+                rollouts_per_example=self.rollouts_per_example,
+                sampling_args=self.sampling_args,
+                max_retries=0,  # TODO: make configurable
+            )
+        )
+        self.inflight_requests[run_group_task] = InflightRolloutInfo(off_policy_steps=0, client_config=client_config)
+
+    def _drop_group(self, group_id: int) -> int:
```
mikasenghaas: lets make public

mikasenghaas (commented):
nice, yea lets do some testing on this to verify that we dont have any async race conditions anywhere but directionally looks good.

i think mid-term we want to move away from the verifiers env group for training envs and make our abstractions "multi-env" aware by default, e.g. smth like having a buffer and scheduler per env goverened a "scheduling" component on top or smth bc i think we will want more and more fine-grained control over how each env behaves (e.g. here whether or not to use gruop scheduling) and its always awkward to code this in an abstraction that handles multiple cases where you need conditional everywhere

samsja (approved):

samsja:
lets wait for @mikasenghaas review before merging

src/prime_rl/configs/orchestrator.py
```diff
@@ -899,9 +899,7 @@ def resolve_batching(self):
             if self.max_inflight_rollouts is not None and self.oversampling_factor is not None:
                 expected_max_inflight_rollouts = int(self.batch_size * self.oversampling_factor)
                 if self.max_inflight_rollouts != expected_max_inflight_rollouts:
-                    raise ValueError(
-                        "max_inflight_rollouts conflicts with oversampling_factor * batch_size"
-                    )
+                    raise ValueError("max_inflight_rollouts conflicts with oversampling_factor * batch_size")
```
mikasenghaas, line 887: imo, could just deprectea `oversampling_factor`, no?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -169,6 +169,29 @@ async def orchestrate(config: OrchestratorConfig):
         env_names=train_env_names,
         map_kwargs=dict(writer_batch_size=1),  # set defensively to not error on map operations on large datasets
     )
+    verification_enabled = not config.buffer.skip_verification
```
mikasenghaas: wait i never knew abt this arg, what is it for?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -169,6 +169,29 @@ async def orchestrate(config: OrchestratorConfig):
...
+
+    def task_uses_group_scoring(task_name: str) -> bool:
```
mikasenghaas: would prefer to not have this logic in orch, could maybe put into `vf_utils`?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -82,14 +80,24 @@ def __init__(
...
+
         # Track in-flight requests: task -> info
-        self.inflight_group_rollouts: dict[asyncio.Task, InflightRolloutInfo] = {}
+        self.inflight_requests: dict[asyncio.Task, InflightRolloutInfo] = {}
+
+        # Track in-progress groups while rollouts are generated independently.
+        self.next_group_id = 0
+        self.group_examples: dict[int, dict] = {}
+        self.group_rollouts_to_schedule: dict[int, int] = {}
+        self.completed_group_rollouts: dict[int, list[vf.RolloutOutput]] = defaultdict(list)
```
mikasenghaas: wonder if we need all of these data structures or if some can be merged. e.g. `group_rollouts_to_schedule[group_id]` seems redundant with `group_size - len(completed_group_rollouts[group_id]`

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -134,27 +144,79 @@ async def _select_least_loaded_client(self) -> vf.ClientConfig:
...
                 sampling_args=self.sampling_args,
                 max_retries=0,  # TODO: make configurable
             )
         )
-        self.inflight_group_rollouts[run_group_task] = InflightRolloutInfo(0, client_config)
+        self.inflight_requests[run_rollout_task] = InflightRolloutInfo(
+            off_policy_steps=0, client_config=client_config, group_id=group_id
+        )
+
+    def _inflight_rollout_count(self) -> int:
```
mikasenghaas: can make this public imo, also nice to decorate thes with `@property` imo

tests/unit/orchestrator/test_scheduler.py
```diff
@@ -0,0 +1,53 @@
...
+    scheduler.wait_for_ckpt_time = 0.0
+    scheduler.update_weights_time = 0.0
+    scheduler.step = 7
+    scheduler.ckpt_step = 5
+    scheduler.cancelled_rollouts_count = 4
+    scheduler.inference_pool = SimpleNamespace(get_metrics=lambda: {"pool/servers": 1.0})
+    return scheduler
+
+
+def test_inflight_sample_count_includes_pending_group_rollouts():
```
mikasenghaas: not sure these tests are super useful as is haha

src/prime_rl/configs/orchestrator.py
faresobeid: i agree but @samsja mentioned to keep for backwards compatibility (although im hating this lol)

src/prime_rl/orchestrator/orchestrator.py
faresobeid: added it for on-policy distillation as u dont need to verify the rollout technically. Can also be useful for synthetic datagen

mikasenghaas: why is it a buffer arg tho?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -229,59 +319,76 @@ async def generate_batch(self, step: int) -> list[vf.RolloutOutput]:
...
-        # Manually check the async barrier before starting the step, the re-create the update policy loop
-        # This ensures that we respect max_async_level, while still listening for policy updates mid-step
+        # Check the async barrier before starting, then re-create the update policy loop.
+        # This ensures we respect max_async_level while still listening for policy updates mid-step.
         await self.maybe_update_policy()
         self.update_policy_task = asyncio.create_task(self.update_policy_loop())
 
         batch_start_time = time.perf_counter()
 
+        self.logger.debug("Starting to generate batch rollouts")
```
mikasenghaas, line 329: any reason to fill before the loop and at the end of each loop? is there a problem with filling at the start of the loop only?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -229,59 +319,76 @@ async def generate_batch(self, step: int) -> list[vf.RolloutOutput]:
...
         while batch_progress < self.batch_target:
-            # Fill up
-            while len(self.inflight_group_rollouts) < self.max_inflight_group_rollouts:
-                await self.schedule_group_rollout()
+            inflight_tasks = list(self.inflight_requests.keys())
+            if not inflight_tasks:
+                await self._fill_inflight_requests()
+                inflight_tasks = list(self.inflight_requests.keys())
+                if not inflight_tasks:
+                    raise RuntimeError("No in-flight rollout requests and could not schedule new requests.")
```
mikasenghaas: when would this happen?

mikasenghaas (commented):
getting there:)

src/prime_rl/orchestrator/scheduler.py
faresobeid: ya fair i dont think this would ever realistically happen

samsja (approved):

mikasenghaas (approved):

PR #1872, opened 2026-02-24
add output_dir safety check to prevent accidental overwrites

mikasenghaas:
Training now raises FileExistsError when output_dir already contains files from a previous run, unless explicitly resuming (ckpt.resume_step) or opting in to clean (clean_output_dir=true). This encourages using unique per-experiment output directories and prevents silent data loss. The behavior is now consistent across single-node and multi-node experiments (previously, multi-node experiments had a somewhat weird flag that required the output dir to be manually set). Notably this check is done before deploying (locally or via SLURM) so that the error loop is quick (doesn't depend on SLURM scheduling)

---

> [!NOTE]
> **Medium Risk**
> Changes training entrypoints and default output directory behavior, which can block runs or delete existing outputs when `clean_output_dir` is enabled; guarded by explicit flags but affects core launch paths.
> 
> **Overview**
> Adds a preflight `output_dir` safety check for both `rl` and `sft`: runs now raise `FileExistsError` if the target directory already contains checkpoints, unless resuming (`ckpt.resume_step`) or explicitly opting into deletion via the new `clean_output_dir` flag.
> 
> Updates configs and docs to support this workflow: RL `output_dir` is now always a `Path` defaulting to `"outputs"`, removes the old RL `clean` behavior that implicitly wiped logs/rollouts/broadcasts, and documents the updated SLURM guidance for `output_dir`. Includes unit tests for the new `validate_output_dir` logic.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9991b9fe87a22e8acf0fd45a0d153a95b40dd42c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/rl_config.py
```diff
@@ -176,11 +176,18 @@ class RLConfig(BaseSettings):
...
         ),
-    ] = None
+    ] = Path("outputs")
+
+    clean_output_dir: Annotated[
+        bool,
+        Field(
+            description="If true, delete the output directory before starting training. Required to overwrite an output directory that contains checkpoints from a previous run when not resuming.",
+        ),
+    ] = False
```
samsja, line 190: we should never delete ckpt folder in slurm imo, too dangerous

mikasenghaas: HMM, its false by default tho. i think if the user sets this explicitly we should allow him to do that. i think its weird to make a distinction between local and slurm. ALL EXPERIMENTS ARE MADE EQUAL

src/prime_rl/entrypoints/rl.py
```diff
@@ -473,6 +434,11 @@ def rl_slurm(config: RLConfig):
 
 
 def rl(config: RLConfig):
+    resuming = config.ckpt is not None and config.ckpt.resume_step is not None
+    clean = config.clean_output_dir and not os.environ.get("NEVER_CLEAN_OUTPUT_DIR")
+    validate_output_dir(config.output_dir, resuming=resuming, clean=clean)
+    config.output_dir.mkdir(parents=True, exist_ok=True)
+
```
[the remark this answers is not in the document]
mikasenghaas: this is fine

samsja (approved):

PR #1875, opened 2026-02-24
fix: per-component seq_len overrides shared seq_len

mikasenghaas:
This PR is motivated by the fact the setting `--trainer.model.seq-len` to a different value than `--orchestrator.seq-len` is not allowed when a shared `seq_len` is configured on the RL entrypoint. Currently, we unconditionally overwrite whenever a shared value is set. However, the better behavior would be to have the following precedence:

Custom Per-Component > Custom Shared > Default Per-Component

This can be done in Pydantic by checking whether a custom-per component config is set via `model_fields_set`. However, this only works when `nested_model_default_partial_update=False`. Our base config had this to true but we don't need this, so this PR reverts this to allow proper config precedence logic.

## Summary

- **Remove `nested_model_default_partial_update=True`** from `BaseSettings` — it was unnecessary (all nested defaults are `SomeConfig()` with zero args) and it broke `model_fields_set` on nested models by merging defaults into the raw dict before construction
- **Bump pydantic-settings >=2.12.0** for improved nested model handling
- **Replace `normalize_deployment` workaround** with a generic `_default_discriminator_types` before-validator on `BaseSettings` that injects the `type` discriminator tag from a field's default instance when missing from the config
- **Guard `auto_setup_seq_len` with `model_fields_set` checks** so explicitly set per-component `seq_len` values take precedence over the shared `seq_len`

---

> [!NOTE]
> **Medium Risk**
> Touches core config parsing/validation and training entrypoint behavior, which could change how existing TOML/CLI configs resolve and when runs refuse to start due to `output_dir` state.
> 
> **Overview**
> Ensures **per-component `seq_len` overrides the shared `RLConfig.seq_len`** by only propagating the shared value to components that didn’t explicitly set `seq_len` (using `model_fields_set`).
> 
> Changes config parsing by removing `nested_model_default_partial_update` and adding a `BaseSettings` before-validator that auto-injects missing discriminator `type` fields from defaults; bumps `pydantic-settings` accordingly and adds unit tests covering nested overrides and discriminated unions.
> 
> Makes `output_dir` handling safer and more consistent: `RLConfig.output_dir` now defaults to `"outputs"` even under SLURM, adds `clean_output_dir` (RL+SFT), removes RL’s implicit `clean` behavior, and adds `validate_output_dir` to block accidental reuse of checkpointed directories unless resuming or explicitly cleaning (with tests). Also updates docs and adds a new `examples/hendrycks_sanity` RL+SLURM example.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3bdf1feaf4f7a48be60dd2a165965f3c167ebc48. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/utils/pydantic_config.py
```diff
@@ -85,6 +85,25 @@ def settings_customise_sources(
...
+        """
+        if not isinstance(data, dict):
+            return data
+        for field_name, field_info in cls.model_fields.items():
+            val = data.get(field_name)
+            if isinstance(val, dict) and "type" not in val:
+                default = field_info.default
+                if isinstance(default, BaseModel) and hasattr(default, "type"):
+                    val["type"] = default.type
+        return data
```
samsja, line 105: is this backpropagated everywhere ?

samsja (approved):
lgtm

src/prime_rl/utils/pydantic_config.py
mikasenghaas: yeap, we need it bc otherwise the discriminated union requires the discriminator field being set instead of using the default

mikasenghaas: added some tests for this to be sure its correct

PR #1882, opened 2026-02-24
chore: move tool_call_parser resolution to server startup

samsja:
## Summary
- Move `MODEL_TOOL_CALL_PARSER` dict and auto-detection logic from `configs/inference.py` to `inference/vllm/server.py`
- Resolve tool call parser at server startup instead of config validation time, so it always sees the final model name
- Remove the reset+re-resolve hack in `RLConfig.auto_setup_model()`

## Test plan
- [x] `uv run pytest tests/unit/test_tool_call_parser.py` — 20 passed
- [x] `uv run pytest tests/unit/test_configs.py` — 59 passed
- [x] `uv run ruff check && ruff format --check` — clean

---

> [!NOTE]
> **Medium Risk**
> Changes inference server startup argument resolution for tool calling; mis-resolution could silently disable/enable tool parsing and affect runtime behavior, but the change is localized and covered by unit tests.
> 
> **Overview**
> Tool-call parser selection is now resolved at vLLM server startup: `MODEL_TOOL_CALL_PARSER` and a new `resolve_tool_call_parser()` helper move from `configs/inference.py` into `inference/vllm/server.py`, enabling `tool_call_parser = "auto"` to be interpreted using the final model name and setting `enable_auto_tool_choice` based on whether a parser is resolved.
> 
> Config-layer auto-resolution is removed (drops `enable_auto_tool_choice` from `ModelConfig`, removes the Pydantic validator that inferred the parser, and deletes the reset/re-resolve workaround in `RLConfig.auto_setup_model()`), and example/config TOMLs stop setting `enable_auto_tool_choice` explicitly. Unit tests are updated to cover the new startup-time resolution behavior, including unknown-model and `None` cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 98a0804d03b0f27fe04ee1f79531cccf4c83c847. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/configs/env_mix/env_mix.toml
+++ b/configs/env_mix/env_mix.toml
@@ -48,6 +48,5 @@ api_server_count = 8
 gpu_memory_utilization = 0.8
 
 [inference.model]
-enable_auto_tool_choice = true
 tool_call_parser = "hermes"
 max_model_len = 8192
--- a/configs/math_python/math_python.toml
+++ b/configs/math_python/math_python.toml
@@ -25,6 +25,5 @@ args = { sandbox_client_max_workers = 128, sandbox_timeout_minutes = 10, sandbox
 api_server_count = 4
 
 [inference.model]
-enable_auto_tool_choice = true
 tool_call_parser = "hermes"
 max_model_len = 4096
--- a/configs/wiki_search/rl.toml
+++ b/configs/wiki_search/rl.toml
@@ -40,5 +40,4 @@ online_difficulty_filtering = true
 id = "primeintellect/wiki-search"
 
 [inference.model]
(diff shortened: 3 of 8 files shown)


mikasenghaas (commented):
i think it would be nicer to have this sort of auto resolution in the configs. its the pattern we use in general and helps to e.g. have the configs that are logged to W&B actually be the ones that were used. if it's an issue of precedence (e.g. ig right now we first resolve inference config and then rl config) which means that a model set at the rl config level will not get resolved, we can defo solve this at the pydantic level

PR #1902, opened 2026-02-26
fix off policy off by one

mikasenghaas:
naturally, `max_off_policy_steps=n` should mean that at most `n` _different_ policy versions are allowed to generate a rollout. with `n=0` this means that a rollout has to be generated by a single policy, no mid-rollout policy update is allowed. our current is off by one from this logic. this pr fixes it and handles the edge case of all in-flight rollouts being cancelled in at once when setting aggressively `max_off_policy_steps=0`. e.g in the run below we do not allow any off-policyness, yet the purple (current main) includes rollouts that were generated by 2 distinct policies. the red line (this pr) fixes this

Screenshot 2026-02-26 at 4 27 30 PM


mikasenghaas:
making this part of #1901

PR #1513, opened 2025-12-30
fix: handle skipped eval intervals when ckpt_step jumps

JannikSt:
## Problem

When `strict_async_level=False` (default), the scheduler can jump checkpoint steps. If ckpt_step jumps over a multiple of the eval interval, that eval is silently skipped.

## Example

With `eval.interval=10`, here's what happens when ckpt_step jumps:

```
Step 10 | Async Level: 1 -> ckpt_step = 9  (9 % 10 != 0, no eval)
Step 11 | Async Level: 0 -> ckpt_step = 11 (11 % 10 != 0, no eval)
```

Checkpoint 10 was skipped, so eval at interval=10 never ran. Only the base model eval at step 0 executed.

## Root cause

In `scheduler.py`, when `strict_async_level=False`:
```python
next_ckpt_step = max(async_away_ckpt_step, latest_ckpt_step)
```

This allows jumping from ckpt_step 9 directly to 11 if checkpoints 10 and 11 are both available.

The eval check in `orchestrator.py` only tested the current ckpt_step:
```python
if ckpt_step % config.eval.interval == 0:  # misses skipped steps
```

## Fix

Track previous ckpt_step and find the highest interval step in range (prev, current] that should trigger eval:
```python
highest_interval_step = (ckpt_step // interval) * interval
if highest_interval_step > prev_ckpt_step and highest_interval_step > last_eval_step:
    eval_ckpt_step = highest_interval_step
```

Now if ckpt_step jumps from 9->11, we detect that step 10 was skipped and run eval at step 10.

New log message when this happens:
```
Running evals for skipped checkpoint step 10 (current ckpt_step=11)
```

---

> [!NOTE]
> **Medium Risk**
> Changes the orchestrator’s online-eval scheduling logic and resume behavior; mistakes here could cause evals to run too often/too rarely or be labeled against the wrong checkpoint step, but it doesn’t touch training data or model updates directly.
> 
> **Overview**
> Fixes missed online evaluations when `scheduler.ckpt_step` jumps over multiples of `eval.interval` by tracking `prev_ckpt_step` and triggering an eval if any interval boundary was crossed since the prior loop iteration.
> 
> Adjusts resume behavior so `skip_eval_on_resume` correctly suppresses the immediate eval at the resumed checkpoint, and adds logging when an interval step was skipped but eval is run at the current checkpoint (while still labeling results with the current `ckpt_step`).
> 
[...]

PR #1903, opened 2026-02-26
fix: handle skipped eval intervals when ckpt_step jumps

JannikSt:
Supersedes #1513 (rebased on main + additional fix).

Two bugs with eval scheduling when `strict_async_level=False`:

- **Evals permanently skipped**: `ckpt_step` can jump over interval boundaries (e.g. 24→26 skipping 25) because `update_policy()` sets it to `max(async_away, latest_ckpt_step)`. The old exact modulo check `ckpt_step % interval == 0` misses the jump. Replaced with a range check that finds the highest interval step in `(prev_ckpt_step, ckpt_step]`.
- **Eval results reported at wrong step**: eval metrics were logged with `progress.step` (orchestrator step) instead of `ckpt_step`, so e.g. eval for checkpoint 50 would show up at step 53 in the API. Fixed to use `ckpt_step`.

Confirmed on run `c14miuyha2yhxkw1z3eqgyub` — step 25 eval was permanently missed, step 50 eval results appeared at step 53.

---

> [!NOTE]
> **Medium Risk**
> Touches the orchestrator’s main loop and changes when online evals run, which can affect training throughput and monitoring cadence if the interval logic is wrong. Covered by new unit tests but still impacts a critical control-flow path.
> 
> **Overview**
> Fixes online eval scheduling to handle `ckpt_step` jumps that previously skipped interval-aligned evals by introducing `compute_eval_ckpt_step()` and using it instead of the strict modulo check.
> 
> Updates the orchestrator loop to track `prev_ckpt_step` (including resume behavior) and to log when an eval is triggered for an interval step different from the current checkpoint step. Adds focused unit tests covering exact hits, jumps across boundaries, base-model step 0 behavior, and regression scenarios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit aa06cc307a25aa517094497b2232eaa01c0f1d3c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/eval_utils.py
+++ b/src/prime_rl/orchestrator/eval_utils.py
@@ -13,6 +13,32 @@ from prime_rl.utils.monitor import get_monitor
 from prime_rl.utils.utils import capitalize
 
 
+def compute_eval_ckpt_step(
+    ckpt_step: int,
+    prev_ckpt_step: int,
+    last_eval_step: int,
+    interval: int,
+    eval_base_model: bool = True,
+) -> int | None:
+    """Determine which checkpoint step (if any) should trigger an eval.
+
+    Handles the case where ckpt_step jumps over interval boundaries by finding
+    the highest interval-aligned step in (prev_ckpt_step, ckpt_step].
+
+    Returns the interval step to eval at, or None if no eval should run.
+    """
+    if ckpt_step <= prev_ckpt_step:
+        return None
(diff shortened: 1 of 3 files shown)


samsja (approved):
lgtm and to claude

PR #1905, opened 2026-02-26
SFT distillation with teacher endpoint

willccbb:
> [!NOTE]
> **Medium Risk**
> Touches core training/orchestration flow (inference pool setup, checkpoint/policy update behavior, tokenization) and introduces a new loss path, which could affect training stability and rollout correctness if misconfigured.
> 
> **Overview**
> Adds an *external teacher rollout* mode for RL runs: new `orchestrator.teacher_rollout_model` switches rollout generation to an OpenAI-compatible endpoint and requires `trainer.loss.type = "sft"`, no `[inference]`, and `use_token_client = false` (validated in `RLConfig` and surfaced in the `rl` entrypoint logs).
> 
> Implements the new SFT-style loss (`trainer.loss.type = "sft"`) and updates the orchestrator/scheduler to disable policy updates/weight broadcast when using an external rollout model, while reconstructing missing rollout token data from prompt/completion messages (including tool calls and multimodal `data:image/...`) via shared `chat_template` utilities. Adds docs, example configs, and unit tests covering the new mode and token reconstruction behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5644df3ee12b46d981e7140088c2854dc20a64f5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/configs/rl.py
```diff
@@ -326,6 +326,29 @@ def validate_teacher_model(self):
...
+                "inference must be omitted when orchestrator.rollout_model is configured. "
+                "External rollout mode does not use the local inference server."
+            )
+
+        if self.orchestrator.use_token_client:
+            warnings.warn(
+                "orchestrator.rollout_model is configured. Setting orchestrator.use_token_client=false "
+                "for text-level reconstruction in SFT distillation mode."
+            )
+            self.orchestrator.use_token_client = False
```
samsja: lets rather fail here ?

src/prime_rl/configs/orchestrator.py
```diff
@@ -669,6 +683,17 @@ class OrchestratorConfig(BaseConfig):
         ),
     ] = None
 
+    # External rollout model configuration (optional)
+    rollout_model: Annotated[
```
samsja: I think this name should be more specific, rollout_model ppl will think its the model doing rollout during normal rl

src/prime_rl/configs/orchestrator.py
```diff
@@ -861,6 +886,12 @@ def nccl_max_async_level(self):
                 raise ValueError("max_async_level must be 1 for NCCL broadcast")
         return self
 
+    @model_validator(mode="after")
+    def validate_rollout_model_with_weight_broadcast(self):
+        if self.rollout_model is not None and self.weight_broadcast.type == "nccl":
+            raise ValueError("rollout_model is not compatible with NCCL weight broadcast.")
+        return self
```
samsja: not sure if we need this ? technically nor nccl nor filesystem are compatible

src/prime_rl/configs/orchestrator.py
willccbb: teacher_rollout_model ?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -103,14 +103,28 @@ async def orchestrate(config: OrchestratorConfig):
...
+    rollout_model_name = config.model.name
+    enable_policy_updates = config.rollout_model is None
+
+    if config.rollout_model is not None:
+        rollout_client_config = config.rollout_model.client
+        rollout_model_name = config.rollout_model.model.name
+        logger.info(
+            f"Using external rollout model (base_url={', '.join(rollout_client_config.base_url)}, "
+            f"model={rollout_model_name})"
+        )
```
samsja: can we move this to a setup_external_rollout_model function ? (or any name)

src/prime_rl/orchestrator/trajectories.py
```diff
@@ -35,8 +38,146 @@ def _align_routed_experts(
...
+    return {
+        "prompt_ids": prompt_ids,
+        "prompt_mask": [False] * len(prompt_ids),
+        "completion_ids": completion_ids,
+        "completion_mask": completion_mask,
+        "completion_logprobs": completion_logprobs,
+        "routed_experts": None,
+        "prompt_prefix_len": split_idx,
+    }
+
```
samsja, line 125: we have this logic define in the sft code, is there a way to share it ?

src/prime_rl/orchestrator/trajectories.py
```diff
@@ -74,10 +216,42 @@ def interleave_rollout(
     # this field should be guaranteed because we set temperature in get_sampling_args
     temperature = output["sampling_args"]["temperature"]
```
samsja, line 188: it doesn't feel natural to have the tokenizer in the interleave_rollout case, shouldn't call a different function in the case of on policy sft ?

src/prime_rl/orchestrator/utils.py
```diff
@@ -37,22 +37,37 @@ async def get_semaphore() -> AsyncContextManager:
...
+            "top_k": -1,
+            "min_p": 0.0,
+            "min_tokens": min_tokens,
+            "repetition_penalty": repetition_penalty,
+        }
+    else:
+        if min_tokens > 0:
+            extra_body["min_tokens"] = min_tokens
+        if repetition_penalty != 1.0:
+            extra_body["repetition_penalty"] = repetition_penalty
```
samsja: I don't think we should have different argument here base on the tito client, this seems dangerous to have different behavior

src/prime_rl/trainer/rl/loss.py
```diff
@@ -174,6 +186,13 @@ def loss_fn(inputs: LossInputs) -> LossOutputs:
 
         return loss_fn
 
+    if isinstance(loss_config, SFTLossConfig):
+
+        def loss_fn(inputs: LossInputs) -> LossOutputs:
+            return sft_loss_fn(inputs)
+
+        return loss_fn
```
samsja:
```suggestion
        return setup_loss_fn
```

isn't enough just to do this ?

src/prime_rl/orchestrator/utils.py
willccbb: not all endpoints support the full args so it's a bit tricky, the extra_body schema we're using is vLLM-specific

willccbb: can make the branching cleaner

tests/unit/orchestrator/test_trajectories.py
```diff
@@ -29,6 +30,63 @@ def _decode_pixels(pixel_bytes: bytes, shape: list[int]) -> list[list[float]]:
     return np.frombuffer(pixel_bytes, dtype=np.float32).reshape(shape).tolist()
 
 
+class SimpleChatTokenizer:
+    """Minimal tokenizer mock for testing message-based rollout reconstruction."""
+
+    def __init__(self):
```
samsja: can we move the sft test to a test_sft_trajectory file, want to isolated the two behavior, easier to think about

tests/unit/test_configs.py
```diff
@@ -148,3 +148,66 @@ def test_cli_overrides_toml(tmp_path):
...
+            "use_token_client": False,
+            "rollout_model": {
+                "client": {"base_url": ["https://example.com/v1"], "skip_model_check": True},
+                "model": {"name": "gpt-4o-mini"},
+            },
+        },
+        inference=None,
+    )
+    assert config.orchestrator.rollout_model is not None
+    assert config.inference is None
```
samsja: we need a openai key to run this test right ? would mark it as skip or spawn a vllm server instead

tests/unit/test_configs.py
```diff
@@ -148,3 +148,66 @@ def test_cli_overrides_toml(tmp_path):
...
+            trainer={"loss": {"type": "sft"}},
+            orchestrator={
+                "use_token_client": False,
+                "rollout_model": {
+                    "client": {"base_url": ["https://example.com/v1"], "skip_model_check": True},
+                    "model": {"name": "gpt-4o-mini"},
+                },
+            },
+            inference={},
+        )
```
samsja, line 165: I dont think we need this test, its redundant with the config definition

tests/unit/test_configs.py
```diff
@@ -148,3 +148,66 @@ def test_cli_overrides_toml(tmp_path):
...
+        orchestrator={
+            "use_token_client": True,
+            "rollout_model": {
+                "client": {"base_url": ["https://example.com/v1"], "skip_model_check": True},
+                "model": {"name": "gpt-4o-mini"},
+            },
+        },
+        inference=None,
+    )
+    assert config.orchestrator.use_token_client is False
```
samsja: also redundant

tests/unit/test_configs.py
```diff
@@ -148,3 +148,66 @@ def test_cli_overrides_toml(tmp_path):
...
+        RLConfig(
+            trainer={"loss": {"type": "default"}},
+            orchestrator={
+                "rollout_model": {
+                    "client": {"base_url": ["https://example.com/v1"], "skip_model_check": True},
+                    "model": {"name": "gpt-4o-mini"},
+                },
+            },
+            inference=None,
+        )
```
samsja: also redundant

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -385,6 +406,9 @@ async def orchestrate(config: OrchestratorConfig):
 
         # Capture ckpt_step once for consistency (it's updated inside the scheduler)
         ckpt_step = scheduler.ckpt_step
+        if not enable_policy_updates:
+            ckpt_step = progress.step
+            scheduler.ckpt_step = progress.step
```
samsja: not sure to understand why we don't need this variable when not doing policy update, I feel we can jsut remove the if statement here ?

src/prime_rl/orchestrator/trajectories.py
willccbb: a lot of the logic is shared between the interleaving + our conversion from teacher rollouts to SFT batches -- we're still going from trajectorystep to packed/masked sequences, but the trajectory steps don't already have token_ids/masks so we need to retokenize from messages.

think we wanna reuse as much as possible, but will see if there's a good way we can make the tokenizer usage more encapsulated

samsja (commented):
added minor comment to the pr.

Generally looks good, I would like to not put the sft logic inside the interleave_rollout one, this  rather seems like two level of abstraction one deal with token the other one with chat template for sft. 

Also if we can reuse the chat template logic for the sft entrypoint would be great

src/prime_rl/configs/orchestrator.py
samsja: hmm wondering if its not confusing with real OPD then who also has a teacher_rollout_model, what about sft_rollout_model ?

src/prime_rl/orchestrator/trajectories.py
willccbb: ah okay yeah we can pre-tokenize before interleaving

src/prime_rl/configs/orchestrator.py
willccbb: teacher isn't doing the rollouts there though? just for prefill logprobs

src/prime_rl/orchestrator/utils.py
```diff
@@ -37,22 +37,35 @@ async def get_semaphore() -> AsyncContextManager:
...
+        extra_body["repetition_penalty"] = repetition_penalty
+
+    if use_token_client:
+        sampling_args["logprobs"] = True
+        extra_body = {
+            **extra_body,
+            "return_token_ids": True,  # Always return token IDs
+            "top_k": -1,
+            "min_p": 0.0,
+        }
```
samsja, line 64: still unsure of only passing topk and minp for token client only, can we make sure both are pass ?

mikasenghaas:
alr merged

PR #1915, opened 2026-02-28
migrate config system to pydantic_config

samsja:
## Summary
- Replace the custom `pydantic-settings` wrapper (`BaseSettings`, `parse_argv`, `toml_files` inheritance) with the external [`pydantic_config`](https://github.com/samsja/pydantic_config) library backed by tyro
- Delete `src/prime_rl/utils/pydantic_config.py` (302 lines) and replace with a thin 50-line wrapper in `src/prime_rl/utils/config.py`
- Replace `allow_extras` pattern in inference server with a `vllm_extra: dict` field on `InferenceConfig`
- Remove `toml_files` inheritance from example SLURM configs (use `@ base.toml @ slurm.toml` composition instead)

### Boolean negation flags changed location

The `--no-` prefix now attaches to the field name, not the full dotted path:

```bash
# Before
uv run trainer @ config.toml --no-model.compile

# After
uv run trainer @ config.toml --model.no-compile
```

### vLLM extra arguments use `vllm_extra` dict instead of passthrough CLI args

The old system captured unknown CLI args and forwarded them to vLLM. The new system uses an explicit `vllm_extra` field:

```bash
# Before (unknown args passed through)
uv run inference @ config.toml --some-vllm-arg value

# After (use vllm_extra in TOML)
```

```toml
# In config TOML
vllm_extra = { some_vllm_arg = "value" }
```

## Test plan
- [x] `uv sync --all-extras` resolves successfully
- [x] `uv run pytest tests/unit/test_configs.py -v` — all 58 tests pass
- [x] `uv run python -c "from prime_rl.utils.config import BaseConfig, cli"` — imports work
- [x] No remaining imports of `prime_rl.utils.pydantic_config` in codebase
- [x] No remaining imports of `pydantic_settings` in codebase


src/prime_rl/configs/inference.py
```diff
@@ -237,6 +237,13 @@ class InferenceConfig(BaseSettings):
         ),
     ] = False
 
+    vllm_extra: Annotated[
```
mikasenghaas, line 240: should we just call this `vllm` and deprecate all vllm fields we alr put in our inference config?

PR #1928, opened 2026-03-02
feat: add async filesystem weight broadcast (type=async_filesystem)

shayonj:
During each training step, the filesystem weight broadcast gathers the full model state dict from GPU to CPU and then serializes it to disk (safetensors) before the trainer can proceed. In the synchronous path, the trainer is blocked for the full gather + write duration, which scales with model size and filesystem throughput. On shared/network filesystems, this can become a significant source of step latency.

We introduce `AsyncFileSystemWeightBroadcast`, a drop-in option activated through config:

```toml
[weight_broadcast]
type = "async_filesystem"
```

It splits broadcast into three phases:

1. **Synchronous FSDP gather with non-blocking DMA.** In `gather_weights_on_master`, when `non_blocking=True`, each tensor is copied into pinned CPU memory (`torch.empty(..., pin_memory=True)` + `.copy_(..., non_blocking=True)`) instead of blocking `.to("cpu")`. This queues DMA on the default CUDA stream without host blocking. A `torch.cuda.Event` is recorded after gather and used as a fence before background write reads tensor data.

2. **Main-thread state preparation.** Run metadata (paths, step counters, ready flags) is snapshotted on the main thread before background submission to avoid races with trainer-side state changes.

3. **Background serialization and disk I/O.** A single-worker `ThreadPoolExecutor` receives the pinned state dict, CUDA event, and run metadata; it calls `cuda_event.synchronize()` to drain in-flight DMA, then performs `save_state_dict` + STABLE marker notification off the training critical path. Backpressure is preserved: if a previous write is still in flight, `_wait_for_pending()` blocks the next broadcast, so worst-case behavior matches the synchronous path.

A `shutdown()` method on `WeightBroadcast` ensures in-flight async writes are drained before final checkpointing in `train.py`.

Wall-clock time for steps 1–19:
- sync = 67s (3.72s/step)
- async = 58s (3.22s/step) -> **13.4% faster**

Broadcast blocking time per step:
- sync ~= 1.5s (gather + write inline)
- async ~= 0.16s (gather-only blocking; write runs in background at ~1.5s with zero backpressure across all 20 steps)

These measurements are from a 0.6B model (~1.2 GB state dict).

---

> [!NOTE]
> **Medium Risk**
> Introduces a new threaded weight-broadcast path and non-blocking GPU→CPU transfers, which can surface race conditions or incomplete writes if mis-synchronized. Changes touch the trainer’s step loop and broadcast validation/config wiring, but are scoped to the filesystem broadcast family and include unit tests.
> 
> **Overview**
> Adds a new `async_filesystem` weight broadcast type that gathers weights synchronously but performs serialization and disk writes asynchronously on the trainer to reduce step blocking time.
> 
> Wires the new type through config (`SharedWeightBroadcastConfig`, trainer `WeightBroadcastConfig` union) and broadcast setup, including logic to keep inference/orchestrator using the `filesystem` protocol and to treat `filesystem`/`async_filesystem` as compatible during shared-config validation.
> 
> Updates training to treat both filesystem-based broadcasts equivalently for final-step behavior and cleanup, and adds `WeightBroadcast.shutdown()` so training drains any in-flight async writes before final checkpointing; includes unit tests covering config dispatch and async broadcast lifecycle helpers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e12ba55ce88ae78bfadf8dc29cf8c80561e919ba. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/rl/broadcast/__init__.py
+++ b/src/prime_rl/trainer/rl/broadcast/__init__.py
@@ -3,23 +3,20 @@ from pathlib import Path
 import torch
 
 from prime_rl.configs.trainer import LoRAConfig, WeightBroadcastConfig
+from prime_rl.trainer.rl.broadcast.async_filesystem import AsyncFileSystemWeightBroadcast
 from prime_rl.trainer.rl.broadcast.base import WeightBroadcast
+from prime_rl.trainer.rl.broadcast.filesystem import FileSystemWeightBroadcast
+from prime_rl.trainer.rl.broadcast.nccl import NCCLWeightBroadcast
 
 
 def setup_weight_broadcast(
     output_dir: Path, config: WeightBroadcastConfig, lora_config: LoRAConfig | None = None
 ) -> WeightBroadcast:
     if config.type == "nccl":
-        from prime_rl.trainer.rl.broadcast.nccl import NCCLWeightBroadcast
-
         return NCCLWeightBroadcast(output_dir, config, torch.cuda.current_device())
     elif config.type == "async_filesystem":
-        from prime_rl.trainer.rl.broadcast.async_filesystem import AsyncFileSystemWeightBroadcast
-
         return AsyncFileSystemWeightBroadcast(output_dir, config, lora_config)
     elif config.type == "filesystem":
-        from prime_rl.trainer.rl.broadcast.filesystem import FileSystemWeightBroadcast
-
         return FileSystemWeightBroadcast(output_dir, config, lora_config)
     else:
         raise ValueError(f"Invalid weight broadcast type: {config.type}")
--- a/src/prime_rl/trainer/rl/broadcast/__init__.py
+++ b/src/prime_rl/trainer/rl/broadcast/__init__.py
@@ -4,16 +4,22 @@ import torch
 
 from prime_rl.configs.trainer import LoRAConfig, WeightBroadcastConfig
 from prime_rl.trainer.rl.broadcast.base import WeightBroadcast
-from prime_rl.trainer.rl.broadcast.filesystem import FileSystemWeightBroadcast
-from prime_rl.trainer.rl.broadcast.nccl import NCCLWeightBroadcast
 
 
 def setup_weight_broadcast(
     output_dir: Path, config: WeightBroadcastConfig, lora_config: LoRAConfig | None = None
 ) -> WeightBroadcast:
     if config.type == "nccl":
+        from prime_rl.trainer.rl.broadcast.nccl import NCCLWeightBroadcast
+
         return NCCLWeightBroadcast(output_dir, config, torch.cuda.current_device())
+    elif config.type == "async_filesystem":
+        from prime_rl.trainer.rl.broadcast.async_filesystem import AsyncFileSystemWeightBroadcast
+
+        return AsyncFileSystemWeightBroadcast(output_dir, config, lora_config)
     elif config.type == "filesystem":
+        from prime_rl.trainer.rl.broadcast.filesystem import FileSystemWeightBroadcast
+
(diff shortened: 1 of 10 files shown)


src/prime_rl/trainer/rl/broadcast/async_filesystem.py
```diff
@@ -0,0 +1,242 @@
...
+                    save_lora_config(model, save_dir, rank=rank, alpha=alpha, dropout=dropout)
+
+                self._notify_orchestrator(save_dir)
+
+                # Avoid zombie run directories that get recreated by mkdir
+                # after the orchestrator deletes them during broadcast
+                if self.multi_run_manager.get_orchestrator_config(self.multi_run_manager.idx_2_id[idx]) is None:
+                    shutil.rmtree(run_dir)
+            except FileNotFoundError:
+                self.logger.warning(f"[async] Run {idx} directory deleted during broadcast, skipping")
```
mikasenghaas: nit: no need for `[async]` prefix. in debug mode, we append the file and line num we log from

src/prime_rl/trainer/rl/broadcast/async_filesystem.py
```diff
@@ -0,0 +1,242 @@
...
+            except Exception as e:
+                self.logger.error(f"[async] Error broadcasting weights for run {idx}: {e}")
+
+        self._last_broadcast_time = time.perf_counter() - write_start
+        self.logger.info(f"[BENCH] async background write completed in {self._last_broadcast_time:.3f}s")
```
mikasenghaas: lets not info log bench logic

src/prime_rl/utils/validation.py
```diff
@@ -100,19 +100,33 @@ def validate_shared_max_async_level(
         )
 
 
+def _broadcast_type_family(type_name: str) -> str:
```
mikasenghaas: can inline this function and keep public imo

src/prime_rl/trainer/rl/broadcast/__init__.py
```diff
@@ -4,16 +4,22 @@
...
 from prime_rl.trainer.rl.broadcast.base import WeightBroadcast
-from prime_rl.trainer.rl.broadcast.filesystem import FileSystemWeightBroadcast
-from prime_rl.trainer.rl.broadcast.nccl import NCCLWeightBroadcast
 
 
 def setup_weight_broadcast(
     output_dir: Path, config: WeightBroadcastConfig, lora_config: LoRAConfig | None = None
 ) -> WeightBroadcast:
     if config.type == "nccl":
+        from prime_rl.trainer.rl.broadcast.nccl import NCCLWeightBroadcast
```
mikasenghaas: any reason to not import module level?

mikasenghaas (commented):
i remember we had some silent races with async fs back in the days. today, we just use nccl for proper sized models anyways

src/prime_rl/trainer/rl/broadcast/async_filesystem.py
shayonj: secret is out! :D

src/prime_rl/trainer/rl/broadcast/__init__.py
shayonj: ah! nope, just old habit. Better to be consistent. 1s

shayonj:
that makes sense! Happy to sit on this or close this too given we rely on nccl. WDYT @mikasenghaas ?

shayonj:
chatted, gonna close given the nccl benefits. apreesh the review 🙏🏾

PR #1938, opened 2026-03-03
fix: validate prompt length in TITO endpoint before computing max_tokens

mikasenghaas:
## Summary

- In multi-turn TITO rollouts (turns > 1), `prompt_token_ids` are injected *after* `render_chat_request`, bypassing vLLM's built-in context length validation
- When the prompt exceeds `max_model_len`, `get_max_tokens` returns a negative value (e.g. `-17`), causing a `BadRequestError` (`"max_tokens must be at least 1"`)
- This error message isn't recognized by the verifiers `handle_openai_overlong_prompt` decorator, so the rollout is **aborted** instead of gracefully **truncated**
- Adds an explicit prompt length check before `get_max_tokens` that raises a `VLLMValidationError` with the same "context length" message format as vLLM's MITO path, so verifiers catches it as an `OverlongPromptError`

## Example

```bash
# start inference with --max-model-len 256

uv run vf-eval alphabet-sort -d -v -n1 -r1 -m Qwen/Qwen3-4B-Instruct-2507 -p local -a '{"min_turns": 5, "max_turns": 5}' --api-client-type openai_chat_completions_token -t 128
```

The rollout fails on an error (and would be masked out)

Screenshot 2026-03-03 at 2 13 48 PM

Now, we correctly catch the overlong prompt and treat the rollout as truncated but valid

Screenshot 2026-03-03 at 2 45 08 PM

```txt
is_truncated: avg - 1.000, std - 0.000
stop_conditions: prompt_too_long: 1.000
```

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit b1605daf704e3e95a1f597ad9dcc108ae9b955c7. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/inference/vllm/serving_chat_with_tokens.py
+++ b/src/prime_rl/inference/vllm/serving_chat_with_tokens.py
@@ -10,6 +10,7 @@ from vllm.entrypoints.openai.chat_completion.serving import OpenAIServingChat
 from vllm.entrypoints.openai.engine.protocol import ErrorResponse, RequestResponseMetadata
 from vllm.entrypoints.openai.engine.serving import GenerationError
 from vllm.entrypoints.utils import get_max_tokens
+from vllm.exceptions import VLLMValidationError
 from vllm.logger import init_logger
 from vllm.outputs import RequestOutput
 from vllm.reasoning import ReasoningParser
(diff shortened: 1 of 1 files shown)


samsja (approved):

hallerite (approved):
lgtm

PR #1939, opened 2026-03-03
fix env server task cancellation

mikasenghaas:
This PR bumps `verifiers` to include PR [#974](https://github.com/PrimeIntellect-ai/verifiers/pull/974) which properly cancels rollut and group tasks on the server if the client task is cancelled. This avoids resource waste on the env server. Validated by running a simple run with 0 max off policy steps which will trigger rollout cancellation with every policy update. From the env  server logs we can see that we always have <= max concurrent rollouts, so tasks are correct cancelled and rescheduled on the client and server.

```bash
uv run rl @ examples/alphabet_sort/rl.toml --clean-output-dir --orchestrator.max-off-policy-steps 0  
```

Screenshot 2026-03-03 at 5 06 49 PM

Also found a small race between the `update_policy_loop` and `generate_batch`. Because `drop_group` yields, partially initializes group rollout requests can be scheduled to cancel with N in-flight rollout requests, but actually cancel >N rollouts because in the meantime we may have scheduled new rollouts due to the yield. In this way, we may log cancellation of >max_inflight_requests which is unintuitive.

---

> [!NOTE]
> **Medium Risk**
> Touches async cancellation/scheduling logic and updates a core external dependency (`verifiers`), which can affect rollout lifecycle behavior and resource usage under load.
> 
> **Overview**
> Improves orchestrator rollout cleanup by cancelling *all* stale rollout groups concurrently in `_update_off_policy` (via `asyncio.gather`), reducing a race where cancellations/logging could exceed the intended in-flight limit.
> 
> Bumps the `verifiers` git revision and refreshes `uv.lock` (including new transitive deps like `connect-python`/`pyqwest`, and updates to `prime-sandboxes`/`prime-tunnel`) to pull in env-server fixes that properly cancel server-side rollout/group tasks when client tasks are cancelled.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e69a435213cb94b5f7a92cc82bee0dd2e37beef8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,7 @@ override-dependencies = [
 [tool.uv.sources]
 torch = { index = "pytorch-cu128" }
 math-env = { index = "primeintellect" }
-verifiers = { git = "https://github.com/PrimeIntellect-ai/verifiers.git", rev = "b35d0c7" }
+verifiers = { git = "https://github.com/PrimeIntellect-ai/verifiers.git", rev = "bb73f78" }
 torchtitan = { git = "https://github.com/pytorch/torchtitan", rev = "a1fdd7e" }
 dion = { git = "https://github.com/samsja/dion.git", rev = "d891eeb" }
 transformers = { git = "https://github.com/huggingface/transformers.git", rev = "609e3d5" }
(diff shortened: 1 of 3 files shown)


faresobeid:
lgtm!

rasdani (approved):

PR #1959, opened 2026-03-05
fix: modality-aware batch distribution to prevent FSDP deadlocks

cdreetz:
## Summary

- Fixes FSDP all-gather deadlocks when training multimodal (VLM) models with mixed text-only and image batches across ranks
- Separates micro batches by modality (multimodal vs text-only), pads each group independently to be divisible by `num_train_workers`, and distributes in strided order so all ranks see the same modality at each step index
- Extracts `_make_dummy_batch` and `_pad_group_for_distribution` helpers for clarity

## Problem

FSDP requires all ranks to execute the same collective operations at each training step. When a batch with images lands on one rank (triggering the vision encoder's all-gather) while another rank gets a text-only batch, the all-gather hangs indefinitely.

The previous code distributed micro batches contiguously without regard to modality, so mixed-modality batches could end up misaligned across ranks.

## Test plan

- [x] Existing `test_batch.py` tests pass (text-only path unchanged)
- [x] Verify with multimodal training run that FSDP no longer deadlocks on mixed batches

---

> [!NOTE]
> **Medium Risk**
> Changes batch scheduling/distribution logic that affects multi-GPU training behavior; mistakes could cause subtle training skew or remaining hangs, though the change is localized to `prepare_batch`.
> 
> **Overview**
> Updates `prepare_batch` to **separate micro-batches by modality** (multimodal vs text-only), **pad each modality group independently** with zero-loss dummy batches, and then **distribute batches in a strided order** so every rank sees the same modality at a given step index (avoiding FSDP deadlocks on mixed-modality steps).
> 
> Extracts helper functions (`_make_dummy_batch`, `_pad_group_for_distribution`) to encapsulate dummy-batch creation and divisibility padding logic.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ae20df4aa3f1f7d7ab6efd1163c99807bae8c1cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/batch.py
+++ b/src/prime_rl/trainer/batch.py
@@ -163,6 +163,23 @@ def pad_micro_batch(micro_batch: MicroBatch, pad_to_multiple_of: int) -> MicroBa
     return micro_batch
 
 
+def _make_dummy_batch(source: MicroBatch) -> MicroBatch:
+    """Create a zero-loss dummy batch from an existing batch, preserving its modality."""
+    dummy = copy.deepcopy(source)
+    dummy.advantages = [0.0] * len(dummy.input_ids)
+    dummy.loss_mask = [False] * len(dummy.input_ids)
+    return dummy
+
+
+def _pad_group_for_distribution(group: list[MicroBatch], num_train_workers: int) -> list[MicroBatch]:
+    """Pad a group of micro batches so its length is divisible by num_train_workers."""
+    num_padding = -len(group) % num_train_workers
+    if num_padding > 0 and len(group) > 0:
+        dummy = _make_dummy_batch(group[0])
+        group.extend([dummy] * num_padding)
+    return group
+
+
 def prepare_batch(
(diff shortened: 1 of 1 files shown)


src/prime_rl/trainer/batch.py
```diff
@@ -163,6 +163,23 @@ def pad_micro_batch(micro_batch: MicroBatch, pad_to_multiple_of: int) -> MicroBa
     return micro_batch
 
 
+def _make_dummy_batch(source: MicroBatch) -> MicroBatch:
+    """Create a zero-loss dummy batch from an existing batch, preserving its modality."""
+    dummy = copy.deepcopy(source)
```
samsja, line 168: do we need the deepcopy here ? seems that it can make the trainer quite slow

samsja (approved):

PR #1961, opened 2026-03-05
make the loss equal weight per unmaked token

sapiosaturn:
warning: not tested yet, do not merge yet.

There is a likely algorithmic weighting bug in PrimeRL SFT because each rank computes a masked-token loss mean on its local shard (loss[loss_mask].mean()), backpropagates that scalar divided only by grad_accum_steps, and then relies on FSDP/HSDP gradient averaging over the hsdp mesh, which includes context-parallel ranks when model.cp > 1; that means the optimized objective is effectively an equal-weight average of per-rank, per-microstep local means, not a global mean over all unmasked tokens in the step. This is harmless only in the special case where every participating rank and every gradient-accumulation microstep has the same number of supervised tokens, which is roughly true for pretraining-style dense loss but not for SFT, because loss_mask is sparse and variable, and because the default SFT packing mode is pack_function = "cat" in prime_rl/configs/sft.py, with CatDataset in prime_rl/trainer/sft/data.py concatenating variable numbers of examples into fixed-length windows; as a result, windows, DP ranks, and GA microsteps can contain very different counts of trainable tokens, yet still contribute equal optimization weight. Context parallelism makes this more severe, because the same local-mean reduction now happens on half-sequences or smaller sequence shards, so token weighting depends on how supervised tokens happen to be distributed across CP partitions, making CP SFT optimize a different objective from both globally token-normalized SFT and non-CP SFT.

---

> [!NOTE]
> **Medium Risk**
> Changes the effective training objective and gradient scaling across distributed meshes, which can materially affect convergence/metrics. Touches core training-loop math (DP/CP/FSDP), so any mismatch in token counting or scaling could destabilize training.
> 
> **Overview**
> Fixes SFT loss/gradient weighting to be **per unmasked token globally** instead of averaging per-rank local means.
> 
> `compute_loss` now returns `(loss_sum, token_count)` over unmasked tokens, validation aggregates these across the `dp_cp` mesh to log a token-weighted mean loss, and the training loop accumulates per-microstep loss sums then **rescales gradients** after backprop using the global token count (undoing FSDP’s `fsdp_gradient_divide_factor`) so the optimized objective matches a true global token mean under DP+CP+GA.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a8fdc138c5b1e72f1f3832d851ef81637e12fb18. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/sft/train.py
+++ b/src/prime_rl/trainer/sft/train.py
@@ -304,38 +304,13 @@ def train(config: SFTConfig):
         step_start_time = time.perf_counter()
         forward_backward_start_time = time.perf_counter()
 
-        # Pre-fetch micro-batches and compute the global supervised-token count so that each
-        # backward pass can use the correct normalization factor directly, avoiding a post-hoc
-        # param.grad traversal.
-        micro_batches = []
-        step_local_token_count = torch.tensor(0, dtype=torch.int64, device="cuda")
-        for _ in range(grad_accum_steps):
-            micro_batch = next(dataiter)
-            micro_batches.append(micro_batch)
-            loss_mask = micro_batch["loss_mask"].to("cuda")
-            if cp_enabled:
-                loss_mask = shard_for_cp(loss_mask, cp_rank=cp_rank, cp_world_size=cp_size)
-            step_local_token_count += loss_mask.sum(dtype=torch.int64)
-
-        global_step_token_count = step_local_token_count.clone()
-        dist.all_reduce(global_step_token_count, op=dist.ReduceOp.SUM, group=dp_cp_group)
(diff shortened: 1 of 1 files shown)


sapiosaturn:
manual gradient div is very ugly, but should work. need to test

hallerite:
Thanks @sapiosaturn ! I made some changes & resolved merge conflicts. 
image
Purple is prime-rl main & blue is this PR, so lgtm!

sapiosaturn:
awesome, would test at high batch size (maybe with grad acc) + large seq len just to make sure it's not too numerically unstable with all the summation haha

PR #1981, opened 2026-03-07
fp8 broadcast

S1ro1:
> [!NOTE]
> **High Risk**
> Touches the training↔inference weight-update pipeline (NCCL broadcast, in-place parameter updates, and optional FP8 quantization), where shape/format mismatches can break live inference or silently degrade quality. Also changes multi-node SLURM orchestration and dependency pinning, which can affect cluster runs and reproducibility.
> 
> **Overview**
> Adds an opt-in **vLLM kernel-format weight broadcast** path for NCCL, including optional block-wise FP8 quantization, to enable direct in-place updates on inference workers without converting through HF checkpoint format.
> 
> Plumbs new config flags (e.g. `use_vllm_format_transfer`, `quantize_fp8`) through shared/trainer/orchestrator configs and the `/init_broadcaster` RPC; updates the vLLM NCCL worker to `copy_()` received params (with EP expert slicing and MLA absorbed-weight recomputation) and introduces model-side layer conversion for `glm_moe_dsa`.
> 
> Extends multi-node RL deployment to support multiple inference replicas (`num_infer_replicas`/`total_infer_nodes`) and updates the SLURM template for per-replica head selection and headless nodes; also relaxes inference `api_server_count` to allow `0` for headless mode and adjusts model-name propagation/validation so orchestrator matches inference.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d685dfc85328fe059c21919086e1cafe03cb9078. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/configs/inference.py
```diff
@@ -226,6 +226,28 @@ class InferenceConfig(BaseConfig):
...
+            description="Starting DP rank for this node in multi-node EP. Passed to vLLM as `--data-parallel-start-rank`.",
+        ),
+    ] = None
+
+    headless: Annotated[
+        bool,
+        Field(
+            description="Run in headless mode (no API server). Passed to vLLM as `--headless`.",
+        ),
+    ] = False
```
samsja: can pass all of this via vllm_extras instead imo

src/prime_rl/configs/rl.py
```diff
@@ -117,6 +117,22 @@ class SharedWeightBroadcastConfig(BaseConfig):
     port: Annotated[int, Field(description="The port to use for NCCL weight broadcast.")] = 29501
     timeout: Annotated[int, Field(description="The timeout in seconds for NCCL weight broadcast.")] = 1200
 
+    use_kernel_format_transfer: Annotated[
+        bool,
+        Field(
+            description="Transfer weights in vLLM kernel format instead of HF checkpoint format. "
+            "Avoids the HF conversion intermediate step and allows direct in-place weight updates."
+        ),
+    ] = False
```
samsja:
```suggestion
    use_vllm_format_transfer: Annotated[
        bool,
        Field(
            description="Transfer weights in vLLM kernel format instead of HF checkpoint format. "
            "Avoids the HF conversion intermediate step and allows direct in-place weight updates."
        ),
    ] = False
```

lets rename to this

src/prime_rl/configs/orchestrator.py
```diff
@@ -627,6 +627,13 @@ class NCCLWeightBroadcastConfig(BaseModel):
     port: Annotated[int, Field(description="The port to use for the NCCL broadcast.")] = 29501
     timeout: Annotated[int, Field(description="The timeout in seconds to use for the NCCL broadcast.")] = 1200
 
+    use_kernel_format_transfer: Annotated[
```
samsja: can we move this to vllm broadcast config instead of having it on the orch ?

src/prime_rl/configs/rl.py
```diff
@@ -280,6 +296,16 @@ class RLConfig(BaseConfig):
...
 
+    allow_different_inference_model: Annotated[
+        bool,
+        Field(
+            description="Allow the inference server to use a different model name than the trainer. "
+            "When enabled, the orchestrator uses the inference model name for querying. "
+            "Useful for kernel format weight transfer where the trainer uses a bf16 model "
+            "and inference uses a quantized (e.g. FP8) variant.",
+        ),
+    ] = False
```
samsja: does that mean we always need to load the fp8 model on the vllm side ? 

I think we should just allow this by default and not have this param imo, should still translate from model.name to trainer.model.name and infer.model.name but user should be able to override it via trainer.model.name basically

src/prime_rl/templates/multi_node_rl.sbatch.j2
```diff
@@ -106,15 +106,15 @@ if [ "$SLURM_PROCID" -lt "$NUM_INFER_NODES" ]; then
...
                 2>&1 | tee $OUTPUT_DIR/slurm/latest_infer_node_rank_${INFER_NODE_RANK}.log $OUTPUT_DIR/slurm/job_${SLURM_JOB_ID}_infer_node_rank_${INFER_NODE_RANK}.log
         else
             uv run inference \
                 @ $CONFIG_DIR/inference.toml \
-                --data-parallel-start-rank $INFER_DP_START_RANK \
-                --data-parallel-address $INFER_HEAD_HOST \
-                --data-parallel-rpc-port $INFERENCE_DATA_PARALLEL_RPC_PORT \
+                --data_parallel_start_rank $INFER_DP_START_RANK \
+                --data_parallel_address $INFER_HEAD_HOST \
+                --data_parallel_rpc_port $INFERENCE_DATA_PARALLEL_RPC_PORT \
```
samsja: should  this be always enabled ?

PR #2008, opened 2026-03-10
Add built-in PPO clip and REINFORCE loss, normalized and reinforce ad…

nevasini1:
Adds built-in loss and advantage options configurable via TOML.
**Loss (`trainer.loss`):**
- `ppo_clip`: token-level PPO clipped surrogate; `clip_eps`, `kl_coef`, `entropy_coef`
- `reinforce`: policy gradient -log π * A; `entropy_coef`
**Advantage (`orchestrator.advantage`):**
- `normalized`: zero-mean, unit-variance per problem; `eps`
- `reinforce`: raw rewards (no baseline)
Tests and docs (bring-your-own-algorithms.md) updated.

### DIFF
--- a/docs/bring-your-own-algorithms.md
+++ b/docs/bring-your-own-algorithms.md
@@ -36,7 +36,36 @@ class LossOutputs:
     metrics: dict[str, Tensor]      # Metrics to log
 ```
 
-### Example: PPO Clipped Loss
+### Built-in loss types
+
+You can use these without writing custom code by setting `type` in `[trainer.loss]`:
+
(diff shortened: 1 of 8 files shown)


hallerite:
appreciate the contribution, but we currently do not plan on officially supporting any other losses besides the one we maintain. if a user wants to use a custom loss, they can simply use the bring-your-own-loss feature. we do intend on improving the ergonomics on bringing a custom loss, as well as custom credit assignment into prime-rl, but this is currently not a priority

PR #2013, opened 2026-03-10
fix: reconcile VLM image tokens with pixel_values after prompt truncation

cdreetz:
## Summary

In multi-turn VLM conversations that exceed `seq_len`, vLLM left-truncates the prompt to fit in `max_model_len`, removing image_pad tokens from early turns. But `pixel_values`/`image_grid_thw` are computed independently by the orchestrator's image processor from all conversation images. This causes a mismatch — the trainer crashes with `ValueError: Image features and image tokens do not match`.

The fix runs after seq_len truncation in `prepare_sample`: compare image_pad tokens in `input_ids` against `image_grid_thw`, drop images whose placeholder tokens were truncated, and remove any orphaned partial image tokens so both sides agree on the same set of complete images. No-op for text-only samples and VLM samples where tokens already match.

## Test plan

- [x] Unit tests for `_trim_multimodal_to_match`: noop cases (text-only, already matching), partial image drop, all images dropped
- [x] Integration tests for `prepare_sample`: seq_len truncation path, vLLM pre-truncation path
- [x] End-to-end: 5-step VLM tic-tac-toe training run with Qwen3-VL-4B-Instruct completes without errors

---

> [!NOTE]
> **Medium Risk**
> Changes core sample-prep logic for multimodal training and can alter which images/features are fed to the model when truncation occurs, though the behavior is narrowly scoped and covered by new tests.
> 
> **Overview**
> Fixes Qwen3-VL batch preparation when prompts are truncated by ensuring image placeholder tokens in `input_ids` stay consistent with `pixel_values`/`image_grid_thw`.
> 
> `prepare_sample` now calls a new `_trim_multimodal_to_match` helper that drops orphaned/partial image tokens and trims (or removes) trailing image features/grids accordingly, while leaving text-only or already-matching samples unchanged. Adds unit tests covering noop, partial/all-image drops, and both `seq_len` truncation and vLLM left-truncation scenarios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f4db3bbd5e470af896fe89ff7ca07073dbbe6a0d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/batch.py
+++ b/src/prime_rl/trainer/batch.py
@@ -6,47 +6,8 @@ from prime_rl.transport.types import MicroBatch, TrainingSample
 _IMAGE_PAD_TOKEN_ID = 151655
 
 
-def _compute_keep_mask_for_multimodal(
-    input_ids: list[int],
-    image_grid_thw: list[list[int]],
-    merge_size: int = 2,
-) -> tuple[list[bool], int, list[list[int]]]:
-    """Compute a per-position keep mask that drops orphaned image_pad tokens.
-
-    After seq_len truncation, the last image may have only partial placeholder tokens.
-    This computes which image_pad tokens belong to complete images (keep=True) and which
-    are orphaned partials (keep=False). Non-image tokens are always kept.
-
-    Returns (keep_mask, kept_image_count, kept_grids).
-    """
-    num_image_tokens = sum(1 for t in input_ids if t == _IMAGE_PAD_TOKEN_ID)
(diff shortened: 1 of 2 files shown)


src/prime_rl/trainer/batch.py
```diff
@@ -2,6 +2,66 @@
 
 from prime_rl.transport.types import MicroBatch, TrainingSample
 
+# Qwen3-VL image placeholder token ID
+_IMAGE_PAD_TOKEN_ID = 151655
```
hallerite, line 6: need to add this to the model registry, should def not be a magic number

PR #2023, opened 2026-03-12
Add sample_ratio to LogExtrasConfig for gradual rollout control

shayonj:
## Summary

- Adds `sample_ratio` field (0.0–1.0, optional float) to `LogExtrasConfig` to control what fraction of rollouts are logged per step
- When set, the effective sample cap is computed from the actual rollout count at runtime: `int(len(rollouts) * sample_ratio)`
- `1.0` = all rollouts, `0.5` = half, `0.0` = disable, `None` (default) = current behavior (log all)

### How it interacts with existing defaults

The existing `interval` gate (`step % interval != 0`, default `interval=10`) is untouched and always runs first. `sample_ratio` only applies **after** the interval gate passes. When `sample_ratio` is `None` (the default), the new code is a complete no-op.

**Current behavior (unchanged):** every 10th step, all rollouts are logged.

| `interval` | `sample_ratio` | batch_size=128 | What happens |
|---|---|---|---|
| 10 | `None` (default) | step 10: 128 rollouts | Current behavior, no change |
| 10 | 1.0 | step 10: 128 rollouts | Equivalent to current — all rollouts |
| 10 | 0.5 | step 10: 64 rollouts | Half the rollouts, randomly sampled |
| 1 | 1.0 | every step: 128 rollouts | Full data every step |
| 1 | 0.1 | every step: 12 rollouts | 10% of rollouts every step |
| 10 | 0.0 | step 10: skip | Disables sample logging entirely |

### Usage

In the orchestrator TOML config:

```toml
[orchestrator.prime_monitor.log_extras]
sample_ratio = 0.5
interval = 1
```

Or via CLI args:

```bash
--prime-monitor.log-extras.sample-ratio 0.5 --prime-monitor.log-extras.interval 1
```

For hosted platform runs, this can be passed through `run_config`:

```toml
[run_config]
prime-monitor.log-extras.sample-ratio = 1.0
prime-monitor.log-extras.interval = 1
```

### DIFF
--- a/src/prime_rl/configs/shared.py
+++ b/src/prime_rl/configs/shared.py
@@ -269,6 +269,18 @@ class LogExtrasConfig(BaseConfig):
         ),
     ] = 10
 
+    sample_ratio: Annotated[
+        float | None,
+        Field(
+            ge=0.0,
+            le=1.0,
+            description="Fraction of rollouts to log per step (0.0–1.0). "
+            "When set, the effective sample cap is len(rollouts) * sample_ratio. "
+            "1.0 = all rollouts, 0.5 = half, 0.0 = none. "
+            "None (default) = log all rollouts (current behavior).",
+        ),
+    ] = None
+
 
 class WandbConfig(BaseConfig):
     """Configures logging to Weights and Biases."""
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Documenting changes which affect configuration usage patterns (added/moved/removed/renamed fields, notable logic changes).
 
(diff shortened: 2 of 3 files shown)


src/prime_rl/configs/shared.py
```diff
@@ -269,6 +269,18 @@ class LogExtrasConfig(BaseConfig):
         ),
     ] = 10
 
+    sample_percentage: Annotated[
```
mikasenghaas: maybe we use ratio, not percentage here. i dont think we have any other arg in the configs that takes percentage

shayonj: yeah good call, looks like there is a pattern already for ratios. updated

PR #2030, opened 2026-03-16
feat: disaggregated prefill-decode inference with multi-replica support

samsja:
## Summary
- Extracted PD disaggregation changes from #2010 into a standalone PR
- Adds disaggregated prefill/decode deployment type where prefill and decode nodes run separately, connected via NIXL for KV cache transfer and a vllm-router for request routing
- Adds vLLM kernel format weight transfer path that converts weights to vLLM's internal format (fused projections, stacked experts, optional FP8 quantization) before NCCL broadcast, avoiding HF checkpoint round-trip
- Adds pause/resume lifecycle to drain in-flight requests before weight updates in disaggregated mode
- Adds multi-replica inference support for both multi-node and disaggregated templates
- Includes install scripts for DeepEP kernels, NIXL, and vllm-router

## Test plan
- [ ] Verify disaggregated deployment config validation with sample TOML
- [ ] Test multi-replica inference with `num_infer_replicas > 1`
- [ ] Validate weight broadcast in both checkpoint and kernel format paths
- [ ] End-to-end test with disaggregated sbatch template on a multi-node cluster

---

> [!NOTE]
> **High Risk**
> High risk because it changes multi-node inference/SLURM orchestration and NCCL weight-broadcast initialization/rank math, which can break distributed launches and live weight updates across many GPUs.
> 
> **Overview**
> Adds a new **disaggregated prefill/decode inference** deployment mode, including config schema (`deployment.type = "disaggregated"`) with prefill/decode node counts and ports, plus auto-configuration for EP/DP sizing and optional `VLLM_USE_DEEP_GEMM` FP8 kernels.
> 
> Extends multi-node RL/inference launch paths to support **multiple inference replicas** and router-based request routing (new `vllm-router` dependency), updating SLURM templates to generate per-replica router URLs, distinct admin URLs that bypass routers, and the role-specific env needed for NIXL/UCX/NVSHMEM-based KV transfer.
> 
> Hardens online weight updates by adding `/pause` and `/resume` endpoints and making `update_weights()` pause/resume all engines around the update, and refactors NCCL broadcast init to use `rank_offset` + `inference_world_size` (+ per-server GPU counts) with updated worker rank computation.
> 
> Includes install scripts for DeepEP kernels, UCX+NIXL-from-source, DeepGEMM, and a patched `vllm-router`, plus new docs and an example `GLM-5-FP8` disaggregated config.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b7f7bd026da8e68205a5f73dd4542822022a78c9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/configs/inference.py
```diff
@@ -298,6 +320,21 @@ def validate_multi_node_requires_slurm(self):
             raise ValueError("Must use SLURM for multi-node deployment.")
         return self
 
+    @model_validator(mode="after")
+    def auto_setup_disaggregated(self):
+        """Auto-configure inference for disaggregated P/D: force TP=1, enable EP."""
+        if self.deployment.type == "disaggregated":
+            self.parallel.tp = 1
```
samsja: lets rather failed here if tp is not 1

src/prime_rl/configs/orchestrator.py
```diff
@@ -643,6 +643,14 @@ class NCCLWeightBroadcastConfig(BaseModel):
     port: Annotated[int, Field(description="The port to use for the NCCL broadcast.")] = 29501
     timeout: Annotated[int, Field(description="The timeout in seconds to use for the NCCL broadcast.")] = 1200
 
+    inference_world_size: Annotated[
+        int,
+        Field(
+            ge=1,
+            description="Total number of inference GPUs across all servers. Used by init_nccl_broadcast to compute per-server rank offsets.",
+        ),
+    ] = 1
```
samsja, line 659: why do we need this ?

src/prime_rl/configs/rl.py
```diff
@@ -163,9 +163,16 @@ class MultiNodeDeploymentConfig(BaseDeploymentConfig):
...
+            description="Number of inference nodes per replica. Set to 0 to skip inference and orchestrator (requires fake data).",
         ),
     ]
+    num_infer_replicas: Annotated[
+        int,
+        Field(
+            ge=1,
+            description="Number of independent inference replicas. Total inference nodes = num_infer_nodes * num_infer_replicas.",
+        ),
+    ] = 1
```
samsja, line 204: why do we need this ?

src/prime_rl/configs/rl.py
```diff
@@ -442,12 +453,26 @@ def auto_setup_wandb(self):
...
+        explicitly set inference.model.name (allowing e.g. an FP8 variant for inference).
+        The orchestrator always uses the inference model name so it queries the
+        correct model on the inference server.
+        """
         if self.model is not None:
             self.trainer.model.name = self.model.name
-            self.orchestrator.model.name = self.model.name
             if self.inference is not None:
-                self.inference.model.name = self.model.name
+                inference_model_explicitly_set = "name" in self.inference.model.model_fields_set
```
samsja, line 533: why do we need this ? I guess its to support different model on the trainer and inference because of fp8 right ? if yes  don't think it belong to this pr

src/prime_rl/configs/rl.py
```diff
@@ -442,12 +453,26 @@ def auto_setup_wandb(self):
...
+                if not inference_model_explicitly_set:
+                    self.inference.model.name = self.model.name
+
+            # Orchestrator must use the inference model name so it queries the correct model
+            if self.inference is not None:
+                self.orchestrator.model.name = self.inference.model.name
+            else:
+                self.orchestrator.model.name = self.model.name
```
samsja: not need I think for pd

src/prime_rl/trainer/rl/broadcast/nccl.py
```diff
@@ -112,33 +112,48 @@ def __init__(
...
+
+    def _broadcast_checkpoint_format(
+        self, model: nn.Module, state_dict: dict[str, Tensor], num_layers: int, layer_prefix: str
+    ) -> None:
+        """Broadcast weights in HF checkpoint format (original path)."""
+        for layer_id, layer_sd in filter_state_dict_by_layers(state_dict, num_layers, layer_prefix):
+            layer_sd = self._resolve_dtensors(layer_sd)
+
+            if isinstance(model, PreTrainedModelPrimeRL) and model.is_prime_state_dict(layer_sd):
+                model.convert_layer_to_hf(layer_sd, layer_id)
```
samsja: I don't think we need this this is for fp8 stuff

src/prime_rl/utils/client.py
```diff
@@ -218,6 +221,31 @@ async def _check_health(admin_client: AsyncClient) -> None:
...
+async def _resume_engines(admin_clients: list[AsyncClient]) -> None:
+    """Resume all inference engines after weight update."""
+    logger = get_logger()
+
+    async def _resume(client: AsyncClient) -> None:
+        response = await client.post("/resume")
+        response.raise_for_status()
+
+    await asyncio.gather(*[_resume(client) for client in admin_clients])
+    logger.info("All inference engines resumed")
```
samsja, line 247: I am not sure to understand how this works

src/prime_rl/utils/client.py
```diff
@@ -337,7 +388,7 @@ async def _init_nccl_broadcast(
 
     await asyncio.gather(
         *[
-            _init_nccl_broadcast(admin_client, host, port, client_num, timeout)
+            _init_nccl_broadcast(admin_client, client_num * gpus_per_server)
             for client_num, admin_client in enumerate(admin_clients)
         ]
     )
```
samsja, line 397: I don't understand the inference_world_size logic

src/prime_rl/configs/inference.py
samsja: Done — now raises a `ValueError` if `tp != 1` instead of silently overriding.

src/prime_rl/configs/orchestrator.py
samsja: This is needed for multi-GPU-per-server NCCL rank computation. The old code assumed 1 GPU per admin client (`server_rank = client_num`), but with multi-node inference each vLLM server manages multiple GPUs. `inference_world_size` lets `init_nccl_broadcast` compute the correct `rank_offset` and `gpus_per_server` so every inference GPU gets a unique rank in the NCCL broadcast group. Without it, multi-node NCCL broadcast assigns overlapping ranks.

src/prime_rl/configs/rl.py
samsja: `num_infer_replicas` is needed for PD disaggregation — each replica is an independent prefill+decode group. For example, with `num_infer_nodes=3` (1 prefill + 2 decode) and `num_infer_replicas=2`, you get 6 total inference nodes split into 2 independent PD groups. Without this field there's no way to express "run N copies of the PD topology".

src/prime_rl/configs/rl.py
samsja: You're right, this was for FP8 inference model variants. Reverted `auto_setup_model` back to main's version in 914729c.

src/prime_rl/configs/rl.py
samsja: Removed along with the rest of the `auto_setup_model` changes in 914729c.

src/prime_rl/trainer/rl/broadcast/nccl.py
samsja: Agreed — the method extraction was only needed for the kernel format branching. Reverted `broadcast_weights` back to main's inline style in 914729c.

src/prime_rl/utils/client.py
samsja: This is the pause/resume lifecycle for disaggregated weight updates:

1. **`_pause_engines`** — calls `POST /pause` on each vLLM server, which drains all in-flight requests so every DP worker is idle
2. **Weight update** — with all workers idle, they can safely join the NCCL broadcast collective to receive new weights
3. **`_resume_engines`** — calls `POST /resume` to start accepting new requests again

This is needed because in disaggregated mode, DP workers can't participate in a collective NCCL broadcast while they're busy serving requests. Also added try/finally to ensure engines are always resumed even if the weight update fails.

src/prime_rl/utils/client.py
samsja: The old code assumed 1 GPU per admin client (`server_rank = client_num, num_inference_server = len(admin_clients)`). With multi-GPU servers this breaks because each server has multiple GPUs that need unique NCCL ranks.

The new logic:
- `inference_world_size` = total inference GPUs across all servers (e.g., 48 for 6 nodes × 8 GPUs)
- `gpus_per_server = inference_world_size // len(admin_clients)` = GPUs each server manages (e.g., 8)
- `rank_offset = client_num * gpus_per_server` = starting rank for each server (0, 8, 16, ...)

Each server then assigns ranks `rank_offset + local_rank` to its GPUs, so all 48 GPUs get unique ranks in the NCCL broadcast group (with the trainer on rank 0).

src/prime_rl/configs/rl.py
samsja: why not call it dp tho ?

src/prime_rl/utils/validation.py
```diff
@@ -34,16 +34,11 @@ def validate_shared_model_name(
...
-            f"Trainer model name ({trainer.model.name}) and orchestrator model name ({orchestrator.model.name}) are not the same. Please specify the same model name for both."
-        )
-
+    # Orchestrator must match inference (it queries the inference server)
     if inference and inference.model.name != orchestrator.model.name:
         raise ValueError(
-            f"Inference model name ({inference.model.name}) and orchestrator model name ({orchestrator.model.name}. Please specify the same model name for both."
+            f"Inference model name ({inference.model.name}) and orchestrator model name ({orchestrator.model.name}) are not the same. "
+            "The orchestrator queries the inference server and must use the same model name."
         )
```
samsja: lets not revert this for now

src/prime_rl/templates/disaggregated_rl.sbatch.j2
```diff
@@ -0,0 +1,259 @@
...
+
+if [ "$NODE_RANK" -lt "$TOTAL_INFER_NODES" ]; then
+    # ── Inference node (prefill or decode) ──
+    export VLLM_WORKER_MULTIPROC_METHOD=spawn
+    export PYTORCH_CUDA_ALLOC_CONF="expandable_segments:False"
+    export GLOO_SOCKET_IFNAME=bond0
+    export VLLM_ENGINE_READY_TIMEOUT_S=4200
+    export UCX_TLS=all
+    export UCX_NET_DEVICES=all
+    export LD_LIBRARY_PATH="$PROJECT_DIR/third_party/ucx/lib:$PROJECT_DIR/third_party/ucx/lib/ucx:${LD_LIBRARY_PATH:-}"
```
S1ro1: Delete this

src/prime_rl/templates/inference.sbatch.j2
```diff
@@ -60,18 +96,86 @@ echo "INFER_URLS=${INFER_URLS}"
...
+    # Reconstruct hostnames array
+    read -ra HOSTNAMES <<< "$HOSTNAMES_STR"
+
+    export VLLM_WORKER_MULTIPROC_METHOD=spawn
+    export PYTORCH_CUDA_ALLOC_CONF="expandable_segments:False"
+    export GLOO_SOCKET_IFNAME=bond0
+    export VLLM_ENGINE_READY_TIMEOUT_S=4200
+    export UCX_TLS=all
+    export UCX_NET_DEVICES=all
+    export LD_LIBRARY_PATH="$PROJECT_DIR/third_party/ucx/lib:$PROJECT_DIR/third_party/ucx/lib/ucx:${LD_LIBRARY_PATH:-}"
```
S1ro1, line 137: delete

src/prime_rl/templates/inference.sbatch.j2
```diff
@@ -60,18 +96,86 @@ echo "INFER_URLS=${INFER_URLS}"
...
-    export NCCL_IB_HCA=$IB_HCA
+    IB_HCA=$(ibv_devinfo 2>/dev/null | sed -n -e "/hca_id/p" -e "/link_layer:/p" | grep -B1 InfiniBand | grep hca_id | sed -e "s/^hca_id://g" | tr -d "[[:blank:]]" | paste -sd,)
+    [ -n "$IB_HCA" ] && export NCCL_IB_HCA=$IB_HCA
 
     INFER_NODE_RANK=$SLURM_PROCID
     LOCAL_IP=$(hostname -I | awk "{print \$1}")
+
+{% if disaggregated -%}
+    # NVSHMEM libs for DeepEP — ensure the rpath target exists on all nodes
+    if [ ! -d /tmp/deepep_build/nvshmem/lib ] && [ -d "$OUTPUT_DIR/nvshmem/lib" ]; then
```
S1ro1, line 120: maybe move out

src/prime_rl/templates/inference.sbatch.j2
```diff
@@ -60,18 +96,86 @@ echo "INFER_URLS=${INFER_URLS}"
...
+        echo "Starting vllm-router on $LOCAL_IP:$ROUTER_PORT for replica $REPLICA_IDX" | tee $ROUTER_LOG
+
+        REPLICA_ROUTER_ARGS=$(echo "$ALL_ROUTER_ARGS" | cut -d"|" -f$((REPLICA_IDX + 1)))
+
+        vllm-router \
+            --policy round_robin \
+            --vllm-pd-disaggregation \
+            $REPLICA_ROUTER_ARGS \
+            --host 0.0.0.0 \
+            --port $ROUTER_PORT \
```
S1ro1, line 200: --local-data-parallel size or something like that, check vllm-router logs

S1ro1 (commented):
https://docs.vllm.ai/en/latest/serving/data_parallel_deployment/#hybrid-load-balancing and https://github.com/vllm-project/router/blob/main/scripts/llama3.1/start_router.sh

configs/hendrycks_math/rl_glm_air.toml
```diff
@@ -0,0 +1,57 @@
+seq_len = 2048
+
```
samsja: I will delete this before merge

configs/glm4_air_disagg_hendrycks_math/rl.toml
```diff
@@ -0,0 +1,76 @@
+seq_len = 2048
+output_dir = "/data/outputs/glm4-air-disagg-hendrycks-math"
```
samsja: I will delete this before mrging

src/prime_rl/utils/validation.py
```diff
@@ -34,16 +34,11 @@ def validate_shared_model_name(
     orchestrator: OrchestratorConfig,
     inference: Optional[InferenceConfig] = None,
 ) -> None:
-    if trainer.model.name.startswith("Jackmin108/"):  # The TT MoE models will have a different name on the orchestrator
-        return
-    if trainer.model.name != orchestrator.model.name:
-        raise ValueError(
-            f"Trainer model name ({trainer.model.name}) and orchestrator model name ({orchestrator.model.name}) are not the same. Please specify the same model name for both."
-        )
```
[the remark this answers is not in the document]
S1ro1: This is intended to allow for FP8 inference <-> BF16 trainer

src/prime_rl/utils/client.py
```diff
@@ -242,23 +270,23 @@ async def update_weights(
...
+            if weight_dir is not None:
+                nccl_ready_file = weight_dir / NCCL_READY_MARKER
+                nccl_ready_file.parent.mkdir(parents=True, exist_ok=True)
+                nccl_ready_file.touch()
+                logger.debug(f"Created NCCL_READY marker at {nccl_ready_file}")
 
-        await asyncio.gather(*[_update_weights(admin_client, weight_dir_posix) for admin_client in admin_clients])
+            await asyncio.gather(*[_update_weights(admin_client, weight_dir_posix) for admin_client in admin_clients])
+        finally:
+            await _resume_engines(admin_clients)
```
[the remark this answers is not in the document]
S1ro1: This is also intended, the old way could introduce a hang in update-weights but its probability was way too low to ever notice it.

src/prime_rl/templates/disaggregated_rl.sbatch.j2
```diff
@@ -0,0 +1,260 @@
...
+
+else
+    # ── Training node ──
+    TRAIN_NODE_RANK=$((NODE_RANK - TOTAL_INFER_NODES))
+
+    if [ "$TRAIN_NODE_RANK" -eq 0 ]; then
+        export VF_LOG_LEVEL=DEBUG
+        uv run orchestrator \
+            @ $CONFIG_DIR/orchestrator.toml \
+{%- if weight_broadcast_type == "nccl" %}
```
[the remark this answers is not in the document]
S1ro1: maybe valid?

scripts/install_deep_gemm.sh
S1ro1, line 1: Let's remove given we now do from wheels? Or rename to `build_x`?

src/prime_rl/configs/inference.py
```diff
@@ -204,8 +229,8 @@ class InferenceConfig(BaseConfig):
     api_server_count: Annotated[
         int,
         Field(
-            ge=1,
-            description="The number of API servers to use. Passed to vLLM as `--api-server-count`",
+            ge=0,
```
S1ro1, line 232: We don't do headless anywhere anymore, was just an artifact from early internal load balancing, should revert

src/prime_rl/configs/inference.py
```diff
@@ -332,6 +383,10 @@ def auto_setup_api_server_count(self):
         size. Unless LoRA is enabled, in which case only one API server is
         supported (vLLM limitation).
         """
+        if self.vllm_extra.get("headless", False):
```
S1ro1, line 386: Likewise here, this should not be needed.

src/prime_rl/configs/inference.py
```diff
@@ -254,6 +279,13 @@ class InferenceConfig(BaseConfig):
         ),
     ] = False
 
+    use_deep_gemm: Annotated[
+        bool,
+        Field(
+            description="Force DeepGEMM FP8 kernels via VLLM_USE_DEEP_GEMM=1. Only works with per-tensor FP8 quantization (e.g. GLM-5-FP8).",
```
S1ro1, line 285: block-scale FP8 quantization, not per-tensor

src/prime_rl/configs/inference.py
```diff
@@ -298,6 +330,25 @@ def validate_multi_node_requires_slurm(self):
             raise ValueError("Must use SLURM for multi-node deployment.")
         return self
 
+    @model_validator(mode="after")
+    def auto_setup_disaggregated(self):
+        """Auto-configure inference for disaggregated P/D: enable EP and compute DP."""
+        if self.deployment.type == "disaggregated":
+            if "enable_expert_parallel" not in self.model_fields_set:
+                self.enable_expert_parallel = True
+            if "enable_eplb" not in self.model_fields_set:
```
S1ro1, line 339: We should crash on eplb=True or just override to false - the weight transfer then doesn't work.

src/prime_rl/configs/inference.py
```diff
@@ -298,6 +330,25 @@ def validate_multi_node_requires_slurm(self):
...
+                self.enable_eplb = False
+            gpus_per_node = self.deployment.gpus_per_node
+            tp = self.parallel.tp
+            dp_per_node = gpus_per_node // tp
```
S1ro1, line 343: Don't need to care about TP here since we explicitly crash on tp now, no?

src/prime_rl/configs/rl.py
```diff
@@ -685,11 +701,38 @@ def auto_setup_deployment(self):
...
+        if self.deployment.num_infer_nodes != expected_infer_nodes:
+            raise ValueError(
+                f"deployment.num_infer_nodes ({self.deployment.num_infer_nodes}) must equal "
+                f"inference.deployment.num_prefill_nodes ({infer_deploy.num_prefill_nodes}) + "
+                f"inference.deployment.num_decode_nodes ({infer_deploy.num_decode_nodes}) = {expected_infer_nodes}"
+            )
+
+        total_infer_gpus = self.deployment.total_infer_nodes * self.deployment.gpus_per_node
+        if self.weight_broadcast is not None and self.weight_broadcast.type == "nccl":
+            assert self.trainer.weight_broadcast.type == "nccl"
```
S1ro1, line 807: Maybe add messages to these asserts?

src/prime_rl/entrypoints/inference.py
```diff
@@ -32,15 +32,38 @@ def write_slurm_script(config: InferenceConfig, config_path: Path, script_path:
...
+            num_prefill_nodes=config.deployment.num_prefill_nodes,
+            num_decode_nodes=config.deployment.num_decode_nodes,
+            prefill_port=config.deployment.prefill_port,
+            decode_port=config.deployment.decode_port,
+            router_port=config.deployment.router_port,
+            data_parallel_rpc_port=config.data_parallel_rpc_port,
+            use_deep_gemm=config.use_deep_gemm,
+        )
+    elif is_multi_node:
+        template_vars.update(
```
S1ro1, line 60: Shouldn't `use_deep_gemm` also be passed in?

src/prime_rl/utils/validation.py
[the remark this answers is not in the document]
samsja: yeah but shouldn't be done in this pr then

src/prime_rl/utils/client.py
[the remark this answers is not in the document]
samsja: sg

src/prime_rl/templates/inference.sbatch.j2
```diff
@@ -60,18 +102,146 @@ echo "INFER_URLS=${INFER_URLS}"
...
+    read -ra HOSTNAMES <<< "$HOSTNAMES_STR"
+
+    export VLLM_WORKER_MULTIPROC_METHOD=spawn
+{%- if use_deep_gemm %}
+    export VLLM_USE_DEEP_GEMM=1
+{%- endif %}
+    export PYTORCH_CUDA_ALLOC_CONF="expandable_segments:False"
+    export GLOO_SOCKET_IFNAME=bond0
+    export VLLM_ENGINE_READY_TIMEOUT_S=4200
+    export UCX_TLS=all
```
S1ro1, line 135: This probably needs adapting

pyproject.toml
```diff
@@ -70,11 +70,15 @@ envs = [
     "aime2024",
     "aime2025",
 ]
+disagg = [
```
S1ro1, line 74: Maybe give proper group name?

S1ro1 (approved):
Last minor nits, but looks good overall now

src/prime_rl/configs/inference.py
samsja: we don't crash on tp and we allow tp with PD

S1ro1 (approved):
lgtm, enjoy the conflict

PR #2043, opened 2026-03-18
feat: collect and log per-engine inference server metrics to W&B

mikasenghaas:
Poll vLLM Prometheus /metrics endpoints from the orchestrator to track `num_requests_running`, `num_requests_waiting`, and `kv_cache_usage_perc` per server and per engine. Values are averaged over a sliding window of the last 20 fetches. Also switch EventLoopLagMonitor to use deque for improved performance.

**Single Node**

Screenshot 2026-03-18 at 4 42 30 PM

**Multi Engine**

Screenshot 2026-03-18 at 6 27 50 PM

**Multi Node & Multi Engine**

---

> [!NOTE]
> **Medium Risk**
> Adds background polling of inference servers’ `/metrics` endpoints and injects the results into orchestrator W&B logging, which could affect training-loop performance or introduce failures if metrics endpoints are slow/flaky (though failures are logged and skipped). Also changes HTTPX admin client connection limits and event-loop lag tracking implementation.
> 
> **Overview**
> Adds optional (`orchestrator.collect_inference_metrics`, default `True`) collection of vLLM Prometheus `/metrics` from all inference servers and logs **per-server, per-engine running averages** (e.g. running/waiting requests, KV cache usage, plus token throughput rates) to W&B under `inference/*`.
> 
> Implements a new `InferenceMetricsCollector` that polls in the background (5s interval, 20-sample window), handles dynamic pool membership, and is started/stopped by the orchestrator; the step loop snapshots and merges these metrics into the main log payload.
> 
> Separately, `EventLoopLagMonitor` now uses a bounded `deque` instead of list trimming, and admin HTTP clients increase connection pool limits to better support concurrent admin/metrics requests. Changelog and unit tests are updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e4adb5885804bebbd222936976e48b837e0e8701. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


eexwhyzee:
@mikasenghaas i think you should be able to configure wandb sdk to scrape openmetrics/prometheus endpoints, see: https://wandb.ai/dimaduev/dcgm/reports/Monitoring-GPU-cluster-performance-with-NVIDIA-DCGM-Exporter-and-Weights-Biases--Vmlldzo0MDYxMTA1

so essentially you would only need to update the [wandb monitor init](https://github.com/PrimeIntellect-ai/prime-rl/blob/main/src/prime_rl/utils/monitor/wandb.py#L45) to allow passing in the settings config that defines the openmetrics endpoints and metrics filters

edit: ahh the downside with this direct wandb approach is that i don't think there's a way to apply any calculations to the metric values (like averaging over a sliding window), it just scrapes the raw values and sends it to wandb

mikasenghaas:
@eexwhyzee ah nice, i acc didn't know abt this. i think there is a couple of reasons we might want to keep programmatic control over this logging tho rather than being boxed in by the wandb sdk:
- elastic inference pools
- log with custom keys
- as you say, apply calculations/ aggregations (altho i acc much prefer a time x-axis for sys metrics over an ema logged per step)

PR #2049, opened 2026-03-19
Add router_url support for elastic inference pool

JannikSt:
## Summary
- Adds `router_url` field to `ClientConfig` for routing inference requests through a vllm-router
- When set alongside elastic mode, inference requests go through the router for load-aware balancing while admin ops (weight updates, LoRA loading) still go directly to discovered pods
- Needed for multi-tenant hosted RL where multiple orchestrators share inference pods

---

> [!NOTE]
> **Medium Risk**
> Changes how elastic inference clients are constructed by optionally routing all inference traffic through a single `router_url`, which can affect availability/traffic flow if misconfigured. Admin operations still target individual pods, but the new branching logic could change behavior in multi-server readiness edge cases.
> 
> **Overview**
> Adds an optional `router_url` to `ClientConfig` to support sending *inference* requests through a vLLM router while keeping elastic DNS discovery for *admin* operations.
> 
> Updates `ElasticInferencePool` to accept/propagate `router_url` from config and, when configured and at least one backend is `ready`, exposes the router as the sole client URL instead of directly using discovered pod `/v1` endpoints.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 47833682a06e17edb1c554aafb60a51c1d629710. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/configs/shared.py
+++ b/src/prime_rl/configs/shared.py
@@ -196,6 +196,13 @@ class ClientConfig(BaseConfig):
         ),
     ] = None
 
+    router_url: Annotated[
+        str | None,
+        Field(
+            description="URL of a vllm-router for load-aware inference routing. When set with elastic mode, inference requests go through the router while admin operations (weight updates, LoRA loading) still go directly to discovered pods.",
+        ),
+    ] = None
+
     @property
     def is_elastic(self) -> bool:
         """Check if elastic mode is enabled."""
--- a/src/prime_rl/utils/elastic.py
+++ b/src/prime_rl/utils/elastic.py
@@ -168,7 +168,9 @@ class ElasticInferencePool:
         # When a router URL is configured, route inference requests through it
         # instead of directly to discovered pods. Admin operations still use
         # individual pod IPs via admin_clients.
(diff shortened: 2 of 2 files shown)


hallerite (approved):

mikasenghaas (commented):
hm can we avoid prime-rl local and hosted rl diverging even more here? eg. would it make sense to deploy the router for local setups as well? maybe this would also remove the need for static vs. elastic inference pools as discovery could happen in the shared router component?

JannikSt:
> hm can we avoid prime-rl local and hosted rl diverging even more here? eg. would it make sense to deploy the router for local setups as well? maybe this would also remove the need for static vs. elastic inference pools as discovery could happen in the shared router component?

I only consider this as temporary fix for a bigger problem in hosted-rl. We still need elastic as a control plane. The router URL serves as data plane. 
Matej and Sami are working anyways on including the router in static environments.

samsja (approved):

PR #2052, opened 2026-03-19
quick fix for sft on single run

Jackmin801:


src/prime_rl/trainer/weights.py
```diff
@@ -174,6 +175,9 @@ def get_adapter_state_dict(model: nn.Module, is_master: bool) -> dict[str, Tenso
             # Add PEFT-expected prefix
             peft_key = f"base_model.model.{clean_key}"
 
+            # Strip multi-adapter index from ParameterList keys (e.g. lora_A.0 -> lora_A)
+            peft_key = re.sub(r"(lora_[AB])\.\d+", r"\1", peft_key)
+
```
Jackmin801, line 180: Im not sure if this is necessary after the fix above where we use multi_run_managers state dict saving

PR #2064, opened 2026-03-21
fix: skip multimodal samples that exceed seq_len instead of truncating

hallerite:
## Summary

Prevents corrupt/crashing VLM training by **skipping multimodal samples whose image tokens were truncated** by vLLM.

**Problem:** vLLM truncates prompts to `max_model_len` (= `seq_len`), cutting `<|image_pad|>` tokens from `input_ids` while `pixel_values` and `image_grid_thw` are unchanged. This causes `ValueError: Image features and image tokens do not match` in the vision encoder.

**Fix:** The orchestrator now validates each multimodal sample by comparing the actual `<|image_pad|>` count in `input_ids` against the expected count derived from `image_grid_thw`. Mismatched samples are skipped before serialization/transport. A clear `ValueError` is raised if all samples are skipped, telling the user to increase `seq_len`.

**Key changes:**
- Filtering moved to orchestrator (avoids wasting transport bandwidth on doomed samples)
- Image token ID resolved from tokenizer at startup
- Unified skip logging with per-reason breakdown: `Skipped 3/256 samples (3 multimodal with truncated image tokens)`
- `prepare_sample` in the trainer is simplified back to a plain function (no more `None` returns)

## End-to-end validation (Qwen3-VL-4B, color-codeword env)

| Branch | seq_len | Result |
|--------|---------|--------|
| `main` | 128 | **Crash:** `ValueError: Image features and image tokens do not match, tokens: 16, features: 128` |
| `fix` | 128 | **Clear error:** `All 4 training samples were skipped (4 multimodal with truncated image tokens). Increase seq_len (currently 128)` |
| `fix` | 4096 | **Success:** Both training steps complete, no samples skipped |

## Test plan
- [x] `test_prepare_sample_truncates_multimodal_exceeding_seq_len` — multimodal truncation works
- [x] `test_prepare_sample_keeps_multimodal_within_seq_len` — multimodal within seq_len preserved
- [x] `test_prepare_sample_still_truncates_text_only` — text-only truncation unchanged
- [x] All existing batch/trajectory tests pass
- [x] End-to-end VLM RL validated on 2x RTX PRO 6000 (1 infer + 1 train GPU)

---

> [!NOTE]
> **Medium Risk**
> Changes the orchestrator’s sample filtering and can drop training data when prompts are truncated, which may affect training throughput/behavior if misconfigured. Low security impact but touches core training data preparation for VLM runs.
> 
> **Overview**
> Prevents corrupt VLM training data by **skipping multimodal samples where vLLM truncation breaks image-token ↔ `pixel_values` alignment** (detected by comparing counted image placeholder tokens vs. the expected count derived from `image_grid_thw`). If all samples are skipped, the orchestrator now raises a clear error instructing to increase `seq_len`, and logs per-reason skip counts.
> 
> Adds `model.vlm.image_token_id` (optional override) and extends the VLM registry to include per-family `image_token_id`, with orchestrator resolution via config or registry at startup. The trainer’s `prepare_sample` now **refuses to truncate multimodal sequences** (raises `ValueError`), while keeping existing truncation behavior for text-only samples; new unit tests cover these cases and docs are updated with the new limitation/guidance.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e7f3da3fb6a737744628d3aa0999b09e3d5509e7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/trainer/batch.py
```diff
@@ -25,6 +29,16 @@ def prepare_sample(training_example: TrainingSample, seq_len: int) -> MicroBatch
     routed_experts = training_example.routed_experts
 
     if len(input_ids) > seq_len:
+        # Multimodal samples cannot be truncated: the number of image_pad tokens
+        # in input_ids must match the pixel_values, and truncation would break
+        # that alignment. Skip the sample instead of producing garbage.
+        if training_example.pixel_values is not None:
```
mikasenghaas, line 28: i think if we alr have warning log in the `prepare_batch` we don't need a per-sample warning log here?

mikasenghaas: imo this would only make sense if there are multiple conditions for skipping a smaple but even then i would make it a debug log to not flood the console

PR #2073, opened 2026-03-22
feat: GPT-OSS support with LoRA + DP fixes

eligotts:
## Summary

Adds support for serving GPT-OSS MoE models (e.g., `unsloth/gpt-oss-20b-BF16`) with LoRA enabled under data parallelism. Fixes two vLLM 0.17.0 bugs that caused corrupted model output when combining LoRA + MoE + DP>1.

### Bug 1: Harmony stop token propagation

vLLM's `ChatCompletionRequest.to_sampling_params()` ignores `default_sampling_params["stop_token_ids"]`, which contains the harmony stop tokens (`<|call|>` 200012 and `<|return|>` 200002). The model generates past tool call boundaries without stopping.

**Fix:** Monkey-patch `to_sampling_params()` to merge default stop tokens into per-request `SamplingParams`.

Upstream: https://github.com/vllm-project/vllm/issues/22519

### Bug 2: FusedMoE LoRA + DP produces corrupted output

Two interacting issues:

1. **`supports_internal_mk` lie:** LoRA injection creates a `FusedMoEModularMethod` with `moe_kernel` set (not None), making `supports_internal_mk=True`. The MoE runner trusts this and skips DP dispatch/combine. But the LoRA kernel uses `NoDPEP` — it doesn't handle DP. Each rank processes only local tokens without coordination. Result: corrupted output.

2. **Full-batch vs chunk mismatch:** The LoRA decorators (`act_decorator`, `moe_sum_decorator`) capture tensors from the kernel's `apply()` kwargs (full-batch), but fire inside the kernel's chunk loop (chunk-sized workspace). At DP≥3, the dispatched batch exceeds `VLLM_FUSED_MOE_CHUNK_SIZE` (16384), causing shape assertion failures in the LoRA triton kernel.

**Fix:** Replace `_inject_lora_into_fused_moe` with a version that:
- Sets `moe_kernel=None` so the runner correctly applies DP dispatch/combine
- Makes decorators chunk-aware by tracking `_chunk_offset` and slicing captured tensors to the current chunk

The patch is applied in `vllm/worker/__init__.py` so it runs inside every DP worker subprocess (not just the main API server process).

Upstream: https://github.com/vllm-project/vllm/issues/23244

### Additional setup notes

One item that is NOT a code change but required for GPT-OSS:

- **`generation_config.json` fix for unsloth model:** `unsloth/gpt-oss-20b-BF16` is missing `<|call|>` (token 200012) from `eos_token_id`. Patch the cached HF file:
  ```python
  # eos_token_id should be [200002, 199999, 200012] (not [200002, 199999])
  ```

## Test plan

- [x] DP=6 + LoRA: plain text completion produces coherent output ("Paris")
- [x] DP=6 + LoRA: tool calling produces valid tool calls with `stop_reason=200012`
- [x] DP=6 + LoRA: 30 concurrent requests — 0 corrupted, 0 errors
- [x] DP=6 + LoRA: verifiers ToolEnv multi-turn rollout succeeds (both client paths)
- [x] DP=6 + LoRA: full RL training run with wiki-search environment — training steps complete with reward ~0.5
- [x] DP=1 + LoRA: no regression (works as before)
- [x] DP=6 without LoRA: no regression
- [ ] Multi-LoRA with DP>1 (not tested on this hardware, but fix is designed to preserve multi-LoRA routing)


examples/wiki_search/rl_gpt_oss_20b.toml
```diff
@@ -0,0 +1,52 @@
...
+
+[deployment]
+num_train_gpus = 2
+num_infer_gpus = 6
+
+[model]
+name = "unsloth/gpt-oss-20b-BF16"
+
+[wandb]
+project = "wiki-search"
```
samsja: lets remove the config

samsja (approved):

PR #2074, opened 2026-03-23
feat: accept `messages` column for SFT

rasdani:
allows `messages` column in an SFT dataset instead of only `prompt` and `completion` columns.
more convenient for multi-turn SFT.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk: changes how SFT samples are parsed/tokenized and which fields take precedence, which can affect training loss masking and data compatibility, but is localized to dataset preprocessing with added unit coverage.
> 
> **Overview**
> Adds support for SFT datasets that provide a raw `messages` column (full conversation) in addition to the existing `prompt`/`completion` split format, with `messages` taking precedence when both are present.
> 
> Updates SFT preprocessing to normalize/deserialize/strip message content for either input form before building the incremental token mask, and expands unit tests and docs to cover the new format and precedence/equivalence behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d7d7e1960672259c0354d1b43a134ba28216e744. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/trainer/sft/data.py
```diff
@@ -157,14 +162,28 @@ def _process(self, example: dict) -> dict | None:
         if self.tokenizer is None:
             return example
 
-        # Assert that the example has a 'prompt' and 'completion' column
-        if "prompt" not in example or "completion" not in example:
-            raise ValueError("All examples in the dataset must have a 'prompt' and 'completion' column for SFT")
+        def resolve_messages(example: dict) -> tuple[list[dict], list[dict]]:
```
mikasenghaas: i think it would be cleaner if this function resolves to `list[dict]`. in the prompt-completion case, we already return the concatenated lists.

src/prime_rl/trainer/sft/data.py
```diff
@@ -157,14 +162,28 @@ def _process(self, example: dict) -> dict | None:
...
+            # `messages` takes precedence over explicit split fields and is interpreted
+            # as a whole-chat training sample with an empty prompt.
+            if "messages" in example:
+                return [], normalize_messages(example["messages"], default_role="assistant")
+
+            if "prompt" not in example or "completion" not in example:
+                raise ValueError(
```
mikasenghaas, line 175: code style: cleaner imo to do 1) if message 2) elif promt-completion 3) else raise ValueError

PR #2092, opened 2026-03-25
fix: terminate process on fatal orchestrator error

JannikSt:
Saw this on hosted training on the platform - a run was stuck in RUNNING but showing the error: `ValueError: dataset is not set` leading to an orphan run which was never cleaned up.

- Replace raise with sys.exit(1) in the clean_exit decorator
- In async context, re-raising an exception does not kill the process - the event loop swallows it and the process hangs forever
- sys.exit raises SystemExit so the finally block still runs and cleans up the distributed process group

---

> [!NOTE]
> **Medium Risk**
> Changes the `clean_exit` error path to call `sys.exit(1)` instead of re-raising, which affects how orchestrator/trainer processes terminate and could cause earlier-than-expected shutdown if exceptions were previously handled upstream.
> 
> **Overview**
> Ensures processes terminate on fatal exceptions wrapped by `clean_exit` by replacing exception re-raises with `sys.exit(1)` and importing `sys`.
> 
> This applies to both async and sync wrappers, while preserving cleanup behavior (`wandb.finish` and `torch.distributed` process group teardown) in the `finally` block to avoid hung runs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 911c935efa7f2dabbcaca11122dfd4a3cf1d3b04. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/utils/utils.py
+++ b/src/prime_rl/utils/utils.py
@@ -130,7 +130,7 @@ def clean_exit(func: Callable) -> Callable:
                 ret = await func(*args, **kwargs)
                 wandb.finish()
                 return ret
-            except Exception as e:
+            except Exception:
                 get_logger().opt(exception=True).error(f"Fatal error in {func.__name__}")
                 wandb.finish(exit_code=1)
                 # sys.exit raises SystemExit so the finally block still runs.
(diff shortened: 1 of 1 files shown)


JannikSt:
@codex review

samsja (approved):

PR #2111, opened 2026-03-27
Fix: run metrics/health server in separate process (GIL starvation)

JannikSt:
## Problem

The health/metrics HTTP server runs as a Python thread sharing the GIL with the training loop. During long forward/backward passes on large models (397B with EP=8, CP=4), Python holds the GIL for extended periods during checkpoint saves, tensor reshaping, and gradient sync. The HTTP thread can't respond, K8s liveness probes get `connection reset by peer`, and the pod gets killed.

On gai-single (GB200 NVL72), this caused the entire 8-pod trainer group to restart every ~30 minutes, preventing any training progress.

## Fix

Replace `threading.Thread` with `multiprocessing.Process`. The server process has its own GIL and always responds instantly.

Metrics are shared via `multiprocessing.Array` (lock-free shared memory):
- Trainer calls `update()` which writes floats to shared arrays
- Server process reads from the same arrays to serve `/metrics`
- `/health` returns `ok` with zero shared state

## API

Fully backwards compatible — same `HealthServer` and `MetricsServer` classes, same `update()` and `update_runs()` signatures. Only the internal implementation changes from thread to process.

---

> [!NOTE]
> **Medium Risk**
> Changes the liveness/metrics server execution model from a thread to a separate process with lock-free shared memory, which can introduce multiprocessing lifecycle/port-binding and metric consistency issues if mis-handled.
> 
> **Overview**
> Moves `HealthServer`/`MetricsServer` off a background thread and into a dedicated `multiprocessing.Process` so `/health` and `/metrics` remain responsive even when the trainer holds the GIL.
> 
> Replaces `prometheus_client` gauges/registry with a fixed shared-memory layout (`multiprocessing.Array`/`Value`) written by `update()`/`update_runs()` and read by the server process to emit Prometheus text format (including per-run labeled metrics, capped at 32 runs).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1c04d6d422a8b5f5937a77c4283530989ecbb585. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/models/nemotron_h/configuration_nemotron_h.py
+++ b/src/prime_rl/trainer/models/nemotron_h/configuration_nemotron_h.py
@@ -170,10 +170,13 @@ class NemotronHConfig(PretrainedConfig):
         self.mamba_proj_bias = mamba_proj_bias
         self.chunk_size = mamba_chunk_size
 
-        # Zamba2MambaMixer compat aliases (read by parent __init__ before NemotronHMamba2Mixer overrides)
+        # Zamba2MambaMixer compat aliases (read by parent __init__ before NemotronHMamba2Mixer overrides).
+        # mamba_expand must give the correct intermediate_size = mamba_num_heads * mamba_head_dim
+        # when Zamba2 computes int(mamba_expand * hidden_size); the config's raw "expand" field
+        # does not satisfy this for all model sizes (e.g. Nemotron-3-Nano-30B).
         self.mamba_d_state = ssm_state_size
         self.mamba_d_conv = mamba_d_conv
-        self.mamba_expand = mamba_expand
+        self.mamba_expand = (mamba_num_heads * mamba_head_dim) / hidden_size
         self.mamba_ngroups = mamba_n_groups
         self.mamba_headdim = mamba_head_dim
         self.n_mamba_heads = mamba_num_heads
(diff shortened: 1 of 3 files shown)


vominh1919:
Hi! I've analyzed this issue and have a proposed implementation plan.

## Proposed Solution

Replace `threading.Thread` with `multiprocessing.Process` and use shared memory for metrics:

### Key Changes

1. **HealthServer/MetricsServer** → Use `multiprocessing.Process` instead of `threading.Thread`
2. **Metrics storage** → Use `multiprocessing.Array` and `multiprocessing.Value` for shared memory
3. **Backward compatibility** → Keep same API (`update()`, `update_runs()` signatures)
4. **Cleanup** → Proper process lifecycle management (start/stop/join)

### Implementation Approach

```python
# Shared memory layout for metrics
class SharedMetrics:
    def __init__(self):
        # Scalar metrics
        self.step = multiprocessing.Value('i', 0)  # int
        self.loss = multiprocessing.Value('d', 0.0)  # double
        self.throughput = multiprocessing.Value('d', 0.0)
        # ... etc
        
        # Per-run metrics (up to 32 runs)
        self.run_steps = multiprocessing.Array('i', 32)  # int array
        self.run_tokens = multiprocessing.Array('i', 32)
        # ... etc
```

### Benefits

- ✅ Server responds instantly during GIL-heavy operations
- ✅ Lock-free shared memory for metrics
- ✅ Backward compatible API
- ✅ No dependency changes (still uses stdlib multiprocessing)

### Risks

- ⚠️ Port binding lifecycle changes
- ⚠️ Process cleanup on crashes
- ⚠️ Shared memory synchronization

## Next Steps

Would you like me to:
1. **Implement the full fix** — I can create a PR with the multiprocessing-based implementation
2. **Start with a minimal POC** — Just HealthServer first, then expand to MetricsServer
3. **Discuss design first** — Review the shared memory layout before implementation

Let me know how you'd like to proceed!

PR #2140, opened 2026-03-30
fix ODF logging and add min_abs_adv

faresobeid:
> [!NOTE]
> **Medium Risk**
> Changes which rollouts count toward training batches based on computed advantages, which can alter effective batch composition and learning dynamics. Adds new config validation/deprecation paths that could break misconfigured runs but is covered by unit tests.
> 
> **Overview**
> Introduces `orchestrator.buffer.min_abs_adv` to **filter which rollouts are sent to the trainer based on `|advantage|`**, while still keeping all rollouts for orchestrator logging and analysis.
> 
> Deprecates `buffer.online_difficulty_filtering` as an alias for `min_abs_adv=0.0` (emits `FutureWarning` and rejects conflicting values), updates validation so `verification.enabled=false` and GR³ `advantage.length_shaping_alpha` cannot be used without an explicit `min_abs_adv` setting.
> 
> Moves filtering responsibility out of `Buffer` (and removes its `filtered_rollouts/*` buffer metrics), updates the `Scheduler`/`orchestrator` loop to track a per-rollout keep mask, only emit selected rollouts into training batches, and logs new `filtered_rollouts/*` and `passes_min_abs_adv` metrics. Example configs are updated to use `min_abs_adv = 0.0`, and unit tests are added/updated for the new behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d1c8b858e87b99624f8eb748dba08f7b0fc5227f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/configs/orchestrator.py
```diff
@@ -514,10 +515,20 @@ class BufferConfig(BaseConfig):
         ),
     ] = 0.0
 
+    adv_filter: Annotated[
+        float | None,
+        Field(
+            description=(
+                "Only send rollouts with advantage greater than this threshold to training. "
```
mikasenghaas: this is absolute advantage right?

mikasenghaas: i think could have more descriptive name here. maybe `--minimum-absolute-advantage` (or `--min-abs-adv`)

mikasenghaas: ah wait its greater than so this is still not quite clear but smth along those lines maybe

src/prime_rl/configs/orchestrator.py
```diff
@@ -541,6 +552,30 @@ def validate_env_ratios(self):
             assert all(ratio > 0 for ratio in self.env_ratios), "All env_ratios must be positive."
         return self
 
+    @model_validator(mode="after")
+    def resolve_adv_filter(self):
+        if not self.online_difficulty_filtering:
+            return self
+
+        warnings.warn(
```
mikasenghaas, line 560: should we use prime-rl logger?

src/prime_rl/orchestrator/advantage.py
```diff
@@ -97,3 +97,26 @@ def compute_advantages(
...
+        samples_per_problem=samples_per_problem,
+        advantage_config=advantage_config,
+    )
+    keep_mask = [should_keep_rollout(advantage, adv_filter) for advantage in advantages]
+    return advantages, keep_mask
+
+
+def should_keep_rollout(advantage: float, adv_filter: float | None) -> bool:
+    """Returns whether a rollout should be sent to training."""
+    return adv_filter is None or advantage > adv_filter
```
mikasenghaas: oh its not absolute?!

src/prime_rl/orchestrator/buffer.py
```diff
@@ -231,14 +231,6 @@ def update(self, rollouts: list[vf.RolloutOutput]):
                 target_pool.append(example)
 
             self.num_examples_per_step[env_name][pool] += 1
-            if self.config.online_difficulty_filtering:
```
mikasenghaas, line 234: does this not fuck up the buffer metrics?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -553,11 +553,12 @@ def _cleanup_env_processes():
         num_unique_examples = len(set(example_ids))
         rewards = [r["reward"] for r in train_rollouts]
         completion_lens = [get_completion_len(r) for r in train_rollouts]
-        advantages = compute_advantages(
-            rewards,
-            completion_lens,
-            config.rollouts_per_example,
-            config.advantage,
+        advantages, rollout_keep_mask = compute_advantages_and_keep_mask(
```
mikasenghaas: ouf, yea those 2 things should not be coupled. can we not compute advantages and then apply the filter?

mikasenghaas: or what is the difficulty with decoupling?

mikasenghaas (commented):
would be really cool if we could extend the `Filter` abstraction to also operate on groups?

src/prime_rl/orchestrator/advantage.py
faresobeid: oh oops idk how that happened

src/prime_rl/configs/orchestrator.py
faresobeid: hmm not sure, also config validation happens before the prime-rl logger is set up

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -553,11 +553,12 @@ def _cleanup_env_processes():
         rewards = [r["reward"] for r in train_rollouts]
         completion_lens = [get_completion_len(r) for r in train_rollouts]
         advantages = compute_advantages(
-            rewards,
-            completion_lens,
-            config.rollouts_per_example,
-            config.advantage,
+            rewards=rewards,
```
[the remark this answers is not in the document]
faresobeid: we want to log rollout metrics based on them all, not just the ones going to the trainer

src/prime_rl/configs/orchestrator.py
mikasenghaas: we fixed this so that `get_logger()` always works and alr do this in other places in the configs, let's do it for consistency

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -609,18 +612,28 @@ def process_rollout(rollout: vf.RolloutOutput, rollout_idx: int) -> list[Trainin
                     sample_prefill_tokens = len(sample.prompt_ids) + len(sample.completion_mask) - sample_decode_tokens
                     rollout_decode_tokens += sample_decode_tokens
                     rollout_prefill_tokens += sample_prefill_tokens
-                    train_examples.append(sample)
+                    if selected_for_training:
```
mikasenghaas, line 615: oh but this doesn't work, no? bc now all rollouts from the scheduler might get filtered and we get an empty batch? imo, the filtering should happen inside the scheduler so that we are guaranteed to get the specified number of training rollouts

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -453,18 +472,21 @@ async def generate_batch(self, step: int) -> list[vf.RolloutOutput]:
 
                 self.buffer.update(completed_rollouts)
                 accepted_rollouts = self.buffer.sample_rollouts(n=self.rollouts_per_example)
```
mikasenghaas, line 474: does the sample rollout buffer method still do smth if we dont do odf there?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -453,18 +472,21 @@ async def generate_batch(self, step: int) -> list[vf.RolloutOutput]:
...
+                selected_group_mask, progress_increment = self.select_rollouts(
+                    accepted_rollouts,
+                    remaining_batch_target=self.batch_target - batch_progress,
+                )
 
                 batch_rollouts.extend(accepted_rollouts)
-                progress_increment = self.get_batch_progress_increment(accepted_rollouts)
+                selected_rollout_mask.extend(selected_group_mask)
```
mikasenghaas, line 481: yea, imo we should not track what was selected but just only store the rollouts which were selected. seems way cleaner bc then we don't have to do any re-filtering on the orch

mikasenghaas (changes requested):

PR #2144, opened 2026-03-31
[fix] dont use ep enabled variant of clip gradnorm

Jackmin801:
This seems to not actually work but without it its fine

---

> [!NOTE]
> **Low Risk**
> Low risk: a small change limited to the SFT training loop’s gradient clipping call; main risk is behavior differences in gradient norm computation/clipping under expert-parallel configurations.
> 
> **Overview**
> Removes use of the `ep_enabled` argument when calling `clip_grad_norm_` in the SFT training loop, always using the default gradient-norm clipping path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 501b23709caec5e372036537b7973e055a1abe55. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/sft/train.py
+++ b/src/prime_rl/trainer/sft/train.py
@@ -389,9 +389,7 @@ def train(config: SFTConfig):
         nan_loss_count = nan_loss_count.item()
 
         logger.debug(f"Clipping gradients with max norm {config.optim.max_norm}")
-        grad_norm = clip_grad_norm_(
-            model.parameters(), max_norm=config.optim.max_norm
-        )
+        grad_norm = clip_grad_norm_(model.parameters(), max_norm=config.optim.max_norm)
         if grad_norm.device.type == "cpu":
             grad_norm = grad_norm.to(torch.device("cuda"))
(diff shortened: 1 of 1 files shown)


src/prime_rl/trainer/sft/train.py
```diff
@@ -389,9 +389,7 @@ def run_validation(step: int) -> None:
         nan_loss_count = nan_loss_count.item()
 
         logger.debug(f"Clipping gradients with max norm {config.optim.max_norm}")
-        grad_norm = clip_grad_norm_(
-            model.parameters(), max_norm=config.optim.max_norm, ep_enabled=parallel_dims.ep_enabled
-        )
+        grad_norm = clip_grad_norm_(model.parameters(), max_norm=config.optim.max_norm)
```
S1ro1, line 392: Imo this will be broken with standard EP, so it should be just `en_enabled=use_deepep` and the same should be added to rl? Though I'd swear it worked before with deepep

PR #2161, opened 2026-04-01
moe_act target and better docs

faresobeid:
Adds moe_act selective AC target and adds some docs on selective AC tuning

---

> [!NOTE]
> **Medium Risk**
> Changes the trainer’s default activation-checkpointing behavior for custom models (now implicitly enables selective AC), which can affect memory/throughput characteristics and recomputation. Adds a new selective AC hook in MoE expert code paths, so correctness/perf should be validated on representative MoE workloads.
> 
> **Overview**
> **Selective activation checkpointing is expanded and made easier to use for custom models.** When `trainer.model.ac` is unset, the trainer now *implicitly enables* selective activation checkpointing (`mode="selective"`, `targets=["norm"]`) for the custom implementation, while HF models still default to AC disabled; explicitly setting `[trainer.model.ac]` (or `--model.ac`) continues to mean full-layer checkpointing.
> 
> **Adds a new selective target `moe_act` for MoE layers.** The selective AC system can now checkpoint only the routed expert activation function, and it is automatically skipped when `routed_experts` is also enabled to avoid nested/double checkpointing; MoE expert implementations were refactored to expose `moe_act` as a hookable method.
> 
> Docs and changelog were updated to describe the new defaulting behavior and provide selective-AC tuning guidance, and unit tests were added/updated to cover the new defaults and `moe_act` patching/subsumption behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2b0abaaad8846ce36bfa75260bda726c75578bb8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


docs/memory_usage.md
```diff
@@ -29,20 +29,86 @@ max_inflight_activations = 1
 
 Activation checkpointing discards intermediate activations during the forward pass and recomputes them during the backward pass, trading compute for memory.
 
-To enable it, use:
+If `trainer.model.ac` is unset, supported custom implementations default to selective AC on the cheapest target:
```
samsja, line 32: huh ? where is it define, I don't like this tbh

samsja (changes requested):
I don't like the magic logic that auto select the ac even when the cli doesn't enable it, lets just enable ac by default instead ??

docs/memory_usage.md
mikasenghaas: would agree here. a config unset should not default do doing it anyway

mikasenghaas: we can ofc default to selective ac if this is reasonable, but it should be explicit in the configs imo (e.g. an agent should see ac.mode = selective instead of ac.mode = None)

PR #2162, opened 2026-04-01
Make fused SFT loss default

faresobeid:
Makes fused SFT loss default.
Now only exposes standard and fused with fused as default, in the back we choose quack if available, if not we fallback to liger

---

> [!NOTE]
> **Medium Risk**
> Breaking config change removes the old `loss_impl` options and changes the default SFT loss path, which can affect training correctness/perf across GPU types and dependency availability.
> 
> **Overview**
> **Makes fused SFT loss the default and simplifies configuration.** The SFT `loss_impl` setting is reduced to `standard` vs `fused` (default), with the changelog documenting required config migrations.
> 
> When `loss_impl="fused"`, SFT now auto-selects the fused backend at runtime: prefer `quack-kernels` on supported GPUs (sm90+), otherwise fall back to Liger; Gemma softcapping no longer hard-errors and instead falls back to Liger with a warning. Docs are updated with optional `quack-kernels` install guidance and hardware requirements.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d2a1f27409def4fcf579d865d4964e6a78a6a194. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/prime_rl/trainer/models/layers/lm_head.py
```diff
@@ -299,14 +320,20 @@ def inject_prime_lm_head(
 
     logger = get_logger()
 
+    if fused_cross_entropy == "quack":
+        fused_cross_entropy, fallback_reason = resolve_fused_cross_entropy_backend()
```
mikasenghaas: lets not have `str` side effects for logging from our funcs. you can `get_logger` inside all utils anytime

faresobeid: oh ya cursor bugbot just did that on its own, not even linked to the issue i resolved lol

samsja (changes requested):
I don't like this auto magic. We shouldn't have a param that can dispatch to two completely different implementation on the same thing base on the dependency install.

Things should stay explicit

PR #2169, opened 2026-04-01
feat: reward-gated length shaping for correct rollouts

hallerite:
## Summary

- Replace GR³ length scaling with a simpler correctness-gated approach: correct rollout rewards are attenuated by `L_min / L_i`, where `L_min` is the shortest correct completion per problem
- Shortest correct rollout keeps full reward (1.0), longer correct ones get proportionally less, incorrect rollouts are untouched
- Remove `length_shaping_alpha: float` in favor of `length_shaping: bool` (no hyperparameter needed) and drop the online difficulty filtering requirement

## Breaking changes

- `orchestrator.advantage.length_shaping_alpha` → `orchestrator.advantage.length_shaping` (float|None → bool)
- `length_shaping` no longer requires `online_difficulty_filtering = true`

### DIFF
--- a/configs/length_shaping_ablation/base.toml
+++ b/configs/length_shaping_ablation/base.toml
@@ -1,0 +1,92 @@
+max_steps = 10000
+seq_len = 40960
+
+[slurm]
+job_name = "length-penalty"
+
+[deployment]
+type = "multi_node"
+num_train_nodes = 1
+num_infer_nodes = 3
+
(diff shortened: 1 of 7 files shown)


hallerite:
Superseded by #2377, which builds on the length shaping that landed in #2265 and adds weighted shaping over tool-response tokens for the agentic length-penalty ablations.

PR #2180, opened 2026-04-02
chore: strip [default0] prefix from trainer logs

mikasenghaas:
## Summary
- Strip the `[default0]:` prefix that torchrun's `--tee` flag prepends to every log line (note, that these can still be seen for in the per-rank logs)
- Uses `sed -u` in the tail/tee pipelines to remove the prefix in real-time
- Also sets `--role=trainer` for clearer per-rank log files in the torchrun log directory
- Applied to both local launchers (rl.py, sft.py) and SLURM templates

**Before:**
```
[default0]:00:19:20    INFO Starting training loop (max_steps=100)
[default0]:00:20:21 SUCCESS Step 0 | Time: 60.98s | Loss: 0.0000 | ...
```

**After:**
```
00:19:20    INFO Starting training loop (max_steps=100)
00:20:21 SUCCESS Step 0 | Time: 60.98s | Loss: 0.0000 | ...
```

---

> [!NOTE]
> **Low Risk**
> Logging-only changes that adjust `torchrun` args and shell pipelines; main risk is portability/quoting issues when spawning `shell=True` or running on systems without compatible `sed`.
> 
> **Overview**
> Trainer launch commands now pass `torchrun --role=trainer` so per-rank log files are labeled more clearly.
> 
> For both local (`rl.py`, `sft.py`) and multi-node SLURM templates, the displayed/tee’d trainer log stream is piped through `sed -u` to strip the leading `[...]` rank prefix (e.g., `[default0]:`) while keeping the underlying per-rank torchrun logs intact.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0332018eac7615dca88e27f84050a9bcef803fa6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/prime_rl/entrypoints/rl.py
+++ b/src/prime_rl/entrypoints/rl.py
@@ -296,6 +296,7 @@ def rl_local(config: RLConfig):
         # Start training process
         trainer_cmd = [
             "torchrun",
+            "--role=trainer",
             f"--rdzv-endpoint=localhost:{get_free_port()}",
             f"--rdzv-id={uuid.uuid4().hex}",
             # Pipe all logs to file, and only master rank logs to stdout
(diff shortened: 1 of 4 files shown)


samsja (changes requested):
nah we need this for when we have multi rank trainer writting to the same file. Also I am against doing too many magic beyond tail. The file is the file

samsja (approved):
approved

PR #2198, opened 2026-04-05
fix: handle chat template errors in eval sample logging

eligotts:
## Summary

`apply_chat_template` in `log_eval_samples` crashes when completions contain tool calls serialized as strings (from verifiers) but the model's Jinja template expects tool call objects with `.name`/`.arguments` attributes. This kills the orchestrator after eval.

Observed with gpt-oss + wiki-search environment — the gpt-oss chat template strictly accesses `tool_call.name`, but verifiers serializes tool calls as JSON strings in `RolloutOutput`.

Introduced in #2143.

## Fix

Wrap `apply_chat_template` in try/except, fall back to raw string representation. Logging code should never crash training.

---

> [!NOTE]
> **Low Risk**
> Low risk: only affects eval sample logging to W&B, adding defensive formatting/parsing with a safe fallback if chat templating fails.
> 
> **Overview**
> Prevents `WandbMonitor.log_eval_samples` from crashing when formatting list-based `completion` messages that include tool calls.
> 
> The monitor now runs completions through `deserialize_tool_calls()` before `tokenizer.apply_chat_template()`, and wraps templating in `try/except` to fall back to a raw string representation on any error so eval logging can't kill training.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d18503e626530d00a54266c5b8fcb7224978a838. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/utils/monitor/wandb.py
+++ b/src/prime_rl/utils/monitor/wandb.py
@@ -196,14 +196,9 @@ class WandbMonitor(Monitor):
                 continue
             if isinstance(completion, list):
                 try:
-                    completion = self.tokenizer.apply_chat_template(completion, tokenize=False)
+                    completion = self.tokenizer.apply_chat_template(deserialize_tool_calls(completion), tokenize=False)
                 except Exception:
-                    try:
-                        completion = self.tokenizer.apply_chat_template(
-                            deserialize_tool_calls(completion), tokenize=False
-                        )
-                    except Exception:
-                        completion = str(completion)
+                    completion = str(completion)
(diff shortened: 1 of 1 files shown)


mikasenghaas (commented):
should we do one more fallback to json decode the tool call args before just string logging? i remember there was some other chat templates that had problems with this, and we acc do this on the sft entrypoint as well

mikasenghaas (approved):

PR #2214, opened 2026-04-07
feat: make gradient clipping optional, error with DeepEP

samsja:
## Summary
- `optim.max_norm` can now be set to `null` to disable gradient clipping (default remains `1.0`)
- Config validation errors when DeepEP (`model.ep_comm_backend='deepep'`) is used with grad clipping enabled, since they are incompatible
- Applied to both RL and SFT trainers

---

> [!NOTE]
> **Medium Risk**
> Touches core RL/SFT training loops and optimizer configuration; disabling gradient clipping (and conditionally skipping grad_norm metrics) can change training stability/observability if configs rely on prior behavior.
> 
> **Overview**
> **Gradient clipping can now be disabled** by setting `optim.max_norm: null` (config type widened to `float | None`).
> 
> Both RL and SFT trainers now *conditionally* call `clip_grad_norm_` and only log/export `grad_norm` when clipping is enabled.
> 
> Adds config-level validation in `TrainerConfig` and `SFTConfig` to warn and automatically turn off gradient clipping when `model.ep_comm_backend='deepep'`, and updates the Prometheus `MetricsServer.update()` API to accept an optional `grad_norm`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8aef65b8735fc246f90e40f2c0a61497acf70aaf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/configs/sft.py
+++ b/src/prime_rl/configs/sft.py
@@ -272,6 +272,15 @@ class SFTConfig(BaseConfig):
     ### Validate configs (e.g. raise for unsupported (combinations of) configs)
 
     @model_validator(mode="after")
+    def deepep_incompatible_with_grad_clipping(self):
+        if self.model.ep_comm_backend == "deepep" and self.optim.max_norm is not None:
+            raise ValueError(
+                "Gradient clipping (optim.max_norm) is not compatible with DeepEP (model.ep_comm_backend='deepep'). "
+                "Set optim.max_norm to null to disable gradient clipping."
+            )
+        return self
+
+    @model_validator(mode="after")
     def validate_deployment(self):
         if self.deployment.type == "multi_node" and self.slurm is None:
             raise ValueError("Must use SLURM for multi-node deployment.")
--- a/src/prime_rl/configs/sft.py
+++ b/src/prime_rl/configs/sft.py
@@ -1,3 +1,4 @@
+import warnings
 from pathlib import Path
 from typing import Annotated, Literal, TypeAlias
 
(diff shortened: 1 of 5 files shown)


src/prime_rl/trainer/sft/train.py
```diff
@@ -468,12 +473,13 @@ def run_validation(step: int) -> None:
         monitor.log(perf_metrics, step=progress.step)
 
         # Log optimizer metrics
-        optim_metrics = {
+        optim_metrics: dict = {
```
samsja: remove the dict here

src/prime_rl/configs/sft.py
```diff
@@ -271,6 +271,15 @@ def normalize_deployment(cls, data):
 
     ### Validate configs (e.g. raise for unsupported (combinations of) configs)
 
+    @model_validator(mode="after")
+    def deepep_incompatible_with_grad_clipping(self):
```
mikasenghaas: should we not auto set this to `None` for the user?

mikasenghaas: maybe with a warning log

samsja: idk seems off no ?

mikasenghaas: i think if there is not decision left for the user, we could auto set this. diff ofc if idk 1 out of 3 options doesn't work, then we should raise and tell the user about the 2 options left

mikasenghaas (approved):

PR #2232, opened 2026-04-08
chore: improve multi-node SFT defaults

mikasenghaas:
## Summary
- Move `mamba-ssm` from `[dependency-groups]` to a proper `[mamba]` optional extra (NOTE: this means that `uv sync --all-extras` will fail if the `mamba-ssm` package is not compatible with the hardware, but `uv sync` or `uv sync --extra all` will still work)
- Add `UV_NO_SYNC=1` to multi-node SFT template to avoid unnecessary syncs on worker nodes
- Remove `HF_HUB_OFFLINE=1` from SFT template (was preventing model downloads when needed)


samsja:
> * Remove `HF_HUB_OFFLINE=1` from SFT template (was preventing model downloads when needed)

this should be a param no ? we defo should always have hf hub offline for large run, the first time we dl we just need to enable it

> Add UV_NO_SYNC=1 to multi-node SFT template to avoid unnecessary syncs on worker nodes
can we also do this on the rl side ? who is in charge of uv sync ? user ? top level command in the sbatch ?

PR #2242, opened 2026-04-09
fix: prepend base_model.model. prefix to broadcast LoRA keys

JannikSt:
## Problem

LoRA training on MoE models (Qwen3.5-35B, etc.) crashes with vLLM 0.19 (v0.5.1.dev48+) when the orchestrator tries to load the LoRA adapter into inference:

```
ValueError: While loading .../broadcasts/step_1, expected target modules in
{'down_proj', 'gate_proj', 'up_proj', ...} but received
['language_model.model.layers.0.mlp.experts.0.down_proj', ...]
```

## Root cause

Two code paths save LoRA weights:

1. **Checkpoint path** (ckpt.py:322) - prepends `base_model.model.` to keys
2. **Broadcast path** (filesystem.py:62-67) - saves raw keys without prefix

vLLM's `parse_fine_tuned_lora_name()` expects keys starting with `base_model.model.` so it can strip the prefix and match the short module name (e.g. `down_proj`) against `expected_lora_modules`. Without the prefix, the full path doesn't match and vLLM rejects it.

This worked on vLLM 0.17 (dev2) because it didn't have `check_unexpected_modules()` validation. vLLM 0.19 added this check.

## Fix

Prepend `base_model.model.` in the broadcast path, matching what the checkpoint path already does. Backward compatible with dev2/vLLM 0.17.

## Impact

- all MoE model LoRA training broken on dev48 (Qwen3.5-35B confirmed)
- dense models unaffected
- backward compatible

---

> [!NOTE]
> **Medium Risk**
> Overrides vLLM’s `LoRAModel.from_local_checkpoint` validation/parsing via monkeypatching, which can affect whether adapters load or are rejected across models. Risk is moderate because it changes inference-time weight loading logic but is scoped to LoRA adapter ingestion.
> 
> **Overview**
> Fixes LoRA adapter loading failures in vLLM by installing a new `_patch_lora_key_prefix()` that monkeypatches `LoRAModel.from_local_checkpoint` to tolerate adapter tensor keys that *don’t* include the `base_model.model.` prefix.
> 
> The patch also relaxes/adjusts unexpected-module validation for MoE expert weights by matching against the last path component (e.g. `down_proj`) rather than full expert-qualified names, and it is now invoked from `transformers_v5_compat()` so it runs in all vLLM processes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f99fd82f777af2615469d6c9b417ec0429a87408. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/rl/broadcast/filesystem.py
+++ b/src/prime_rl/trainer/rl/broadcast/filesystem.py
@@ -59,12 +59,16 @@ class FileSystemWeightBroadcast(WeightBroadcast):
             if adapter_only:
                 # For adapter-only, MultiRunManager creates state dict directly for each run
                 # All ranks must participate in DTensor gathering, but only master saves
-                state_dict = self.multi_run_manager.get_state_dict_for_run(idx)
-                for key, value in state_dict.items():
+                raw_state_dict = self.multi_run_manager.get_state_dict_for_run(idx)
+                state_dict = {}
+                for key, value in raw_state_dict.items():
                     if isinstance(value, DTensor):
                         value = value.full_tensor()
                     if self.world.is_master:
-                        state_dict[key] = value.to("cpu", non_blocking=False)
+                        # Prepend base_model.model. prefix for PEFT/vLLM compatibility.
+                        # vLLM's parse_fine_tuned_lora_name() expects this prefix to
+                        # correctly strip it and match module names against target_modules.
+                        state_dict[f"base_model.model.{key}"] = value.to("cpu", non_blocking=False)
 
             # TODO: Broadcast ready to update in sync, then we dont need to gather on not ready
             if self.world.is_master:
--- a/src/prime_rl/trainer/rl/broadcast/filesystem.py
+++ b/src/prime_rl/trainer/rl/broadcast/filesystem.py
@@ -59,16 +59,12 @@ class FileSystemWeightBroadcast(WeightBroadcast):
             if adapter_only:
                 # For adapter-only, MultiRunManager creates state dict directly for each run
                 # All ranks must participate in DTensor gathering, but only master saves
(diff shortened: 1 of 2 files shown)


JannikSt:
@codex review

Jackmin801 (commented):
do we need both changes? would think you either
1. patch on vllm side
2. rewrite state dict key on trainer side

And either one would work. Oh but i guess we need the patch for BC reasons.

src/prime_rl/trainer/rl/broadcast/filesystem.py
```diff
@@ -59,12 +59,16 @@ def broadcast_weights(self, model: nn.Module, step: int) -> None:
...
+                state_dict = {}
+                for key, value in raw_state_dict.items():
                     if isinstance(value, DTensor):
                         value = value.full_tensor()
                     if self.world.is_master:
-                        state_dict[key] = value.to("cpu", non_blocking=False)
+                        # Prepend base_model.model. prefix for PEFT/vLLM compatibility.
+                        # vLLM's parse_fine_tuned_lora_name() expects this prefix to
+                        # correctly strip it and match module names against target_modules.
+                        state_dict[f"base_model.model.{key}"] = value.to("cpu", non_blocking=False)
```
Jackmin801: I think this might duplicate the weights so we end up with both keys in the sate dict. 

Also, I would think that a change to multi_run_manager / multi lora layers is the better way to do it. So make it that `self.multi_run_manager.get_state_dict_for_run(idx)` already returns the right keys and you dont need to do this key fix here.

samsja (approved):

PR #2246, opened 2026-04-10
feat: add num_epochs config to SFT trainer

hallerite:
## Summary

- Adds a `num_epochs` field to `SFTConfig` that stops training after N full passes over the dataset
- Can be combined with `max_steps` — whichever limit is reached first wins
- Wires up the existing `max_epochs` support already present in `SFTDataset` (no data layer changes needed)

### Packing caveat

Because `CatDataset`/`StackDataset` pack tokens into fixed-length chunks, epoch boundaries fall mid-chunk. This means the total step count per epoch can vary by ±1 compared to a naive `num_epochs × steps_per_epoch` estimate. For example, with `PrimeIntellect/Reverse-Text-SFT` (1000 examples), `batch_size=2`, `seq_len=1024`:

| num_epochs | steps | steps/epoch |
|------------|-------|-------------|
| 1          | 45    | 45          |
| 2          | 91    | 45.5        |

This is inherent to token packing — the partial chunk at the epoch boundary gets filled with tokens from the next epoch. A warning log is emitted at training start when `num_epochs` is set.

### Usage

```toml
num_epochs = 2

[model]

[data]
name = "PrimeIntellect/Reverse-Text-SFT"
batch_size = 4
seq_len = 1024
```

Or via CLI: `--num_epochs 2`

---

> [!NOTE]
> **Medium Risk**
> Adds a new early-stop path that pre-tokenizes data and uses a distributed `all_reduce` to derive `max_steps`, which can affect training length and introduces startup-time compute/sync that must be correct across ranks.
> 
> **Overview**
> Adds a new `single_epoch` flag to `SFTConfig` to stop SFT training after one full dataset pass.
> 
> Implements this by adding `SFTDataset.count_tokens()` and, at trainer startup, pre-tokenizing each rank’s shard to estimate how many packed micro-batches it can produce; the trainer then `all_reduce`s the minimum across ranks and sets `config.max_steps` accordingly (only when it’s smaller or unset).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb0e87bf2010a167e23c370dc9627e3460d27e67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/configs/sft.py
+++ b/src/prime_rl/configs/sft.py
@@ -211,11 +211,9 @@ class SFTConfig(BaseConfig):
     max_epochs: Annotated[
         int | None,
         Field(
-            description="Maximum number of epochs to train for. Training stops when the dataset has been exhausted this many times. "
-            "Can be combined with max_steps — whichever limit is reached first wins. "
-            "If both are None, training runs indefinitely. "
-            "Note: with token packing (pack_function='cat' or 'stack'), epoch boundaries fall mid-chunk, "
-            "so the total step count may vary by ±1 compared to a naive (max_epochs * steps_per_epoch) estimate."
+            description="Maximum number of epochs to train for. Converted to max_steps at startup by pre-tokenizing "
+            "the dataset and synchronizing across ranks. Can be combined with max_steps — whichever limit is "
+            "reached first wins. If both are None, training runs indefinitely."
         ),
     ] = None
 
--- a/src/prime_rl/configs/sft.py
+++ b/src/prime_rl/configs/sft.py
@@ -208,14 +208,14 @@ class SFTConfig(BaseConfig):
         Field(description="Maximum number of steps to run training for. If None, will run indefinitely."),
     ] = None
 
-    num_epochs: Annotated[
+    max_epochs: Annotated[
         int | None,
         Field(
-            description="Number of epochs to train for. Training stops when the dataset has been exhausted this many times. "
+            description="Maximum number of epochs to train for. Training stops when the dataset has been exhausted this many times. "
             "Can be combined with max_steps — whichever limit is reached first wins. "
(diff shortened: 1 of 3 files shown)


src/prime_rl/configs/sft.py
```diff
@@ -208,6 +208,17 @@ class SFTConfig(BaseConfig):
         Field(description="Maximum number of steps to run training for. If None, will run indefinitely."),
     ] = None
 
+    num_epochs: Annotated[
```
mikasenghaas: nice, can we rename to `max_epochs` for consistency? also makes clear that max epochs may not be reached if max steps is exhausted first

src/prime_rl/trainer/sft/train.py
```diff
@@ -334,7 +341,22 @@ def run_validation(step: int) -> None:
...
+            # when num_epochs is reached. Because packing operates on a continuous token
+            # stream, different ranks may exhaust at different micro-steps. We must
+            # synchronize before any collective ops (FSDP forward/backward) to avoid hangs.
+            try:
+                micro_batch = next(dataiter)
+            except StopIteration:
+                micro_batch = None
+
+            if config.num_epochs is not None:
+                exhausted = torch.tensor(float(micro_batch is None), device="cuda")
```
mikasenghaas: each rank has the global information and _could_ look ahead to see if any rank is exhausting the dataset which would avoid the dist op

not sure if it's worth the hassle tho

mikasenghaas (commented):
if tested, lgtm. but would rename the config arg bc we cannot guarantee num epochs exactly

src/prime_rl/configs/sft.py
hallerite: good point

PR #2266, opened 2026-04-13
feat: add config to disable prefix cache reset after weight updates

samsja:
## Summary
- Adds `inference.experimental.reset_prefix_cache_after_update` (default: `false`) — optionally reset the entire prefix cache after weight updates for strict KV correctness at the cost of extra prefill
- Adds `orchestrator.experimental.use_prefix_cache_salt` (default: `true`) — sets a `cache_salt` on inference requests that changes with each weight update, partitioning the prefix cache by policy version without resetting it
- Removes the unconditional `reset_prefix_cache()` calls from `/update_weights` and `/load_lora_adapter` server endpoints

## Usage
```toml
# Salt is on by default — partitions prefix cache by policy version
# Off-policy rollouts keep their cached KV, new rollouts get fresh computation
[orchestrator.experimental]
use_prefix_cache_salt = true  # default

# Full cache reset is off by default — enable for strict correctness
[inference.experimental]
reset_prefix_cache_after_update = false  # default
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes how prefix-cache KV state is invalidated across weight/LoRA updates by defaulting away from full cache resets and introducing per-checkpoint cache salting; misconfiguration could lead to stale KV reuse or throughput regressions.
> 
> **Overview**
> Adds new experimental knobs to control prefix-cache invalidation after policy updates: `inference.experimental.reset_prefix_cache_after_update` (default `false`) and `orchestrator.experimental.use_prefix_cache_salt` (default `true`).
> 
> Inference server `/update_weights` and `/load_lora_adapter` now **only** call `reset_prefix_cache()` when configured, while the orchestrator/scheduler can attach a per-checkpoint `cache_salt` to rollout/eval sampling args so new policies avoid reusing prefix-cached KV from older weights without flushing the entire cache.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d1e4cca98ede2e1e7b0e427a7e6266e11a9bf0b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2270, opened 2026-04-14
[feat] use cache salt in request

Jackmin801:
This removes the need to reset kv cache to prevent overly stale cache

we currently reset the kv cache on all inference engines after a policy update. This is because we had a bug that was caught in nightly a few moons ago where the kl mismatch would go crazy after a few hundred steps because the system prompts kv cache would never be reset: https://discord.com/channels/1383922374964936786/1464376927459606681/1464438352303362149

But this causes quite some prefill work everytime we update the policy. So we should instead tag rollouts with the step they started at and make it that requests from different steps dont share kv cache.

That way, it gets refreshed for new rollouts. But running rollouts can keep using the old kv cache they had

This tagging is exposed as cache_salt in vllm. The original intention was to prevent users from being able to use each others prefix cache and be exposed to timing attacks

---

> [!NOTE]
> **Medium Risk**
> Changes how KV/prefix cache is invalidated across policy updates by relying on `cache_salt` propagation rather than server-side cache resets, which could affect rollout correctness and performance if not applied consistently.
> 
> **Overview**
> Switches cache invalidation for policy/adapter updates from server-side prefix-cache resets to **per-request KV cache segregation** using vLLM’s `cache_salt`.
> 
> The orchestrator now threads an optional `cache_salt` (set to the current `ckpt_step` when policy updates are enabled) through `Env.run_rollout`/`run_group` by injecting it into `sampling_args.extra_body`, so rollouts from different checkpoint steps don’t share prefix/KV cache.
> 
> On the inference API side, the `/update_weights` and `/load_lora_adapter` endpoints no longer call `reset_prefix_cache`; the LoRA endpoint docs are updated to note that cache invalidation is handled client-side via `cache_salt`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba1ef18c2491be11db3db770a047208142e57ec4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/inference/vllm/server.py
+++ b/src/prime_rl/inference/vllm/server.py
@@ -195,23 +195,20 @@ async def resume(request: Request):
 async def update_weights(request: Request):
     data = await request.json()
     await engine_client(request).collective_rpc("update_weights_from_path", args=(data.get("weight_dir"),))
-    await engine_client(request).reset_prefix_cache()
     return {"status": "ok"}
 
 
 @router.post("/load_lora_adapter")
 async def load_lora_adapter(lora_request: LoadLoRAAdapterRequest, raw_request: Request):
-    """Load a LoRA adapter and reset the prefix cache.
+    """Load a LoRA adapter.
 
-    Wrapper around vLLM's /v1/load_lora_adapter that also resets the prefix cache
-    to invalidate KV states computed with old weights.
+    Wrapper around vLLM's /v1/load_lora_adapter. KV cache invalidation for old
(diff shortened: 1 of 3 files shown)


mikasenghaas (commented):
code lgtm! would send a `hendrycks-sanity` run for 5k steps on the cluster to make sure we dont have any regression re stale caches somehow

Jackmin801:
Done in https://github.com/PrimeIntellect-ai/prime-rl/pull/2266

PR #2271, opened 2026-04-14
feat: per-batch loss type override for hosted SFT hard distill

eligotts:
Add loss_type field to TrainingSample and MicroBatch transport types so the orchestrator can stamp samples with a loss override (e.g. "sft") that the trainer dispatches per-batch. This allows a shared multi-tenant trainer to handle both RL and SFT distill runs on the same cluster.

When teacher_rollout_model is configured, the orchestrator sets loss_type="sft" on each sample. The trainer's compute_loss checks the field and swaps to sft_loss_fn for that batch, falling back to the configured default when None (fully backward compatible).

---

> [!NOTE]
> **Medium Risk**
> Changes training sample transport and loss dispatch to allow per-batch switching to `sft_loss_fn`, which can alter optimization behavior and packing semantics if misconfigured. Adds stricter config validation around external teacher rollout mode that may break existing configs until updated.
> 
> **Overview**
> Enables hosted hard-distill runs to **force SFT loss per micro-batch** via a new `orchestrator.use_sft_loss` flag, with the orchestrator stamping `TrainingSample.sft_loss` and the trainer dispatching `compute_loss` to `sft_loss_fn` when set.
> 
> Tightens config validation by requiring `orchestrator.teacher_rollout_model` and `orchestrator.use_sft_loss` to be configured together (and `use_token_client=false` in that mode), and removes the old coupling to `trainer.loss.type="sft"`.
> 
> Propagates the new `sft_loss` field through `TrainingSample`/`MicroBatch`, prevents packing sequences with mixed `sft_loss` values into the same micro-batch, and updates docs, example config, changelog, and unit tests accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a56ae24b7f01f31d2c006fb913ff09f19e100f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/batch.py
+++ b/src/prime_rl/trainer/batch.py
@@ -76,7 +76,7 @@ def prepare_sample(training_example: TrainingSample, seq_len: int) -> MicroBatch
         pixel_values=training_example.pixel_values,
         pixel_values_shape=training_example.pixel_values_shape,
         image_grid_thw=training_example.image_grid_thw,
-        loss_type=training_example.loss_type,
+        sft_loss=training_example.sft_loss,
     )
 
 
--- a/src/prime_rl/trainer/batch.py
+++ b/src/prime_rl/trainer/batch.py
@@ -76,6 +76,7 @@ def prepare_sample(training_example: TrainingSample, seq_len: int) -> MicroBatch
         pixel_values=training_example.pixel_values,
         pixel_values_shape=training_example.pixel_values_shape,
         image_grid_thw=training_example.image_grid_thw,
+        loss_type=training_example.loss_type,
     )
 
 
--- a/src/prime_rl/trainer/batch.py
+++ b/src/prime_rl/trainer/batch.py
@@ -116,7 +116,10 @@ def packed_samples_into_micro_bs(
             if _is_multimodal_sample(bin_content):
                 continue
             # Check if sequence fits in this bin
-            if len(bin_content.input_ids) + len(sample.input_ids) <= max_seq_len:
+            if (
+                len(bin_content.input_ids) + len(sample.input_ids) <= max_seq_len
+                and bin_content.sft_loss == sample.sft_loss
+            ):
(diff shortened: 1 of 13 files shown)


samsja (commented):
would actually turn it into a boolean instead of string. "need_sft_loss"

Jackmin801 (approved):
wondering if since you define loss type as string that it should just use string throughout instead of having this is_sft bool. would make future extension to more loss functions simpler to add as well. fine to merge as is to get sft out there though

samsja:
> wondering if since you define loss type as string that it should just use string throughout instead of having this is_sft bool. would make future extension to more loss functions simpler to add as well. fine to merge as is to get sft out there though

would actually turn it into a boolean instead of string. "need_sft_loss" lets do bool instead no need fors tring here

samsja (changes requested):
lets move to bool instead of string as mention above

src/prime_rl/trainer/batch.py
```diff
@@ -115,7 +116,10 @@ def packed_samples_into_micro_bs(
             if _is_multimodal_sample(bin_content):
                 continue
             # Check if sequence fits in this bin
-            if len(bin_content.input_ids) + len(sample.input_ids) <= max_seq_len:
+            if (
+                len(bin_content.input_ids) + len(sample.input_ids) <= max_seq_len
+                and bin_content.sft_loss == sample.sft_loss
```
samsja, line 121: what happen if this is not the case we just skip silently ? isn't an issue ?

eligotts: then we just create a new micro batch? we shouldn't ever hit this in normal single tenant runs (mixed loss types) but in hosted could we have the situation where trainer is processing in same step samples from different runs with different loss types?

samsja (approved):

PR #2278, opened 2026-04-14
feat: per-model client-side renderers for RL rollouts (+ /v1/generate)

hallerite:
## Summary

Adds **client-side renderer tokenization** for RL rollouts plus a new vLLM **`POST /v1/generate`** (tokens-in / tokens-out) endpoint. Training samples, between-turn bridges, and teacher-logprob prefill go through deterministic per-model Python renderers from the [`renderers` package](https://github.com/PrimeIntellect-ai/verifiers/tree/renderers), so the trainer sees the exact token ids the sampler saw.

**Full design + concrete examples of the failure modes this fixes**: [`packages/renderers/README.md`](https://github.com/PrimeIntellect-ai/verifiers/blob/renderers/packages/renderers/README.md) (verifiers repo, §4 "Breaking behaviors you might not expect" is the motivating part).

### Renderer matrix

Qwen3 / Qwen3.5 / Qwen3-VL · GLM-5 / GLM-5.1 / GLM-4.5 · Kimi K2 / K2.5 / K2.6 · Nemotron 3 (Nano / Super) · DeepSeek V3 · GPT-OSS · MiniMax-M2 · `DefaultRenderer` fallback. Render parity (`render_ids` == `apply_chat_template`) and render → parse round-trip are enforced by a 15-model test matrix in the verifiers package.

### New config knobs (`model.*`, propagate to trainer/orchestrator/inference)

- `renderer` — `auto` (detect from tokenizer `name_or_path` by exact match) or an explicit renderer name.
- `tool_parser` / `reasoning_parser` — pluggable parsers consumed by `DefaultRenderer` (e.g. `qwen3`, `qwen3.5`, `glm`, `deepseek_v3`, `think`). Model-specific renderers bake their own parsing in.
- `renderer_pool_size` — concurrent renderer slots for long multi-turn prompts.

### Orchestrator routing policy

Driven by two mutually-exclusive `OrchestratorConfig` toggles (`use_renderer`, `use_token_client`); config-level validators reject invalid combinations.

- **External teacher rollout** → MITO (`openai_chat_completions`), no renderer; toggles forced off.
- **`use_renderer=True`** → renderer client over `/v1/generate` (TITO via tokens-in / tokens-out). Not allowed for VLMs.
- **`use_token_client=True`** → legacy TITO via `/v1/chat/completions/tokens` (kept for back-compat; default for VLMs since they need server-side chat templating + image preprocessing).
- **Both False** → MITO (`openai_chat_completions`).

The renderer path (`use_renderer=True`) is the new recommended path for plain LMs; `use_token_client` and the `/v1/chat/completions/tokens` endpoint stay supported for existing pipelines and as the VLM fallback.

### Correctness fixes that landed on the way

- **Hand-coded `bridge_to_next_turn` per renderer** (ChatML / GLM / harmony / Kimi / DeepSeek / Nemotron / MiniMax) — replaces the earlier shared `chatml_bridge` / `glm_bridge` dummy-assistant tricks. Each renderer's bridge reuses the same `_render_tool` / per-role emit helpers that `render()` uses, so the two paths can never silently diverge when a template does something position-dependent (GLM-5.1 last-assistant thinking wrap, harmony channel selection, Kimi auto-system). `DefaultRenderer`'s bridge returns `None` and falls back to a fresh render.
- **`samples_per_rollout` regression on multi-turn RL**: the incremental-prompt anchor skipped truncated steps, so `synthesize_close_on_truncation` never fired. Fix + the per-renderer synth-close brought wordle from `~1.36` back to `1.00`, matching main. Hand-coded renderers default `synthesize_close_on_truncation = True`.
- **Hermes-parser fallback on malformed tool-call JSON** (verifiers `dba2e97`) — `parse_qwen3` returns the raw completion as content when `<tool_call>` JSON fails to decode, matching vLLM's `hermes_tool_parser`. Prevents `EmptyModelResponseError` cascades on hermes tool envs with untrained models.
- **FA3 kernel kwarg rename** (PR #2352, cherry-picked into this branch): `flash_attn_3` 3.0.0 renamed `causal` → `is_causal` and split `window_size` into `_left` / `_right`. The FA3 ring-attn wrappers in `ring_attn.py` still passed the old keys, crashing the first backward on any `cp>1` + FA3 + `impl=custom` config (hit live on Qwen3.5-35B-A3B + mini-swe-agent-plus).
- **BPE boundary bugs** in Nemotron3 / MiniMax tool-call and consecutive tool-response emission (split `emit_text` calls were fragmenting merges vs. the Jinja single-pass output).
- **Index translation** when renderers auto-inject system messages (Nemotron3 empty system, Kimi default system) — keeps `build_supervised_sample` masks aligned with the caller's original message list.
- Teacher logprob prefill switched from `/v1/chat/completions/tokens` to `/v1/generate`.

### Removed

- `OpenAIServingChatWithRoutedExperts` (merged into `serving_generate.py`).
- `build_incremental_prompt_ids` + `bridges.py` dummy-assistant helpers (per-renderer `bridge_to_next_turn` replaces them).

`use_token_client` and `OpenAIServingChatWithTokens` are **kept for back-compat**; the new renderer path is opt-in via `use_renderer=True`.

### Verifiers dependency

Pinned to [`c58da4a`](https://github.com/PrimeIntellect-ai/verifiers/commit/c58da4a) on the `renderers` branch (PR [PrimeIntellect-ai/verifiers#1068](https://github.com/PrimeIntellect-ai/verifiers/pull/1068)). Includes the parse_qwen3 hermes fallback, break-category telemetry, and the expanded package README.

### Risk / test plan

- Hard-distill is the supported path for renderer rollouts (docs updated).
- Unit coverage: `tests/unit/inference/test_serving_generate.py`, `tests/unit/orchestrator/test_orchestrator_setup.py`, `tests/unit/orchestrator/test_teacher_logprobs.py`, `tests/unit/utils/test_client.py`, `tests/unit/utils/test_elastic.py`.
- E2E validated:
  - **wordle** multi-turn RL (`wordle-renderers-default-synth` vs main's `wordle-main-200-retry`): `samples_per_rollout` matches at 1.00, reward / KL curves track main.
  - **opencode-math + opencode-science** on GLM-4.5 Air: ~8 breaks/step (floor is the opencode AI-SDK `invalid`-rewrite, not fixable client-side).
  - **mini-swe-agent-plus** on Qwen/Qwen3.5-35B-A3B (cp=2, FA3, custom impl): 0 extension breaks per step vs. main's 32 in the same step. Forward + backward + NCCL weight broadcast clean across multiple steps. See verifiers README §4 for the concrete break modes this comparison surfaces.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new inference endpoint and a new rollout client mode that changes how rollouts are tokenized and how logprobs are fetched, which can affect training correctness and throughput if misconfigured. Validators and added unit tests reduce risk, but the change touches orchestrator/inference integration paths.
> 
> **Overview**
> Enables an opt-in *client-side renderer* rollout path that tokenizes and parses trajectories via the `renderers` package, and routes rollout inference to a new vLLM `POST /v1/generate` endpoint (tokens-in/tokens-out) while keeping existing TITO (`/v1/chat/completions/tokens`) and MITO (`/v1/chat/completions`) modes.
> 
> Adds `RendererConfig` and orchestrator toggles (`use_renderer`, updated `use_token_client`) with validators enforcing mutual exclusivity, disallowing renderer mode for VLMs and external teacher rollouts, and rejecting renderer-specific knobs unless enabled. Orchestrator initialization now chooses the correct client mode via `setup_rollout_inference_pool`, passes renderer metadata through `setup_inference_pool`/static+elastic client construction, and uses the renderer for (re)tokenizing trajectories when tokens are missing.
> 
> Inference server wiring is extended to register `/v1/generate` with a new `OpenAIServingGenerate` handler that forwards LoRA/trace/DP-rank metadata, computes default `max_tokens` from model context, and returns token IDs + logprobs (including prompt logprobs); teacher-logprob prefill is switched to call `/generate` instead of the token chat endpoint. Trainer VLM forward pass now derives `mm_token_type_ids` automatically for Qwen3-VL inputs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ece39ff4d68ed07be10f490e535952b4344b155. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/configs/orchestrator.py
```diff
@@ -1057,10 +1057,25 @@ class OrchestratorConfig(BaseConfig):
     use_token_client: Annotated[
```
mikasenghaas, line 1061: should we not make this a literal config now? seems more future proof?

src/prime_rl/configs/orchestrator.py
```diff
@@ -1057,10 +1057,25 @@ class OrchestratorConfig(BaseConfig):
...
+        bool,
+        Field(
+            description="Whether to use the renderer client (client-side tokenization via the ``renderers`` package, "
+            "served by ``/v1/generate``). Mutually exclusive with ``use_token_client``. When True, the "
+            "``model.renderer`` / ``model.tool_parser`` / ``model.reasoning_parser`` / "
+            "``model.renderer_pool_size`` knobs apply; when False they must be left at their defaults. "
+            "Not supported for VLMs — VLMs must use the token client (TITO) so image preprocessing and chat "
+            "templating stay server-side."
+        ),
+    ] = False
```
samsja, line 1080: can we use tito and renderes at the same time ? should we unify these while keeping backward  compatabiltiy with tito args somehow ?

src/prime_rl/configs/orchestrator.py
mikasenghaas: and avoids all this mutually exclusive ugliness

src/prime_rl/configs/shared.py
```diff
@@ -124,6 +124,51 @@ class BaseModelConfig(BaseConfig):
...
+        int | None,
+        Field(
+            ge=1,
+            description=(
+                "Number of renderer slots shared across concurrent rollouts. "
+                "None keeps the verifiers default (1). Bump for long multi-turn "
+                "prompts where client-side jinja tokenization serializes."
+            ),
+        ),
+    ] = None
```
samsja: I dont think this belongs to shared model info tbh. cant we have a chat tempalte sub config or smth ?

src/prime_rl/inference/vllm/server.py
```diff
@@ -178,10 +178,57 @@ def models(request: Request) -> OpenAIServingModels:
 }
 
 
+def generate_handler(request: Request):
```
samsja, line 184: do we need this function ?

src/prime_rl/utils/client.py
```diff
@@ -62,10 +62,23 @@ def __init__(
         self,
         client_config: ClientConfig,
         model_name: str,
-        train_client_type: str = "openai_chat_completions_token",
+        train_client_type: str = "renderer",
```
samsja: why rendere by default here btw ?

src/prime_rl/inference/vllm/serving_generate.py
```diff
@@ -0,0 +1,225 @@
...
+    prompt_token_ids: list[int]
+    choices: list[GenerateChoiceResponse]
+    usage: dict
+    prompt_logprobs: list[float | None] | None = None
+
+
+# ── Handler ──────────────────────────────────────────────────────────
+
+
+class OpenAIServingGenerate:
```
mikasenghaas, line 86: hm i thought this was built-in?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -479,33 +480,53 @@ async def orchestrate(config: OrchestratorConfig):
         # Convert rollouts to training samples
         parallel_preprocess_start = time.perf_counter()
 
-        # Pretokenize before VLM image cache build (which strips image data from messages)
-        for rollout in train_rollouts:
-            pretokenize_rollout_trajectory(rollout, tokenizer, processor=processor)
+        # Stage 1: pretokenize + (for VLM) build image cache concurrently.
```
mikasenghaas, line 484: im confused why this is on the orch. i thought tokenization is on the env workers

src/prime_rl/inference/vllm/server.py
hallerite: it's just FastAPI-style Dependency Injection, same as `chat_with_tokens()`

src/prime_rl/configs/shared.py
hallerite: yeah, we should have a renderer subconfig

src/prime_rl/configs/orchestrator.py
hallerite: breaks hosted training, no?

src/prime_rl/inference/vllm/serving_generate.py
samsja: yeah also wondering

src/prime_rl/configs/orchestrator.py
hallerite: this is not about TITO vs Renderers, but how we achieve TITO, i.e. whether to use the old token client or the new renderers. both are TITO

src/prime_rl/orchestrator/orchestrator.py
hallerite: this is for the non renderer paths

src/prime_rl/configs/orchestrator.py
mikasenghaas: GODDAMN, yea after dogfooding phase we should defo break hosted tho. cant be keeping backwards compat like this all the time

mikasenghaas (approved):

samsja (approved):

PR #2288, opened 2026-04-15
fix: log inference metrics on time axis instead of step

samsja:
## Summary

- Fix inference metrics being plotted against training step instead of wall time
- The wildcard `define_metric("*", step_metric="step")` in WandbMonitor was overriding the inference-specific pattern
- Use `commit=False` on `wandb.log` to avoid advancing the global step counter
- Explicitly define `_timestamp` as a hidden metric so W&B uses it as the x-axis

### DIFF
--- a/src/prime_rl/orchestrator/inference_metrics.py
+++ b/src/prime_rl/orchestrator/inference_metrics.py
@@ -4,10 +4,10 @@ import asyncio
 import time
 from collections import deque
 
+import wandb
 from httpx import AsyncClient
 from prometheus_client.parser import text_string_to_metric_families
 
-import wandb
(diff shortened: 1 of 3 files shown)


vominh1919:
Hi! I'd like to help with this issue.

I can analyze the problem and create a fix. Here's my approach:

1. **Understand the issue** - Read the description and reproduce the bug
2. **Find the root cause** - Locate the problematic code
3. **Implement a fix** - Create a minimal, targeted solution
4. **Test the fix** - Verify it works as expected
5. **Submit a PR** - Ready for your review

Would you like me to proceed?

---
*Offering help with open issues to improve the project.*

PR #2205, opened 2026-04-06
Fix Prime monitor public API flow

d42me:
## Summary
Fix the remaining Prime monitor public API mismatches by keeping bearer auth for uploads, normalizing `/samples/presign` responses that come back as nested camelCase data, avoiding the duplicate completion update after a successful `/finalize`, and dropping non-finite metric/distribution values before JSON upload so monitor requests do not fail on `NaN`. The payload normalization helpers were also simplified into small pure helpers to keep the monitor flow easier to reason about.

## Verification
- Local synthetic E2E on macOS against a temporary server that exercised registration, metrics, distributions, samples presign/upload/confirm, and finalize with the public `/api/v1/rft` response shape
- Local synthetic E2E on macOS that verified `NaN` values are dropped from metrics, distributions, and final summary payloads before upload
- Synthetic E2E rerun on the RTX 6000 machine against a temporary local server there; registration, metrics, distributions, samples presign/upload/confirm, and finalize all succeeded on the target machine
- Direct production auth check on the RTX 6000 machine with the provided token: `x-api-key` returned 403 Not authenticated and bearer/both returned 401 Invalid token, so full production validation is currently blocked by credentials rather than code
- `VIRTUAL_ENV=$PWD/.venv PATH=$PWD/.venv/bin:$PATH uv run --active --no-project ruff check src/prime_rl/utils/monitor/prime.py`

---

> [!NOTE]
> **Medium Risk**
> Changes request headers, endpoints, and finalize flow for PrimeMonitor uploads; mistakes could break run registration, sample uploads, or completion status updates. Adds sanitization that drops `NaN`/`inf` values, which changes what data is persisted if metrics are non-finite.
> 
> **Overview**
> Fixes Prime public API mismatches in `PrimeMonitor` by standardizing request headers (Bearer + `x-api-key`) and using consistent base URL paths for run registration/status updates and sample presign/confirm calls.
> 
> Normalizes the `/samples/presign` response shape (nested `data` with camelCase keys) and updates finalize behavior to avoid a redundant status update when `/finalize` succeeds.
> 
> Adds JSON payload sanitization to **drop non-finite floats** (`NaN`/`inf`) before uploads (metrics, distributions, finalize summary), logging the dropped field paths; includes a unit test covering the sanitization behavior.
> 
[...]

PR #2315, opened 2026-04-18
fix: complete Prime monitor v1 finalize and samples flow

d42me:
Builds on the bearer-auth fix by normalizing public v1 presign responses and avoiding duplicate completion updates so PrimeMonitor works end to end with the production monitoring API.


d42me:
Superseded by the follow-up pushed onto #2205.

PR #2328, opened 2026-04-19
feat: replace max_async_level with on_policy debug flag

samsja:
## Summary

- Removes `max_async_level` and `strict_async_level` from the orchestrator/trainer/rl configs and from the scheduler.
- Async training now always serves rollouts from the latest available policy (no upper bound on the step gap between trainer and inference). `max_off_policy_steps` continues to bound rollout staleness.
- Adds an `on_policy: bool` flag (default `false`) on the orchestrator, trainer, and rl configs. When `true`, the orchestrator blocks until the trainer has produced a checkpoint for the current step, i.e. fully synchronous on-policy RL. This is **debug-only**: it will be significantly slower than async training in this setup.
- NCCL weight broadcast validation now checks `on_policy=false` instead of `max_async_level=1`.
- Filesystem weight-broadcast cleanup now keeps the two most recent broadcasts unconditionally (previously computed from `max_async_level`).
- Updates `docs/async.md` and removes `max_async_level` from debug and CI configs.

---

> [!NOTE]
> **Medium Risk**
> Changes rollout scheduling and checkpoint/broadcast cleanup semantics in the RL training loop, which can affect throughput and correctness of policy updates despite being conceptually straightforward.
> 
> **Overview**
> **Removes the `max_async_level`/`strict_async_level` configuration knobs** across RL (`trainer`, `orchestrator`, and shared `rl` config) and deletes the associated cross-config validation.
> 
> **Updates the orchestrator scheduler to a fixed async policy**: it now serves rollouts from the latest available checkpoint, caps itself to at most *one step ahead* of the trainer, and relies on `max_off_policy_steps` to drop overly stale rollouts.
> 
> **Adds `orchestrator.on_policy` (debug-only)** to force fully synchronous on-policy RL by blocking rollout generation until the trainer checkpoint for the current step is available, and adjusts log messaging/docs and debug/CI configs accordingly.
> 
> **Simplifies weight broadcast tail-step and cleanup behavior**: trainer broadcast skipping is now based on the final step (not async level), and filesystem broadcast cleanup keeps the two most recent broadcasts unconditionally.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c7817bda59e55cd33998530cc6caebae0c8a0c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mikasenghaas (commented):
nice, love it. would check tests and allow nccl weight broadcast with `on_policy=True` as well

src/prime_rl/configs/rl.py
```diff
@@ -330,10 +330,10 @@ class RLConfig(BaseConfig):
         ),
     ] = None
 
-    max_async_level: Annotated[
-        int | None,
+    on_policy: Annotated[
```
mikasenghaas: i think we can remove this from the shared  config? the trainer doesn't need to know about the async level anymore. thus far, it needed to to know at which point it can start cleaning broadcast checkpoints. since now async level is implicitly <=1, we are fine with cleaning dirs >=2 steps away

src/prime_rl/configs/rl.py
```diff
@@ -608,13 +608,13 @@ def auto_setup_max_steps(self):
         return self
 
     @model_validator(mode="after")
-    def auto_setup_async_level(self):
-        """Auto-setup shared async level for trainer and orchestrator."""
-        if self.max_async_level is not None:
-            self.trainer.max_async_level = self.max_async_level
-            self.orchestrator.max_async_level = self.max_async_level
+    def auto_setup_on_policy(self):
```
mikasenghaas: can remove this as well

src/prime_rl/configs/trainer.py
```diff
@@ -784,13 +784,12 @@ class TrainerConfig(BaseConfig):
         ),
     ] = None
 
-    max_async_level: Annotated[
-        int,
+    on_policy: Annotated[
```
mikasenghaas: can remove

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -274,17 +272,15 @@ async def update_policy_loop(self):
 
     def _compute_next_ckpt_step(self) -> int:
         latest_ckpt_step = get_latest_ckpt_step(get_broadcast_dir(self.config.output_dir)) or 0
-        async_away_ckpt_step = max(self.step - self.max_async_level, 0)
-        if self.strict_async_level:
-            return async_away_ckpt_step
-        return max(async_away_ckpt_step, latest_ckpt_step)
+        if self.on_policy:
```
mikasenghaas, line 275: we still need the `async_away_ckpt_step` but hardcoded with `max_async_level=1`. otherwise the orchestrator will race away from the trainer

src/prime_rl/utils/validation.py
```diff
@@ -94,13 +94,13 @@ def validate_shared_max_steps(
         )
 
 
-def validate_shared_max_async_level(
+def validate_shared_on_policy(
```
mikasenghaas: can remove

src/prime_rl/configs/orchestrator.py
```diff
@@ -1136,7 +1121,6 @@ def resolve_batching(self):
     def auto_setup_bench(self):
         if self.bench:
             self.max_steps = 4  # Run for 1 warmup step + 3 evaluation steps
-            self.max_async_level = int(1e9)  # Never wait for RL weight checkpoints
```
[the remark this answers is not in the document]
mikasenghaas: oh this is valid i think. if we hardcode async level 1 then the benchmark mode will stop at the async barrier of 1. maybe we can circumvent this by setting `enable_policy_updates=False`

PR #2332, opened 2026-04-19
Add orchestrator-side replay buffer support for RL training

### LINKED ISSUE #2273: Replay buffer support
Add support for a replay buffer. Naive implementation would be random sampling from a limited size buffer.
Can support priority sampling based on absolute advantage, and off policyness

taivu1998:
## Summary

Implements replay buffer support for RL training in the orchestrator, with replay-aware batching, checkpoint persistence, metrics, and focused unit coverage.

Closes #2273.

## What Changed

- added a new orchestrator-side replay subsystem that stores immutable rollout groups and samples them back into future training batches
- added replay configuration for capacity, replay fraction, replay sampling seed, and replay-specific off-policy age limits
- persisted replay state in orchestrator checkpoints and added `skip_replay` so replay restore can be disabled independently of the main buffer
- captured `policy_step` at rollout schedule time so replay age and off-policy accounting use the policy that actually generated the rollout
- mixed replayed groups into the train batch without changing trainer transport, while keeping the current fresh-buffer path intact
- updated per-step metrics to use `problem_idx` instead of assuming `example_id` is unique within a step, and added replay-specific progress/age/utilization metrics
- added focused tests for replay target calculation, replay immutability, replay checkpoint round-tripping, `skip_replay`, scheduler `policy_step` capture, and replay config validation
- updated the config skill to document the new `[orchestrator.replay]` surface and checkpoint skip behavior

## Design Notes

This keeps replay as an orchestrator-side feature rather than turning the current transient rollout queue into a random bounded buffer or widening trainer transport.

Key design choices:

- replay stores whole rollout groups so problem-level grouping and advantage semantics stay aligned with the existing `rollouts_per_example` contract
- replay entries are deep-copied on insert and on sample so later in-place rollout mutation, especially on the VLM path, does not corrupt stored replay state
- replay is sampled before fresh generation so the fresh target is based on the actual replay payload selected for the step rather than only an estimate
- replay age is tracked from the rollout's schedule-time `policy_step`, not the completion step, which matches the actual generating policy

## Validation

Validated in the isolated worktree with:

- `python3 -m py_compile` on all changed Python files
- `git diff --check`
- dependency-free runtime harnesses covering replay logic, replay checkpoint round-trip behavior, and scheduler schedule-time `policy_step` capture

I also attempted to run the targeted pytest slice:

- `uv run pytest tests/unit/orchestrator/test_replay.py tests/unit/orchestrator/test_scheduler.py tests/unit/test_configs.py -q`
- `./.venv/bin/python -m pytest tests/unit/orchestrator/test_replay.py tests/unit/orchestrator/test_scheduler.py tests/unit/test_configs.py -q`

Those remain blocked on this macOS host because the repo lockfile only supports Linux environments and the local `.venv` does not have `pytest` installed.

---

> [!NOTE]
> **Medium Risk**
> Touches core orchestrator batching/metrics and checkpoint save/load paths; bugs could change training data mix or resume behavior, though the feature is gated behind explicit `replay` config and has unit coverage.
> 
> **Overview**
> Adds an orchestrator-side replay buffer that stores immutable rollout *groups* and can mix sampled replayed groups into each training batch based on a new `replay` config (`capacity`, `replay_fraction`, seed, and replay-specific off-policy age).
> 
> Extends checkpointing to persist/restore replay state and introduces `ckpt.skip_replay` to independently bypass replay restoration on resume.
> 
> Updates scheduling/batching to record `policy_step` at rollout schedule time, annotates rollouts with `data_source`/`problem_idx`, and adjusts step metrics/aggregations to use `problem_idx` (plus new replay utilization/age and requested-vs-actual replay fraction metrics). Includes focused unit tests and config-skill documentation for the new replay surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01180b616fc080d67feef3d50f995b31dbde75f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skills/config/SKILL.md
+++ b/skills/config/SKILL.md
@@ -168,6 +168,24 @@ In TOML, an empty section header does the same:
 [ckpt]  # enables checkpointing with defaults
 ```
 
+### Replay config
+
+Replay is configured on the orchestrator. In a standalone orchestrator TOML, use `[replay]`. In an `rl` config, nest it under `[orchestrator.replay]`. It stays disabled unless both `capacity` and `replay_fraction` are set:
+
+```toml
+[orchestrator.replay]
+capacity = 1024
+replay_fraction = 0.25
+max_replay_off_policy_steps = 8
+```
+
+Replay entries are stored as full rollout groups, so `capacity` counts groups rather than individual rollouts. When resuming from a checkpoint, you can skip replay restoration independently of the main buffer:
+
+```toml
+[ckpt]
+skip_replay = true
+```
+
 ## Key files
 
 - `src/prime_rl/utils/config.py` — re-exports `BaseConfig` and `cli` from pydantic_config
--- a/src/prime_rl/configs/orchestrator.py
+++ b/src/prime_rl/configs/orchestrator.py
@@ -605,6 +605,13 @@ class CheckpointConfig(BaseConfig):
         ),
     ] = False
 
+    skip_replay: Annotated[
+        bool,
+        Field(
+            description="Whether to skip loading the replay buffer from checkpoint.",
+        ),
+    ] = False
+
 
 class BufferConfig(BaseConfig):
     """Configures the buffer for the orchestrator."""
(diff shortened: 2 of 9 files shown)


faresobeid:
sorry for the late reply! We've since then gone against supporting replay buffers in this sense for now

PR #2336, opened 2026-04-21
feat: add orch2 — async-native orchestrator redesign

samsja:
## Summary

Adds an experimental orchestrator at `src/prime_rl/orch2` that replaces the current `scheduler` + `buffer` + `orchestrator` loop with a short async pipeline. Opt-in via `use_orch2 = true` on `RLConfig`; the default remains the existing orchestrator. Nothing in the current code path is touched.

The goal is a design that is shorter, more testable, and easier to extend (eval, difficulty filtering, multi-endpoint, etc.) without touching the engine.

## Architecture — orch2

```
               ┌────────────────────────────────────────────────────────────┐
               │                   WeightWatcher (task)                     │
               │   polls broadcasts/step_N/STABLE, notifies observers in    │
               │   order: admin → engine → scheduler                        │
               └────────────────┬───────────────────────────────────────────┘
                                │ on_new_version(step)
             ┌──────────────────┼────────────────────────────┐
             ▼                  ▼                            ▼
      InferenceAdmin        RolloutEngine                 Scheduler
  resolves step→path    bumps policy_version,       (placeholder today;
  pause / update /      cancels stale inflight      eval cadence / ratios
  resume on vLLM        groups                      in the future)

          Scheduler.next_task() ─────────► (Task, example)
                                                 │
                                                 ▼
                                      ┌────────────────────┐
                                      │   RolloutEngine    │
                                      │   (task)           │
                                      │  • single dispatch │
                                      │    loop, global    │
                                      │    semaphore       │
                                      │  • fan-out N       │
                                      │    rollouts/group  │
                                      │  • cancel stale    │
                                      │  • tag with        │
                                      │    env_id + kind   │
                                      └─────────┬──────────┘
                                                │ Group(rollouts, policy_version, env_id, kind)
                                                ▼
                                      ┌────────────────────┐
                                      │   TrainBatcher     │   (task)
                                      │   per group:       │
                                      │   ─ Advantage      │   GRPO scoring
                                      │   ─ apply_filters  │   annotation only
                                      │   ─ BatchingStrat  │   step/samples/tokens
                                      │   ─ PostProcessor  │   convert + send + log
                                      └─────────┬──────────┘
                                                ▼
                                        filesystem sender
                                                │
                                                ▼
                                             trainer
```

Three tasks in one `asyncio.TaskGroup`: `engine`, `batcher`, `watcher` — plus background `_run_group` coroutines the engine spawns for in-flight groups.

## Components

| Component | Responsibility |
|---|---|
| `Task` (dataclass) | `id`, `env` (verifiers), `sampling_args`, `kind` — everything the engine needs to run one rollout |
| `Scheduler` | Owns the per-env datasets, picks next `(Task, example)`; round-robin today, designed to grow into weighted/cadence mixing |
| `RolloutEngine` | Single dispatch loop, global `asyncio.Semaphore`, off-policy cancel on weight updates. Holds the shared `client`, `model`, `rollouts_per_group`, `max_off_policy` |
| `Group` (dataclass) | Engine → batcher payload: `rollouts`, `policy_version`, `env_id`, `kind` |
| `PolicyState` (Protocol) | Read-only surface engine exposes to the batcher — batcher doesn't import the engine |
| `Advantage` | GRPO advantage scoring, mutates rollouts in place |
| `apply_filters` | Annotates each rollout with `is_filtered` + per-filter detection bits; does not drop |
| `BatchingStrategy` (Protocol) + `StepStrategy` / `SamplesStrategy` / `TokensStrategy` | Decides when a batch is ready. Three modes, same Protocol — see *Batching modes* below |
| `PostProcessor` | Off-thread tokenize + interleave → `TrainingSample`s, wraps in `TrainingBatch`, sends, emits per-step loguru + wandb (trainable view + pre-filter cohort view + per-filter rate) |
| `TrainBatcher` | Thin orchestrator: pull group → filter eval → Advantage.score → apply_filters → strategy.add → (if has_batch) wait_barrier + pop + PostProcessor.process |
| `WeightWatcher` | Polls broadcast dir, notifies a `list[VersionObserver]` in order |
| `VersionObserver` (Protocol) | Just `async on_new_version(step: int)` — admin, engine, scheduler all implement it |
| `InferenceAdmin` | Health / check model / on_new_version (pause → update_weights → resume on vLLM); resolves `broadcast_dir/step_N` internally |

## Batching modes

`batch_size` is a discriminated union, not an int. Scalar `batch_size = N` is sugar for `type="step", size=N` (orch1 semantics).

| Mode | Flush predicate | Trainer batch size | Notes |
|---|---|---|---|
| **`step`** (default) | `len(buf) ≥ size` | variable (≤ size, minus filter drops) | orch1 semantics: fix engine cost per batch, trainer sees whatever survives filters |
| **`samples`** | `trainable_count(buf) ≥ size` | fixed = `size` | oversamples until N trainable accumulate, filtered rollouts ride along for metrics |
| **`tokens`** | `trainable_tokens(buf) ≥ size` | variable | fixed token budget per batch, stable compute time regardless of seq-len variance |

TOML:
```toml
batch_size = 128              # sugar — step mode, size=128

[batch_size]                  # explicit step
type = "step"
size = 128

[batch_size]                  # samples
type = "samples"
size = 128

[batch_size]                  # tokens
type = "tokens"
size = 32768
```

`token_batch_size` + `oversampling_factor` XOR dance is gone; one mode-dispatched resolver handles defaulting `max_inflight_rollouts` for sample-counting modes, required for tokens.

## Rollout filters

Reuses `src/prime_rl/orchestrator/filters.py` unchanged. Filters **annotate**, they don't drop. Drop happens when `BatchBuilder.pop()` respects `is_filtered`. Two benefits:

- Monitor-mode filters (`enforce=False`) still participate in per-step metrics without affecting training.
- PostProcessor can log both the trainer-facing view (`train/*`) and the full pre-filter cohort (`rollouts/*`, `filters/<name>/rate`, `filters/drop_rate`).

Default filter list is unchanged from orch1 (gibberish + repetition in monitor mode, zero-advantage enforcing). Set `filters = []` to disable entirely.

## What's different

| | old | new |
|---|---|---|
| Main control flow | synchronous `for step in ...` driving everything | `asyncio.TaskGroup` supervising coroutines, no step loop |
| Env mixing | `Buffer` with difficulty classes + hashing | `Scheduler` picks next task; round-robin today, designed for cadence/weighting |
| Off-policy handling | scheduler increments counters, cancels stale, may reschedule | engine snapshots `policy_version` per group; cancels stale via `gather.cancel()` on `on_new_version` |
| Weight broadcast | inline in scheduler, gated by `checkpoint_ready` event | separate `WeightWatcher` + observer fan-out to admin/engine/scheduler |
| Inference admin | list-based multi-endpoint helpers | single `InferenceAdmin`, one endpoint |
| Batcher | advantage + batching + send + log interleaved | small classes wired by a thin `TrainBatcher`; batching policy is a swappable Strategy |
| `batch_size` | `int` xor `token_batch_size: int`, with validator | discriminated union (`step` / `samples` / `tokens`); scalar int = step sugar |
| Filters | apply-then-drop at generation time, with retry-on-empty | annotate in place at batch time; drop happens at pop |
| Engine coupling to batcher | direct object ref | read-only `PolicyState` protocol |
| Rollout type | `vf.RolloutOutput` everywhere | unchanged (no translation) |
| LOC | ~1400 across orch/scheduler/buffer | ~500 across engine/scheduler/batcher/watcher/admin/orchestrator |

## Key invariants preserved

- Filesystem transport layout (`rollouts/step_N/train_rollouts.bin` and `broadcasts/step_N/STABLE`) unchanged — the existing trainer consumes orch2 output byte-for-byte.
- `OrchestratorConfig` reused as-is; same TOMLs parse (scalar `batch_size = N` still works).
- Advantage computation reuses `default_advantage_fn` from `prime_rl.orchestrator.advantage`, so GRPO numerics match.
- Tokenization path reuses `pretokenize_rollout_trajectory` and `interleave_rollout`.
- Filter implementations reused from `prime_rl.orchestrator.filters` — same detection logic.

## How the off-policy + cancellation works

1. Each group snapshots `policy_version` at start.
2. `gather` over N rollout coroutines is kept in `self._inflight` with the snapshot version.
3. When trainer broadcasts new weights, `WeightWatcher` fires observers in order: admin updates vLLM → engine bumps version + walks `_inflight`, cancelling any `gather` where `step - version > max_off_policy` → scheduler notified (cadence hook for later).
4. Cancellation propagates down to the httpx request, so vLLM stops generating tokens for stale rollouts.
5. A post-await defensive check catches the race where `gather` completed just before cancel arrived.

## Feature status vs full orch

- [x] Single or multi-env training with `Scheduler` (round-robin or weighted by `env.ratio`)
- [x] Filesystem weight broadcast + observer fan-out
- [x] GRPO default advantage (`DefaultAdvantageConfig`)
- [x] Single inference endpoint
- [x] Wandb + prime-monitor per-step metrics (trainable + cohort + per-filter views)
- [x] Graceful shutdown on `max_steps`
- [x] `tasks_per_minute` rate limiting (AsyncLimiter on engine)
- [x] `env_install_prerelease` (installs verifiers envs at startup)
- [x] `strict_async_level` (batcher gates on exact lead, not ≤)
- [x] `max_async_level` barrier
- [x] Rollout filters (gibberish / repetition / zero-advantage) — annotation-based, drop at batch pop
- [x] Batching modes (`step` / `samples` / `tokens`) via discriminated union
- [x] Eval — `Scheduler.on_new_version` triggers an exclusive eval epoch at `eval.interval` boundaries (and at step 0 when `eval_base_model=true`); batcher aggregates eval groups by trigger step, logs `eval/<env>/{reward,seq_len,pass_rate,pass_at_k,avg_at_k,n_rollouts,n_examples}` + overall `eval/*` to wandb at the trigger step. Eval inflight exempt from off-policy cancellation. Filters annotate eval rollouts (never drop). Per-env `rollouts_per_example` honored.
- [x] Per-group wall-clock cap — `max_rollout_time_minutes` wraps the gather in `asyncio.wait_for`. Train timeouts drop silently; eval timeouts emit empty groups so the batcher's expected-count check still flushes.
- [x] Shared train/eval metrics — `_rollout_metrics` produces flat per-rollout stats; `_eval_metrics` adds per-example pass_at_k / avg_at_k. Train metrics top-level (no kind prefix); only `eval/` is kind-prefixed.
- [x] Checkpointing — `<output_dir>/checkpoints/step_N/orchestrator/state.pt`. State = `(batcher.step, last_eval_step)`. Periodic save at `ckpt.interval`. Resume via `resume_step` (`-1` = latest) restores state, primes vLLM with resumed weights, suppresses eval_at_zero, honors `skip_eval_on_resume`. `keep_last` / `keep_interval` pruning scoped to the orchestrator subdir.
- [x] NCCL weight broadcast — `InferenceAdmin` gains a `mode` flag; at startup posts `/init_broadcaster` to set up vLLM's NCCL receiver, then on each `on_new_version` touches the `NCCL_READY` marker before `/update_weights` so the trainer's sender joins the collective at the right moment. Smoke-tested single-node 1-GPU on reverse-text, weight updates flowing through NCCL.
- [x] `seed` + `env.ratio` weighted env sampling — Scheduler uses a local `random.Random(seed)` for env selection. When all train envs have a `ratio` (config validator enforces all-or-none), `next_task` weighted-samples via `random.choices`; round-robin is the default when no ratios are set. Verified deterministic across runs at fixed seed.
- [x] Barrier-stall warning — batcher emits a one-shot warning if `_wait_barrier` blocks for >30s without policy version advancing (defensive log so trainer-down / barrier-deadlock failure modes are immediately visible).
- [ ] LoRA — `model.lora.*` ignored
- [ ] Teacher models — no distillation path
- [ ] Buffer / difficulty filtering — `buffer.*` ignored
- [ ] `collect_inference_metrics` — no inference metrics poller
- [ ] Non-default advantage — `CustomAdvantageConfig` not handled
- [ ] Inference / trainer crash fail-fast — orch retries httpx forever if vLLM dies; should health-check and exit
- [ ] Long-run stability validation — only smoke-tested ≤25 steps
- [ ] Prometheus / Grafana metrics — designed but not implemented; engine-level rollout counters (success / cancel / error / filtered per env_id) for ops dashboards, independent from wandb

### Deprecated (not porting)

- Multi-endpoint inference client — orch2 assumes a single endpoint (`base_url[0]`); multi-endpoint config is dropped, use an upstream load balancer if you need it

## How to try it

```bash
uv run rl @ examples/reverse_text/rl.toml --use_orch2
```

The only change in `rl.py` is swapping the subprocess binary (`"orchestrator"` → `"orch2"`) when the flag is set.

## Verification

Reverse-text, Qwen3-0.6B, 20 steps, same reverse-text config:
- **step mode** (default, pre-filter size=128): reward 0.19 → 0.79
- **samples mode** (post-filter size=128): reward 0.16 → 0.80

Both match the baseline curve. Async Level 1, Max Off-Policy 0 throughout.

---

> [!NOTE]
> **High Risk**
> High risk because it rewires the core orchestration loop (batching, env execution, checkpoint/resume, weight updates, and metrics) and deletes the prior `scheduler`/`buffer` implementation, so behavior and performance regressions could impact training stability.
> 
> **Overview**
> Introduces a new async-native orchestrator pipeline that replaces the prior monolithic step loop: env work is produced by new `EnvSampler` implementations, aggregated and shipped by a new `TrainBatcher` with pluggable batching strategies, and policy updates are driven by a `WeightWatcher` plus a single-endpoint `InferenceAdmin` for health checks and weight/LoRA updates.
> 
> Reworks `OrchestratorConfig` batching to a discriminated-union `batch_size` policy (`step`/`samples`/`tokens`) with validation/defaulting changes, adds `max_rollout_time_minutes`, and adds upfront validators that drop unsupported orch1 behaviors (multi-endpoint inference, renderer, teacher rollouts, LoRA with NCCL).
> 
> Simplifies checkpointing to persist minimal orchestrator state (`OrchState`) under the unified checkpoint tree with pruning support, removes the old difficulty-filtering flag (`online_difficulty_filtering`) from configs/code, and switches inference metrics collection to an on-demand `/metrics` snapshot model.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6379545d671df0263b42f0971bbaf2154bf1e47b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/orch2/main.py
```diff
@@ -0,0 +1,277 @@
...
+    logger.info(f"Tokenizer ready: {cfg.tokenizer.name}")
+
+    num_envs = len(cfg.train.env)
+    per_env_concurrency = max(1, cfg.max_inflight_rollouts // num_envs)
+    pools = [build_pool(env_cfg, cfg, per_env_concurrency) for env_cfg in cfg.train.env]
+    for pool in pools:
+        logger.info(
+            f"Pool '{pool.id}' ready | concurrency={pool.concurrency} | "
+            f"rollouts_per_group={pool.rollouts_per_group} | max_off_policy={pool.max_off_policy}"
+        )
```
samsja: lets do a setup_pools function

src/prime_rl/orch2/main.py
```diff
@@ -0,0 +1,277 @@
...
+        for r in rollouts:
+            pretokenize_rollout_trajectory(r, self.tokenizer)
+            out = interleave_rollout(r)
+            if out is None:
+                continue
+            for s in out:
+                s.advantage = r.get("advantage")
+                s.reward = r.get("reward")
+            samples.extend(out)
+        return samples
```
samsja: lets move to its own file

src/prime_rl/orch2/main.py
```diff
@@ -0,0 +1,277 @@
...
+        sampling_args=env_cfg.sampling.to_sampling_args(),
+    )
+    return Pool(
+        id=env_cfg.resolved_name,
+        dataset=cycle_forever(env.get_dataset()),
+        env=worker,
+        concurrency=concurrency,
+        rollouts_per_group=cfg.rollouts_per_example,
+        max_off_policy=cfg.max_off_policy_steps,
+    )
```
samsja: lets moev to its own file with all pool realted stuff

src/prime_rl/orch2/engine.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+@dataclass
+class Pool:
+    id: str
+    dataset: Iterator[dict]
+    env: "EnvWorker"
+    concurrency: int
+    rollouts_per_group: int
+    max_off_policy: int
```
samsja: lets move to pool file

src/prime_rl/orchestrator/engine.py
```diff
@@ -0,0 +1,117 @@
...
+    pool_id: str
+    rollouts: list[vf.RolloutOutput]
+    policy_version: int
+
+
+@dataclass
+class Inflight:
+    version: int
+    max_off_policy: int
+    gather: asyncio.Future
```
samsja: lets move to engine ? does group makes snes if not let keep it here

src/prime_rl/orch2/engine.py
```diff
@@ -0,0 +1,117 @@
...
+
+    async def rollout(self, example: dict, policy_version: int) -> vf.RolloutOutput:
+        del policy_version  # tracked on the Group, not the rollout
+        return await self.env.run_rollout(
+            vf.RolloutInput(**example),
+            client=self.client,
+            model=self.model,
+            sampling_args=self.sampling_args,
+            state_columns=["trajectory", "sampling_args"],
+        )
```
samsja: eh can move this to utils I guess ? not sure why we need this like this

src/prime_rl/orch2/main.py
```diff
@@ -0,0 +1,277 @@
...
+        models = r.json().get("data", [])
+        if not any(m["id"] == model_name for m in models):
+            raise ValueError(f"Model '{model_name}' not found on {self.base_url}")
+
+    async def update_weights(self, weight_dir: Path) -> None:
+        (await self.client.post("/pause", params={"mode": "keep", "clear_cache": "false"})).raise_for_status()
+        try:
+            (await self.client.post("/update_weights", json={"weight_dir": weight_dir.as_posix()})).raise_for_status()
+        finally:
+            (await self.client.post("/resume")).raise_for_status()
```
samsja: lets move to inference_admin file

src/prime_rl/orch2/main.py
```diff
@@ -0,0 +1,277 @@
...
+            pretokenize_rollout_trajectory(r, self.tokenizer)
+            out = interleave_rollout(r)
+            if out is None:
+                continue
+            for s in out:
+                s.advantage = r.get("advantage")
+                s.reward = r.get("reward")
+            samples.extend(out)
+        return samples
+
```
samsja: lets move to train batcher

src/prime_rl/orch2/main.py
```diff
@@ -0,0 +1,277 @@
...
+
+def cycle_forever(ds) -> Iterator[dict]:
+    i = 0
+    while True:
+        for row in ds:
+            r = dict(row)
+            r["example_id"] = i
+            i += 1
+            yield r
```
samsja: move to uitls

src/prime_rl/orch2/orchestrator.py
```diff
@@ -0,0 +1,277 @@
...
+            f"Pool '{pool.id}' ready | concurrency={pool.concurrency} | "
+            f"rollouts_per_group={pool.rollouts_per_group} | max_off_policy={pool.max_off_policy}"
+        )
+
+    groups_q: asyncio.Queue[Group] = asyncio.Queue()
+
+    engine = RolloutEngine(pools, groups_q)
+    training_sender = setup_training_batch_sender(cfg.output_dir, cfg.rollout_transport)
+    batcher = TrainBatcher(groups_q, tokenizer, training_sender, engine, cfg.batch_size, cfg.advantage)
+    logger.info(f"Training batch sender ready ({cfg.rollout_transport.type})")
```
samsja: can we move all into the setup_training_batch_sender incl log ?

src/prime_rl/configs/orchestrator.py
```diff
@@ -791,6 +784,36 @@ class ZeroAdvantageFilterConfig(BaseModel):
 ]
 
 
+class SamplesBatching(BaseConfig):
```
mikasenghaas, line 773: i quite like this naming for its brevity but it's also not super intuitive for newbies, e.g. `step` vs `samples` batch is much less clear than `rollouts` vs `trainable_rollouts` 

again, im not against this naming but would defo need very good docs on this both in prime-rl and hosted

src/prime_rl/configs/orchestrator.py
```diff
@@ -791,6 +784,36 @@ class ZeroAdvantageFilterConfig(BaseModel):
...
+    """Ship a batch every `size` trainable completion tokens."""
+
+    type: Literal["tokens"] = "tokens"
+    size: Annotated[int, Field(ge=1, description="Trainable completion tokens to ship per batch.")]
+
+
+class StepBatching(BaseConfig):
+    """Ship the first `size` rollouts the engine produced, pre-filter. Filtered
+    rollouts count toward `size` but don't reach the trainer — the trainer gets
+    the trainable subset (≤ `size`). Matches orch1 semantics: fixed engine
```
mikasenghaas, line 790: nit: remove the orch1 mention

src/prime_rl/configs/orchestrator.py
```diff
@@ -956,27 +980,21 @@ class OrchestratorConfig(BaseConfig):
...
-    ] = None
+    ] = StepBatching()
 
     oversampling_factor: Annotated[
         float | None,
         Field(
             ge=1,
             description=(
-                "Rollout-mode batching only. Multiplier used to derive max_inflight_rollouts from batch_size "
+                "Samples-mode only. Multiplier used to derive max_inflight_rollouts from batch_size.size "
```
mikasenghaas, line 979: samples or step, right? this is only disabled if we count tokens in the batch iiuc

src/prime_rl/configs/orchestrator.py
```diff
@@ -1030,6 +1048,20 @@ class OrchestratorConfig(BaseConfig):
         ),
     ] = 8
 
+    max_rollout_time_minutes: Annotated[
```
mikasenghaas, line 1033: this is personal preference but i kinda dislike having units in literal config fields. imo this can be put in the description. esp if we are consistent abt e.g. all user-facing time units being minutes (smth i have been trying to push for in verifiers as well, altho we seem to be converging on seconds there)

src/prime_rl/configs/orchestrator.py
```diff
@@ -1174,6 +1216,48 @@ def validate_client_mode(self):
...
+    @model_validator(mode="after")
+    def validate_single_endpoint(self):
+        """The async orchestrator supports a single inference endpoint;
+        multi-endpoint config from orch1 is dropped (use an upstream load
+        balancer if you need it)."""
+        if len(self.client.base_url) != 1:
+            raise ValueError(
+                f"orchestrator.client.base_url must contain exactly one endpoint, got {len(self.client.base_url)}"
+            )
+        return self
```
mikasenghaas, line 1205: ah let's drop it then. would not try to maintain both in parallel

src/prime_rl/configs/orchestrator.py
```diff
@@ -1174,6 +1216,48 @@ def validate_client_mode(self):
...
+    def validate_renderer_not_implemented(self):
+        """The async orchestrator doesn't yet support per-model client-side
+        renderers (the `[orchestrator.renderer]` block + ``/v1/generate`` path).
+        Raise up front so users don't get a confusing failure later."""
+        if self.use_renderer:
+            raise ValueError(
+                "orchestrator.use_renderer is not yet supported by the new async orchestrator. "
+                "Set use_renderer=false (use_token_client=true for TITO, both false for MITO)."
+            )
+        return self
```
mikasenghaas, line 1217: need this

mikasenghaas: in fact, i want to deprecate MITO+TITO(at least TITO) asap

src/prime_rl/configs/orchestrator.py
```diff
@@ -1222,34 +1306,24 @@ def nccl_max_async_level(self):
...
-        if has_token_batch:
-            if self.oversampling_factor is not None:
-                raise ValueError("oversampling_factor can only be set when batch_size is set")
-            if self.max_inflight_rollouts is None:
-                raise ValueError("max_inflight_rollouts must be set when token_batch_size is set")
-        else:
-            assert self.batch_size is not None
-            if self.batch_size % self.rollouts_per_example != 0:
-                raise ValueError("Batch size must be divisible by the number of samples per problem")
+        if isinstance(self.batch_size, (SamplesBatching, StepBatching)):
```
mikasenghaas, line 1286: imo a `.type` check is cleaner than `isinstance` here?

src/prime_rl/configs/rl.py
```diff
@@ -684,11 +684,13 @@ def auto_setup_weight_broadcast(self):
...
+            from prime_rl.configs.orchestrator import SamplesBatching
+
             self.trainer.bench = BenchConfig()
             self.orchestrator.bench = True
-            self.trainer.data.fake = FakeDataLoaderConfig(
-                batch_size=self.orchestrator.batch_size or 32,
-            )
+            bs = self.orchestrator.batch_size
+            fake_bs = bs.size if isinstance(bs, SamplesBatching) else 32
+            self.trainer.data.fake = FakeDataLoaderConfig(batch_size=fake_bs)
```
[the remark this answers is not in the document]
mikasenghaas: this is valid. should only default to 32 for token batching

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+from prime_rl.orchestrator.trajectories import interleave_rollout, pretokenize_rollout_trajectory
+from prime_rl.orchestrator.vf_utils import get_completion_len
+from prime_rl.transport import TrainingBatch, TrainingSample, setup_training_batch_sender
+from prime_rl.transport.base import TrainingBatchSender
+from prime_rl.utils.heartbeat import Heartbeat
+from prime_rl.utils.logger import get_logger
+from prime_rl.utils.monitor import get_monitor
+
+
+class Done(Exception):
```
mikasenghaas, line 28: feels weird to raise on done

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+    """Small surface the batcher needs from the scheduler: how many eval
+    groups to expect for a given trigger step, plus the most recent eval
+    trigger step for ckpt persistence."""
+
+    last_eval_step: int
+
+    def expected_eval_count(self, step: int) -> int | None: ...
+
+
+def _split(rollouts: list[vf.RolloutOutput]) -> tuple[list[vf.RolloutOutput], list[vf.RolloutOutput]]:
```
mikasenghaas, line 37: we should consider making an interanl `RolloutOutput` class which is a superset of `vf.RolloutOutput` (e.g. which has fields like `is_filtered` or `env_id` to be explicit about the types. eg. i just tripped over this `is_filtered` field which is not vf api

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+        self.size = size
+        self._buf: list[vf.RolloutOutput] = []
+
+    def add(self, rollouts: list[vf.RolloutOutput]) -> None:
+        self._buf.extend(rollouts)
+
+    def has_batch(self) -> bool:
+        return len(self._buf) >= self.size
+
+    def pop(self) -> tuple[list[vf.RolloutOutput], list[vf.RolloutOutput]]:
```
mikasenghaas, line 64: should we raise if size is not available instead of doing the slicing that might silently pop `<size` rollouts?

src/prime_rl/configs/orchestrator.py
mikasenghaas: there's a bunch of mentions all over the pr

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+    """Ship when `size` trainable rollouts (post-filter) have accumulated.
+    Oversamples the engine: filtered rollouts are kept in the buffer for
+    metric aggregation but don't count toward `size`."""
+
+    def __init__(self, size: int):
+        self.size = size
+        self._buf: list[vf.RolloutOutput] = []
+
+    def add(self, rollouts: list[vf.RolloutOutput]) -> None:
+        self._buf.extend(rollouts)
```
mikasenghaas, line 77: could also just only accumulate trainable rollouts here?

mikasenghaas: ahhh i see, for metrics. nice

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+        for i, r in enumerate(self._buf):
+            if r.get("is_filtered"):
+                filtered.append(r)
+            else:
+                trainable.append(r)
+                if len(trainable) == self.size:
+                    cut = i + 1
+                    break
+        self._buf = self._buf[cut:]
+        return trainable, filtered
```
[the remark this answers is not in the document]
mikasenghaas: seems valid

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+                tokens += get_completion_len(r)
+                if tokens >= self.size:
+                    cut = i + 1
+                    break
+        self._buf = self._buf[cut:]
+        return trainable, filtered
+
+
+def build_strategy(cfg: BatchingConfig) -> BatchingStrategy:
+    if isinstance(cfg, StepBatching):
```
mikasenghaas: would use `.type` here as well

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+    for k in _TIMING_FIELDS:
+        v = t.get(k)
+        if isinstance(v, dict):
+            out[k] = float(v.get("duration", 0.0))
+        elif v is not None:
+            out[k] = float(v)
+    return out
+
+
+def _rollout_metrics(prefix: str, rollouts: list[vf.RolloutOutput]) -> dict:
```
mikasenghaas, line 160: we need to find a clean abstraction for metrics as well. been unhappy with this for a while. and if it only means shoving all the ugly code into a separate class/file haha

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+        filtered: list[vf.RolloutOutput],
+        step: int,
+    ) -> None:
+        t0 = time.perf_counter()
+        samples = await asyncio.to_thread(self._convert, trainable)
+        convert_time = time.perf_counter() - t0
+        if not samples:
+            # Trainer needs at least one sample per step. Step-mode can produce
+            # fully-filtered cohorts on small models; surface clearly.
+            self.logger.warning(
```
mikasenghaas, line 222: hmm this is a regression no? why would we want to allow crashing the trainer? i think rn we retry the batch 3 times and then raise loudly

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+    def _convert(self, rollouts: list[vf.RolloutOutput]) -> list[TrainingSample]:
+        samples: list[TrainingSample] = []
+        for r in rollouts:
+            pretokenize_rollout_trajectory(r, self.tokenizer)
+            out = interleave_rollout(r)
+            if out is None:
+                continue
+            for s in out:
+                s.advantage = r.get("advantage")
+                s.reward = r.get("reward")
```
mikasenghaas, line 247: can we make this part of `interleave_rollout`?

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+        advs_t = [r.get("advantage") or 0.0 for r in trainable]
+        adv_abs = sum(abs(a) for a in advs_t) / len(trainable) if trainable else 0.0
+        async_level = step - self.policy.policy_version
+        max_off_policy_level = self.policy.max_off_policy_level()
+
+        metrics: dict = {
+            # Trainable subset (post-filter) — top level, no kind prefix.
+            **_rollout_metrics("", trainable),
+            # Full pre-filter cohort — grouped under rollouts/.
+            **_rollout_metrics("rollouts", cohort),
```
mikasenghaas: i dont like the no prefix for trainable and `rollouts` prefix for all rollouts. i think we need super descriptive names here

src/prime_rl/orchestrator/batcher.py
```diff
@@ -0,0 +1,529 @@
...
+    cfg: OrchestratorConfig,
+    *,
+    in_q: asyncio.Queue[GroupOutput],
+    tokenizer: Any,
+    policy: PolicyState,
+    eval_counter: EvalCounter | None = None,
+    ckpt_manager: CkptManager | None = None,
+    buffer: DifficultyBuffer | None = None,
+    inference_metrics: InferenceMetricsCollector | None = None,
+) -> TrainBatcher:
```
mikasenghaas, line 459: hmm bit of code smell that the batcher requires global config and so many heavy classes for init

src/prime_rl/orchestrator/ckpt.py
```diff
@@ -1,96 +1,84 @@
...
-    total_tokens: int = 0
-    total_samples: int = 0
-    total_problems: int = 0
+    last_eval_step: int = 0
+    buffer_state: dict[str, Any] = field(default_factory=dict)
 
 
-class CheckpointManager:
-    """Utility class to save and load orchestrator checkpoints to resume orchestrator."""
+class CkptManager:
```
mikasenghaas, line 30: hm, let's only rename if we also rename on trainer

src/prime_rl/orchestrator/ckpt.py
```diff
@@ -1,96 +1,84 @@
...
+    """Minimal orchestrator state needed to resume cleanly. Engine version +
+    watcher cursor are recoverable from disk on the next watcher tick, so we
+    don't persist them. Round-robin scheduler index resets to 0 on resume —
+    minor reordering, no correctness loss."""
+
     step: int = 0
-    total_tokens: int = 0
-    total_samples: int = 0
-    total_problems: int = 0
+    last_eval_step: int = 0
```
mikasenghaas, line 26: why is eval step required for resume?

src/prime_rl/orchestrator/ckpt.py
```diff
@@ -1,96 +1,84 @@
...
+    `<output_dir>/checkpoints/step_N/orchestrator/state.pt` so trainer + orch
+    ckpts at the same step share a parent dir."""
 
     def __init__(self, output_dir: Path, config: CheckpointConfig):
         self.config = config
         self.ckpt_dir = get_ckpt_dir(output_dir)
         self.logger = get_logger()
 
-    def get_ckpt_path(self, step: int) -> Path:
+    def _path(self, step: int) -> Path:
```
mikasenghaas, line 40: this is a worse name than before

src/prime_rl/orchestrator/ckpt.py
mikasenghaas, line 1: not sure i get why this file has such major changes in general

src/prime_rl/orchestrator/engine.py
```diff
@@ -0,0 +1,149 @@
...
+    """Lightweight tracker for the max_off_policy_level metric. Not used for
+    cancellation — once a group is dispatched, it runs to completion or times
+    out, and gets filtered post-hoc by per-sample max_off_policy."""
+
+    version: int
+    kind: Kind
+
+
+@dataclass
+class EngineInputs:
```
mikasenghaas: not sure we gain anything from this config class?

src/prime_rl/orchestrator/engine.py
```diff
@@ -0,0 +1,149 @@
...
+        self.policy_version = 0
+        self._inflight: list[Inflight] = []
+        self.logger = get_logger()
+
+    async def run(self) -> None:
+        sem = asyncio.Semaphore(self.concurrency)
+        while True:
+            await sem.acquire()
+            dispatch = self.scheduler.next_task()
+            if dispatch is None:
```
mikasenghaas: when can this be `None`?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
+class Task:
+    """What the engine needs to run a rollout. No dataset — that's the scheduler's job."""
 
-    off_policy_steps: int
-    client_config: vf.ClientConfig
-    env_name: str
-    group_id: int | None = None
-    rollout_count: int = 1
+    id: str
+    env: vf.Environment
```
mikasenghaas: should we use the resolved `env_name` here, instead of the full class? and can then do `train_envs[task.env_name]` to retrieve the class for dispatch inside the scheduler

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
+    sampling_args: dict
```
mikasenghaas: do we have any sampling args that change per-task? otherwise would make this `Env` concern or even leave global (which would allow us to not track per-task)

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
+    kind: Kind
+    rollouts_per_group: int
```
mikasenghaas: thinking more abt it, i kinda like `group_size` more and more here. we can still allow both `rollout_per_example` and `group_size` in the public configs but i think for internal read `group_size` is more concise and matches our `Group` abstractions better

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
 
 
 @dataclass
-class GroupState:
-    """Tracks the state of a rollout group (one example × N rollouts)."""
+class Dispatch:
+    """One unit of work handed to the engine."""
 
+    task: Task
     example: dict
```
mikasenghaas: why is `example` not in `Task`?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
-    rollouts_to_schedule: int
-    completed_rollouts: list[vf.RolloutOutput] = field(default_factory=list)
-    pinned_client: vf.ClientConfig | None = None
+    eval_step: int | None = None  # trigger step for eval dispatches; None for train
```
mikasenghaas: also wondering if we should not just make a exlicit `TrainTask` and `EvalTask`. i have tried multiple times in the past to get away with a boolean discriminator on a single type but then had to pull those apart bc the logic ended up being too different. if we split the types in the first place, we'd be more future proof

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
-    """
-    Asynchronously manages scheduling of rollout requests and policy updates.
-    Keeps a constant number of rollouts in-flight (continuous batching) and
-    updates the policy as soon as it becomes available.
+@dataclass
+class SchedulerInputs:
+    """Pre-built inputs for the Scheduler. `setup_scheduler` produces this from
+    config; tests construct it directly with handcrafted tasks/datasets."""
+
+    train_tasks: list[Task]
```
mikasenghaas: ohh i think i completely misunderstoof the `Task` dataclass bc in vf this is per-rollout information. would try to avoid the name `Task` in this context then

mikasenghaas: this type then reads more like all the stuff you need to run an example on a train or eval env. we should aim to have all information alr populated on `TrainEnv` and `EvalEnv` (in fact, i was under the impression we alr did this but maybe not fully)

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
+    for env_cfg in cfg.train.env:
+        env = vf.load_environment(env_cfg.stripped_id, **env_cfg.args)
+        train_tasks.append(
+            Task(
+                id=env_cfg.resolved_name,
+                env=env,
+                sampling_args=env_cfg.sampling.to_sampling_args(),
+                kind="train",
+                rollouts_per_group=cfg.rollouts_per_example,
             )
```
mikasenghaas: ya i really think we dont need this. we alr iterate over the train env configs to construct `TrainEnvs` (same for eval). the scheduler should just consume those

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
-            )
-            self.checkpoint_ready.clear()
-            wait_for_ckpt_start_time = time.perf_counter()
-            await wait_for_path(get_step_path(get_broadcast_dir(self.config.output_dir), next_ckpt_step) / "STABLE")
-            self.wait_for_ckpt_time = time.perf_counter() - wait_for_ckpt_start_time
-            self.logger.info(
-                f"Orchestrator resumed: checkpoint {next_ckpt_step} ready (after {self.wait_for_ckpt_time:.2f}s)"
+        train_datasets.append(_cycle_forever(env.get_dataset(), env_id=env_cfg.resolved_name, buffer=buffer))
+
+    ratios = [env_cfg.ratio for env_cfg in cfg.train.env]
```
mikasenghaas: same with the ratios, they can alr be retrieved from `TrainEnvs`

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
+    env_weights: list[float] | None = ratios if all(r is not None for r in ratios) else None  # type: ignore[assignment]
+    if env_weights is not None:
+        named = ", ".join(f"{e.resolved_name}={e.ratio:.2f}" for e in cfg.train.env)
+        logger.info(f"Sampling train envs by ratio ({named})")
+    else:
+        logger.info(f"Sampling train envs round-robin ({len(cfg.train.env)} env(s))")
```
mikasenghaas: btw, what does round-robin mean here? round-robin the env and then the tasks per env. or concat all tasks and then iterate over that?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
+    def expected_eval_count(self, step: int) -> int | None:
+        """How many eval groups were dispatched for this eval epoch. None if
+        there was no epoch triggered at that step."""
+        return self._eval_expected.get(step)
+
+    def next_task(self) -> Dispatch | None:
+        if self._eval_queue:
+            task, example, eval_step = self._eval_queue.popleft()
+            if not self._eval_queue:
+                self._dispatching_eval_step = None
```
mikasenghaas: not sure i get this conditional

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
+
+        if self._env_weights is not None:
+            # Weighted: pick by ratio; loop only as a defensive guard if an
+            # iterator is exhausted (shouldn't happen with _cycle_forever).
+            for _ in range(len(self.tasks)):
+                i = self._rng.choices(range(len(self.tasks)), weights=self._env_weights, k=1)[0]
+                try:
+                    example = next(self._datasets[i])
+                    return Dispatch(task=self.tasks[i], example=example)
+                except StopIteration:
```
mikasenghaas: no defensive guard

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -1,554 +1,256 @@
...
+                    continue
+            return None
+
+        for _ in range(len(self.tasks)):
```
mikasenghaas: this is changing the default behavior:
- current main: randomly pick across all tasks in all envs
- now: cycle tasks per env and round-robin envs

the now is implicitly 1:1:...:1 ratios which whereas on curret main no ratio means we respect the natural task distribution across envs. this is a soft regression imo. 

if we don't want to support the natural task distribution anymore, we should just make ratios non-optional and default to 1:1:...:1 and then we can remove this extra code path

src/prime_rl/orchestrator/envs.py
mikasenghaas, line 1: wait wuut, why we delete all this?;( i think this is one of the more useful abstractions on orch atm

samsja: huhuhuhu good point

PR #2337, opened 2026-04-21
fix(trainer): wait for NCCL_READY on all ranks before broadcast

mikasenghaas:
## The bug

In `src/prime_rl/trainer/rl/broadcast/nccl.py`, `NCCLWeightBroadcast.broadcast_weights` only waits for the orchestrator's `NCCL_READY` marker on the master rank:

\`\`\`python
if self.world.is_master:
    notified_runs = self._notify_orchestrator()
    self._wait_for_nccl_ready(notified_runs)   # master only
self.nccl_broadcast_sender.broadcast_weights(model, step)  # all ranks
\`\`\`

The non-master ranks skip the wait and enter \`nccl_broadcast_sender.broadcast_weights\` immediately. That function's preparation path (DTensor resolution in \`_resolve_dtensors\`, plus optional quantization) runs on every rank and enqueues NCCL all_gathers on PG 1 (FSDP). Those collectives sit in-flight with no counter-party until the orchestrator has paused inference and the master joins.

If the orchestrator takes more than 10 minutes to create the \`NCCL_READY\` marker (which happens when eval and rollout generation are contending for sandbox/IB capacity), NCCL's watchdog kills the process on a non-master rank — hard SIGABRT, ~18 minutes of wall-clock lost per incident.

## Symptom

We've hit this repeatedly on multi-node RL runs. Always the same signature:

- Master (rank 0) logs \`Waiting for path '.../broadcasts/step_N/NCCL_READY'\`
- ~10 minutes later, a non-master rank's stderr shows \`Watchdog caught collective operation timeout: WorkNCCL(SeqNum=..., OpType=COALESCED, NumelIn=25M, NumelOut=402M, Timeout(ms)=600000)\` on PG 1
- Default-PG rank 0 receives the flight-recorder dump signal from that non-master rank and the trainer dies

## Fix

Split \`_notify_orchestrator\` into:

- \`_compute_notified_runs\` — pure function of SPMD-replicated \`multi_run_manager\` state, runs on every rank and returns the same list everywhere.
- \`_notify_orchestrator(notified_runs)\` — master-only, does the filesystem side-effects (\`mkdir\`, \`touch STABLE\`) and clears \`ready_to_update[idx]\`.

Call \`_compute_notified_runs\` unconditionally, \`_notify_orchestrator\` only on master, and then have every trainer rank wait for the \`NCCL_READY\` marker before entering the broadcast path. The wait is a pure filesystem poll, no NCCL involvement, so a slow orchestrator now just blocks the trainer safely instead of tripping the watchdog.

---

> [!NOTE]
> **Medium Risk**
> Changes distributed synchronization around NCCL weight broadcasts; mistakes could deadlock broadcasts or regress multi-run update coordination, but scope is limited to the broadcast/notification path.
> 
> **Overview**
> Prevents non-master ranks from entering the NCCL broadcast preparation path before inference is ready by making **all ranks** compute the same `notified_runs` list and wait for the `NCCL_READY` marker.
> 
> Splits the previous `_notify_orchestrator` flow into `_compute_notified_runs` (pure, SPMD-consistent) and `_notify_orchestrator(notified_runs)` (master-only filesystem `STABLE` writes + clearing `ready_to_update`), so only rank 0 performs side effects while every rank blocks safely before broadcasting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae6aef48553b9703632fc21e58b7984e68a65d37. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/rl/broadcast/nccl.py
+++ b/src/prime_rl/trainer/rl/broadcast/nccl.py
@@ -198,43 +198,61 @@ class NCCLWeightBroadcast(WeightBroadcast):
         """Broadcast the state dict of a model into the inference pool using NCCL and notifies the orchestrator."""
         self.logger.debug("Starting broadcasting weights to inference engine via NCCL")
         start_time = time.perf_counter()
-        notified_runs: list[tuple[int, Path]] = []
+        # `_compute_notified_runs` is a pure function of SPMD-replicated state on
+        # multi_run_manager, so every trainer rank derives the same list. Only
+        # the master touches the filesystem to notify the orchestrator, but all
+        # ranks must wait on NCCL_READY before entering the broadcast path:
+        # the broadcast preparation (DTensor resolution, quantization) enqueues
+        # collectives on non-master ranks, and if those ranks start prep before
+        # the orchestrator has paused inference, the collectives sit unmatched
+        # until NCCL's watchdog kills the process after 10 min.
+        notified_runs = self._compute_notified_runs()
         if self.world.is_master:
-            notified_runs = self._notify_orchestrator()
-            # Wait for inference workers to signal readiness before starting NCCL broadcast
-            self._wait_for_nccl_ready(notified_runs)
+            self._notify_orchestrator(notified_runs)
+        self._wait_for_nccl_ready(notified_runs)
         self.nccl_broadcast_sender.broadcast_weights(model, step)
         self.logger.debug(f"Weights broadcasted in {time.perf_counter() - start_time:.2f}s")
 
-    def _notify_orchestrator(self) -> list[tuple[int, Path]]:
-        """Notify the orchestrator to initiate weight broadcast.
+    def _compute_notified_runs(self) -> list[tuple[int, Path]]:
+        """Derive the list of (run_idx, save_dir) pairs that need broadcasting.
 
-        Returns:
-            List of (run_idx, save_dir) tuples for runs that were notified.
+        Pure function of `multi_run_manager` state, which is replicated across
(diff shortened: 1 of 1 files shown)


src/prime_rl/trainer/rl/broadcast/nccl.py
```diff
@@ -198,44 +198,62 @@ def broadcast_weights(self, model: nn.Module, step: int) -> None:
...
+            try:
+                save_dir.mkdir(parents=True, exist_ok=True)
+                stable_file = save_dir / "STABLE"
+                stable_file.touch()
+            except FileNotFoundError:
+                self.logger.warning(f"Run {idx} is deleted, skipping")
+            except Exception as e:
+                self.logger.error(f"Error broadcasting weights for run {idx}: {e}")
+            finally:
+                self.multi_run_manager.ready_to_update[idx] = False
```
samsja, line 255: can remove the try expect here

Jackmin801 (approved):
yea lgtm. im wondering if this can be simpler since we dont really support multi-run nccl broadcast. so there is always only one run, and you always notify that run

src/prime_rl/trainer/rl/broadcast/nccl.py
Jackmin801: this try except was probably grafted from the one at filesystems which is used by multi-run. The issue was that  if a run was stopped after dataloader and before broadcast, its directory would be cleaned and not having this try except would crash the trainer when it tried to broadcast to this phantom run.

samsja (approved):

PR #2351, opened 2026-04-24
fix: correct multi-env example_id collision in /all/ metrics

jfeldstein:
# Context

When multiple environments each auto-assign integer `example_ids` from `range(len(dataset))` — the default for every built-in environment — `groupby("example_id")` silently merges rollouts from different problems across environments into invalid groups. 

# Issue

This corrupts `solve_all/all`, `solve_none/all`, `effective_batch_size/all`, and `all seq_len/reward/all/*` metrics logged to W&B in any multi-environment training run. 

# Impact

If anyone is manually using `solve_all/all` or `reward/all/mean` as a stopping signal or to make decisions about hyperparameters mid-run, those decisions could be wrong. But **the automated training loop itself is clean**.

# Fix

- Extract compute_solve_rates into a new, minimal metrics.py module (no heavy deps, directly unit-testable without torch/transformers)
- Group by ["env_name", "example_id"] instead of "example_id" alone in both compute_solve_rates and the by_example groupby
- Add regression tests covering single-env baseline, the collision scenario, and three-env mixed cases

The per-env metrics path (env_df.groupby("example_id")) was already correct because env_df is pre-filtered to a single environment.

---

> [!NOTE]
> **Medium Risk**
> Changes how global training metrics are grouped and computed, which can alter W&B dashboards and any human-in-the-loop stopping heuristics, but does not affect the training data sent to the trainer.
> 
> **Overview**
> Fixes corrupted `/all/` metrics in multi-environment training runs by grouping rollouts by `("env_name", "example_id")` instead of `"example_id"` alone, preventing cross-environment ID collisions from skewing logged `reward/*`, `seq_len/*`, and solve-rate metrics.
> 
> Extracts `compute_solve_rates` into a new lightweight `orchestrator/metrics.py` and adds unit tests covering single-env baselines and multi-env collision regression scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf3d67f9c3075788ec885648beaf812c2c087232. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/metrics.py
+++ b/src/prime_rl/orchestrator/metrics.py
@@ -1,0 +1,44 @@
+"""Pure-pandas metric helpers for the RL orchestrator.
+
+Kept in a separate module with no heavy imports so they can be unit-tested
+without pulling in torch / transformers / verifiers.
+"""
+
+from __future__ import annotations
+
+import pandas as pd
+
+
+def compute_solve_rates(
+    df: pd.DataFrame,
+    rollouts_per_example: int,
+) -> tuple[float, float, float]:
+    """Compute solve_none, solve_all, and effective_batch_size for a set of rollouts.
+
+    Groups by **(env_name, example_id)** so that problems from different
+    environments with the same integer example_id are never merged.  Grouping
+    by example_id alone causes cross-environment collisions when every
+    environment auto-assigns IDs from ``range(len(dataset))``, which is the
+    default behaviour for all built-in environments.
+
+    Args:
+        df: DataFrame with at minimum columns ``env_name``, ``example_id``,
+            and ``reward``.  Must contain at least one row.
+        rollouts_per_example: Number of rollouts generated per problem.
+            A problem is "fully solved" when the sum of its binary rewards
+            equals this value (i.e. every rollout scored 1).
+
+    Returns:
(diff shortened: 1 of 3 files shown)


src/prime_rl/orchestrator/metrics.py
```diff
@@ -0,0 +1,44 @@
...
+Kept in a separate module with no heavy imports so they can be unit-tested
+without pulling in torch / transformers / verifiers.
+"""
+
+from __future__ import annotations
+
+import pandas as pd
+
+
+def compute_solve_rates(
```
mikasenghaas, line 12: lets inline this for the time being pls

tests/unit/orchestrator/test_metrics.py
mikasenghaas, line 1: no need for tests here

mikasenghaas (commented):
thx for the pr! pls address the 2 comments and then we will get this in asap

tests/unit/orchestrator/test_metrics.py
jfeldstein: Why no tests for this?

Happy to clean up, but they do cover a bug in code that previously lacked coverage

PR #2353, opened 2026-04-24
orchestrator: set up early stopping status for zero advantages filtering

eexwhyzee:
> [!NOTE]
> **Medium Risk**
> Changes orchestrator termination semantics and run lifecycle reporting by introducing an early-stop path, which could prematurely end training or alter platform status if the zero-advantage condition is misdetected.
> 
> **Overview**
> Adds an *early stopping* path in the orchestrator when batch generation repeatedly yields fully-filtered rollouts **and** all computed advantages are `0.0`: the run writes `control/early_stopped.txt`, exits the main loop cleanly, and propagates `early_stopped` into `monitor.save_final_summary()`.
> 
> Extends monitor APIs (`Monitor`, `MultiMonitor`, `PrimeMonitor`, `WandbMonitor`) to accept an `early_stopped` flag; `PrimeMonitor` now finalizes platform runs with `status=early_stopped` (including in the `/finalize` payload) when applicable.
> 
> Updates trainer run discovery to treat both `control/evicted.txt` and `control/early_stopped.txt` as *inactive* runs so they are ignored in future `discover_runs()` scans.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a48b85345a74c722f2c82e84bdb78d7bb6aadc9a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -77,8 +77,8 @@ from prime_rl.utils.utils import (
 SHUTDOWN_TIMEOUT_S = 300
 
 # Maximum number of times to attempt generating a training batch when all
-# rollouts are filtered out. After this many attempts, the orchestrator
-# crashes (or early stops for zero advantages) rather than silently skipping training steps.
+# rollouts are filtered out. After this many attempts, the orchestrator crashes
+# (or early stops for zero advantages) rather than silently skipping training steps.
 MAX_EMPTY_BATCH_ATTEMPTS = 3
 
 
@@ -104,6 +104,10 @@ async def orchestrate(config: OrchestratorConfig):
     # Save configs to output directory
     config_dir = config.output_dir / "control"
     config_dir.mkdir(parents=True, exist_ok=True)
+    evicted_path = config_dir / "evicted.txt"
+    early_stopped_path = config_dir / "early_stopped.txt"
+    if early_stopped_path.exists():
+        early_stopped_path.unlink()
     with open(config_dir / "orch.toml", "wb") as f:
         tomli_w.dump(config.model_dump(exclude_none=True, mode="json"), f)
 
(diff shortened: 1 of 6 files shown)


eexwhyzee:
@codex review

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -106,6 +106,10 @@ async def orchestrate(config: OrchestratorConfig):
     # Save configs to output directory
     config_dir = config.output_dir / "control"
     config_dir.mkdir(parents=True, exist_ok=True)
+    evicted_path = config_dir / "evicted.txt"
```
mikasenghaas, line 107: i still dont get why we need to write a file for eviction. somehow the trainer has to learn about unexpected orch crashes as well no?

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -439,24 +443,32 @@ async def orchestrate(config: OrchestratorConfig):
...
-                )
                 reason = (
                     f"All {num_rollouts} rollouts were filtered out on "
                     f"{MAX_EMPTY_BATCH_ATTEMPTS} consecutive attempts at step {progress.step}"
                 )
-                evicted_path = config.output_dir / "control" / "evicted.txt"
-                evicted_path.parent.mkdir(parents=True, exist_ok=True)
+
+                # Check for zero advantages for early stopping
+                all_zero_advantage = len(train_rollouts) > 0 and all(r.get("advantage") == 0.0 for r in train_rollouts)
```
mikasenghaas, line 451: there is just one reason right? why this check

eexwhyzee: the rollouts can also get filtered for gibberish and repetition, so this check just verifies that all of the rollouts are of zero advantage filtered

src/prime_rl/orchestrator/orchestrator.py
eexwhyzee: yeah i am not too sure either, looks like the trainer only looks for the evicted.txt file during `discover_runs()`, but that doesn't cover the scenario of unexpected orch crashes like you mentioned

seems like right now the launcher's `cleanup_processes` is what handles unexpected crashes but that only prob works in single node setups and not multi node?

PR #2354, opened 2026-04-25
fix(multi_ckpt): run maybe_clean on all ranks to keep ckpt_steps in sync

JannikSt:
## Summary

`MultiCheckpointManager.maybe_clean()` returned early on non-master ranks. Only master's in-memory `ckpt_steps` list was updated when old checkpoint directories were deleted; non-master ranks retained the stale entries. A subsequent `_should_save()` call on the same step could then return divergent values across ranks, causing the trainer to deadlock at the next save.

## Reproducer

A trainer in multi-run mode (`max_concurrent_runs > 1`) starts up with multiple per-run checkpoint directories on disk — say `step_440/` and `step_450/` — both showing up in the filesystem scan that initializes `CheckpointManager.ckpt_steps`. The `step_450` checkpoint is selected as the resume target, so `progress[idx].step = 450`.

```
all ranks: ckpt_steps = [440, 450],  progress.step = 450
```

After the first training iteration, `maybe_clean` runs with `keep_last=1`:

```
master:     CheckpointManager.maybe_clean() runs -> ckpt_steps = [450]
non-master: MultiCheckpointManager.maybe_clean() returns early -> ckpt_steps = [440, 450]   ← drift
```

A few iterations later, the multi-run save loop reaches the resumed step that's still inside the in-memory list. `_should_save(idx, 450)`:

| rank       | `step in ckpt_steps?` | `_should_save` |
| ---------- | --------------------- | -------------- |
| master     | `450 in [450]` → no   | True (saves)   |
| non-master | `450 in [440, 450]` → yes | False (skips)  |

Master enters the save try block in `MultiCheckpointManager.save()` and calls `dist.barrier()` (line 155). Non-master ranks took the `continue` path, never participate in that barrier. Master hangs forever — the whole trainer is dead, no rank files get written, the orchestrator's broadcast wait never resolves.

## Root cause

```python
# Before
def maybe_clean(self) -> None:
    if not self.world.is_master:
        return                               # ← non-master ranks bail before updating ckpt_steps
    for idx in self.multi_run_manager.used_idxs:
        if self.managers[idx] is None:
            continue
        self.managers[idx].maybe_clean()
```

The file deletion inside `CheckpointManager.maybe_clean()` is already master-gated. The early return in the wrapper was redundant for filesystem safety but caused the in-memory `ckpt_steps` list to drift across ranks.

## Fix

Remove the early return so every rank executes the cleanup logic. The actual `_try_rmtree` call inside `CheckpointManager.maybe_clean()` is still master-only, so the on-disk behavior is unchanged. Only the in-memory list update now happens uniformly.

## Why this hasn't bitten more stacks

- Single-run mode (`max_concurrent_runs == 1`) doesn't go through `MultiCheckpointManager` and isn't affected.
- In steady-state multi-run, each run usually has exactly one checkpoint dir on disk at startup, so `maybe_clean` has nothing to delete and `ckpt_steps` never drifts.
- The bug only manifests when the filesystem scan picks up multiple checkpoint dirs (e.g., orphan dirs from a prior crash, or fresh-restart-mid-rotation), AND the resumed step is one that gets cleaned out.

## Regression test

`tests/unit/train/test_multi_ckpt.py::test_multi_ckpt_maybe_clean_keeps_ckpt_steps_in_sync_across_ranks` simulates a non-master rank (via `monkeypatch` on `World.is_master`), sets up two pre-existing checkpoint dirs, and asserts `ckpt_steps` is updated to `[450]` after `maybe_clean`. Pre-fix it stayed `[440, 450]`.

---

> [!NOTE]
> **Medium Risk**
> Touches distributed checkpoint coordination logic; if incorrect, it can still cause hangs or inconsistent save/cleanup behavior across ranks, though changes are narrowly scoped and regression-tested.
> 
> **Overview**
> Fixes a multi-rank deadlock/inconsistency in multi-run checkpointing by running `MultiCheckpointManager.maybe_clean()` on **all ranks** so in-memory `ckpt_steps` stays synchronized (file deletions remain master-gated inside `CheckpointManager`).
> 
> Also ensures `MultiCheckpointManager.save()` maintains the `ckpt_steps` sorted invariant by inserting steps with `bisect.insort` instead of appending, avoiding crashes when resuming with orphan future-step checkpoint directories.
> 
> Adds unit regression tests covering both the cross-rank `maybe_clean` synchronization and the sorted `ckpt_steps` behavior on save.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81e0434f1051331a2d30567c2422b228c6b52cd4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/multi_ckpt.py
+++ b/src/prime_rl/trainer/multi_ckpt.py
@@ -219,8 +219,10 @@ class MultiCheckpointManager:
             return False
 
     def maybe_clean(self) -> None:
-        if not self.world.is_master:
-            return
+        # Run on all ranks: file deletion inside CheckpointManager.maybe_clean is
+        # already master-gated, but the in-memory ckpt_steps update must happen on
+        # every rank, otherwise _should_save() diverges across ranks and the next
+        # save can hit the dist.barrier mismatch in save() -> deadlock.
         for idx in self.multi_run_manager.used_idxs:
             if self.managers[idx] is None:
                 continue
--- a/src/prime_rl/trainer/multi_ckpt.py
+++ b/src/prime_rl/trainer/multi_ckpt.py
@@ -4,6 +4,7 @@ MultiCheckpointManager owns per-run CheckpointManagers and AppStates,
 each saving to its own run directory.
 """
 
+import bisect
 import shutil
 from dataclasses import asdict
 from pathlib import Path
@@ -154,7 +155,12 @@ class MultiCheckpointManager:
                         )
                 dist.barrier()
                 manager.mark_stable(step)
-                manager.ckpt_steps.append(step)
+                # Use insort (like CheckpointManager.save_to_path does) so the
+                # list stays sorted. Plain append would break the
+                # `assert list(ckpt_steps) == sorted(ckpt_steps)` invariant in
+                # maybe_clean when saving a step lower than ones already in the
+                # list (happens on resume with orphan future-step ckpt dirs).
+                bisect.insort(manager.ckpt_steps, step)
             except FileNotFoundError:
                 self.logger.warning(f"Run {idx} deleted during checkpoint, skipping")
             except Exception as e:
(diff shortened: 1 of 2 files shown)


JannikSt:
@codex review

PR #2368, opened 2026-04-28
feat: debug.force_balanced_routing for MoE smoke tests

mikasenghaas:
## Summary
- Adds `debug.force_balanced_routing` to `DebugModelConfig`. When true, `TokenChoiceTopKRouter` overrides its top-k expert assignment with a round-robin pattern so every expert receives `num_tokens * top_k / num_experts` tokens.
- Motivation: fake-data smoke tests produce severely skewed routing, which under EP causes per-rank memory blowup and OOM (hit this on a 16-node GLM-5.1 run). Forced balance lets us stress-test memory and throughput without that variance.
- Gating scores are gathered from the override indices, so the forward pass stays self-consistent. Gradients through `gate.weight` are broken — this is **only** valid for benchmarking, not for any optimization signal. The applied path logs a warning calling that out.

## Implementation
- `configs/trainer.py`: `force_balanced_routing: bool = False` on `DebugModelConfig`.
- `models/layers/moe.py`: `TokenChoiceTopKRouter.force_balanced` flag (default False) + a new branch in `forward` that emits `idx = (arange(N*top_k) % num_experts).view(N, top_k)`.
- `trainer/model.py`: `apply_force_balanced_routing(model)` walks `language_model.layers`, flips the flag on each `MoE`/`LatentMoE` router, raises if the model has no custom-impl MoE. Called from `setup_model` immediately after `freeze_moe_router`.

## Caveats
- Custom-impl MoE only (`isinstance(mlp, (MoE, LatentMoE))`); HF-impl path is not affected.
- Per-rank balanced. With EP the resulting traffic is uniform across all experts, which is the intended profile for memory benchmarking.

---

> [!NOTE]
> **Medium Risk**
> Low-impact by default (flag off), but when enabled it changes MoE expert assignment and intentionally breaks router gradient flow, which could invalidate training runs if misused.
> 
> **Overview**
> Adds a new `debug.force_balanced_routing` config that, when enabled, **overrides MoE token-choice routing** to assign tokens to experts in a round-robin pattern for fake-data/throughput smoke tests.
> 
> Implements the override by adding a `force_balanced` flag to `TokenChoiceTopKRouter` and wiring `setup_model` to enable it across custom `MoE`/`LatentMoE` layers, with a warning and an error if no eligible routers are found.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26a7c5c2a1924f836eac156d296f73d8b73fc131. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/configs/trainer.py
+++ b/src/prime_rl/configs/trainer.py
@@ -183,6 +183,13 @@ class DebugModelConfig(BaseConfig):
         ),
     ] = False
 
+    force_balanced_routing: Annotated[
+        bool,
+        Field(
+            description="If True, override MoE token-choice routing with a round-robin assignment so every expert receives an equal share of tokens. Intended for fake-data smoke tests where untrained routing would otherwise produce severe expert imbalance and OOM. Gating scores are still gathered from the override indices so the forward pass stays consistent.",
+        ),
+    ] = False
+
 
 class ModelConfig(BaseModelConfig):
     """Configures the model for training."""
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -351,6 +351,27 @@ def freeze_moe_router(model: nn.Module) -> None:
     logger.info(f"Froze {num_frozen} MoE router parameters")
 
 
+def apply_force_balanced_routing(model: nn.Module) -> None:
+    """Force MoE token-choice routers into round-robin assignment for fake-data smoke tests."""
+    logger = get_logger()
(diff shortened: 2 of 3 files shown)


Jackmin801 (approved):
nice! lgtm

S1ro1 (approved):
Don't see an issue, long overdue anyway o7

PR #2373, opened 2026-04-28
fix(orchestrator): bound monitor history and drop duplicate samples list

samsja:
## Summary

The orchestrator process retains two unbounded structures for the full lifetime of a run, both accumulating per orchestrator step. For long prod runs with sample logging enabled this is the dominant memory growth in the orchestrator and a plausible source of orchestrator-side OOMs.

- **`WandbMonitor.history` / `PrimeMonitor.history`** — full metrics dict per step, only ever read at end of run (orchestrator's `--bench` print, `PrimeMonitor.save_final_summary`'s `history[-1]`).
- **`WandbMonitor.samples`** — full prompt/completion text + stringified token ids per logged sample, only ever read once in `log_final_samples()` to package a redundant `final-samples` wandb table. The same rows are already streamed live to the incremental `samples` wandb table via `samples_table.add_data(...)` + `wandb.log({"samples": ...})`, so the second table was a duplicate.

## Changes

- Add `Monitor.prune_history()` that truncates `history` to its last entry — preserves `history[-1]` access for `save_final_summary`. `MultiMonitor` delegates to each wrapped monitor.
- Call `monitor.prune_history()` from the orchestrator loop, gated on `not config.bench` so benchmark mode still gets the full history table for `print_benchmark`.
- Drop `WandbMonitor.samples` and the duplicate `final-samples` end-of-run table. `log_final_samples` is now a no-op for `WandbMonitor`. Drops the now-unused `pandas` import.

## Verification

Ran `examples/reverse_text/rl.toml` (20 orchestrator steps, wandb on, `log_extras.interval=1`) with a temporary `[history-probe]` log measuring `len(monitor.history)` each step before pruning. Steady-state was `len=2` for every step from step 1 onward (1 carried over from previous prune + 1 just appended by the current step's `monitor.log()`); without the fix this grows linearly to `len=N` at step N. Wandb samples table received rows every step and synced cleanly. Probe log was removed before commit.

---

> [!NOTE]
> **Medium Risk**
> Changes monitoring behavior by limiting in-memory metric history outside `--bench` mode and removing end-of-run `final-samples` W&B logging, which could affect downstream expectations of logged artifacts but is otherwise isolated to observability/memory usage.
> 
> **Overview**
> Prevents orchestrator-side memory growth from monitors by adding a `keep_full_history` option to `setup_monitor`/monitors and defaulting it to **only keep the latest metrics entry** when not running with `--bench`.
> 
> Removes the end-of-run `log_final_samples()` flow (and the extra in-memory `samples` list in `WandbMonitor`), relying solely on the incremental samples table logging, and drops the unused `pandas` dependency from the W&B monitor path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 799fbd149056c65882493e88c772569cb49a9825. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -736,6 +736,13 @@ async def orchestrate(config: OrchestratorConfig):
             step=progress.step,
         )
 
+        # The orchestrator only reads monitor.history for the end-of-run benchmark
+        # table; outside --bench, holding every step's metrics dict is a slow leak
+        # that grows linearly with run length. Keep only the most recent entry so
+        # PrimeMonitor.save_final_summary's `history[-1]` access still works.
+        if not config.bench:
+            monitor.prune_history()
+
         if usage_reporter and run_id:
             usage_reporter.report_training_usage(
                 run_id=run_id,
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -162,6 +162,7 @@ async def orchestrate(config: OrchestratorConfig):
         output_dir=config.output_dir,
         tokenizer=tokenizer,
         run_config=config,
+        keep_full_history=config.bench,
     )
 
     # Read run_id AFTER setup_monitor so that newly registered runs are captured
@@ -736,13 +737,6 @@ async def orchestrate(config: OrchestratorConfig):
             step=progress.step,
         )
 
-        # The orchestrator only reads monitor.history for the end-of-run benchmark
-        # table; outside --bench, holding every step's metrics dict is a slow leak
-        # that grows linearly with run length. Keep only the most recent entry so
-        # PrimeMonitor.save_final_summary's `history[-1]` access still works.
-        if not config.bench:
-            monitor.prune_history()
-
         if usage_reporter and run_id:
             usage_reporter.report_training_usage(
                 run_id=run_id,
@@ -792,8 +786,6 @@ async def orchestrate(config: OrchestratorConfig):
                 save_rollouts, eval_rollouts, step_path / "eval_rollouts.jsonl", exclude_keys={"trajectory"}
             )
 
-    # Log final (immutable) samples and distributions to monitor(s)
-    monitor.log_final_samples()
     monitor.save_final_summary()
 
     # Write final checkpoint
--- a/src/prime_rl/utils/monitor/multi.py
+++ b/src/prime_rl/utils/monitor/multi.py
@@ -70,7 +70,4 @@ class MultiMonitor(Monitor):
 
     def prune_history(self) -> None:
         for monitor in self.monitors:
-            try:
-                monitor.prune_history()
-            except Exception as e:
-                self.logger.warning(f"Failed to prune history on {monitor.__class__.__name__}: {e}")
+            monitor.prune_history()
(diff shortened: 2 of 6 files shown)


src/prime_rl/utils/monitor/wandb.py
```diff
@@ -216,21 +213,10 @@ def log_eval_samples(self, rollouts: list[vf.RolloutOutput], env_name: str, step
         wandb.log({"eval/samples": self.eval_samples_table, "step": step})
 
     def log_final_samples(self) -> None:
```
mikasenghaas: let's just remove this method all together. i think its a no-op everywhere

src/prime_rl/utils/monitor/multi.py
```diff
@@ -67,3 +67,10 @@ def close(self) -> None:
                 monitor.close()
             except Exception as e:
                 self.logger.warning(f"Failed to close {monitor.__class__.__name__}: {e}")
+
+    def prune_history(self) -> None:
+        for monitor in self.monitors:
+            try:
+                monitor.prune_history()
+            except Exception as e:
+                self.logger.warning(f"Failed to prune history on {monitor.__class__.__name__}: {e}")
```
samsja: no need for try expoect here

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -736,6 +736,13 @@ def compute_solve_rates(df):
             step=progress.step,
         )
 
+        # The orchestrator only reads monitor.history for the end-of-run benchmark
```
mikasenghaas: hm i dont like that only need this only bc we have the history for the bench

S1ro1: let's remove this useless comment

mikasenghaas: can we not just only fill the history in bench mode instead? which has bounded steps anyways

src/prime_rl/utils/monitor/wandb.py
```diff
@@ -216,21 +213,10 @@ def log_eval_samples(self, rollouts: list[vf.RolloutOutput], env_name: str, step
...
-            or not self.config.log_extras
-            or not self.config.log_extras.samples
-        ):
-            return
-
-        self.logger.info("Logging final samples to W&B table")
-        df = pd.DataFrame(self.samples)
-        table = wandb.Table(dataframe=df)
-        wandb.log({"final-samples": table})
+        # Per-step samples are already streamed to the incremental `samples` table,
```
S1ro1: also this, just no need here.

S1ro1 (approved):

samsja:
Addressed @mikasenghaas:

- Dropped `log_final_samples` from the Monitor interface and all implementations (it was a no-op everywhere after the prior commit). Removed the orchestrator call site and the `MultiMonitor` delegator.
- Replaced `prune_history()` + per-step orchestrator call with a constructor `keep_full_history` flag on monitors. The orchestrator passes `keep_full_history=config.bench`, so non-bench runs never accumulate history at all. Trainer keeps the default `True`. `Monitor.log()` either appends or replaces in place based on the flag.

Pushed in `799fbd1`.

PR #2377, opened 2026-04-29
feat: agentic length-penalty ablations — weighted shaping over tool-response tokens

hallerite:
## Summary

Refactors `length_penalty` into a **discriminated sub-config** and generalizes token shaping to a weighted blend of model completion + tool-response tokens:

```toml
[orchestrator.advantage.length_penalty]
type = "tokens"
completion_weight = 1.0     # default
tool_response_weight = 1.0  # default — total context

# or for tool-only shaping:
# completion_weight = 0.0
# tool_response_weight = 1.0
```

```toml
[orchestrator.advantage.length_penalty]
type = "turns"
```

The discriminator (`type: "tokens" | "turns"`) mirrors the existing `AdvantageConfig` pattern (`default` | `custom`). Weight semantics live inside the variant they apply to, so there's no cross-field validation needed — the type system enforces correctness.

This unblocks the agentic length-penalty ablations: callers can penalize model output tokens, tool-response tokens, or any weighted mix.

Tool-response tokens are read from a harness-emitted rollout metric (any key matching `*_total_tool_response_tokens`, e.g. RLM's `rlm_total_tool_response_tokens` from [rlm-harness#71](https://github.com/PrimeIntellect-ai/rlm-harness/pull/71), which dedupes across turns / branches / sub-RLMs). Returns `0` when no harness emits the metric — so for envs without tool accounting, only the completion term contributes regardless of weight.

## Changes

- `packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py` — new `TokensLengthPenaltyConfig` (with `completion_weight`, `tool_response_weight`, both `ge=0`, `allow_inf_nan=False`) and `TurnsLengthPenaltyConfig`. `DefaultAdvantageConfig.length_penalty` now takes the discriminated `LengthPenaltyConfig | None`.
- `src/prime_rl/orchestrator/advantage.py` — `default_advantage_fn` dispatches on the sub-config type. Rewards cast to `float32` so fractional weights don't truncate (also fixes a latent crash when rollout rewards arrive as ints). Zero-cost groups fall back to plain GRPO instead of producing NaN advantages.
- `src/prime_rl/orchestrator/vf_utils.py` — added `get_tool_response_len(rollout)` reading from `rollout["metrics"]`.

All 18 advantage tests pass: 12 prior + 1 turns + 5 new (tool-response weighting, fractional/int rewards, zero-cost guard, default-weights equivalence, missing-metric no-op).

## Migration

```diff
-length_penalty = "tokens"
+[orchestrator.advantage.length_penalty]
+type = "tokens"
```

## Reconciliation note

Originally written against the `length_shaping: bool` API. Reworked atop #2384's `length_penalty` enum, then promoted to a sub-config so the weight knobs aren't free-floating fields with cross-validators.

## Supersedes

Closes #2169 (older draft for length shaping; functionally superseded by #2265 which already landed on main; this PR builds on top).


packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -707,6 +707,42 @@ class DefaultAdvantageConfig(BaseModel):
...
+    @model_validator(mode="after")
+    def validate_token_weights(self):
+        if self.length_penalty != "tokens" and (
+            self.length_penalty_completion_weight != 1.0 or self.length_penalty_tool_response_weight != 0.0
+        ):
+            raise ValueError(
+                "`length_penalty_completion_weight` and `length_penalty_tool_response_weight` "
+                "only apply when `length_penalty='tokens'`."
+            )
+        return self
```
samsja: can we get a length penalty sub config instead

samsja (approved):

PR #2384, opened 2026-05-01
feat(advantage): add num_turns option to efficiency shaping

samsja:
## Summary

Adds a `num_turns` option to correctness-gated efficiency shaping and refactors the `AdvantageInputs` API to give custom advantages full access to rollout metadata.

### Config change (breaking)

`orchestrator.advantage.length_shaping: bool` → `orchestrator.advantage.length_penalty: Literal["tokens", "turns"] | None` (default: `None`).

```toml
[orchestrator.advantage]
type = "default"
length_penalty = "turns"   # or "tokens", or omit for no shaping
```

Migration: `length_shaping = true` → `length_penalty = "tokens"`; `length_shaping = false` → omit.

### `AdvantageInputs` change (breaking)

`AdvantageInputs` now holds a single `rollouts: list[list[vf.RolloutOutput]]` (grouped by problem) instead of separate `rewards` / `completion_lengths` / `num_turns` tensors. Custom advantage functions extract whatever metadata they need directly:

```python
rewards = torch.tensor([[r["reward"] for r in g] for g in inputs.rollouts])
```

### Other changes

- `_efficiency_length_shaping` → `_efficiency_shaping(rewards, costs)`: same correctness-gated logic, generic over per-rollout cost.
- `get_num_turns` helper added in `vf_utils.py`.
- Regression test for per-problem advantage isolation (catches cross-group leakage, slicing/transpose bugs, and out-of-order writeback).
- CHANGELOG entries for both breaking changes.
- `docs/bring-your-own-algorithms.md` updated for the new `AdvantageInputs` shape.


src/prime_rl/configs/orchestrator.py
```diff
@@ -691,16 +691,18 @@ class DefaultAdvantageConfig(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
     type: Literal["default"] = "default"
-    length_shaping: Annotated[
-        bool,
+    shaping_metric: Annotated[
```
snimu: i don't like the name `shaping_metric` ; something l like `cost_penalty` or `efficiency_shaping_metric`  or something like that would make it clearer what is being shaped

mikasenghaas: ya its horrible haha, why not just `length_penalty: None : Literal["tokens", "turns"]`

src/prime_rl/orchestrator/advantage.py
```diff
@@ -1,22 +1,25 @@
...
 from prime_rl.configs.orchestrator import AdvantageConfig, CustomAdvantageConfig
-from prime_rl.orchestrator.vf_utils import get_model_completion_len
+from prime_rl.orchestrator.vf_utils import get_model_completion_len, get_num_turns
 from prime_rl.utils.utils import import_object
 
+ShapingMetric = Literal["length", "num_turns"]
+
 
 @dataclass
 class AdvantageInputs:
```
mikasenghaas, line 17: we should prob just make this input be the full `list[vf.RolloutOutput]` so in the future we always have access to all kinds of metadata. only rewards feels too restrictive

mikasenghaas: but its also breaking but whatever

docs/bring-your-own-algorithms.md
```diff
@@ -86,10 +86,15 @@ def my_custom_advantage(inputs: AdvantageInputs, **kwargs) -> AdvantageOutputs:
 ```python
 @dataclass
 class AdvantageInputs:
-    rewards: Float[Tensor, "num_examples rollouts_per_example"]
-    completion_lengths: Int[Tensor, "num_examples rollouts_per_example"]
+    # Rollouts grouped by problem: rollouts[i][j] is the j-th rollout for problem i.
+    rollouts: list[list[vf.RolloutOutput]]
+
+    @property
+    def rewards(self) -> Float[Tensor, "num_examples rollouts_per_example"]: ...
```
samsja: do we still need rewards here ?

docs/bring-your-own-algorithms.md
```diff
@@ -86,10 +86,12 @@ def my_custom_advantage(inputs: AdvantageInputs, **kwargs) -> AdvantageOutputs:
 ```python
 @dataclass
 class AdvantageInputs:
-    rewards: Float[Tensor, "num_examples rollouts_per_example"]
-    completion_lengths: Int[Tensor, "num_examples rollouts_per_example"]
+    # Rollouts grouped by problem: rollouts[i][j] is the j-th rollout for problem i.
+    rollouts: list[list[vf.RolloutOutput]]
```
mikasenghaas, line 90: i think we should move towards `list[list[...]]` as this allows dynamic group sizes (e.g just train on 6/8 rollouts if 2 failed) but right now our datastrcuture is always flat so wondering if we should keep it like this?

mikasenghaas: ah nvm we alr had a 2d tensor, all good then

src/prime_rl/orchestrator/advantage.py
```diff
@@ -127,12 +142,8 @@ def compute_advantages(
         return
 
     advantage_fn = setup_advantage_fn(advantage_config)
-    completion_lengths = [get_model_completion_len(r) for r in rollouts]
-
-    inputs = AdvantageInputs(
-        rewards=torch.tensor(rewards).view(-1, samples_per_problem),
-        completion_lengths=torch.tensor(completion_lengths).view(-1, samples_per_problem),
-    )
+    grouped = [rollouts[i : i + samples_per_problem] for i in range(0, len(rollouts), samples_per_problem)]
```
mikasenghaas, line 145: i think this is correct but lets send an agent over it one more time to make sure we dont compute advantages cross group lol

mikasenghaas (approved):

PR #2389, opened 2026-05-02
[Weight Transfer] NIXL + MX Integration

S1ro1:
Consolidates the NIXL + MX weight-transfer work into one implementation, now rebased on current `main`.

Recent branch state:
- Keeps the NIXL/MX transport path, Model Express rendezvous wiring, and GLM/Qwen transfer conversion specs.
- Includes the GLM NIXL cast stabilization via explicit `bf16_cast` / `fp32_cast` conversion handling.
- Preserves main's DeepGEMM MoE env-var behavior in the generated inference launch scripts.
- Keeps `inference_model_name` scoped to the NIXL trainer weight-broadcast config, where the transfer planner uses it.

## Design

image

image

## Results

image

BF16 <-> BF16 + BF16 <-> FP8 Qwen3 30B on dummy task.

image

Qwen3 235B A22B BF16 <-> BF16 transfer time comparison

## Verification

- `uv sync --all-extras`
- `uv run ruff check src/prime_rl/inference/vllm/server.py src/prime_rl/inference/vllm/worker/nixl_mx.py src/prime_rl/orchestrator/orchestrator.py src/prime_rl/orchestrator/scheduler.py src/prime_rl/trainer/models src/prime_rl/trainer/rl/broadcast/nixl_mx.py src/prime_rl/transport src/prime_rl/utils/client.py src/prime_rl/utils/elastic.py tests/unit/train/models tests/unit/transport`
- `uv run pytest tests/unit/train/models/conversions/test_qwen3_moe.py tests/unit/train/models/slots/test_qwen3_moe.py`
- `uv run ruff check src/prime_rl/entrypoints/rl.py packages/prime-rl-configs/src/prime_rl/configs/trainer.py`
- `uv run rl @ configs/glm51_math_nixl_dpep16_10node/rl.toml --dry-run`

---

> [!NOTE]
> **High Risk**
> High risk because it adds a new distributed RDMA-based weight transfer path (NIXL + Model Express) and changes orchestrator/trainer/inference coordination, touching CUDA memory allocation, network rendezvous, and weight update lifecycle.
> 
> **Overview**
> Adds a new **`nixl_mx` weight broadcast transport** that pushes weights via NIXL RDMA with Model Express as the rendezvous/metadata plane, wiring it through configs, trainer broadcast setup, orchestrator scheduling, and the vLLM inference server.
> 
> The trainer now builds per-parameter transfer *slots* from model-defined `ConversionSpec`s (including BF16/FP32 casts and FP8 blockwise quantization with scale buffers), publishes tensor descriptors + layout over MX, and executes a validated per-peer RDMA write plan; inference workers register destination tensors with NIXL, publish expert maps, and apply in-place updates after the push.
> 
> Operational support is included: a docker-compose stack to run Model Express + Redis, SLURM templates that auto-launch/tear down MX when `nixl_mx` is enabled, early CUDA runtime preload shims for allocator stability, and unit tests covering conversion resolution, slot allocation, and MX rendezvous end-to-end.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 79ea824d85182bc66a0c725e3fa637bd01f17d5e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/templates/multi_node_rl.sbatch.j2
```diff
@@ -20,6 +20,9 @@
 #SBATCH --exclusive
 #SBATCH --output={{ output_dir }}/job_%j.log
 #SBATCH --error={{ output_dir }}/job_%j.log
+{%- if use_nixl_mx_broadcast %}
+#SBATCH --signal=B:TERM@30
```
samsja, line 24: what is this about ?

samsja (commented):
pretty complex pr but seems well design and scope. I think that adding a .md with all the detail is more than welcome here. Also wondering if we can merge the mx_rendezvous.p and nixl_agent.py under the same file ?

src/prime_rl/transport/transport_plan.py
```diff
@@ -0,0 +1,232 @@
...
+        self.local_preps: dict[str, Any] = {}
+        self.remote_preps: dict[tuple[str, str], Any] = {}
+
+        self._validate_descriptors(slots, peers, peer_tensor_sizes, peer_tensor_dtypes)
+
+        for slot in slots:
+            self.writes.extend(slot.build_writes(peers))
+
+        for peer in self.peers:
+            name = agent.add_remote_agent(peer.agent_metadata)
```
KavinKrishnan, line 56: On GCP GB200 (and any multi-NIC IB fabric where rdma-0..N are separate L3 subnets), this unconditional full-mesh `add_remote_agent` over every `peer.agent_metadata` produces `NIXL_ERR_REMOTE_DISCONNECT` on the off-subnet pairs as soon as the first write fires. We've reproduced this on Qwen3-30B-A3B with 4 trainer x 4 inference ranks.

The runtime patch we applied (`patch_nixl_mx.py` in our configmap) filters `peer_metadata` to same-`worker_rank` peers before this loop. The clean fix is to plumb a `same_rank_only: bool` (or `peer_filter: Callable[[PeerInfo], bool]`) into `TransportPlan` and let the trainer config control it. On single-subnet fabrics it's a no-op; on multi-NIC fabrics it's a hard requirement.

We've codified the same fix at the rendezvous layer on `KavinKrishnan/prime-rl:kavink/post-2389-phase2-rendezvous-fixes` — happy to send a follow-up PR threading it through to `TransportPlan` once you tell us the API shape you prefer.

src/prime_rl/transport/mx_rendezvous.py
```diff
@@ -0,0 +1,185 @@
...
+            poll_interval: Seconds between ``ListSources`` polls.
+        """
+        import logging
+
+        _log = logging.getLogger("prime_rl.transport.mx_rendezvous")
+        deadline = time.monotonic() + timeout
+        peer_id = self._identity(self.peer_role)
+        _logged = False
+        while True:
+            resp = self.client.list_sources(peer_id, status_filter=status)
```
KavinKrishnan, line 117: `list_sources` returns every catalog row matching the role identity, including stale ones left by previous `mx_source_id` rotations (one of which is guaranteed on any pod restart). When two rows share `worker_rank` and the freshness signal isn't checked, `add_remote_agent` downstream may pick the stale one and immediately hit `NIXL_ERR_NOT_ALLOWED`.

We've reproduced this on GB200 across two pod restart cycles. The fix is to dedup by `(worker_rank,)`, keeping the entry with the largest `updated_at`. Our [post-2389-phase2 branch](https://github.com/KavinKrishnan/prime-rl/blob/kavink/post-2389-phase2-rendezvous-fixes/src/prime_rl/transport/mx_rendezvous.py) ships this as `_freshest_per_rank(instances, *, metas)` along with a `_collect_updated_at` helper that does the GetMetadata fan-out. Happy to PR if you like the shape.

src/prime_rl/transport/mx_rendezvous.py
```diff
@@ -0,0 +1,185 @@
...
+            if len(resp.instances) >= self.peer_world_size:
+                return list(resp.instances)
+            if time.monotonic() >= deadline:
+                raise TimeoutError(
+                    f"timed out after {timeout}s waiting for {self.peer_world_size} "
+                    f"{self.peer_role!r} peers (saw {len(resp.instances)})"
+                )
+            time.sleep(poll_interval)
+
+    def wait_for_all_peers_ready(
```
KavinKrishnan, line 135: The `MxRendezvous` is missing a `modelexpress.metadata.heartbeat.HeartbeatThread` lifecycle. Without one, the MX server's reaper can't transition crashed workers to `STALE` — crashed pods leave permanent `READY` rows that wedge subsequent restarts.

This was the second of the two GB200 runtime patches. Suggested shape: spawn the thread inside `publish()` keyed on `(mx_source_id, worker_id, worker_rank)`, expose a `close()` method to stop it on graceful shutdown. `nixl_manager` arg can be `None` since prime-rl drives NIXL outside the MX manager.

We've shipped this on our follow-up branch with 11 new unit tests (`tests/unit/transport/test_mx_rendezvous_phase2.py`) — happy to send a PR.

src/prime_rl/inference/vllm/worker/nixl_mx.py
```diff
@@ -0,0 +1,95 @@
...
+
+        logger.info(
+            f"NIXL+MX init: rank={global_rank} tensors={len(self._descriptors)} "
+            f"experts={sum(len(v) for v in expert_map.values())} model={inference_model_name}"
+        )
+
+    @torch.no_grad()
+    def update_weights_from_path(self, weight_dir: str | None = None) -> None:
+        """Block until the trainer's RDMA push completes, then recompute the MLA absorbed weights and return, orchestrator can then call `/resume`"""
+        self.rendezvous.wait_for_all_peers_ready(timeout=1200)
```
KavinKrishnan, line 92: `timeout=1200` is hardcoded here but the trainer side uses `self.config.timeout` ([trainer/rl/broadcast/nixl_mx.py:148](https://github.com/PrimeIntellect-ai/prime-rl/blob/79ea824d85182bc66a0c725e3fa637bd01f17d5e/src/prime_rl/trainer/rl/broadcast/nixl_mx.py#L148)). On a slow first-step compile (we've seen 30B-A3B FP8 first-step ~18 min in cold-start), the inference side times out before the trainer does, and the next refit is racey.

Suggest: read `timeout` from the same config the trainer reads (or fall back to env var `PRIME_RL_NIXL_MX_TIMEOUT`) so both sides time out in lockstep.

src/prime_rl/inference/vllm/worker/nixl_mx.py
```diff
@@ -0,0 +1,95 @@
...
+        torch.cuda.synchronize(self.device)
+        update_mla_absorbed_weights(self.raw_model)
```
KavinKrishnan, line 94: `update_mla_absorbed_weights` is called unconditionally, but it only applies to DeepSeek-style MLA models. Running it on Qwen3-30B-A3B (no MLA) is harmless today only because the function early-returns when it can't find the absorbed-weight names — that early-return is silently load-bearing. If someone tightens the MLA codepath this becomes a runtime crash on non-MLA models.

Suggest gating on `hasattr(self.raw_model, ...)` or an explicit `model_kind` flag from the inference config so the call is opt-in for MLA-class models only.

src/prime_rl/trainer/rl/broadcast/nixl_mx.py
```diff
@@ -0,0 +1,179 @@
...
+            self.lazy_init(model)
+            self.rendezvous.set_status(p2p_pb2.SOURCE_STATUS_INITIALIZING)
+
+        if self.world.is_master:
+            for idx in self._multi_run_manager.used_idxs:
+                if self._multi_run_manager.ready_to_update[idx]:
+                    self._multi_run_manager.ready_to_update[idx] = False
+            self.rendezvous.wait_for_all_peers_ready(role="orchestrator", timeout=self.config.timeout)
+
+        dist.barrier()
```
KavinKrishnan, line 152: On HSDP setups where only the primary HSDP rank does the publish, `dist.barrier()` at line 152 is reached by non-primary ranks before `lazy_init` has completed on the primary. If `lazy_init` takes longer than the NCCL barrier timeout (Qwen3-30B-A3B first-cycle observed ~140s on GB200), non-primary ranks emit a `Timeout(s) waiting for response` and abort. Same pattern at line 179.

Two options: (a) move `self.lazy_init(model)` outside the `is_primary_hsdp_rank` guard so every rank does the (idempotent) initialisation, or (b) bump the trainer-side NCCL barrier timeout to match `self.config.timeout`. Option (a) matches what we do in NeMo-RL's MX integration.

KavinKrishnan (commented):
**Summary** (kavink @ NVIDIA — validation feedback from GB200 Qwen3-30B-A3B bring-up)

We've validated this PR on a 4-trainer × 4-inference GCP GB200 cluster after applying two runtime patches (same-rank-only peer filter + freshest-per-rank dedup) that the inline comments below recommend codifying. With patches: ~80 ms NIXL push per step, end-to-end refit ~12.5 s, 10/10 RL steps clean. Without patches: `NIXL_ERR_REMOTE_DISCONNECT` from cross-subnet routing + `NIXL_ERR_NOT_ALLOWED` from stale-READY peers, every time.

The 6 inline comments fall into three buckets:

1. **Multi-NIC fabric handling** (`transport_plan.py:56`, `mx_rendezvous.py:117`, `mx_rendezvous.py:135`) — the GB200/EFA path. Same-rank-only filter, freshest-per-rank dedup, and heartbeat lifecycle. We've codified all three on [`KavinKrishnan/prime-rl:kavink/post-2389-phase2-rendezvous-fixes`](https://github.com/KavinKrishnan/prime-rl/tree/kavink/post-2389-phase2-rendezvous-fixes) with 11 unit tests; happy to PR once you ack the API shape.

2. **Lifecycle / timeout hygiene** (`inference/vllm/worker/nixl_mx.py:92`, `:94`, `trainer/rl/broadcast/nixl_mx.py:152`) — hardcoded 1200s timeout that races with trainer-side configurable timeout; unconditional `update_mla_absorbed_weights` that depends on a silent early-return on non-MLA models; HSDP barrier ordering hazard.

3. **Follow-up RFC** — we have a separate plan at [`docs/proposals/post-pr2389-kernel-compile-plan.md`](https://github.com/KavinKrishnan/prime-rl/blob/kavink/post-2389-kernel-compile-plan/docs/proposals/post-pr2389-kernel-compile-plan.md) covering (a) graduating the in-tree `MxRendezvous` reimplementation onto the published ModelExpress fat clients (`MxV2TrainingPublisher` / `MxV2RefitReceiver`, –185 LOC), (b) a compile-target registry so trainer-side cutlass compile becomes one option of several (Matej's concern from last sync), and (c) mixed-TP slice discovery. None of it fights this PR — it builds on top.

For the immediate merge: nothing on this list blocks; the runtime patches keep #2389 working on GB200 today, and a single-NIC fabric should hit none of bucket 1. Phase-2 PRs from us would be additive.

src/prime_rl/transport/transport_plan.py
S1ro1: This will though require multiple changes to the transfer - if we can only do rank N to rank N transfer, the trainer/inference world shapes have to match? Which is not the case usually.

src/prime_rl/transport/mx_rendezvous.py
S1ro1: We can integrate this eventually, our current slurm setup produces a fresh MX server per restart. Would target fault-tolerance and being restart tolerant later

src/prime_rl/transport/mx_rendezvous.py
S1ro1: https://github.com/PrimeIntellect-ai/prime-rl/pull/2389#discussion_r3313978032 as here, would add later

PR #2391, opened 2026-05-02
[DYNAMO] fix: NFS-safe filesystem broadcast for LoRA adapters

AmeenP:
## Summary

Fixes a visibility-ordering race in the filesystem-based weight broadcast that surfaces on shared filesystems (NFS / CephFS / Kubernetes PVC volumes). On a shared filesystem, a write on the trainer node is not immediately visible on the orchestrator/inference node. The previous code touched `STABLE` immediately after `save_state_dict`, so the orchestrator could observe the sentinel before `adapter_model.safetensors` / `adapter_config.json` were readable cluster-wide — causing `LoRAAdapterNotFoundError` when `/v1/rl/load_lora_adapter` was called.

Fix: fsync each adapter file, fsync the directory's dentries, touch `STABLE`, then fsync the directory once more so the sentinel itself is durable.

`_fsync_path` is a small helper. `OSError` is swallowed deliberately — fsync is a hint to the kernel, not the correctness gate on its own. The actual correctness gate is the *ordering*: write all data → flush → publish sentinel. On filesystems that don't support directory fsync, the `OSError` is harmless; on filesystems that do, the call is load-bearing.

## Scope

One file changed, 32 lines added. No new config knobs, no behavior change for local-filesystem deployments.

A companion PR (#2395) adds the `weight_broadcast.keep_recent` config knob that addresses a *separate* LoRA failure mode on the same code path (the trainer racing ahead and `rmtree`-ing an adapter dir while a vLLM generate request is still lazily loading it). The two are independent and can land in either order. Together they cover both classes of `LoRAAdapterNotFoundError` we've seen on shared-filesystem deployments.

### DIFF
--- a/src/prime_rl/trainer/rl/broadcast/filesystem.py
+++ b/src/prime_rl/trainer/rl/broadcast/filesystem.py
@@ -1,3 +1,4 @@
+import os
 import shutil
 import time
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


samsja:

PR #2393, opened 2026-05-02
[DYNAMO] feat(helm): tolerations, imagePullSecrets, configMap mount

AmeenP:
## Summary

Adds three commonly-needed knobs to the prime-rl Helm chart, all opt-in. Verified with `helm template`: rendered output is byte-identical to current main when defaults are used.

- **`<component>.tolerations`** on orchestrator / inference / trainer — schedule onto tainted nodes (dedicated GPU pools, spot/preemptible nodes, etc.) without forking the chart.
- **`<component>.imagePullSecrets`** so private-registry images can be pulled without baking secrets into the namespace's default ServiceAccount.
- **`orchestrator.configMap`** mounts an existing ConfigMap at `/configs` in the orchestrator container, so TOML configs can be shipped without rebuilding the image.

Also quotes the `command` arg in `autoStart` mode so multi-line shell strings don't break YAML parsing — this was a bug waiting to happen.

The orchestrator's `volumeMounts` / `volumes` blocks are now gated on `storage.enabled OR configMap` so we don't emit empty lists when both are off (the naive version would render `volumes:` with no items, which kubectl rejects).

## Verification

```
helm template t k8s/prime-rl > /tmp/defaults.yaml                # 421 lines, identical to main
helm template t k8s/prime-rl --set 'orchestrator.configMap=cm' \
  --set 'orchestrator.tolerations[0].key=nvidia.com/gpu' \
  --set 'orchestrator.imagePullSecrets[0].name=regcred' > /tmp/over.yaml  # 433 lines, fields land correctly
```

### DIFF
--- a/k8s/prime-rl/templates/deployment.yaml
+++ b/k8s/prime-rl/templates/deployment.yaml
@@ -26,6 +26,14 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.orchestrator.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.orchestrator.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
       - name: prime-rl-orchestrator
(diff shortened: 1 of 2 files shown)


samsja:

PR #2395, opened 2026-05-02
[DYNAMO] feat: weight_broadcast.keep_recent retention knob for LoRA

AmeenP:
## Summary

Adds an optional `weight_broadcast.keep_recent` knob that decouples broadcast-dir retention from `max_async_level`. Default behavior is unchanged (when unset, retention falls back to `max_async_level` exactly as before).

## Problem

`maybe_clean` previously deleted `step_{N - max_async_level - 1}`. That's the right bound for full-weight broadcasts, but **not** for LoRA. vLLM lazy-loads the adapter inside `_prepare_inputs`, so a generate request admitted under an older adapter can page that adapter dir in *after* the orchestrator has moved on. The trainer can race ahead and `rmtree` the adapter directory just before the inference worker tries to read it.

## Fix

New optional `weight_broadcast.keep_recent: int | None` (default `None`):
- `None` → fall back to `max_async_level` (no behavior change for existing users).
- Set to `max_async_level + 1` or `+2` for LoRA + filesystem broadcast.

A `model_validator` rejects `keep_recent < max_async_level` with a message that names the failure mode (`LoRAAdapterNotFoundError`), since a smaller-than-staleness retention window is always a misconfiguration.

`maybe_clean`'s int parameter is renamed `async_level` -> `retention` to reflect that it can now come from either knob.

## Relationship to other PRs

A companion PR (#2391) addresses a *different* LoRA failure mode on the same code path — an NFS visibility race where the `STABLE` sentinel becomes visible before the adapter files it points to. The two are independent and can land in either order. Together they cover both classes of `LoRAAdapterNotFoundError` we've seen on shared-filesystem deployments.

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/trainer.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/trainer.py
@@ -740,6 +740,21 @@ class FileSystemWeightBroadcastConfig(BaseWeightBroadcastConfig):
     save_format: Annotated[
         Literal["safetensors", "torch"], Field(description="The format to save the weight checkpoint in.")
     ] = "safetensors"
+    keep_recent: Annotated[
+        int | None,
+        Field(
+            ge=0,
+            description=(
+                "Number of recent broadcast step directories to keep on disk. At trainer step N, "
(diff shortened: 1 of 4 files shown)


samsja:

PR #1320, opened 2025-11-20
Support broadcasting just lora weights and drop support for merged weights

Jackmin801:
> [!NOTE]
> Adds adapter-only filesystem weight broadcasting and runtime LoRA loading, wires LoRA names through orchestrator/scheduler, removes merged-weight path, and updates tests/CI accordingly.
> 
> - **LoRA broadcasting and loading**:
>   - Add `adapter_only` option to weight broadcast configs (`trainer/rl/config.py`); filesystem broadcaster now saves only adapters and writes `adapter_config.json`; NCCL path disallows adapter-only.
>   - Introduce `get_adapter_state_dict` and adapter filename handling in `trainer/weights.py`; drop merged-LoRA gathering path; simplify `gather_weights_on_master`.
>   - Pass `lora_config` into broadcaster setup and persist with adapter saves.
> - **Orchestrator & Scheduler**:
>   - Add `OrchestratorConfig.lora_name` and thread through `Scheduler`; use LoRA name as `model_name` after updates.
>   - `update_weights` now accepts `lora_name` and uses new `/v1/load_lora_adapter`; avoid base reload when using LoRA.
> - **Inference (vLLM server)**:
>   - Inject hack to bypass adapter already-loaded checks to allow in-place LoRA updates.
> - **Trainer**:
>   - Broadcast weights each step using `adapter_only` flag; checkpoint manager unchanged but aligned with new adapter flow.
> - **Tests/CI**:
>   - Add integration tests for LoRA RL with dynamic adapter loading; separate LoRA tests in CI; capture vLLM/stdout logs for debugging.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7857428f9e02b9314087b5f609c41f135554cc2c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #2396, opened 2026-05-02
[DYNAMO] chore: remove dead unload_lora_adapter

AmeenP:
## Summary

`unload_lora_adapter` in `src/prime_rl/utils/client.py` has been dead code since November 2025. Removing it.

## History

- #1320 (`2aa47ae6b`, 2025-11-21) introduced `unload_lora_adapter` along with a call to it at the end of `orchestrate()` in `orchestrator.py`. The PR also shipped with a `# TODO: first one can fail, subsequent ones should succeed` comment and a commented-out `raise_for_status()` — i.e. the author already knew the call was unreliable.
- `220b50a38` ("dont unload for now"), same day, **same author**, removed the call site and the import. The function has had zero callers ever since.

```bash
$ git log --all --oneline -S'unload_lora_adapter(' -- src/
2aa47ae6b Support broadcasting just lora weights ...   # adds function + caller
220b50a38 dont unload for now                          # removes caller, leaves function
$ rg 'unload_lora' src/ tests/ examples/
src/prime_rl/utils/client.py:412 ...   # only the function itself
```

## Why delete, not fix

The function has two issues that surface only when reactivated:

1. **Silent-failure pattern.** The `raise_for_status()` is commented out. Per `AGENTS.md`: *"Minimal try/except: let errors propagate — silent failures hide bugs."*
2. **Path bug against Dynamo.** Line 418 is `/v1/unload_lora_adapter` (with `/v1/` prefix). Line 403 — its sibling `load_lora_adapter` — is bare `/load_lora_adapter`. Against vLLM, both paths work because vLLM aliases them. Against a Dynamo frontend with `admin_base_url = http://.../v1/rl`, httpx joins the path onto the base, producing `http://.../v1/rl/v1/unload_lora_adapter`, which doesn't exist. The 404 is invisible because `raise_for_status()` is commented out.

Fixing in place would mean making dead code "correct" without any way to validate the fix end-to-end. When end-of-training adapter unload is actually needed, the right move is to re-add it with a real call site and a real integration check, not to maintain a half-working stub in the meantime.

## Scope

13 lines removed in one file. No callers, no imports, no tests reference it. Verified.

### DIFF
--- a/src/prime_rl/utils/client.py
+++ b/src/prime_rl/utils/client.py
@@ -409,19 +409,6 @@ async def load_lora_adapter(admin_clients: list[AsyncClient], lora_name: str, lo
     await asyncio.gather(*[_load_lora_adapter(admin_client) for admin_client in admin_clients])
 
 
-async def unload_lora_adapter(admin_clients: list[AsyncClient], lora_name: str) -> None:
-    """Make a HTTP post request to the vLLM server to unload a LoRA adapter."""
-    logger = get_logger()
-
-    async def _unload_lora_adapter(admin_client: AsyncClient) -> None:
-        logger.debug(f"Sending request to unload LoRA adapter {lora_name}")
-        await admin_client.post("/v1/unload_lora_adapter", json={"lora_name": lora_name})
-        # TODO: The first one can fail, but subsequent ones should succeed.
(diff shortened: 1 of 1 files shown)


samsja:

PR #2397, opened 2026-05-02
[DYNAMO] refactor: extract admin endpoints into AdminAPI Protocol

AmeenP:
## Summary

Pure refactor with no behavior change. Extracts admin-endpoint construction into a typed `AdminAPI` Protocol so a `DynamoAdminAPI` follow-up can drop in without touching every call site or scattering `if backend == "dynamo"` branches across `client.py`.

## What changes

- **`AdminAPI` Protocol** — 7 admin operations (`health`, `list_models`, `pause`, `resume`, `update_weights`, `load_lora_adapter`, `init_broadcaster`).
- **`VLLMAdminAPI`** — single implementation, byte-identical paths/params/timeouts/raise-for-status to the previous inline code.
- **`_DEFAULT_ADMIN: AdminAPI = VLLMAdminAPI()`** — module singleton so existing call sites need zero changes.
- Each free admin function takes a keyword-only `admin: AdminAPI = _DEFAULT_ADMIN` and delegates HTTP construction to it.

Inlined two functions while I was here:
- `_pause_engines` / `_resume_engines` — were one-line `gather()` wrappers around `admin.pause` / `admin.resume`, with a single caller. Inlined into `update_weights`. Net: -12 lines, callsite gets clearer (the pause→try→resume pattern is visible inline).

## What does *not* change

- `orchestrator.py`, `scheduler.py`, `elastic.py` are untouched — they pick up `VLLMAdminAPI` behavior via the default arg.
- `tests/unit/utils/test_client.py` is untouched — `mock_client.post` is called with the same args, just routed through `VLLMAdminAPI.load_lora_adapter`.
- No config schema changes. No new TOML fields.

## Why this shape (and not alternatives)

| Option | Why not |
|---|---|
| Per-endpoint paths in TOML | Every deployment spells out 6+ routes. Doesn't address request-shape differences (only paths). Contract lives in config — drifts on rename. |
| `if backend == "dynamo"` inside each free function | 5+ scattered branches. Same divergence the Protocol localises, but uglier and untyped. |
| Drop the Protocol, just have two classes | Loses the typed seam — callers can't be `admin: AdminAPI`. Implicit inheritance also creates the risk of accidentally inheriting vLLM-specific behavior. |

The Protocol earns its keep *now*, not hypothetically: there are already two backends (vLLM, Dynamo) with materially different surfaces, and the existing `use_token_client`, `skip_model_check`, and bis-branch's `use_prefix_cache_salt` flags are all the same shape ("backend has different conventions, papered over by a config flag").

## Follow-ups (separate PRs)

1. Add `client.backend: Literal["vllm", "dynamo"] = "vllm"` config field; pick the right `AdminAPI` in `setup_admin_clients`.
2. Add `DynamoAdminAPI` with the correct paths for Dynamo's admin frontend (some under `/v1/rl/`, some not).
3. Migrate scattered shape flags into backend defaults.

Each is independently shippable. This PR is the foundation.

## Verification

- `ruff format` + `ruff check` clean.
- Hand-traced `test_load_lora_adapter_succeeds_on_first_attempt`: same `mock_client.post("/load_lora_adapter", json=..., timeout=...)` call via `VLLMAdminAPI.load_lora_adapter`.
- Local `uv run pytest` is unrunnable due to an unrelated `uv.lock` parse issue (`exclude-newer = "7 days"` from #2382 + an older uv binary). CI please verify.

## Stat

`+156 / -75` (one file). The ~30 lines of net growth are the Protocol contract — exactly what the `DynamoAdminAPI` follow-up will need.

### DIFF
--- a/src/prime_rl/utils/client.py
+++ b/src/prime_rl/utils/client.py
@@ -16,6 +16,106 @@ from prime_rl.configs.shared import ClientConfig
 from prime_rl.utils.logger import get_logger
 
 
+class AdminAPI(Protocol):
+    """Admin endpoints for an inference backend.
+
+    Per-method: construct one HTTP call. Per-server parallelism, retry, and
+    raise-for-status policy live in the caller.
+    """
+
+    async def health(self, client: AsyncClient) -> None: ...
+    async def list_models(self, client: AsyncClient) -> list[dict]: ...
+    async def pause(self, client: AsyncClient) -> None: ...
+    async def resume(self, client: AsyncClient) -> None: ...
+    async def update_weights(self, client: AsyncClient, weight_dir: str | None) -> None: ...
+    async def load_lora_adapter(
+        self,
+        client: AsyncClient,
+        lora_name: str,
+        lora_path: str,
+        *,
+        timeout: httpx.Timeout,
+    ) -> None: ...
+    async def init_broadcaster(
+        self,
+        client: AsyncClient,
+        *,
+        host: str,
+        port: int,
+        rank_offset: int,
+        inference_world_size: int,
+        timeout: int,
+        quantize_in_weight_transfer: bool,
+    ) -> None: ...
+
+
+class VLLMAdminAPI:
+    """vLLM admin endpoints."""
+
+    async def health(self, client: AsyncClient) -> None:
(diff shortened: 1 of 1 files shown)


samsja:

PR #2398, opened 2026-05-02
[DYNAMO] feat: DynamoAdminAPI + client.backend selector

AmeenP:
## Summary

Second AdminAPI implementation, foreshadowed by #2397. Default behavior is byte-identical to today; opt-in via `client.backend = "dynamo"`.

**Stacked on #2397** — base branch is `refactor/admin-api-protocol`. Once #2397 merges, rebase against `main`.

## What's actually different about Dynamo

| Method | Dynamo behavior | DynamoAdminAPI |
|---|---|---|
| `health`, `pause`, `resume`, `update_weights`, `load_lora_adapter`, `init_broadcaster` | Same relative paths as vLLM; admin_base_url is what differs (`…/v1/rl` vs `…/v1`) | Inherited from `VLLMAdminAPI` — httpx joins relative paths onto admin_base_url. |
| `list_models` | Served at OpenAI-compat root (`/v1/models`), **not** under `/v1/rl/` | Overridden — sends absolute URL to bypass admin prefix. |

That is it: one method override.

## Why this fixes a real bug

Today, against Dynamo:

```toml
admin_base_url = ["http://...:8000/v1/rl"]
skip_model_check = true   # workaround
```

The `skip_model_check = true` papers over the fact that `VLLMAdminAPI.list_models` resolves to `…/v1/rl/v1/models` (httpx joins, does not host-anchor), which 404s. With `backend = "dynamo"` set, `DynamoAdminAPI.list_models` constructs an absolute URL anchored at the host, so the model check actually runs.

## Wiring

- `client.backend: Literal["vllm", "dynamo"] = "vllm"` added to `ClientConfig`.
- `setup_admin_api(client_config)` selector.
- Static and elastic inference pools build a backend-specific `AdminAPI` and pass it to model readiness and weight/LoRA update paths.

## What's NOT in this PR

- Renderer transport config is split into #2420.
- The `_check_server_health` `/health` call in `elastic.py` is left inline; it has different raise-for-status semantics.
- No path migration plan for Dynamo moving endpoints out of `/v1/rl/`; that would be a method override on `DynamoAdminAPI`.

## Verification

- `git diff --check`
- `uv run --no-sync python -m py_compile src/prime_rl/utils/client.py src/prime_rl/utils/elastic.py src/prime_rl/configs/shared.py`

`uv run --no-sync ruff check src/prime_rl/utils/client.py src/prime_rl/utils/elastic.py src/prime_rl/configs/shared.py` is currently blocked by import sorting in `client.py` and `elastic.py`.

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/shared.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/shared.py
@@ -310,6 +310,19 @@ class ClientConfig(BaseConfig):
         ),
     ] = None
 
+    backend: Annotated[
+        Literal["vllm", "dynamo"],
+        Field(
+            description=(
+                "Inference backend selector. Picks the AdminAPI implementation used for "
+                "pause/resume/update_weights/load_lora_adapter/list_models. Default 'vllm' "
+                "matches prime-rl's bundled vLLM frontend. 'dynamo' targets NVIDIA Dynamo's "
+                "native /v1/rl/* admin routes (DYN_ENABLE_RL=true) and routes /v1/models to "
+                "the OpenAI-compat root rather than under /v1/rl/."
+            ),
+        ),
+    ] = "vllm"
+
     elastic: Annotated[
         ElasticConfig | None,
         Field(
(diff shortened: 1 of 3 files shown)


samsja:

PR #2416, opened 2026-05-04
feat: split prime-rl-configs into separate workspace package

samsja:
## Summary

Split `prime_rl.configs.*` into its own pip-installable package so downstream consumers can depend on the config schema (~12 deps) without pulling the full GPU/ML stack (~329 deps).

- New uv workspace member at `packages/prime-rl-configs/` ships `prime_rl.configs.*` plus the small slim utils (`utils/config.py`, `utils/validation.py`).
- Root `prime-rl` keeps everything else (including `utils/logger.py`) and depends on `prime-rl-configs`.
- Both wheels merge into a single PEP 420 namespace package `prime_rl/` at install time (no `__init__.py` at `prime_rl/` or `prime_rl/utils/`).

End users:
- `pip install git+https://github.com/PrimeIntellect-ai/prime-rl.git#subdirectory=packages/prime-rl-configs` → schema only, ~12 deps.
- `pip install prime-rl` → full stack (transitively gets `prime-rl-configs`).
- `uv sync` from repo root installs both editably as workspace members.

## Using the slim package

Distribution name (`prime-rl-configs`) ≠ import path (`prime_rl.configs.*`) — the package contributes to the shared `prime_rl` namespace.

```python
from pydantic_config import cli
from prime_rl.configs.rl import RLConfig

config = cli(RLConfig, args=["@", "path/to/rl.toml"])
```

Other config classes live alongside `RLConfig` under `prime_rl.configs.*` (`sft`, `inference`, `orchestrator`, `trainer`, `env_server`). `import prime_rl` on its own succeeds but is empty — it's a PEP 420 namespace package with no top-level attributes. Always import a submodule. README at `packages/prime-rl-configs/` covers this.

### Layout
- `src/prime_rl/configs/` → `packages/prime-rl-configs/src/prime_rl/configs/`
- `src/prime_rl/utils/{config,validation}.py` → `packages/prime-rl-configs/src/prime_rl/utils/`
- New `packages/prime-rl-configs/{pyproject.toml,README.md}`

### Code
- Move `rgetattr`/`rsetattr` from `utils/utils.py` → `utils/config.py` — was the only heavy import the configs needed; this breaks the dep.
- Replace 7 `get_logger().warning(...)` sites in `configs/rl.py` + `configs/orchestrator.py` with stdlib `warnings.warn(...)` (deprecation notices use `DeprecationWarning`). This drops the `loguru` dep from the slim wheel and lets `utils/logger.py` stay in the heavy package.
- Replace `Path(prime_rl.__file__).parent / 'templates'` (None on a namespace package) with `find_package_resource('templates')`, which iterates `prime_rl.__path__` to locate the heavy wheel's data dir. Returns None on slim installs.
- Delete `src/prime_rl/__init__.py` so `prime_rl` becomes a namespace package; move its `import _compat` side-effect into the actual entrypoints (`orchestrator.py`, `entrypoints/{rl,sft}.py`, `utils/cp.py`, `trainer/models/layers/ring_attn.py`; `trainer/{rl,sft}/train.py` already had it).

### Packaging / CI / tooling
- Root `pyproject.toml`: declare `[tool.uv.workspace]`, add `prime-rl-configs` as a dep + workspace source, add `[tool.hatch.build.targets.wheel] packages = ['src/prime_rl']` (hatchling can't auto-discover a namespace package).
- Root `pyproject.toml`: add `[tool.ruff.lint.isort] known-first-party = ['prime_rl']`. Without this, ruff classifies `prime_rl.configs.*` as third-party (since it lives outside `src/`) and reorders imports across the codebase.
- Slim `pyproject.toml`: pin `pydantic-config @ git+https://github.com/samsja/pydantic_config.git`. PyPI's `pydantic-config` is a *different project* that doesn't expose `BaseConfig` / `cli`, so without this pin plain `pip install` resolves the wrong package and configs.* imports crash.
- `Dockerfile.cuda`: also `COPY packages/` so the workspace member is available during `uv sync`.
- `.github/workflows/cpu_tests.yaml`: new `slim-configs-install` job that **builds the slim wheel and installs it from /tmp** so uv can't read the root workspace's `[tool.uv.sources]` and silently apply them — this catches dep-resolution bugs that only show up in a real external install. Asserts:
  - All 6 config modules import.
  - `sys.modules` contains zero of `torch / transformers / vllm / wandb / ring_flash_attn / verifiers / prime / datasets / liger_kernel / loguru`.
  - `find_package_resource('templates')` is None (heavy wheel absent).
  - `examples/Intellect-3.1/rl.toml` (which has `[slurm]`) parses; `template_path` stays None.

## Verification

- `uv sync --all-extras` resolves and builds both wheels.
- 333 unit tests pass.
- `uv run ruff check` and `uv run ruff format --check` are clean.
- All 6 entrypoints (`rl` / `sft` / `trainer` / `orchestrator` / `env-server` / `inference`) load and print `--help`.
- Fresh `pip install git+https://github.com/PrimeIntellect-ai/prime-rl.git@feat/separate-configs-package#subdirectory=packages/prime-rl-configs` (plain pip, no uv, no workspace context) installs ~11 packages, imports all configs, parses a real `[slurm]` config, and leaks none of the heavy deps into `sys.modules`.


JannikSt (approved):
lgtm - very excited for this feature. we should still build the wheel with the release IMO. 
also worth double checking the deprecation warning here?

.github/workflows/cpu_tests.yaml
```diff
@@ -38,3 +38,59 @@ jobs:
         run: PYTEST_OUTPUT_DIR=/tmp/outputs uv run pytest tests/unit -m "not gpu"
       - name: Cleanup output_dir
         run: rm -rf /tmp/outputs
+
+  slim-configs-install:
```
mikasenghaas, line 42: hmm why all this?

packages/prime-rl-configs/src/prime_rl/configs/__init__.py
mikasenghaas, line 1: we should prob not have `prime-rl` in the import path for the standalone package?

src/prime_rl/entrypoints/rl.py
```diff
@@ -12,6 +12,7 @@
 import pynvml
 import tomli_w
 
+import prime_rl._compat  # noqa: F401 — patch ring_flash_attn compat before transitive import
```
mikasenghaas, line 15: not sure why this change is needed here now but not before? seems risky to having to remember to put this import everywhere

packages/prime-rl-configs/pyproject.toml
```diff
@@ -0,0 +1,25 @@
+[project]
+name = "prime-rl-configs"
+version = "0.4.0"
+description = "Slim config schema for prime-rl (no GPU/ML deps)."
+readme = "README.md"
+requires-python = "~=3.12.0"
+dependencies = [
+    "pydantic>=1.10.13",
+    # The PyPI package named `pydantic-config` is a different project that
```
mikasenghaas, line 9: can remove comment here imo

.github/workflows/cpu_tests.yaml
mikasenghaas: seems like it shouldn't be in prime-rl's cpu tests

samsja: not its needed to make sure we can install and load the config independently

PR #2438, opened 2026-05-07
feat(inference): fp32 lm_head for stable logprobs under FP8/bf16

samsja:
## Summary
- New `inference.enable_fp32_lm_head` flag that runs the lm_head projection in fp32 instead of the model dtype (bf16). Both `hidden_states` and `lm_head.weight` are cast to float32 before the matmul.
- Wired through vLLM's `additional_config` dict (no env vars). The launcher's `to_vllm()` injects `{"fp32_lm_head": True}` when the flag is set.
- The flag is captured **once**, on `LogitsProcessor.__init__`, onto the instance. The per-forward check then just reads `self._fp32_lm_head_enabled`. This matters because `get_current_vllm_config()` only resolves during model init — calling it from `_get_logits` during serving raises `AssertionError` (we verified empirically that the per-call approach silently no-ops).
- The fp32 weight is cached on the layer, but **invalidated against `Tensor._version`** so RL weight broadcasts (NCCL / filesystem `param.copy_(...)`) don't leave a stale step-0 fp32 copy serving forever.
- Asserts on `tie_word_embeddings=True` since casting `lm_head.weight` in place would also cast `embed_tokens`.

## Why
Under FP8/bf16 inference the only step between hidden states and the sampler that's *not* fp32 is the lm_head matmul itself — vLLM's sampler already promotes to fp32 before softmax/sampling. Casting the operands to fp32 doesn't add precision to the bf16 weight values themselves (zero-padded mantissa), but it does keep the matmul **output** in fp32 instead of truncating to bf16 — which is what the sampler actually needs. SGLang implements the same flag the same way (`logits_processor.py:894-897` in their tree).

Closes the rollout-vs-trainer logprob gap that destabilizes RL importance sampling. Background: [MiniMax-M1 §7.6](https://arxiv.org/pdf/2506.13585), ScaleRL, Meta arxiv:2510.13786. Tracks [vllm-project/vllm#24567](https://github.com/vllm-project/vllm/pull/24567), restricted here to the untied-embeddings case.

## Validation
Two single-node SLURM jobs on Qwen3-8B-Base + hendrycks-math, identical apart from `enable_fp32_lm_head`. The fp32 run shows a measurably lower train/inference logprob mismatch on wandb. Confirmed the patched path actually executes (vs silently falling through) by temporarily injecting `1/0` inside the fp32 branch — the worker crashed with `ZeroDivisionError` on the first forward, proving the init-time capture is correct.

## Usage
```toml
[inference]
enable_fp32_lm_head = true
```

## Notes
- Cost: one extra fp32 copy of `lm_head.weight` (~600MB for Qwen3-30B vocab/hidden), rebuilt only when the underlying weight is mutated. The fp32 GEMM at the end is negligible vs the rest of the model.
- Limitation: no support for `tie_word_embeddings=True` — explicit `NotImplementedError`.
- Removable once `vllm-project/vllm#24567` (or a successor) lands upstream.

### DIFF
--- a/src/prime_rl/inference/patches.py
+++ b/src/prime_rl/inference/patches.py
@@ -1086,3 +1086,90 @@ def monkey_patch_no_moe_lora():
         self.is_lora_enabled = False
 
     FusedMoEConfig.__post_init__ = _patched__post_init__
+
+
+def monkey_patch_fp32_lm_head():
+    """Run the lm_head projection in fp32.
+
+    Casts both hidden_states and lm_head.weight to float32 before the matmul,
+    so logits are computed in fp32 instead of the model dtype (bf16). The
+    sampler already promotes to fp32 for softmax/sampling, so the head matmul
+    itself is the only step where precision is lost — promoting it closes the
+    gap that bites RL importance-sampling under FP8/bf16 inference (cf.
+    MiniMax-M1 §7.6, ScaleRL, Meta arxiv:2510.13786). SGLang exposes the same
+    knob as `--enable-fp32-lm-head`.
+
+    Activated by setting ``additional_config["fp32_lm_head"] = True`` on the
+    vLLM namespace; the launcher does this when ``inference.enable_fp32_lm_head``
+    is set. Reading from ``get_current_vllm_config().additional_config`` keeps
+    the wiring config-driven (no env vars) and survives multiprocess spawns
+    because vLLM serializes its config across worker processes.
+
+    The fp32 weight is materialized lazily on the first call and cached on the
+    layer (``_fp32_lm_head_weight``) so it isn't recomputed each forward.
+
+    Limitation: tied embeddings would require duplicating the embedding table
+    in fp32; we don't try to handle that here. Equivalent of
+    vllm-project/vllm#24567 restricted to the untied case; intended to be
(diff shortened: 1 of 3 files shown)


src/prime_rl/inference/patches.py
```diff
@@ -1086,3 +1086,90 @@ def _patched__post_init__(self: FusedMoEConfig):
...
+        cached = getattr(lm_head, "_fp32_lm_head_weight", None)
+        cached_version = getattr(lm_head, "_fp32_lm_head_weight_version", None)
+        if cached is None or cached_version != cur_version:
+            with torch.no_grad():
+                cached = weight.detach().to(torch.float32)
+            lm_head._fp32_lm_head_weight = cached
+            lm_head._fp32_lm_head_weight_version = cur_version
+
+        bias_fp32 = embedding_bias.to(torch.float32) if embedding_bias is not None else None
+        logits = F.linear(hidden_states.to(torch.float32), cached, bias_fp32)
```
Jackmin801, line 1166: I think the only thing we gain here is removing the epilogue bf16 downcast. we should be able to replace this with torch.mm(hidden_states, lm_head.weight.T, out_dtype=torch.float32) and do this more efficiently

PR #2420, opened 2026-05-05
[DYNAMO] pass renderer transport to verifiers

AmeenP:
## Summary

Passes `renderer_transport="dynamo"` to verifiers renderer clients when `client.backend = "dynamo"`, while preserving `vllm` for the default vLLM path.

Preserves `backend` when `ElasticInferencePool` rebuilds URL-scoped client configs, so elastic Dynamo renderer clients keep the same transport selection.

Updates the shared Dynamo renderer stack contract:

- pins `verifiers` to `9485194f` (PrimeIntellect-ai/verifiers#1287)
- pins `renderers` to `17005dd` (PrimeIntellect-ai/renderers#11)
- stops forcing `return_token_ids` in orchestrator env sampling defaults; token IDs are requested by the renderer/verifiers transport (`nvext.engine_data` for Dynamo)
- adds config coverage that `return_token_ids` is not injected globally

Stacked on #2398.

Companion PRs:
- verifiers: https://github.com/PrimeIntellect-ai/verifiers/pull/1287
- renderers: https://github.com/PrimeIntellect-ai/renderers/pull/11

## Validation

Local:
- `uv lock --check`
- `uvx ruff==0.13.0 check --config=pyproject.toml src/prime_rl/utils/client.py tests/unit/utils/test_client.py`
- `uvx ruff==0.13.0 format --check --config=pyproject.toml src/prime_rl/utils/client.py tests/unit/utils/test_client.py`

GitHub Actions:
- Ruff: pass
- Slim install (prime-rl-configs only): pass
- Unit tests: pass

Local test note:
- `uv run pytest tests/unit/utils/test_client.py` is blocked on macOS by the repo's Linux-only lockfile; `uv run --no-sync pytest tests/unit/utils/test_client.py` exited 139 locally.

PR #2446, opened 2026-05-08
[DYNAMO] wire TITO client to Dynamo renderer transport

AmeenP:
## Summary

Ports the missing prime-rl client-side TITO wiring from Biswa's tested branches on top of #2420 (`feat/dynamo-renderer-transport-config`). The shared renderer/verifiers pins and `return_token_ids` cleanup now live in #2420, so this PR only carries the TITO-specific follow-up.

- treats `openai_chat_completions_token` as token-aware for `client.backend = "dynamo"`
- with #2420, uses `renderer_transport="dynamo"` for both renderer-mode and TITO clients when using Dynamo
- adds unit coverage for Dynamo TITO transport selection

## Biswa Compatibility Notes

Included from Biswa's `rl-sdk-2` where it matches the current Dynamo contract:

- backend-based Dynamo transport selection in this PR
- dropping the global `return_token_ids` default in #2420

Not included yet:

- Biswa's `compute_teacher_logprobs` dispatch that asks Dynamo chat completions for `nvext.extra_fields=["prompt_logprobs"]`. Dynamo #9509 currently exposes `engine_data` / completion token data, not `nvext.prompt_logprobs`, so copying that prime-rl path would require a matching Dynamo change first.

## Validation

Local:
- `uv lock --check`
- `uvx ruff==0.13.0 check --config=pyproject.toml src/prime_rl/utils/client.py tests/unit/utils/test_client.py`
- `uvx ruff==0.13.0 format --check --config=pyproject.toml src/prime_rl/utils/client.py tests/unit/utils/test_client.py`

GitHub Actions:
- Ruff: pass
- Slim install (prime-rl-configs only): pass
- Unit tests: pass

Local test note:
- `uv run pytest tests/unit/utils/test_client.py` is blocked on macOS by the repo's Linux-only lockfile; `uv run --no-sync pytest tests/unit/utils/test_client.py` exited 139 locally.

### DIFF
--- a/src/prime_rl/utils/client.py
+++ b/src/prime_rl/utils/client.py
@@ -24,10 +24,15 @@ class AdminAPI(Protocol):
     """
 
     async def health(self, client: AsyncClient) -> None: ...
+
     async def list_models(self, client: AsyncClient) -> list[dict]: ...
+
     async def pause(self, client: AsyncClient) -> None: ...
+
     async def resume(self, client: AsyncClient) -> None: ...
+
     async def update_weights(self, client: AsyncClient, weight_dir: str | None) -> None: ...
+
     async def load_lora_adapter(
         self,
         client: AsyncClient,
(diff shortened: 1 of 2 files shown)


samsja:

PR #2471, opened 2026-05-10
feat: opt-in renderer-based tokenization for SFT

### LINKED ISSUE #2470: Extend renderers support to SFT training to solve incremental tokenization issues
## Summary

PR #2278 (merged Apr 14) added `renderers` as a tokenization engine for RL, but SFT was not updated and still requires use of `build_incremental_token_mask` + Jinja templates.

The current method leads to issues when working with chat templates that render differently based on conversation state (e.g. Qwen3.5, Qwen3), which `renderers` would fix. 

## Current State

`build_incremental_token_mask` will throw a soft `IncrementalTokenizationError` that causes `SFTDataset._process()` to skip the sample with a warning:

```python
# src/prime_rl/trainer/sft/data.py:210
try:
    input_ids, loss_mask = build_incremental_token_mask(...)
except IncrementalTokenizationError as e:
    self.logger.warning(f"Skipping example {example.get('__index', '')}: {e}")
    return None
```

This causes 2 bugs I've noticed:

1. **Multiturn sample drop:** Chat templates that renders tokens differently based on position in the message list will fail the prefix-consistency check inside `build_incremental_token_mask` on every turn past the first, meaning all multiturn samples are dropped. 

Reproduce with:
```bash
uv run sft @ examples/wordle/sft.toml --model.name Qwen/Qwen3-0.6B --tokenizer.name Qwen/Qwen3-0.6B --max_steps 1
```

2. **Qwen3.5 specific crash:** Qwen3.5's chat template [raises an error](https://huggingface.co/Qwen/Qwen3.5-397B-A17B/blob/main/chat_template.jinja#L43) when there are no non-system messages to be rendered. `build_incremental_token_mask` starts with `messages[:1]` (system only), which triggers the jinja2 `TemplateError` and crashes training immediately.

Reproduce with:
```bash
uv run sft @ examples/wordle/sft.toml --model.name Qwen/Qwen3.5-0.8B --tokenizer.name Qwen/Qwen3.5-0.8B --max_steps 1
```

## Root Cause

`build_incremental_token_mask` renders partial message slices:

```python
cur_ids = render_messages(tokenizer, messages[:idx + 1], ...)
```

Any chat template that computes position-dependent state from the full slice produces different token ids for the same message depending on what follows it. Incremental slicing is fundamentally incompatible with these templates. The SFT CI tests against 'PrimeIntellect/Qwen3-0.6B` which has a modified context independent chat template and is thus not affected. This is fine for some use cases but doesn't account for environments where you want to eschew tokens to reduce context usage. 

## Proposed Solution: Extend Renderers support to SFT Training

PR #2278 added renderer support to the RL rollout path as an opt-in flag (`use_renderer`), leaving the existing path as the default. The latest version of renderers library ships [`build_training_sample`](https://github.com/PrimeIntellect-ai/renderers/blob/b1ecc35b61924179670635159b57df64509bb2f9/renderers/base.py#L537) for exactly this use case. 

Something like:
```python
from renderers.base import build_training_sample

# in src/prime_rl/trainer/sft/data.py, when use_renderer is set
if renderer is not None:
        renderer,
        messages,
        role_to_mask=should_mask,
        tools=tools,
    )
else:
    input_ids, loss_mask = build_incremental_token_mask(...)
```

The renderer object would be constructed the same way as the RL path (`create_renderer(tokenizer, renderer=..., ...)`) and passed into `SFTDataset` alongside the existing tokenizer.

Willing to submit a PR for this if there is interest.

ercbot:
Closes #2470 

Adds `use_renderer` flag to SFTConfig, mirroring the RL path added in #2278. When enabled, SFTDataset tokenizes via `renderers.base.build_training_sample` (single render() + message_indices mask) instead of the incremental Jinja template path. 

Using renderers for tokenization fixes multiturn sample drops and the Qwen3.5 system-only TemplateError crash for chat templates that render position-dependently.

---

> [!NOTE]
> **Medium Risk**
> Changes the SFT data/tokenization path and adds new config validation around renderer usage; misconfiguration could cause training to fail early or change which tokens are masked for loss.
> 
> **Overview**
> Adds `use_renderer` + `renderer` configuration to SFT, including validation to forbid unsupported combos (e.g. VLMs) and to error when renderer-only knobs are set without enabling the feature.
> 
> When enabled, SFT now creates a `renderers`-based tokenizer (`create_renderer`) and routes dataset tokenization through `build_training_sample`, passing the renderer through `setup_dataset` for both train and validation; it also warns once when per-example `chat_template_kwargs` would be ignored and rejects falling back to `DefaultRenderer`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9bb1180bf11ccefee9d25f5eaed3dd69b6fdf672. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hallerite:
thank you @ercbot !

hallerite (approved):
Approved — renderer-based SFT tokenization fixes the multi-turn / Qwen3.5 TemplateError bug as advertised, validators correctly gate VLM and renderer-args misconfig, DefaultRenderer fallback is refused with a helpful error, and `renderer.pool_size` is rejected for SFT (synchronous tokenization, one renderer per DP rank — no in-process pool needed).

Verified end-to-end on Qwen/Qwen3-0.6B + PrimeIntellect/Reverse-Text-SFT for 50 steps: training loop completes cleanly, loss descends monotonically, grad norms track the baseline path.

One follow-up tracked separately: the renderer path currently trains on `<|im_start|>assistant\n` scaffolding because `Qwen3Renderer._render_assistant` attributes the role opener to the assistant message via `message_indices`. RL avoids this via its structural prompt/completion split; SFT (single-render + `build_training_sample`) doesn't. PrimeIntellect-ai/renderers#33 just merged adding `RenderedTokens.sampled_mask` to fix this at the right layer; tracking the prime-rl-side bump as a follow-up issue. Doesn't block this PR — it lands the bug-fix that's actually requested.

hallerite:
@ercbot quick heads-up: CI on #2471 fails because GitHub doesn't expose org secrets to PRs from forks, so wandb crashes the trainer process. I pushed the same head SHA to a different branch and opened #2493 as a mirror so CI can run properly.

I'll make sure to merge that one with `--author="Eric Botti <ericjbotti@gmail.com>"` so you stay the primary author on the resulting commit :)

PR #2476, opened 2026-05-12
feat: support student evals with SFT teacher rollouts

tim0120:
## Summary

When `teacher_rollout_model` is configured for SFT distillation, teacher routing is scoped to training rollout generation only. The normal `inference_pool` remains the student/model-DP pool for online evals, policy weight sync, health checks, metrics, and shutdown.

Teacher rollout inference is represented as raw request clients from `teacher_rollout_model.client`, not as a second managed pool. Teacher rollout uses the plain OpenAI chat-completions/MITO client in this PR to keep the config surface minimal.

## Changes

- Remove the `RLConfig` validator that forbade `[inference]` + `teacher_rollout_model`
- Keep `inference_pool` as the student pool for evals, health, metrics, shutdown, and policy weight updates
- Build teacher rollout request clients from `teacher_rollout_model.client` only when student inference weight updates are enabled
- Keep teacher rollout generation on MITO / `openai_chat_completions`; renderer-backed teacher rollout is deferred
- Apply the teacher client/model override only at `Scheduler.schedule_rollout()` request submission
- Map the selected student logical client onto the teacher client set when the teacher and student client counts differ
- Preserve teacher-only hard-distill configs by default; they keep generating from the teacher and skip student inference weight updates
- Add `orchestrator.update_student_inference_weights` for externally started student inference; RL configs with `[inference]` auto-enable it
- Keep resumed LoRA evals on the resumed adapter name while keeping teacher rollout requests on the teacher model name
- Update config skill docs for the hard-distill teacher-rollout workflow

## W&B / Eval Evidence

These runs exercise the intended SFT hard-distillation path: training rollouts are routed to `teacher_rollout_model`, while online evals and policy weight updates continue to use the student `inference_pool`.

| Run | Eval shown | Result |
| --- | --- | --- |
| [W&B run `9439ba8b`](https://wandb.ai/primeintellect/sft-distill-test/runs/9439ba8b2ecf4e75b228ff57c2459908?nw=nwuserprimetim) | `eval/wiki-search/avg@1` | Improves from about `0.40` at step 0 to about `0.58` peak / `0.57` final by step 100. |
| [W&B run `1913ec66`](https://wandb.ai/primeintellect/sft-distill-test/runs/1913ec66404c46638c1135decb0a9173?nw=nwuserprimetim) | `eval/wordle/avg@1` | Improves from about `0.03` at step 0 to about `0.58` peak / `0.49` final by step 100. |

<a href="https://wandb.ai/primeintellect/sft-distill-test/runs/9439ba8b2ecf4e75b228ff57c2459908?nw=nwuserprimetim">W&B eval/wiki-search avg@1 curve for SFT hard-distill run</a>
<a href="https://wandb.ai/primeintellect/sft-distill-test/runs/1913ec66404c46638c1135decb0a9173?nw=nwuserprimetim">W&B eval/wordle avg@1 curve for SFT hard-distill run</a>

## Testing

- `uv run pytest tests/unit/orchestrator/test_scheduler.py tests/unit/orchestrator/test_orchestrator_setup.py tests/unit/utils/test_client.py`
- `uv run pytest tests/unit/test_configs.py -k 'teacher_rollout_weight_updates_require_inference_config_or_explicit_opt_in or not test_load_configs'`
- `uv run pytest tests/unit/orchestrator/test_scheduler.py tests/unit/test_configs.py -k 'teacher_rollout or lora_policy_update or schedule_rollout_applies_teacher_override'`
- `uv run pytest tests/unit/orchestrator/test_scheduler.py tests/unit/orchestrator/test_orchestrator_setup.py tests/unit/utils/test_client.py tests/unit/test_configs.py -k 'teacher_rollout or lora_policy_update or schedule_rollout_applies_teacher_override or setup_clients'`
- GitHub PR checks are passing, including GPU integration tests for `base`, `rl_lora`, `alphabet_sort`, `rl_multi_run`, and `benchmark_regression`

---

> [!NOTE]
> **Medium Risk**
> Changes rollout routing and policy-weight update behavior when `teacher_rollout_model` is set, which can affect training/eval correctness and weight synchronization across inference servers.
> 
> **Overview**
> Adds an opt-in path for SFT hard-distillation where **training rollouts can be generated via `teacher_rollout_model` while keeping the managed `inference_pool` pointed at the student model** for health checks, online evals, metrics, shutdown, and weight updates.
> 
> Introduces `orchestrator.update_student_inference_weights` (auto-enabled when `[inference]` is configured) to control whether teacher-rollout runs still push student weights; defaults preserve the previous teacher-only behavior (no policy updates) unless explicitly enabled.
> 
> Implements request-time teacher override in `Scheduler.schedule_rollout()` (including mapping student-selected clients onto the teacher client set) and adjusts LoRA resume/model-name handling so evals use `inference_pool.model_name` while teacher rollouts continue using the teacher model name; updates validators/docs and adds targeted unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15fedb5ab24afb11fc878d20476e001482550310. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


willccbb:
wondering if it might make more sense to not have a distinct “student_eval_pool” notion and just treat it as the standard inference pool, but then have the teacher_rollout_model injection more tightly scoped to just the train rollouts? the rollout_teacher doesn’t need a notion of an eval pool in this case, just the ability to pass client info (which we can assume is the same across groups for the teacher_rollout_model pattern, generally will be an external API with its own internal load-balancing, not managed by prime-rl)

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -148,13 +149,30 @@ async def orchestrate(config: OrchestratorConfig):
...
+        teacher_rollout_clients = setup_clients(
+            rollout_client_config,
+            client_type="openai_chat_completions",
+        )
+        teacher_rollout_model_name = rollout_model_name
+        renderer = None
+        inference_pool = await setup_inference_pool(
+            config.client,
+            model_name=config.model.name,
+            eval_client_type="openai_chat_completions",
```
willccbb, line 168: can we have a client_type field here, matching what we expose for evals?

src/prime_rl/orchestrator/scheduler.py
```diff
@@ -166,6 +169,14 @@ async def _select_least_loaded_client(self) -> vf.ClientConfig:
         inflight = Counter(self._client_identity(info.client_config) for info in self.inflight_requests.values())
         return min(clients, key=lambda c: inflight[self._client_identity(c)])
 
+    def _resolve_rollout_request_target(self, client_config: vf.ClientConfig) -> tuple[vf.ClientConfig, str]:
+        if self.teacher_rollout_clients is None:
+            return client_config, self.model_name
+
+        teacher_client = self.teacher_rollout_clients[client_config.client_idx % len(self.teacher_rollout_clients)]
```
willccbb, line 195: can we add a short comment here explaining multiple teachers? def valid but prob not the most common use case

willccbb (changes requested):

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -897,6 +897,11 @@ class TeacherModelConfig(BaseConfig):
 class TeacherRolloutModelConfig(BaseConfig):
     """Configures an external teacher model used to generate rollout text."""
 
+    client_type: Annotated[
```
mikasenghaas: why would we make this configurable here? in my mind, we should do either:
- always mito
- do the same as we do for student (e.g. renderers or mito where renderers should give better prefix hits on inference but mito is "safer" while we are stilil battletesting renderers)

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -961,6 +966,17 @@ class OrchestratorConfig(BaseConfig):
         ),
     ] = False
 
+    enable_policy_updates: Annotated[
```
mikasenghaas: this should not be user configurable? i think this is a derived config arg. if so, can have `@property` on the config

packages/prime-rl-configs/src/prime_rl/configs/rl.py
```diff
@@ -930,6 +900,16 @@ def auto_setup_dp_rank_count(self):
             )
         return self
 
+    @model_validator(mode="after")
+    def auto_setup_teacher_rollout_policy_updates(self):
+        if (
+            self.orchestrator.teacher_rollout_model is not None
+            and self.inference is not None
```
mikasenghaas, line 922: not sure the inference condition is true here? we also want to enable policy updates if inference is started externally, right?

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -897,6 +897,11 @@ class TeacherModelConfig(BaseConfig):
 class TeacherRolloutModelConfig(BaseConfig):
```
mikasenghaas, line 897: i still really dislike this name. imo, this should be `TeacherModelConfig`. can we unify this with the OPD path?

mikasenghaas: (can do in later pr tho)

mikasenghaas: acc, now that i think of it. do we even need a diff config class for this? if they don't diverge, could just do
```py
model: ModelConfig() # alias so that student_model also work
teacher_model: ModelConfig()
```

ig we want to couple model info+clients. in this case, we could do
```py
class RolloutModelConfig(BaseConfig):
  model: ModelConfig()
  client: ClientConfig()

model: RollloutModelConfig() # alias so that student_model also work
teacher_model: RolloutModelConfig()
```
here, rollout model basically means model + client information.

another alternative would be to just fold the client configs into the model configs but this would be a bit more breaking ig

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -148,13 +149,36 @@ async def orchestrate(config: OrchestratorConfig):
...
+            else teacher_client_type
+        )
+        logger.info(f"Using teacher rollout override ({teacher_client_label}, model={rollout_model_name})")
+        teacher_rollout_clients = setup_clients(
+            rollout_client_config,
+            client_type=teacher_client_type,
+        )
+        teacher_rollout_model_name = rollout_model_name
+        renderer = None
+        inference_pool = await setup_inference_pool(
```
mikasenghaas, line 164: wait im a little confused, above we have a `teacher_inference_pool` but here we are overwriting the "student's" pool with teacher configs. yet, the scheduler doesn't just get the inference pool but also the teacher clients and name?

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
tim0120: good point, should be derived

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
tim0120: will keep mito for now

src/prime_rl/orchestrator/orchestrator.py
tim0120: yep, the current naming makes this confusing. `teacher_inference_pool` is for teacher logprobs and OPD, whereas `teacher_rollout_model` is only for hard-distill train rollout generation

tim0120: for hard distill, `inference_pool` stays semantically the student pool, and the teacher client/model is injected only at train rollout request submission, so teacher rollout does not become another managed inference pool, requested a la https://github.com/PrimeIntellect-ai/prime-rl/pull/2476#issuecomment-4427812479

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
tim0120: yes this would be sm cleaner, can look into making this PR

packages/prime-rl-configs/src/prime_rl/configs/rl.py
tim0120: ah true

src/prime_rl/orchestrator/orchestrator.py
tim0120: @willccbb curious why you wanted https://github.com/PrimeIntellect-ai/prime-rl/pull/2476#issuecomment-4427812479 though? i do find that this injection/overwrite a bit patchy but not sure if alternatives (eg separate inference pool) are just relatively less desirable?

PR #2479, opened 2026-05-12
[codex] add Poolside Laguna custom model

samsja:
## Summary
- add a PrimeRL custom model/config/conversion path for Poolside Laguna (`laguna`)
- register Laguna with `AutoConfig` and the custom causal LM mapping
- add a `scripts/mini_moe.py --arch laguna` preset and README support row

## Validation
- `uv run python scripts/mini_moe.py --arch laguna --output-dir /tmp/prime-rl-laguna-tests/mini-laguna` passed (`HF vs PrimeRL max logits diff: 0.000002`; HF -> PrimeRL -> HF weight roundtrip verified)
- `uv run python -m compileall -q src/prime_rl/trainer/models/laguna src/prime_rl/trainer/models/__init__.py scripts/mini_moe.py`
- `uv run ruff check src/prime_rl/trainer/models/laguna src/prime_rl/trainer/models/__init__.py scripts/mini_moe.py`
- `uv run ruff format --check src/prime_rl/trainer/models/laguna src/prime_rl/trainer/models/__init__.py scripts/mini_moe.py`
- reverse-text smoke completed 50 steps with batch size 64; mismatch KL: min `0.0011`, max `0.0314` at step 44, avg `0.0073`, final `0.0040`
- Hendrycks math smoke completed 100 steps with batch size 64, eval disabled, and zero-advantage filtering monitor-only; mismatch KL: min `0.0002`, max `0.0005` at step 40, avg `0.0003`, final `0.0003`

## Notes
- Attempted a Laguna mini reverse-text RL run through vLLM 0.20.2 remote-code/generic Transformers serving. The custom PrimeRL model and conversion passed parity checks, but vLLM loaded the checkpoint without becoming serving-ready, so the Laguna RL smoke did not complete.

# Training sanity check

## hendrycks math

image

image
image

## Reverse text

image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ With `[model] impl = "auto"` (the default), the trainer selects that custom stac
 | Qwen3 MoE (`qwen3_moe`) | `Qwen/Qwen3-30B-A3B`, … | yes | ✅ | ✅ |
 | Qwen3.5 MoE (`qwen3_5_moe`) | `Qwen/Qwen3.5-35B-A3B`, … | yes | ✅ | ✅ |
 | Qwen3 / Qwen3.5 VLMs | [multimodal.md](docs/multimodal.md) (`qwen3_vl`, `qwen3_5`, `qwen3_5_moe`) | MoE only on MoE VLMs | MoE only | ✅ |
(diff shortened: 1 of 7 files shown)


Jackmin801 (approved):
nice! yet more attn impls for the softplus gate but I guess we can always standardize later. lgtm!

Issue #2492, opened 2026-05-14
SFT renderer path trains on `<|im_start|>assistant\n` scaffolding (use renderers `sampled_mask`)


hallerite:
## Summary

The renderer-based SFT tokenization path landed in #2471 includes the assistant's `<|im_start|>assistant\n` role-tag opener in the trainable mask. The model is trained on tokens the chat template emits deterministically at inference and never samples — wasted gradient, and the loss number isn't directly comparable across mask conventions.

## Why it happens

`build_training_sample` builds the loss mask from `RenderedTokens.message_indices` alone:

```python
loss_mask[k] = (msg_idx >= 0) and role_to_mask(messages[msg_idx])
```

`Qwen3Renderer._render_assistant` (and equivalents in every hand-coded renderer) attributes the assistant's `<|im_start|>` + `role\n` opener to the assistant's message index — these get `role_to_mask(assistant) = True`.

RL doesn't have this problem: `trajectories.py` uses `render_ids(prompt, add_generation_prompt=True)` and `_common_prefix_len` to find the split, then `prompt_mask = [False]*N` blanket-masks the scaffolding. SFT has no structural prompt/completion split per turn, so it relied on attribution alone.

## Measured impact

On the CI SFT smoke (`PrimeIntellect/Reverse-Text-SFT`, 1000 single-turn examples), the renderer path's per-example trainable mask includes **7 extra near-zero-CE scaffolding tokens** vs. the baseline `build_incremental_token_mask` path (47 vs. 40 trainable). Across 50 steps this drops the mean loss by ~0.4 without any actual learning advantage — see the side-by-side comparison in #2471.

## Fix

PrimeIntellect-ai/renderers#33 (merged) added `RenderedTokens.sampled_mask: list[bool]` and updated `build_training_sample` to AND it with role attribution. All hand-coded renderers populate it; DefaultRenderer leaves it empty (opt-out).

Once `renderers` cuts a release with the change, bump the pin in `pyproject.toml`:

```toml
renderers==<new_version>
```

No code change needed in prime-rl — `build_training_sample` already does the AND in the upstream library. After the bump, the SFT renderer path will train on exactly the same tokens the RL path does, per-turn.

## Verification plan

Re-run the 50-step comparison from #2471. With the new `renderers` version, the renderer-path loss curve should match the baseline path within step-to-step noise instead of being ~0.4 lower. Both should still show monotonic descent and matching grad norms.

## Context

[...]

PR #2493, opened 2026-05-14
feat: opt-in renderer-based tokenization for SFT

### LINKED ISSUE #2470: Extend renderers support to SFT training to solve incremental tokenization issues

hallerite:
Mirror of #2471 from inside the org so CI has access to `WANDB_API_KEY` and friends (fork PRs can't see org secrets, which is why #2471's integration suite fails at SFT trainer init with `wandb.errors.errors.UsageError: No API key configured`).

Same head commit (`9bb1180bf`), same content, same review approval already granted on #2471.

Closes #2470. Supersedes #2471 for the purposes of CI / merge.

---

Adds `use_renderer` flag to SFTConfig, mirroring the RL path added in #2278. When enabled, SFTDataset tokenizes via `renderers.base.build_training_sample` (single render() + message_indices mask) instead of the incremental Jinja template path.

Using renderers for tokenization fixes multiturn sample drops and the Qwen3.5 system-only TemplateError crash for chat templates that render position-dependently.

Tightening on top of the original PR:
- Typed `renderer: Renderer | None` in `SFTDataset` / `setup_dataset`.
- Raise in `train()` if `create_renderer` returns `DefaultRenderer` — that fallback doesn't fix the position-dependent template bug `use_renderer` is meant to solve.
- One-shot warning when an example carries `chat_template_kwargs` while the renderer path is on (renderers don't forward template kwargs by design).
- Reject `renderer.pool_size` unconditionally in `SFTConfig`. SFT tokenizes synchronously (`num_workers=0`) and already gets one renderer per DP rank — an implicit per-rank pool. An in-process pool adds nothing.

## Known follow-up

The renderer path currently trains on `<|im_start|>assistant\n` scaffolding (about ~15% loss-denominator inflation, no learning impact — gradients on content tokens are identical to the baseline path). PrimeIntellect-ai/renderers#33 merged this morning adds `RenderedTokens.sampled_mask` to fix this at the right layer. Tracked as #2492; no code change needed in prime-rl beyond bumping the `renderers` pin when the next version cuts.

---

> [!NOTE]
> **Medium Risk**
> Changes the SFT data/tokenization path and introduces new config validation/renderer initialization logic, which can affect training behavior and dataset compatibility if misconfigured.
> 
> **Overview**
> Adds an opt-in `use_renderer` mode for SFT that tokenizes samples via the `renderers` library (`build_training_sample`) instead of incremental `apply_chat_template`, with a new `renderer` config block to select renderer/parsers.
> 
> Wires renderer creation into `sft/train.py` and passes it through `setup_dataset` into `SFTDataset`, including guardrails: disallow renderer use with VLMs, reject unused `renderer.pool_size`, error if `use_renderer` resolves to `DefaultRenderer`, and warn once when `chat_template_kwargs` are present but ignored under renderer tokenization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9bb1180bf11ccefee9d25f5eaed3dd69b6fdf672. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hallerite:
Note: SFT with renderers will hereafter have slightly different behavior than one might expect (as outlined in #2492). Will work on a fix once we have multimodal merged, as the machinery necessary to prevent this is only available in https://github.com/PrimeIntellect-ai/renderers/pull/33 of renderers, which has other breaking changes that are already in the process of being merged.

mikasenghaas (commented):
i would still prefer to not make this user-configurable but just default to renderers whenever possible (non-vlm). otherwise adoption and dogfooding will be much slower (see rl) 

otherwise, lgtm!

mikasenghaas (approved):

PR #2394, opened 2026-05-02
[DYNAMO] feat: example k8s manifests + local smoke tests

AmeenP:
## Summary

Adds a worked example for running prime-rl with NVIDIA Dynamo as the inference backend instead of prime-rl's bundled vLLM frontend. The PR is self-contained: it only adds files under `k8s/dynamo-deploy/` and `tools/dynamo/`, plus a formatting fix for the new local admin stub.

### `k8s/dynamo-deploy/` cluster deployment example

| File | Purpose |
|---|---|
| `dynamo-dgd.yaml` | Example `DynamoGraphDeployment` with frontend + vLLM worker. Requires `DYN_ENABLE_RL=true` so `/v1/rl/*` admin endpoints are served natively. |
| `prime-rl-values.yaml` | Helm values overlay that disables prime-rl's own inference component and points `base_url` / `admin_base_url` at Dynamo. |
| `prime-rl-configs.yaml` | ConfigMap mounted at `/configs` in orchestrator and trainer pods. |
| `admin-stub.yaml` | Optional admin-stub Deployment + Service for older Dynamo builds without native `/v1/rl/*`. |

### `tools/dynamo/` local smoke flow

| File | Purpose |
|---|---|
| `admin_stub.py` | Local-dev fallback admin stub using aiohttp. |
| `configs/smoke_rl.toml` / `smoke_rl_long.toml` | Orchestrator configs pointed at local Dynamo on `localhost:8000`. |
| `configs/smoke_trainer.toml` / `smoke_trainer_long.toml` | Matching trainer configs. |
| `run_dynamo.sh` | Launches Dynamo on GPU 0. |
| `run_smoke_test.sh` | Launches orchestrator + trainer on GPU 1, supports `--long`. |

## Notes

- Depends on the Helm chart additions in #2393: `<component>.configMap`, `imagePullSecrets`, and `tolerations`.
- The orchestrator/trainer source code remains backend-agnostic here; it talks to whichever endpoint `client.base_url` and `client.admin_base_url` point at.
- All manifests use placeholders such as `<your-namespace>` and `<your-registry>/...`; no secrets, paths, or registry coordinates are baked in.

## What's not in this PR

Extracted from the same upstream branch as #2391 and #2393. Skipped:

- `experimental.use_prefix_cache_salt` opt-out, because upstream went the other direction in #2314.
- `tools/rl_monitor.py` / `tools/rl_plot.py`, which can be a separate PR if reviewers want it.
[...]

PR #2504, opened 2026-05-15
[DYNAMO] add SLURM smoke config for Dynamo

AmeenP:
## Summary

Adds a single-node SLURM smoke example for running prime-rl against NVIDIA Dynamo as the inference backend.

This is stacked on #2394 so it lives next to the existing Dynamo K8s/local smoke examples.

Files added:

- `tools/dynamo/slurm_smoke/smoke_rl.toml`: RL config for a 2-GPU smoke run with no prime-rl-managed inference process.
- `tools/dynamo/slurm_smoke/dynamo_single_node_rl.sbatch.j2`: custom sbatch template that starts Dynamo on GPU 0 with `--discovery-backend file`, waits for `/health` and `/v1/models`, then runs prime-rl on GPU 1.
- `tools/dynamo/slurm_smoke/README.md`: dry-run, submit, environment override, and log instructions.

The Dynamo side uses file discovery and disables KV event publishing, so the sample does not require etcd or NATS for the single-node smoke path.

## Validation

Local:
- Parsed `tools/dynamo/slurm_smoke/smoke_rl.toml` through `RLConfig` using the config package and temporary validation dependencies.
- Rendered `dynamo_single_node_rl.sbatch.j2` with Jinja2 and checked the rendered script with `bash -n`.
- `git diff --cached --check`

GitHub Actions:
- Ruff: pass
- Slim install (prime-rl-configs only): pass
- Unit tests: pass

Note: full `uv run rl ... --dry-run` cannot run in this macOS workspace because prime-rl's lockfile supports Linux platforms only, and the local environment does not have the project console scripts synced.

### DIFF
--- a/tools/dynamo/slurm_smoke/README.md
+++ b/tools/dynamo/slurm_smoke/README.md
@@ -1,0 +1,87 @@
+# Dynamo SLURM Smoke Test
+
+This example runs a single-node smoke test with Dynamo serving inference and
+prime-rl training against that Dynamo endpoint.
+
+It is intended to mirror the local/Kubernetes Dynamo smoke flow in
+`tools/dynamo/` and `k8s/dynamo-deploy/`, but under one SLURM allocation:
+
+- GPU 0: Dynamo frontend + vLLM worker, using `--discovery-backend file`
+- GPU 1: prime-rl orchestrator + trainer
+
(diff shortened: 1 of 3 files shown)


samsja:

PR #2517, opened 2026-05-16
refactor: propagate shared config via before validator

### LINKED ISSUE #2430: Top-level shared CLI fields (seq_len, max_model_len, model.name) don't propagate when combined with TOML configs
## Summary

When `RLConfig` is built from `@ config.toml` and a top-level shared field is overridden on the CLI, the override updates `RLConfig` itself but fails to reach the nested sub-configs. Flags like `--seq-len`, `--max-model-len`, and `--model.name` are silently ineffective.

## Reproduction

`config.toml`:

```toml
seq_len = 16384
[model]
name = "Qwen/Qwen3-0.6B"
[trainer.model]
name = "Qwen/Qwen3-0.6B"
[inference.model]
max_model_len = 16384
```

Run:

```
rl @ config.toml --seq-len 1024 --max-model-len 1024 --model.name /local/path
```

**Expected:** the nested `trainer.model.seq_len`, `orchestrator.seq_len`, `inference.model.max_model_len`, `trainer.tokenizer.name`, `inference.model.name`, etc. all reflect the CLI values.

**Actual:** they keep the TOML defaults. Notably the trainer tries to load its tokenizer from the original HF id even though `model.name` was overridden, which fails on offline nodes (`HF_HUB_OFFLINE=1`).

Fields with unconditional propagation (`max_steps`, `max_async_level`) are not affected.

## Root cause

`pydantic_config.cli()` first constructs `RLConfig` from the TOML. The `auto_setup_*` validators in `prime_rl/configs/rl.py` propagate the shared values into nested configs and, in doing so, mark those nested fields as "set" in Pydantic's `model_fields_set`. Tyro then applies the CLI overrides on top, but when validators run again the `model_fields_set` guards (e.g. in `auto_setup_seq_len`, `auto_setup_model`) treat the TOML-derived nested values as user-specified and skip propagation.

Additional issues:

- `RLConfig.max_model_len` has no propagation validator at all.
- `TrainerConfig.auto_setup_tokenizer` copies `model.name` → `tokenizer.name` at construction time, so any later override of `trainer.model.name` doesn't reach the tokenizer.

mikasenghaas:
## Summary

Mirror of #2291, rebased onto main after the monorepo restructure (#2507). Fixes #2430.

Reorganizes all shared-config propagation on `RLConfig` into a single `mode="before"` validator (`auto_setup_shared_configs`), so sub-config validators see final values at construction time. This fixes the CLI override bug (#2430): shared fields like `--seq-len` and `--model.name` now correctly propagate when combined with a TOML base config.

**Root cause:** the old `mode="after"` validators used `model_fields_set` guards to detect "unset" fields, but after tyro builds a default from TOML and passes it back through Pydantic, those guards treated TOML-derived values as user-set and skipped propagation. The `mode="before"` approach operates on raw dicts before sub-configs are constructed, so `fill()` (write-if-absent) works correctly regardless of how tyro constructs the final instance.

- Collapses eight per-field after-validators into one `validate_shared_configs` after-validator
- Adds `validate_shared_seq_len` to `validation.py`
- Removes `RLConfig.max_model_len` (declared but never propagated anywhere)
- Moves `auto_setup_session_headers` onto `OrchestratorConfig` (it only touches orchestrator state)
- Adds focused unit tests covering propagation precedence and CLI merge behavior


mikasenghaas:
Superseded by #2565 (merged)

PR #2521, opened 2026-05-16
[codex] make renderers default for rollouts

samsja:
## Summary

- Make the orchestrator default to renderer-backed TITO for text-only RL by setting `use_renderer=true` and `use_token_client=false` by default.
- Keep VLM configs on MITO by setting both `use_renderer=false` and `use_token_client=false`, and keep external teacher-rollout configs explicit about non-renderer mode.
- Update config docs, the config skill, runtime comments, and VLM validation coverage for the new rollout-mode defaults.

## Validation

- `uv run pytest tests/unit/test_configs.py`
- `uv run pytest tests/unit/test_configs.py tests/unit/orchestrator/test_orchestrator_setup.py tests/unit/utils/test_client.py tests/unit/utils/test_elastic.py`

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
@@ -1121,23 +1121,23 @@ class OrchestratorConfig(BaseConfig):
     use_token_client: Annotated[
         bool,
         Field(
-            description="Whether to use the token-in-token-out (TITO) client for training across all environments. "
+            description="Whether to use the server-tokenized token-in-token-out (TITO) client for training across all environments. "
             "WARNING: Only use this if your environment has a linear history and the chat template has the extension "
             "property (i.e. no tokens are ever removed or inserted by the chat template). Mutually exclusive with "
             "``use_renderer``."
         ),
-    ] = True
+    ] = False
 
     use_renderer: Annotated[
         bool,
         Field(
-            description="Whether to use the renderer client (client-side tokenization via the ``renderers`` package, "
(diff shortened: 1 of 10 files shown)


packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -1223,8 +1223,8 @@ def validate_sft_distill_mode(self):
     def validate_client_mode(self):
         """The two client toggles select among three exclusive modes:
 
-        - ``use_token_client=True``  + ``use_renderer=False`` → TITO  (default)
-        - ``use_token_client=False`` + ``use_renderer=True``  → renderer
+        - ``use_token_client=False`` + ``use_renderer=True``  → renderer (default)
+        - ``use_token_client=True``  + ``use_renderer=False`` → TITO
```
hallerite: we should not call it TITO here or clarify that a renderer is also TITO

tests/unit/test_configs.py
```diff
@@ -157,6 +157,43 @@ def test_removed_fused_lm_head_chunk_size_field_is_rejected():
         TrainerModelConfig.model_validate({"fused_lm_head_chunk_size": "auto"})
 
 
+def test_orchestrator_uses_renderer_by_default():
```
hallerite: can we delete this

hallerite (dismissed):
need to let multimodal be MITO by default, otherwise lgtm

tests/unit/test_configs.py
samsja: ah bruh why codex not reading my skills

hallerite (approved):

PR #2523, opened 2026-05-17
[codex] Guard checkpoint disk metrics mkdir

samsja:
## Summary

This hardens checkpoint disk metrics collection so it cannot crash distributed training when the shared checkpoint directory is created concurrently on BeegFS.

## Root Cause

The trainer crash happened after step 0 in `get_ckpt_disk_metrics()`. Multiple ranks reached the disk metrics path and attempted to create/read the shared `<output_dir>/checkpoints` directory. On BeegFS, `Path.mkdir(..., exist_ok=True)` can still surface `FileExistsError` when another rank wins the mkdir and metadata is briefly stale.

## Changes

- Only the master rank attempts to create the checkpoint metrics directory.
- Treat `FileExistsError` from that mkdir as non-fatal.
- Fall back to `output_dir` for disk usage if checkpoint-dir visibility is stale.
- Add regression coverage for stale `FileExistsError` and non-master mkdir skipping.

## Validation

- `uv run pytest tests/unit/train/test_trainer_utils.py tests/unit/train/test_world.py tests/unit/utils/test_pathing.py`
- `uv run ruff check src/prime_rl/trainer/utils.py tests/unit/train/test_trainer_utils.py`
- `git diff --check origin/feat/r3-v3-routed-experts...HEAD`

---

> [!NOTE]
> **Low Risk**
> Low risk: small change limited to metrics collection, only alters error handling around `mkdir` to avoid crashing in distributed environments.
> 
> **Overview**
> Prevents distributed trainers from crashing during checkpoint disk metrics collection by **tolerating `FileExistsError`** from `Path.mkdir(..., exist_ok=True)` when multiple ranks race to create the shared `checkpoints` directory (notably on BeegFS).
> 
> `get_ckpt_disk_metrics()` now wraps the directory creation in a `try/except` and proceeds to `shutil.disk_usage()` even if another rank won the mkdir race.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e275f6868a6975aeb62737f7fb4dbfcf061281cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/unit/train/test_trainer_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    monkeypatch.setattr(trainer_utils, "get_world", lambda: SimpleNamespace(is_master=False))
+    monkeypatch.setattr(
+        trainer_utils,
+        "get_ckpt_dir",
+        lambda output_dir: NonMasterCheckpointPath(output_dir / "checkpoints"),
+    )
+
+    metrics = trainer_utils.get_ckpt_disk_metrics(tmp_path)
+
+    assert metrics["system/ckpt_disk_total_gib"] > 0
```
samsja: lets remove this please

src/prime_rl/trainer/utils.py
```diff
@@ -119,8 +119,19 @@ def get_ckpt_disk_metrics(output_dir: Path) -> dict[str, float]:
     monitor.log(...) call (once per step).
     """
     ckpt_dir = get_ckpt_dir(output_dir)
-    ckpt_dir.mkdir(parents=True, exist_ok=True)
-    usage = shutil.disk_usage(str(ckpt_dir))
+    if get_world().is_master:
+        try:
```
mikasenghaas: wait the bug is that it throws even if `exist_ok=True`? this is wild

mikasenghaas (commented):
this feels a bit overly defensive. feel like doing either of the two would be enough?
- Only the master rank attempts to create the checkpoint metrics directory.
- Treat FileExistsError from that mkdir as non-fatal.

tests/unit/train/test_trainer_utils.py
samsja: Done — removed in 5f89a63.

samsja:
Simplified in e275f68 — kept only the FileExistsError catch and dropped the rank gate / disk_usage fallback. The race on BeegFS surfaces FileExistsError even with exist_ok=True (wild, agreed), so catching it is the targeted fix; every rank then sees the dir and shutil.disk_usage works.

hallerite (approved):

mikasenghaas (approved):

PR #2540, opened 2026-05-18
feat(orchestrator): validate renderer auto-resolution at config time (#2537)

### LINKED ISSUE #2537: Surface unsupported renderer config errors earlier
When using a model that is not in the auto map, prime-rl currently only surfaces the error during rollouts. This should be caught at config validation time instead.

The observed rollout error says that `RendererPool` does not support tools and to choose a model-specific renderer instead of the default fallback.

hallerite:
## Summary

- Adds an `OrchestratorConfig` `@model_validator(mode="after")` that rejects `use_renderer=True` + `renderer.name='auto'` when `model.name` isn't in `MODEL_RENDERER_MAP`, so `--dry-run` catches it (closes #2537).
- Sweeps the public configs to opt into the right renderer explicitly.
- Swaps `PrimeIntellect/Qwen3-0.6B` → `Qwen/Qwen3-0.6B` in the two public **RL** configs that referenced it (gsm8k, alphabet_sort) and sets `renderer.preserve_all_thinking = true` to match PI's template behavior under the auto-resolved `qwen3` renderer.
- **Includes submodule bumps** to pick up the three companion PRs (now merged): renderers#48 (pre-flight overflow check with auto-discovery), renderers#50 (Qwen3-30B-A3B-{Instruct,Thinking}-2507 and GLM-5-FP8 added to `MODEL_RENDERER_MAP`), and verifiers#1408 (renderer-client `OverlongPromptError` translation).

## Why config-time and not orchestrator startup?

The SFT path already rejects this at startup (`src/prime_rl/trainer/sft/train.py:172-179`). Putting the same guard at orchestrator startup wouldn't fire on `--dry-run`. Moving to a config validator catches it in `cli(RLConfig)` — i.e. on dry-run and on every config load — without touching the runtime code path.

## What the error looks like

```
Value error, orchestrator.use_renderer=True with renderer.name='auto' but
'PrimeIntellect/Qwen3-0.6B-Reverse-Text-SFT' is not in
renderers.base.MODEL_RENDERER_MAP, so it would silently fall back to
DefaultRenderer. Pick one:
(a) [orchestrator.renderer] name='default' — for fine-tunes / vendored
    mirrors with custom chat templates (DefaultRenderer calls
    apply_chat_template); pair with tool_parser=<name> if the env uses tools.
(b) [orchestrator.renderer] name=<model-specific renderer> — if <model> is
    template-identical to a mapped family (and ideally also add it upstream
    to renderers.base.MODEL_RENDERER_MAP).
(c) orchestrator.use_renderer=false — opt out of the renderer client entirely.
```

Exits non-zero; `Dry run complete` never prints. The full list of mapped models is omitted from the error (40+ lines was too noisy); users can grep `renderers.base.MODEL_RENDERER_MAP` for it.

## How I classified each unmapped config

After md5-hashing `chat_template.jinja` for each unmapped model against its closest mapped sibling, plus diffing PI mirror templates against upstream Qwen / MiniMax to confirm same-weights-only:

| Action | Configs |
|---|---|
| Swap RL configs: `PrimeIntellect/Qwen3-0.6B` → `Qwen/Qwen3-0.6B`, add `renderer.preserve_all_thinking = true` | `gsm8k/rl`, `ci/integration/alphabet_sort` |
| Keep PI finetune id, set `renderer.name = "qwen3"` + `preserve_all_thinking = true` — chat template byte-identical (md5 `c574d57…`) to `PrimeIntellect/Qwen3-0.6B` base | `examples/wordle/rl.toml` (`PrimeIntellect/Qwen3-1.7B-Wordle-SFT`) |
| `renderer.name = "default"` — PI / Qwen finetune with custom chat template | 9 configs — `multi_reverse_text`, `ci/integration/reverse_text/{start,resume}`, `ci/integration/reverse_text_lora/{start,resume}`, `ci/integration/reverse_text_moe/start`, `ci/integration/reverse_text_multi_run`, `examples/{reverse_text, Intellect-3.1}` |
| `renderer.name = "default" + reasoning_parser = "think"` — DeepSeek-R1 distill | 8 configs — `acereason_math/{stage1,stage2}`, `ci/nightly/acereason_math`, `deepscaler/{stage1,stage2,stage3}`, `hendrycks_math/sanity`, `examples/hendrycks_sanity` |
| Model-specific `renderer.name` — template-identical (md5-confirmed) to a mapped sibling. After the submodule bump for renderers#50, four are auto-resolved and reverted; only the PI-mirror entry stays explicit. | 1 config — `examples/minimax_m2.5_swe` (→ `minimax-m2`, `PrimeIntellect/MiniMax-M2.5-bf16` deliberately not in upstream map) |

### Why we're not swapping all PI mirrors

`PrimeIntellect/*` re-uploads of upstream models mostly exist to carry chat-template patches that the hand-coded renderers ignore anyway. For the two **RL** configs that use the PI Qwen3 mirror, the swap to upstream + `preserve_all_thinking = true` reproduces PI's "always emit prior `<think>`" template patch under the qwen3 renderer.

**SFT configs that reference `PrimeIntellect/Qwen3-{0.6B,1.7B}` are intentionally left untouched** — SFT defaults `use_renderer = false`, so HF `apply_chat_template` runs directly against the model's template; swapping the model id would silently change SFT tokenization (PI emits empty `<think></think>` for every assistant message; upstream Qwen does not).

Genuine PI finetunes with custom chat templates (`Qwen3-0.6B-Reverse-Text-SFT`, `Reverse-Text-SFT`, `INTELLECT-3*`, `INTELLECT-3-Base`) keep the PI id and use `renderer.name = "default"` so `DefaultRenderer` runs the model's own (patched) `apply_chat_template`. **Exception: `Qwen3-1.7B-Wordle-SFT`** ships the standard PI template patch (byte-identical to `PrimeIntellect/Qwen3-0.6B` base), so it gets the same `qwen3` renderer + `preserve_all_thinking = true` treatment as the two RL configs that swapped to upstream Qwen3.

The one exception is `PrimeIntellect/MiniMax-M2.5-bf16`: identical weights to `MiniMaxAI/MiniMax-M2.5` modulo a bf16 dtype cast that prime-rl's trainer requires, and a byte-identical chat template. Keep the PI id (we need the bf16 weights), set `renderer.name = "minimax-m2"` explicitly. The corresponding PI-mirror entry was deliberately removed from PR #50 (no PI ids in the upstream renderer map).

## Slim configs package

`prime-rl-configs` now lists `renderers>=0.1.8.dev4` as a hard dependency. The validator's `from renderers.base import MODEL_RENDERER_MAP` lives inside the function body and only fires on `renderer.name='auto'` configs, so the slim install CI's "no heavy deps in `sys.modules` at import time" check still passes — `renderers` (and its transitive `transformers`) stay out of `sys.modules` until validation actually needs them. Verified locally against both slim CI steps and Bugbot's failure mode (default `OrchestratorConfig()` now resolves instead of raising `ImportError`).

## Submodule bumps

- `deps/renderers`: `17d0584f` → `8704f9d` (`renderers-v0.1.8.dev4`)
  - Includes renderers#48 (overflow check) and renderers#50 (map additions).
- `deps/verifiers`: `dd89b5e9` → `58b119fa`
  - Includes verifiers#1408 (`OverlongPromptError` translation).

## Tests

- 5 new unit tests in `tests/unit/test_configs.py` cover: reject unmapped, accept mapped, explicit `renderer.name` bypass, `use_renderer=false` bypass, explicit `renderer.name="default"` opt-in.
- All 78 parametrized `test_load_configs` cases pass after the sweep.
- All 13 `tests/unit/train/test_runs.py` + `tests/unit/train/rl/test_packer.py` cases pass after adding `use_renderer = False` to test fixtures that use placeholder `"test-model"` (now correctly rejected by the validator).
- Broad `tests/unit/` sweep (excluding pre-existing unrelated CUDA/Laguna failures): 240 passed.

---

> [!NOTE]
> **Medium Risk**
> Medium risk: introduces a new config-time validation that can hard-fail previously working orchestrator configs that relied on `renderer.name='auto'` silently falling back to `DefaultRenderer`, and adds a new `renderers` dependency to `prime-rl-configs`.
> 
> **Overview**
> Adds a new `OrchestratorConfig` validator that **rejects `use_renderer=true` + `renderer.name='auto'` when the configured model isn’t in `renderers.base.MODEL_RENDERER_MAP`**, surfacing misconfiguration during config load / `--dry-run` instead of silently falling back to `DefaultRenderer`.
> 
> Updates public `configs/` and `examples/` to explicitly opt into an appropriate renderer (`name="default"` or a model-specific renderer, plus `reasoning_parser="think"` / `preserve_all_thinking` where needed), and adjusts unit tests/fixtures to either use a mapped model or set `use_renderer=false`. Also adds `renderers>=0.1.8.dev4` to `prime-rl-configs` dependencies and updates the lockfile accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30653f453d3e19f7af5a8ba82606fa5ecd9bf9f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -1284,6 +1284,45 @@ def validate_renderer_args(self):
...
+        ``renderer.tool_parser`` is configured. Surface at config time so
+        ``--dry-run`` reports the error.
+        """
+        if not self.use_renderer or self.renderer.name != "auto":
+            return self
+        try:
+            from renderers.base import MODEL_RENDERER_MAP
+        except ImportError:
+            # Slim configs use without renderers installed — nothing to validate against.
+            return self
```
samsja: hello defense has catched a not real issue

examples/wordle/rl.toml
```diff
@@ -29,3 +29,7 @@ max_completion_tokens = 1024
 
 [inference.parallel]
 dp = 6
+
+# Model not in MODEL_RENDERER_MAP — opt into DefaultRenderer (apply_chat_template).
+[orchestrator.renderer]
+name = "default"
```
mikasenghaas: do we not have a non-default renderer for most of the models here?

samsja (approved):

PR #2071, opened 2026-03-22
add deepEP training

faresobeid:
> [!NOTE]
> **High Risk**
> High risk because it changes core distributed training behavior (MoE expert-parallel communication, checkpointing boundaries, and fused loss/LM head code paths) and adds new CUDA-kernel-backed dependencies that can fail at runtime on unsupported hardware or configs.
> 
> **Overview**
> Adds a new MoE expert-parallel communication backend, `model.ep_comm_backend='deepep'`, wiring DeepEP dispatch/combine into `MoE`/`LatentMoE` and model setup with runtime validation plus new tuning knobs (`model.deepep_num_sms`, `model.deepep_token_chunk_size`).
> 
> Introduces **selective activation checkpointing** via `model.ac.mode`/`model.ac.targets`, with model/runtime validation and layer-method checkpoint patching to checkpoint only chosen subcomponents (with fallback to full checkpointing where unsupported).
> 
> Extends SFT with `loss_impl='quack_fused'` (quack-kernels chunked linear+CE) and adds optional quack-accelerated `RMSNorm` on CUDA; also updates env server orchestration/logging (worker auto-scaling, log directories, verifiers API move), removes trainer `model.tp` usage, adjusts parallel-dims calculations accordingly, and hardens checkpoint loads (`torch.load(..., weights_only=False)`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 43224f80bf44f946f2d92a5950e0a81e43423554. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #2550, opened 2026-05-19
fix(trainer): suppress duplicated trainer logs on non-zero ranks

JannikSt:
## Why

On an 8-GPU trainer pod, the dashboard's Trainer log tab on `/dashboard/training/<id>` shows every log line 8 times - e.g. 8 identical "Starting training loop" lines and 8 near-identical "Step 0 | Time ..." lines (each with slightly different per-rank throughput/memory).

`torchrun --nproc-per-node=N` spawns N processes, each writes to stdout, k8s ships each stdout to Loki as separate streams, and the platform groups by `role=trainer`. The existing `--local-ranks-filter` plumbing in `entrypoints/rl.py` only applies to the launcher-managed single-node path; on k8s the trainer module is invoked directly via torchrun, bypassing that filter.

## What

- `setup_logger` gains a `rank_zero_only` flag. When `True` and `RANK > 0` (read from torchrun's env var, set before dist init), the loguru instance is built with no sinks so every `logger.info` / `.success` / `.debug` / `.warning` call becomes a no-op on non-zero ranks.
- `trainer/rl/train.py` and `trainer/sft/train.py` opt in.
- All other `setup_logger` callers (orchestrator, inference, RL launcher) are unchanged - they're single-process and never had the fan-out.

## Notes

- Uses global `RANK` (not `LOCAL_RANK`) so it works correctly on multi-node setups where only one trainer pod has global rank 0.
- Reads `os.environ["RANK"]` directly rather than going through `torch.distributed.get_rank()`, so it works for any log line emitted before `dist.init_process_group`.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only logging behavior by silencing non-zero distributed ranks, which could reduce per-rank visibility but does not affect training computation.
> 
> **Overview**
> **Reduces duplicated logs in distributed training runs.** `setup_logger` now supports `rank_zero_only`, which detects non-zero `RANK` (from `torchrun`) and creates a loguru logger with no sinks so log calls become no-ops on those ranks.
> 
> Both RL and SFT trainers opt into this (`rank_zero_only=True`) to prevent N-way repeated stdout lines when Kubernetes/Loki merges per-rank streams.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3fe296b87903f294ad3e9bf7bd342a62cd225bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/rl/train.py
+++ b/src/prime_rl/trainer/rl/train.py
@@ -73,9 +73,13 @@ from torchtitan.distributed.utils import clip_grad_norm_
 def train(config: TrainerConfig):
     # Setup world and logger
     world = get_world()
+    # rank_zero_only suppresses logs on non-zero ranks. Every rank's stdout
+    # is merged in k8s/Loki, so without this each line appears N times in the
+    # dashboard's trainer log tab (one per GPU).
     logger = setup_logger(
         config.log.level,
         json_logging=config.log.json_logging,
+        rank_zero_only=True,
     )
     logger.info(f"Starting RL trainer in {world} in {config.output_dir}")
 
--- a/src/prime_rl/trainer/sft/train.py
+++ b/src/prime_rl/trainer/sft/train.py
@@ -62,9 +62,13 @@ from torchtitan.distributed.utils import clip_grad_norm_
 def train(config: SFTConfig):
     # Setup world and logger
     world = get_world()
+    # rank_zero_only suppresses logs on non-zero ranks. Every rank's stdout
+    # is merged in k8s/Loki, so without this each line appears N times in the
+    # dashboard's trainer log tab (one per GPU).
     logger = setup_logger(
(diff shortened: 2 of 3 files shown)


JannikSt:
@codex review

mikasenghaas:
lol @samsja wants this

JannikSt:
Superseded by #2552 (Sami's launcher-flag approach) — better fix because it preserves per-rank logs on the PVC instead of silencing non-zero ranks entirely. Platform-side change to thread the same `--local-ranks-filter=0 --tee=3 --redirect=3 --log-dir=...` flags into the chart's torchrun command lives in PrimeIntellect-ai/platform PR #2071 (commit 36320d073).

PR #2560, opened 2026-05-19
fix(client): revert base_url to single string (drop multi-client)

### LINKED ISSUE #2557: Revert base_url to a single string (drop multi-client)
Roll back `base_url` to a plain `str` on the orchestrator client config. The multi-client/list-of-URLs path isn't needed and adds config + plumbing complexity for no current benefit.

## Scope

* Change `base_url` back to `str` (single endpoint).
* Remove multi-client fan-out / selection logic and any related config knobs.
* Update configs, tests, and skills/docs that reference the list form.

mikasenghaas:
## Summary

- Revert `ClientConfig.base_url` from `list[str]` to `str` and `admin_base_url` from `list[str] | None` to `str | None`. Closes #2557.
- Drop the URL loop in `setup_clients` / rename `setup_admin_clients` → `setup_admin_client` (now returns a single `AsyncClient`). `dp_rank_count` still expands the URL into per-rank logical clients.
- Refactor `ElasticInferencePool._rebuild_clients` to iterate over discovered pod URLs and call `setup_clients` per-URL (instead of round-tripping a list through `ClientConfig`).
- Update `orchestrator.py`, `entrypoints/rl.py`, and rl-config auto-setup to consume a single `base_url` string.
- Update all five `configs/debug/training_modes/*.toml` from `base_url = ["…"]` to `base_url = "…"`.
- Update affected tests (`tests/unit/utils/test_client.py`, `tests/unit/orchestrator/test_orchestrator_setup.py`).

## Multi-client deployment audit

Before reverting I checked what in-repo deployments depend on multiple base URLs.

**Static path:** all five training-mode TOMLs (`configs/debug/training_modes/{opd,opd_lora,sft,sft_lora,sft_external}.toml`) ship a single-element list. No tracked config uses multi-URL fan-out.

**Elastic path** (`configs/elastic/rl.toml`): does **not** set `router_url`, so the orchestrator currently builds one inference client per discovered pod via `_rebuild_clients`. That's multi-client at runtime, but it's internal to `ElasticInferencePool` - the user-facing `client.base_url` field is irrelevant in elastic mode (it's ignored when `elastic` is set). After this PR, elastic still produces multiple per-pod clients internally; the refactor just iterates and calls `setup_clients` per URL.

**`dp_rank_count`** still produces multiple clients per URL for DP-rank pinning - kept, since it's per-engine routing on a single server, not multi-server fan-out.

**`router_url`** is unchanged (it's already a single URL).

Net: no deployment loses functionality. The list form in `base_url` / `admin_base_url` was unused config plumbing.

## Verification

- `pytest tests/unit/utils/test_client.py tests/unit/utils/test_elastic.py tests/unit/orchestrator/test_orchestrator_setup.py tests/unit/orchestrator/test_scheduler.py tests/unit/orchestrator/test_teacher_logprobs.py` → 40 passed.
- Pydantic schema rejects list-form `base_url=[…]` at config load (verified via smoke test).

### DIFF
--- a/configs/debug/training_modes/opd.toml
+++ b/configs/debug/training_modes/opd.toml
@@ -43,7 +43,7 @@ id = "reverse-text"
 name = "PrimeIntellect/Qwen3-0.6B-Reverse-Text-RL"
 
 [orchestrator.teacher.client]
-base_url = ["http://localhost:8001/v1"]
+base_url = "http://localhost:8001/v1"
 
 [trainer.optim]
 lr = 3e-6
--- a/configs/debug/training_modes/opd_lora.toml
+++ b/configs/debug/training_modes/opd_lora.toml
@@ -43,7 +43,7 @@ id = "reverse-text"
 name = "PrimeIntellect/Qwen3-0.6B-Reverse-Text-RL"
 
 [orchestrator.teacher.client]
-base_url = ["http://localhost:8001/v1"]
+base_url = "http://localhost:8001/v1"
 
 [trainer.optim]
 lr = 1e-4
--- a/configs/debug/training_modes/sft.toml
+++ b/configs/debug/training_modes/sft.toml
@@ -40,7 +40,7 @@ id = "reverse-text"
 name = "PrimeIntellect/Qwen3-0.6B-Reverse-Text-RL"
 
 [orchestrator.teacher.client]
-base_url = ["http://localhost:8001/v1"]
+base_url = "http://localhost:8001/v1"
 
 [trainer.optim]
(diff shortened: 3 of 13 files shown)


mikasenghaas:
Closing — see audit comment on #2557. Reverting to single-string `base_url`/`admin_base_url` breaks `multi_node_rl.sbatch.j2`, which passes multi-URL CLI args for both fields in multi-replica / disaggregated-PD setups (added in #2030). Not feasible without first restructuring the slurm template.

PR #2608, opened 2026-05-23
perf(orchestrator): reduce event-loop lag

mikasenghaas:
## Summary

Reduce orchestrator step-boundary event-loop lag, motivated by router-replay runs (Qwen3-30B-A3B + GLM-5.1) where the per-step routed_experts payload (~25 GB across 256 rollouts) caused 5–40 s loop-lag spikes at step boundaries.

Changes are all in the orchestrator and its post-rollout pipeline. The trainer/inference paths are untouched.

> **Coupled with [PrimeIntellect-ai/verifiers#1453](https://github.com/PrimeIntellect-ai/verifiers/pull/1453)** (`perf/r3-serve`). The verifiers PR fixes the env-side counterpart of the same lag (env_worker / env_router), which this PR's verifiers submodule bump pulls in. Land or revert together — without the verifiers side the env worker loop saturates first under 256-rollout concurrency and the orch-side gains here don't surface.

### What's in here

- **Async sender + to_thread advantages/filters + uvloop** (`0ed04c879`)
- **Shared ProcessPoolExecutor** plumbing (`62bbd7b5a`) — kept opt-in via `use_process_pool`; tradeoffs documented in the config field.
- **Eager pool init + p99 lag metric** (`e8ec02c4d`)
- **Defer routed_experts concat in interleave_rollout** (`d0f7b1372`) — per-step `unpack → concat → repack` was O(N²) byte copies (~2.3 GB memcpy / rollout in the worst case). New path keeps a per-sample chunk list and concats once at finalize. O(N) byte work.
- **Per-phase timing + dump_raw_rollouts for offline replay** (`668c18401`)
- **orjson save_rollouts, dedup bool conversion, chunked gather** (`28c3e6d18`):
  - `json.dump` → `orjson.dumps`. The pure-Python serializer held the GIL for the entire 1.4 s of save in replay (1.36 s lag spike); prod hit a 15 s save_rollouts on one step.
  - Dropped redundant `bool()` conversion in `make_sample`/`extend_sample`; switched the remaining one in `prepare_step_tokens` to `list(map(bool, ...))` (~3× faster).
  - New `gather_chunk_size` config option chunks the interleave_rollout fanout with `await sleep(0)` between batches so the loop gets guaranteed slices. Offline bench: chunk=128 cut max loop lag 2.6× under simulated traffic for ~11 % wallclock cost.
- **Pretokenize call-site skip + 10 Hz lag monitor + offline harness** (`afa3a610c`):
  - Pretokenize was a no-op for router-replay (every step has tokens populated), but the 256-way `to_thread` fanout starved the loop for ~2 s per step. Skip the fanout entirely when nothing needs work. Offline attribution: pretokenize lag drops 2050 ms → 11 ms.
  - `EventLoopLagMonitor` sampling 1 Hz → 10 Hz (mirrors verifiers); 1 Hz was missing sub-second spikes.
  - `scripts/perf/` — offline harness used to find these:
    - `replay_orch_step.py` and `replay_orch_attribution.py` replay the step boundary against a `raw_rollouts/step_N.pkl`. The attribution script runs each phase **alone** with 1 ms-granularity lag tracking so blockers are attributed *by phase*, not just by wallclock.
    - `--synthetic-traffic N` spawns N background loop tasks to mimic prod's oversubscription regime (active rollouts, recv handlers, inference replies).
- **Verifiers submodule bump → `445f379`** — pulls in PR 1453 (env_worker / env_router loop fixes). Required for the orch-side gains above to be visible under realistic concurrency.

### How to validate / iterate

Enable raw-rollout dumping:

```toml
[orchestrator]
dump_raw_rollouts = true
gather_chunk_size = 128
```

Then run the attribution against a captured step:

```bash
uv run python scripts/perf/replay_orch_attribution.py \
  --dump /path/to/raw_rollouts/step_8.pkl \
  --gather-chunk-size 128 --synthetic-traffic 200 --quick
```

`README.md` under `scripts/perf/` has the full story.

### What's NOT in here (deferred / open questions)

- Killing base64 on the routed_experts wire. Profile showed `pybase64.b64decode_as_bytearray` is ~73 % of single-thread CPU in `interleave_rollout`. It releases the GIL so it parallelises and doesn't block the loop directly, but it's the dominant wallclock floor. Moving inference→orch transport to msgpack `bin` would eliminate the encode + decode. Cross-repo, deferred.
- Changing `TrainingSample.{prompt,completion}_mask` from `list[bool]` to `list[int]` (or bytes). Would kill the last Python loop in `prepare_step_tokens`. Interface change, deferred until empirical evidence it matters for loop lag.


mikasenghaas:
Superseded by #2609 — same 8 perf commits cherry-picked cleanly onto current main, plus the verifiers submodule bump to PR 1453's head. Closing this one to avoid confusion.

PR #2609, opened 2026-05-23
perf(orchestrator): reduce event-loop lag

mikasenghaas:
## Summary

Reduce orchestrator step-boundary event-loop lag, motivated by router-replay runs (Qwen3-30B-A3B + GLM-5.1) where the per-step routed_experts payload (~25 GB across 256 rollouts) caused multi-second loop-lag spikes at step boundaries.

Changes are all in the orchestrator and its post-rollout pipeline. The trainer/inference paths are untouched.

> **Coupled with [PrimeIntellect-ai/verifiers#1453](https://github.com/PrimeIntellect-ai/verifiers/pull/1453)** (`perf/r3-serve`). The verifiers PR fixes the env-side counterpart of the same lag (env_worker / env_router), which this PR's verifiers submodule bump pulls in. Land or revert together — without the verifiers side the env worker loop saturates first under 256-rollout concurrency and the orch-side gains here don't surface.

> **Follow-up #2612** (`perf/r3-sidecar`) layers a `routed_experts` sidecar-split sender on top of this PR. Use it if the to_thread-only send here isn't enough for the largest router-replay payloads.

### What's in here

7 commits, ~200 LOC. Each commit is a single intervention:

| # | Commit | What |
|---|---|---|
| 1 | `perf(transport): offload TrainingBatch send to a worker thread` | Make `TrainingBatchSender.send` async; encode + disk write run via `asyncio.to_thread` so the orch event loop stays responsive during step transitions. |
| 2 | `perf(orchestrator): await async sender; to_thread advantages+filters+uvloop` | Await the now-async sender. Wrap `compute_advantages` and `apply_filters` in `asyncio.to_thread` — pure-Python work that was running on the loop. Install uvloop in `main()`. |
| 3 | `perf(orchestrator): log p99 event-loop lag in metrics + busy-loop warning` | Add `event_loop_lag/p99` to the wandb dict and the busy-loop warning line. Raise warn thresholds to p90>1s / p99>5s / max>30s — p90 alone undersold the long tail we actually care about. |
| 4 | `perf(orchestrator): defer routed_experts concat in interleave_rollout` | Per-extension `unpack → numpy.concatenate → repack` was O(N²) byte copies (~2.3 GB memcpy per long rollout). New path keeps a per-sample chunk list and concatenates once at finalize. O(N) byte work, ~100× less per rollout. |
| 5 | `perf(orchestrator): orjson save_rollouts + dedup bool conversion` | `json.dump` → `orjson.dumps`. Pure-Python serializer held the GIL for the whole save (multi-second spikes on big steps); orjson serializes in C and releases the GIL. Also dropped the redundant `[bool(i) for i in mask]` re-conversion in `make_sample`/`extend_sample` (use `list(...)` instead) and switched `prepare_step_tokens` to `list(map(bool, ...))`. |
| 6 | `perf(orchestrator): skip pretokenize fanout when nothing needs work; 10Hz lag monitor` | Pretokenize is a no-op for router-replay (every step has `tokens` populated), but the 256-way `to_thread` fanout was firing only to no-op and the GIL stampede blocked the loop. Skip when no rollout needs work. Also `EventLoopLagMonitor` sample rate 1 Hz → 10 Hz to catch sub-second spikes. |
| 7 | `chore(deps): bump verifiers submodule to PR 1453 head` | Pulls in the env_worker / env_router lag fixes from verifiers#1453. |

### What's NOT in here (deferred / dropped)

- **Streaming + routed_experts sidecar in the sender** — moved to follow-up #2612. Real win if the simpler to_thread send here is insufficient under heavy router replay, but adds an on-disk format change and per-sample loop yields.
- **Shared `ProcessPoolExecutor`** — tried for router-replay's GIL residual, didn't help (per-call pickle/IPC cost dominated the small per-rollout payload). Dropped.
- **Per-phase timing instrumentation + `dump_raw_rollouts`** — useful while investigating but not worth keeping in prod.
- **Chunked gather (`gather_chunk_size`)** — sliced `interleave_rollout` waves with `await sleep(0)` between. Real wallclock cost for unclear value once the algorithmic fixes (#4) landed. Reverted.
- **Killing base64 on the routed_experts wire** — profile shows `pybase64.b64decode_as_bytearray` is ~73% of single-thread CPU in `interleave_rollout`. Releases the GIL so it parallelises and doesn't block the loop directly, but it's the dominant wallclock floor. Moving inference→orch transport to msgpack `bin` would eliminate the encode+decode. Cross-repo, deferred.
- **`TrainingSample.{prompt,completion}_mask`: `list[bool]` → `list[int]` or `bytes`** — would kill the last Python loop in `prepare_step_tokens`. Interface change, deferred until empirical evidence it matters.

---

> [!NOTE]
> **Medium Risk**
> Touches the orchestrator post-rollout pipeline and transport `TrainingBatchSender` API, so regressions could impact training throughput or batch delivery ordering, though changes are largely performance-oriented and isolated from model/inference logic.
> 
> **Overview**
> Reduces orchestrator event-loop stalls during step transitions by pushing CPU/GIL-heavy work off the loop and cutting per-rollout copying overhead.
> 
> `compute_advantages`, rollout filtering, and filesystem batch encoding/writes now run via `asyncio.to_thread`, and `TrainingBatchSender.send` becomes `async` (with the orchestrator awaiting it). Pretokenization fanout is skipped when all steps already have tokens, and `uvloop` is installed in `main()`.
> 
> `interleave_rollout` now defers `routed_experts` concatenation/packing until finalization (chunking per step and concatenating once), and rollout JSONL saving switches to `orjson`. Event-loop lag monitoring is sampled at 10Hz and logs/exports `p99` lag with updated warning thresholds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b92f7d3de3374ed0903c570d76fa1df63d6f3189. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -482,22 +482,23 @@ async def orchestrate(config: OrchestratorConfig):
         # Convert rollouts to training samples
         parallel_preprocess_start = time.perf_counter()
 
-        # Pretokenize is a no-op when the renderer client already populated
-        # ``tokens`` on each trajectory step (renderer path); the fallback
-        # tokenizer-only branch handles text-only rollouts whose tokens
-        # were not pre-rendered. Run on threads so CPU work overlaps with
-        # inference for the next batch (via max_async_level >= 2).
-        await asyncio.gather(
-            *(
-                asyncio.to_thread(
-                    pretokenize_rollout_trajectory,
-                    rollout,
-                    tokenizer,
-                    renderer=renderer,
+        # Pretokenize on threads so CPU work overlaps with inference for the
+        # next batch. Skip the fanout entirely when every step already has
+        # ``tokens`` (renderer path) — otherwise the 256-way to_thread fires
+        # only to no-op and the GIL stampede blocks the loop.
+        needs_pretokenize = any(step["tokens"] is None for rollout in train_rollouts for step in rollout["trajectory"])
+        if needs_pretokenize:
+            await asyncio.gather(
+                *(
+                    asyncio.to_thread(
+                        pretokenize_rollout_trajectory,
+                        rollout,
+                        tokenizer,
+                        renderer=renderer,
+                    )
+                    for rollout in train_rollouts
                 )
-                for rollout in train_rollouts
             )
-        )
 
         # Process rollouts in parallel
         results = await asyncio.gather(
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -428,10 +428,10 @@ async def orchestrate(config: OrchestratorConfig):
             # Compute advantages (in-place)
             num_rollouts = len(train_rollouts)
             num_unique_examples = len({(r["env_name"], r["example_id"]) for r in train_rollouts})
-            compute_advantages(train_rollouts, config.advantage)
+            await asyncio.to_thread(compute_advantages, train_rollouts, config.advantage)
 
             # Apply rollout filters — sets rollout["filters"] and rollout["is_filtered"]
-            apply_filters(rollout_filters, train_rollouts)
+            await asyncio.to_thread(apply_filters, rollout_filters, train_rollouts)
 
             n_trainable = sum(1 for r in train_rollouts if not r["is_filtered"])
             if n_trainable > 0:
@@ -564,7 +564,7 @@ async def orchestrate(config: OrchestratorConfig):
             step=progress.step,
         )
 
-        training_batch_sender.send(training_batch)
+        await training_batch_sender.send(training_batch)
 
         step_time = time.perf_counter() - step_start_time
 
@@ -863,6 +863,9 @@ async def orchestrate(config: OrchestratorConfig):
 def main():
     """Main entry-point for orchestrator. Run using `uv run orchestrator`"""
     set_proc_title("Orchestrator")
+    import uvloop
+
+    uvloop.install()
     asyncio.run(orchestrate(cli(OrchestratorConfig)))
 
 
--- a/src/prime_rl/orchestrator/vf_utils.py
+++ b/src/prime_rl/orchestrator/vf_utils.py
@@ -1,7 +1,7 @@
-import json
 import logging
 from pathlib import Path
 
+import orjson
 import verifiers as vf
 from verifiers.utils.save_utils import make_serializable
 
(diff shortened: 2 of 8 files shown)


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -863,6 +864,12 @@ async def _graceful_shutdown() -> None:
 def main():
     """Main entry-point for orchestrator. Run using `uv run orchestrator`"""
     set_proc_title("Orchestrator")
+    try:
+        import uvloop  # type: ignore
+
+        uvloop.install()
+    except ImportError:
+        pass
```
samsja: not need to try expect here imo

src/prime_rl/transport/filesystem.py
```diff
@@ -123,7 +128,6 @@ def __init__(self, output_dir: Path, data_world_size: int, current_step: int = 0
 
     def send(self, micro_batch_grid: list[list[MicroBatch]]) -> None:
         """Send grid of micro batches to the trainers."""
-        # Validation
```
S1ro1, line 126: nit on some of these comments, not sure if intended to remove

src/prime_rl/orchestrator/vf_utils.py
```diff
@@ -82,11 +82,11 @@ def get_tool_response_len(output: vf.RolloutOutput) -> int:
 def save_rollouts(rollouts: list[vf.RolloutOutput], path: Path, exclude_keys: set[str] | None = None) -> None:
     """Save rollouts to a JSONL file using verifiers serialization."""
     path.parent.mkdir(parents=True, exist_ok=True)
-    with open(path, "w") as f:
+    opts = orjson.OPT_APPEND_NEWLINE | orjson.OPT_SERIALIZE_NUMPY
+    with open(path, "wb") as f:
```
S1ro1, line 86: is this supposed to be "wb"?

S1ro1 (dismissed):
love

src/prime_rl/orchestrator/orchestrator.py
mikasenghaas: Agreed — already dropped the try/except in a later commit;  now just does `import uvloop; uvloop.install()`. uvloop is a direct prime-rl dep and the workspace is restricted to linux, so it's always present.

src/prime_rl/transport/filesystem.py
mikasenghaas: Restored both: `# Validation` in `FileSystemMicroBatchSender.send` and the docstrings on `FileSystemMicroBatchReceiver.wait`/`can_receive`/`receive`. They were unrelated drive-by removals from when I was rewriting the sender; not intended.

src/prime_rl/orchestrator/vf_utils.py
mikasenghaas: Yes — `orjson.dumps(...)` returns `bytes` (vs `json.dump(..., f)` which writes `str` directly), so the file needs to be opened in binary mode. That's the swap from `"w"` on main to `"wb"` here.

S1ro1 (approved):

samsja (approved):

PR #2612, opened 2026-05-24
perf(transport): stream-sidecar TrainingBatch sender (routed_experts split)

mikasenghaas:
## Summary

**Draft follow-up to #2609.** Layered on top of `perf/r3-rewrite` — only the transport-side change is new. Merge or rebase this after #2609 lands; or just pick it up directly if the to_thread-only send in #2609 isn't sufficient for our largest router-replay payloads.

### What's in here (one commit)

`perf(transport): stream-sidecar TrainingBatch sender (routed_experts split)`

- Peels `routed_experts.data` out of the msgpack encode and writes it as a single raw-bytes sidecar file (`train_rollouts.routed_experts.bin`). The hot bytes (~85% of payload) skip the base64-style msgpack `bin` framing on encode and the corresponding allocation on decode.
- Streams `TrainingSample` frames one at a time with `await asyncio.sleep(0)` between, so even the encode phase yields to the event loop instead of holding it through a single big `encoder.encode(batch)` call.
- Sidecar must be visible before the main file — receiver watches the main file as the "batch is ready" signal.
- v2 file format: `4-byte manifest_len + manifest + (4-byte frame_len + sample-frame) * n`. Receiver reconstructs samples from the manifest's `offsets`/`shapes`/`dtypes` arrays.

### When to land this

Use as a fallback if #2609 alone doesn't bring step-boundary lag low enough on router-replay runs (Qwen3-30B-A3B + GLM-5.1 territory, ~25 GB routed_experts payload). For non-router-replay workloads the simpler to_thread-only send in #2609 should be sufficient.

### Risk

- New on-disk format (v2). Backward-compat would require keeping both readers around; this PR deletes the v1 reader. Old run artifacts won't be readable by this branch. Acceptable for an experimental knob; not for general release until v1 is migrated away from.
- Per-sample `await sleep(0)` is harmless but adds N loop ticks per batch (small CPU overhead, large lag reduction).

### DIFF
--- a/src/prime_rl/orchestrator/event_loop_lag.py
+++ b/src/prime_rl/orchestrator/event_loop_lag.py
@@ -13,22 +13,22 @@ class EventLoopLagMonitor:
         self,
         interval: float = 1.0,
         max_window_size: int = 10000,
-        warn_med_lag_threshold: float = 0.5,
         warn_p90_lag_threshold: float = 1.0,
-        warn_max_lag_threshold: float = 5.0,
+        warn_p99_lag_threshold: float = 5.0,
+        warn_max_lag_threshold: float = 30.0,
     ):
         assert (
             interval > 0
             and max_window_size > 0
-            and warn_max_lag_threshold > 0
-            and warn_med_lag_threshold > 0
             and warn_p90_lag_threshold > 0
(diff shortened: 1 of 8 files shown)


mikasenghaas:

Issue #2311, opened 2026-04-17
Refactor scheduler to support overlapping train + eval rollouts


samsja:
## Problem

Evals currently run **blocking** — train rollouts are paused while eval rollouts execute, reducing overall throughput.

## Goal

Allow overlapping train and eval rollouts by sharing the same pipeline mechanism (shared dispatcher, concurrency limiter, etc.).

## Proposed Config

Introduce config knobs to control overlap behavior:

* `orchestrator.schedule_train_rollouts_on_finish` — whether to schedule new train rollouts as eval rollouts finish
* `orchestrator.eval.cancel_train_rollouts_on_start` — whether to cancel in-flight train rollouts once we hit the eval step

Setting these to `true` / `false` respectively gives maximum throughput.

**Alternative (cleaner):** a single `orchestrator.eval.overlap_train_rollouts` flag that encapsulates both behaviors (don't cancel on eval start + schedule train rollouts on eval finish). Names TBD.

PR #2639, opened 2026-05-26
feat(orchestrator): refactor to overlap train/eval 

mikasenghaas:
## Why

Three concrete wins over the previous `prime_rl.orchestrator`:

1. **Train + eval rollouts overlap.** No more long tails on online evals. When the orchestrator ships a train batch that lands on an eval interval, it tells the dispatcher to flip to `PREFER_EVAL` — *the dispatcher stops scheduling new train rollouts* but lets the in-flight train tail drain naturally while eval queues up. Conversely, when the eval queue empties the dispatcher starts scheduling train again while the eval tail finishes. Both transitions overlap on the same shared inflight counter — capacity stays saturated.

2. **Shared concurrency limiter across train + eval.** A single `inflight_permits` counter (capacity = `max_inflight_rollouts`) and a single `AsyncLimiter(tasks_per_minute)` govern *both* kinds of rollouts. The pipeline always runs at the rate the inference infra can sustain — no separate budgets to tune and no idle inference capacity while one kind drains.

3. **Natural rollout / group / batch hierarchy amortizes the heavy per-batch post-processing.** The old design did tokenization for an entire batch's worth of rollouts at once (`asyncio.gather` over `interleave_rollout` for all 128 rollouts) at ship time — a synchronization point that stalled the pipeline. The new design splits work by its natural scope:
   - **rollout** level — `interleave_rollout` / `backfill_rollout_tokens` happen *as each rollout completes*, overlapped with the dispatcher producing more rollouts.
   - **group** level — `compute_advantages` runs over one GRPO group at a time.
   - **batch** level — `post_batch_filters` + metrics build runs once per ship, on already-tokenized samples.

## Summary

- Replace the step-driven outer loop in `prime_rl.orchestrator` with an async-pipelined design driven by signal-based sinks. Clean deprecation — no flag, no parallel module, no compatibility shim.
- The atomic unit is a single `Rollout`. Sinks own grouping; **three uniform processing levels** (`process_rollout` / `process_group` / `process_batch`) drive both train and eval flows.
- The **dispatcher schedules both train and eval rollouts** through a single shared `inflight_permits` counter + shared `AsyncLimiter(tasks_per_minute)`. **Drain-switch overlap** on each eval boundary recovers the cost of the previous "blocking eval" without cancelling in-flight work.
- **Every dispatched rollout reaches the sink exactly once.** Success, env-error, empty-trajectory, task-exception, off-policy-cancel — all flow through `out_q` as `Rollout`s; the dispatcher synthesizes minimal `vf.RolloutOutput` markers for the failure paths. The sink owns drop / partial-train policy (filters errored rollouts; drops the whole group when the env `requires_group_scoring` and any failed).
- **Group / batch finalization is sink-derived by counting.** The "last rollout in a group" is whichever straggler finishes last — not knowable at dispatch time — so no flags travel on the `Rollout`; each rollout carries a `group_id` (UUID, generated at dispatch) and the sink finalizes that group on the `group_size`-th arrival.
- **Example sourcing is its own abstraction.** `TrainSource` (infinite weighted-round-robin pull across train envs) lives in `train_source.py`; `EvalSource` (trigger-driven finite-per-epoch pull with round-robin interleaving across fired envs) lives in `eval_source.py`. Both expose a symmetric `next_example(available_permits) -> dict | None`. The dispatcher just asks "what's the next example?" — it doesn't own dataset cursors, shuffling, or eval triggering.
- **Eval triggering happens in the orchestrator, per ship step.** After each train batch ships, `Orchestrator.finalize_train_batch` calls `eval_source.trigger(step)` (which collapses startup, resume, and interval logic into one `first_trigger` flag) and flips the dispatcher to `PREFER_EVAL` if anything fires. The watcher's `on_new_version` is single-purpose: bump off-policy counters and cancel stale rollouts.
- **Pipeline drains at `max_steps` — no out-of-band final eval pass.** When the final train batch ships, the orchestrator cancels all in-flight train rollouts, flips to drain mode, the dispatcher stops scheduling new train, and any interval-aligned eval at the final step completes through the normal pipeline before the main loop exits.
- **One unified `PeriodicLogger` on the orchestrator** drives a single pipeline line: train batch fill (finalized-group survivors, `0→batch_size`) plus a separate `(+N buffered)` count for rollouts sitting in not-yet-complete groups, per-active-eval-epoch progress, and total inflight split as `(train=, eval= | per-env)`. Dispatcher / watcher gauges and event-loop lag go to wandb on the `_timestamp` axis (wall-clock, not step). No central log-loop task that pulls from everyone; no per-component logger to wire up.
- Adds `pre_batch_filters` / `post_batch_filters` config fields, replacing the old `filters` (renamed, no alias — see Breaking). **Filters are train-only** — the eval sink has no filter code by design.

Supersedes the prior brainstorming attempts in #2254 and #2336. Refs #2311.

## Architecture

```
                                ┌──────────────────────────────────┐
                                │     WeightWatcher (task)         │
                                │ polls broadcasts/, on_new_version│
                                └────────────┬─────────────────────┘
                                             │ Policy.version, on_new_version (off-policy cancel)
                                             ▼
   ┌─────────────────┐         ┌──────────────────────────────────┐
   │  TrainSource    │ ──────► │       RolloutDispatcher          │  shared:
   │  (infinite RR)  │         │           (task)                 │   • inflight_permits counter
   └─────────────────┘         │  PREFER_TRAIN ⇄ PREFER_EVAL      │   • AsyncLimiter(tasks_per_minute)
                               │  off-policy cancel               │   • Policy (read at dispatch)
   ┌─────────────────┐         │  drain mode: stop scheduling     │
   │  EvalSource     │ ──────► │     new train (post-max_steps)   │
   │  (trigger-driven│         │  emits every dispatched rollout  │
   │   epoch queue)  │         │  (success or error marker)       │
   └────────▲────────┘         └────────────┬─────────────────────┘
            │ trigger(step)                 │ Rollout (raw.error optional)
            │                               ▼
            │                          asyncio.Queue (out_q)
            │                               │
            │           ┌───────────────────┴────────────────────┐
            │           ▼                                        ▼
            │ ┌────────────────────────┐         ┌───────────────────────────┐
            │ │      TrainSink         │         │       EvalSink            │
            │ │ process_rollout: tok-  │         │ process_rollout: no-op    │
            │ │   enize (interleave)   │         │                           │
            │ │ process_group: filter  │         │ process_group: bucket     │
            │ │   errors, advantages,  │         │   per-example into env    │
            │ │   pre_filters          │         │ process_batch: build per- │
            │ │ process_batch: post_   │         │   env reward/pass@k/      │
            │ │   filters + samples    │         │   turns/truncation        │
            │ │ finalizes by sink-     │         │   (EvalBatchMetrics)      │
            │ │   derived counts       │         │ finalizes by sink-derived │
            │ │                        │         │   counts                  │
            │ └───────────┬────────────┘         └───────────┬───────────────┘
            │             │ TrainBatch                       │ EvalBatch
            │             ▼                                  ▼
            │ ┌────────────────────────────────────────────────────────────────┐
            └─┤ Orchestrator main_loop / finalize_train_batch / finalize_eval_batch │
              │ ckpt save, ship to trainer, monitor.log, heartbeat, step++,        │  + PeriodicLogger
              │ fires eval triggers, drives drain on max_steps                     │    (unified "Pipeline" view:
              └────────────────────────────────────────────────────────────────┘     │     dispatcher + watcher +
                                                                                      │     sinks + event-loop lag)
```

Three background tasks (`dispatcher.start()`, `watcher.start()`, the orchestrator's `PeriodicLogger`) + the in-task `main_loop` under `Orchestrator.start()`. Event-loop lag comes from `verifiers.utils.async_utils.EventLoopLagMonitor` and feeds the pipeline view.

## Verification runs

Run these one-by-one to exercise the new orchestrator end-to-end.

**1. Basic training modes (rl / sft / opd)** — student inference auto-launches on GPU 0; `opd`/`sft` need a teacher on GPU 1 first (`rl` needs no teacher):

```bash
# Teacher — only for opd + sft (skip for rl)
  --model.name PrimeIntellect/Qwen3-0.6B-Reverse-Text-RL \
  --server.port 8001 --gpu-memory-utilization 0.5 --model.enforce-eager

uv run rl @ configs/debug/training_modes/rl.toml
uv run rl @ configs/debug/training_modes/opd.toml
uv run rl @ configs/debug/training_modes/sft.toml
```

**2. Single-turn with resume** (reverse-text; use the same `--output-dir` so `resume_step=15` picks up the checkpoint):

```bash
uv run rl @ configs/ci/integration/reverse_text/start.toml  --output-dir outputs/reverse-text-resume
uv run rl @ configs/ci/integration/reverse_text/resume.toml --output-dir outputs/reverse-text-resume
```

**3. Multi-env + eval** (2 train envs, 2 eval envs with staggered eval intervals):

```bash
uv run rl @ configs/debug/multi_env/reverse_text.toml
```

**4. Long-running basic run** (Hendrycks sanity, 8 GPUs: 4 train / 4 infer):

```bash
uv run rl @ examples/hendrycks_sanity/rl.toml --wandb.project <project> --wandb.name <run>
```

**5. Long-running evals** (RLM-SWE Qwen3.5-4B-Thinking, multi-node SLURM: 1 train × 2 infer node):

```bash
uv run rl @ configs/rlm_swe/qwen35_4b.toml
```

## Curves

`examples/alphabet_sort/rl.toml` ([W&B](https://wandb.ai/primeintellect/alphabet-sort/workspace?nw=nwusermikasenghaas_))

Screenshot 2026-05-30 at 8 14 31 PM

`examples/hendrycks_sanity/rl.toml` ([W&B](https://wandb.ai/primeintellect/hendrycks-sanity?nw=d7deeqfi33v))

Screenshot 2026-05-30 at 8 13 22 PM

`configs/rlm_swe/qwen35_4b.toml` ([W&B](https://wandb.ai/primeintellect/rlm-swe-debug/workspace?nw=nwusermikasenghaas_))

Screenshot 2026-05-30 at 8 16 16 PM

## Config surface

Existing TOMLs in [configs/](configs/) and [examples/](examples/) keep parsing without changes or warnings.

**Additive in [packages/prime-rl-configs/.../orchestrator.py](packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py):**

- `pre_batch_filters: list[FilterConfig]` (default: gibberish, repetition, zero-advantage all in monitor mode).
- `post_batch_filters: list[FilterConfig]` (default unchanged from the prior `filters`: gibberish + repetition monitor, zero-advantage enforce). Replaces the old `filters` key (renamed, no alias — see Breaking).
- `eval.skip_first_step: bool = False` — when `False` (default) the orchestrator runs an eval epoch at step 0 before any train rollouts dispatch. Replaces the old `eval_base_model` (semantics inverted: `False` now means *run* startup eval).
- `LogConfig.interval: float = 10.0` (under `[log]`) — the period for the unified pipeline `PeriodicLogger`.
- `TrainEnvConfig.group_size: int` — per-env override; a validator propagates the top-level `group_size` into each train env that doesn't override it.

Removed / renamed fields are enumerated under **Breaking** below.

## Breaking

The Python module surface inside `prime_rl.orchestrator` changes (the orchestrator class, `RolloutDispatcher`, sinks, sources, etc. move to new files); external users importing those internals must update imports. The `rl` / `orchestrator` / `sft` CLIs are unchanged.

**Removed / renamed config fields — a config that sets these now fails to parse (`extra_forbidden`):**

- `orchestrator.seed` — removed (was only consumed by the deleted buffer; no replacement).
- `orchestrator.filters` → `orchestrator.post_batch_filters` — renamed, **no alias**. Rename the key (and optionally add `pre_batch_filters`).
- `[orchestrator.buffer]` and every key inside it (`seed`, `easy_threshold`, `hard_threshold`, `easy_fraction`, `hard_fraction`, `online_difficulty_filtering`, `hash_keys`) — difficulty pools removed. Delete the block; to keep `online_difficulty_filtering = true` behavior, enforce the zero-advantage pre-batch filter:

  ```toml
  [[orchestrator.pre_batch_filters]]
  type = "zero_advantage"
  enforce = true
  ```
- `orchestrator.eval.eval_base_model` — replaced by `orchestrator.eval.skip_first_step`, with **inverted** semantics: `eval_base_model = true` ⇒ `skip_first_step = false` (the default, i.e. run the step-0 eval).
- `orchestrator.eval.skip_eval_on_resume` (and its alias `skip_eval_on_restart`) — folded into `skip_first_step`; resume doesn't re-run the startup eval (the resumed `EvalSource` starts with `first_trigger = False`).
- `orchestrator.eval.cancel_inflight_rollouts_on_eval` — removed; the drain-switch overlap is the only eval transition mode.
- `ckpt.skip_buffer` — removed (there is no buffer to skip).

**Behavior changes (field unchanged, semantics changed):**

- `orchestrator.max_off_policy_steps` now applies to **eval** rollouts as well as train (eval rollouts past the cap are cancelled).
- Filters apply to **train rollouts only** — eval rollouts are no longer filtered.

---

> [!NOTE]
> **High Risk**
> Large orchestrator rewrite with breaking config (`buffer`, filters rename, eval flags) and changed train/eval scheduling, off-policy cancellation, and checkpoint semantics that can break existing runs and resumes.
> 
> **Overview**
> This PR **replaces the step-driven orchestrator with an async pipeline** that overlaps train and eval on one shared `max_inflight_rollouts` budget. A new **`RolloutDispatcher`** schedules work from **`TrainSource`** / **`EvalSource`**, emits every rollout (including synthetic error/cancel markers) to sinks, and uses **drain-switch** eval transitions (`PREFER_EVAL` ⇄ `PREFER_TRAIN`) instead of blocking or cancelling in-flight train.
> 
> **`TrainSink`** / **`EvalSink`** finalize groups and batches by arrival counts, spreading tokenization, **`assign_advantages`**, and filters across rollout → group → batch instead of batching all tokenization at ship time. The difficulty **`Buffer`** and buffer checkpoints are removed; **`CheckpointManager`** only persists **`Progress`**.
> 
> **Config breaking changes** (documented in `CHANGELOG.md`): drop **`orchestrator.buffer`**, **`orchestrator.seed`**, **`ckpt.skip_buffer`**; rename **`orchestrator.filters`** → **`post_batch_filters`** (add optional **`pre_batch_filters`**); replace eval **`eval_base_model`** / **`skip_eval_on_resume`** with **`eval.skip_first_step`**; remove **`cancel_inflight_rollouts_on_eval`**. **`max_off_policy_steps`** now cancels stale **eval** rollouts too; filters are **train-only**. Repo TOMLs drop buffer blocks and migrate wiki-search / Intellect-style **`online_difficulty_filtering`** to enforced pre-batch zero-advantage filters.
> 
> Also adds **`rlm-swe`** env wiring, new debug configs (multi-env, multimodal), removes/consolidates older multimodal / multi-reverse-text example configs, and renames the entrypoint to **`run_orchestrator`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4842ee08738f296c37ccecd81e018e1d7acbdef6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -1,901 +1,903 @@
...
-        parallel_preprocess_time = time.perf_counter() - parallel_preprocess_start
-        logger.debug(
-            f"Converted {len(train_rollouts)} rollouts ({num_unique_examples} unique examples) "
-            f"to {len(train_examples)} training examples"
+        # Materialize at the I/O boundary so prime-rl metadata travels with
+        # the raw vf payload on disk + in wandb sample tables
+        rollout_dicts = [r.to_dict() for r in batch.rollouts]
+        step_path = get_step_path(get_rollout_dir(config.output_dir), step)
+        await asyncio.to_thread(
+            save_rollouts, rollout_dicts, step_path / "train_rollouts.jsonl", exclude_keys={"trajectory"}
```
samsja, line 563: wondering if we should do this in s subprocess instead ?

src/prime_rl/orchestrator/eval_sink.py
```diff
@@ -0,0 +1,160 @@
...
+    def process_rollout(self, rollout: EvalRollout) -> None:
+        """No-op. Eval rollouts don't need trainer-bound tokenization; the
+        method exists to keep the three-level structure uniform with
+        ``TrainSink``.
+        """
+        return None
+
+    # ── level 2: per-group (move into batch bucket) ───────────────────────
+
+    def process_group(self, group_id: uuid.UUID) -> None:
```
samsja, line 94: sounds like process group in pytorch haha, but no need to change i guess

samsja (dismissed):
pre-approving

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -569,12 +544,27 @@ def _preserve_mito_renderer(self, handler: SerializerFunctionWrapHandler) -> dic
     eval: EvalConfig | None = None
     """Evaluation configuration."""
 
-    buffer: BufferConfig = BufferConfig()
-
     advantage: AdvantageConfig | None = DefaultAdvantageConfig()
 
-    filters: list[FilterConfig] = [GibberishFilterConfig(), RepetitionFilterConfig(), ZeroAdvantageFilterConfig()]
-    """Rollout filters. Each filter can ``monitor`` (default) or ``enforce`` (skip rollouts)."""
+    pre_batch_filters: list[FilterConfig] = [
```
hallerite, line 549: hmm, I don't really understand the distinction of pre and post batch filters here

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -645,6 +632,48 @@ def _preserve_mito_renderer(self, handler: SerializerFunctionWrapHandler) -> dic
 
     experimental: OrchestratorExperimentalConfig = OrchestratorExperimentalConfig()
 
+    @model_validator(mode="before")
+    @classmethod
+    def _alias_filters_to_post_batch_filters(cls, data: Any) -> Any:
```
hallerite: I feel like we should just be breaking here.

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -645,6 +632,48 @@ def _preserve_mito_renderer(self, handler: SerializerFunctionWrapHandler) -> dic
...
+        """``filters`` is accepted as an alias for ``post_batch_filters``."""
+        if isinstance(data, dict) and "filters" in data and "post_batch_filters" not in data:
+            data = dict(data)
+            data["post_batch_filters"] = data.pop("filters")
+        return data
+
+    @model_validator(mode="before")
+    @classmethod
+    def _warn_removed_buffer(cls, data: Any) -> Any:
```
hallerite: same here.

src/prime_rl/orchestrator/advantage.py
```diff
@@ -124,28 +128,22 @@ def advantage_fn(inputs: AdvantageInputs) -> AdvantageOutputs:
...
-    stores them in-place on the rollouts.
+    """Compute and assign advantages for one finished group of rollouts.
 
-    `advantage_fn` is called once per group, so groups may have varying sizes
-    (partial-group training drops failed rollouts rather than rescheduling them).
+    Caller (``TrainSink.process_group``) hands in a single group's
+    post-error-filter survivors, so no grouping logic is needed here.
+    ``advantage_fn=None`` is the trivial case (advantage = reward).
+    Custom advantage functions still receive the raw ``vf.RolloutOutput``\\ s
+    via ``AdvantageInputs.rollouts`` — public API unchanged.
```
hallerite: let's remove this `public API unchanged.`

src/prime_rl/orchestrator/types.py
```diff
@@ -0,0 +1,209 @@
...
+class Progress:
+    """Persistent counters; ``step`` is the trainer-aligned step."""
+
+    step: int = 0
+    total_tokens: int = 0
+    total_samples: int = 0
+    total_problems: int = 0
+
+
+Kind = Literal["train", "eval"]
```
hallerite: very non-descriptive name. how about `RolloutKind`?

src/prime_rl/orchestrator/types.py
```diff
@@ -0,0 +1,209 @@
...
+    group_id: uuid.UUID
+    policy_version: int
+    rollout_count: int
+    client_config: vf.ClientConfig | None = None
+    off_policy_steps: int = 0
+    eval_step: int | None = None
+
+
+@dataclass
+class GroupState:
```
hallerite, line 52: not sure I understand this. why do we pin a client here? do we not do groups via sampling args `n`? if not, why not? the reason to do this is that this way vllm only does 1 prefill

src/prime_rl/orchestrator/types.py
```diff
@@ -0,0 +1,209 @@
...
+                continue
+            out[f.name] = str(val) if isinstance(val, uuid.UUID) else val
+        return out
+
+
+@dataclass
+class TrainRollout(FinishedRollout):
+    samples: list[TrainingSample] = field(default_factory=list)
+    advantage: float | None = None
+    is_filtered: bool = False
```
hallerite, line 114: I feel like we should also have filters active during eval.

src/prime_rl/orchestrator/types.py
```diff
@@ -0,0 +1,209 @@
...
+@dataclass
+class EvalRollout(FinishedRollout):
+    eval_step: int = 0
+
+
+@dataclass
+class TrainBatchMetrics:
+    """Per-batch aggregates from ``TrainSink.process_batch``; consumed by
+    ``MetricsBuilder.build``. ``arrivals_by_env`` / ``errors_by_env`` count
+    rollouts at the sink (errored ones get dropped at the group level before
```
hallerite: I really don't like that we are mentioning `groups` here. We need to think more generally imo and not let prevailing methods of RL seep into our documentation.

src/prime_rl/orchestrator/eval_sink.py
```diff
@@ -0,0 +1,160 @@
...
+        self.process_rollout(rollout)
+        bkey = (env_name, rollout.eval_step)
+        self.pending_groups[rollout.group_id].append(rollout)
+        if len(self.pending_groups[rollout.group_id]) >= self.group_size_for(env_name):
+            self.process_group(rollout.group_id)
+        if len(self.pending_batches[bkey]) >= self.batch_size_for(env_name):
+            return self.process_batch(bkey)
+        return None
+
+    def group_size_for(self, env_name: str) -> int:
```
hallerite, line 48: really don't like `group_size` for eval

hallerite (commented):
really nice work. left some comments that I think need addressing, but nothing big.

src/prime_rl/orchestrator/eval_sink.py
mikasenghaas: i think you're the only person thinking this haha

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: so eg with zero adv filtering enforced:
- pre batch: the sample doesn't count towards the `batch_size` (odf-like, longer inference, more stable training batch)
- post batch: the samples does count towards the `batch_size` (predictable inference, less stable training batch)

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: im acc happy to

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: i think this might nice ux, otherwise they will get a cryptic validation error and it's not clear that this deprecation was intentional

src/prime_rl/orchestrator/types.py
mikasenghaas: OH, no we have never done groups via `n`. i didn't know they could share a prefill that's kinda neat. altho i dont think it matters too much bc sys prompt/user prompt is typically small. 

i think the main reason is that it becomes weird in multi-turn. like for turn 2 you hae to do `n=1` bc the prompts have diverged

src/prime_rl/orchestrator/types.py
mikasenghaas: wdym by this? for monitoring you mean?

src/prime_rl/orchestrator/eval_sink.py
mikasenghaas: what do you like? i kinda think its kind of a fitting name

src/prime_rl/orchestrator/orchestrator.py
mikasenghaas: i think its ok lightweight as-is and not sure if going across process boundary is quicker here. esp if we move it into the pipeline and "stream" completed rollouts to disk (we alr do this for vf-eval btw), this should not cause us any problems

src/prime_rl/orchestrator/types.py
hallerite: It can make quite the difference, especially for single-turn, but yeah you would need to mutate the sampling args properly for multi-turn, so perhaps let's wait on that

src/prime_rl/orchestrator/types.py
hallerite: yes

src/prime_rl/orchestrator/eval_sink.py
hallerite: maybe it's just my taste, but for eval I would actually prefer something like `rollouts_per_example`, as I associate groups with GRPO-style training

hallerite (approved):

samsja (approved):

PR #2644, opened 2026-05-26
feat(sft): default loss_mask to renderer's sampled_mask

hallerite:
## Motivation

Some chat templates put the per-turn stop signal in the *next* message's span rather than inside the assistant turn itself. GLM-4.5 and GLM-5 are the immediate cases: after a tool-calling assistant message, the byte that closes the assistant's turn is the `<|observation|>` (or `<|user|>`) at the **start of the next message** — there is no `<|im_end|>`-style terminator inside the assistant span. vLLM uses that token as a stop id at inference (`get_stop_token_ids`), but the renderer attributes it to the next (tool / user) message structurally.

The previous SFT loss-mask pipeline AND'd the renderer's `sampled_mask` with a per-role filter (`assistant=True, tool=False, user=False, system=False` by default). For the GLM family this meant: the model would-sample `<|observation|>` at inference (`sampled_mask=True`), but it's attributed to the tool message (`role_to_mask(tool)=False`), so it was masked out of the loss and the model never learned to emit it. At inference the natural continuation of `</tool_call>` is another `<tool_call>`, so the model dumps parallel tool_calls instead of closing its turn. In a recent GLM-4.5-Air-Base SFT run, 26% of assistant turns had >1 tool call (max observed: **283 parallel calls in a single turn**); on SWE-bench Verified the model's `pass@1` measurably suffered — rollouts with extreme multi-tool dumps solved at roughly half the rate of the bounded ones.

The companion renderer fix ([PrimeIntellect-ai/renderers#66](https://github.com/PrimeIntellect-ai/renderers/pull/66)) updates `glm45.py` / `glm5.py` to mark that stop opener as `is_sampled=True`. This PR teaches `build_training_sample`'s caller (`SFTDataset`) to actually use that signal — by making the `sampled_mask`-only path the default rather than always AND-ing with a role filter.

## Changes

Refactor `LossMaskConfig` from "a class of per-role booleans" to a discriminated union, mirroring prime-rl's existing `Literal | dict-class` pattern (cf. `fused_lm_head_token_chunk_size: int | Literal[\"auto\", \"disabled\"]` in `configs/trainer.py`):

```python
LossMaskConfig: TypeAlias = Literal["sampled", "all"] | LossMaskRolesConfig
```

- `\"sampled\"` (new default, recommended with renderers): every token the renderer marks `is_sampled=True` is trainable, regardless of role. Stop tokens that the chat template attributes to the next message's span are correctly trained.
- `\"all\"`: every renderer token contributes to the loss. Mostly useful for debugging.
- `LossMaskRolesConfig` (renamed from the old per-role `LossMaskConfig` body): AND the renderer's `is_sampled` with per-role booleans. Use this when you specifically want to restrict supervision to a subset of roles even when other roles' tokens are model-sampled.

`SFTDataset` dispatches:
- `\"sampled\"` → `role_to_mask=None` (relies on `build_training_sample`'s new `role_to_mask` default added in renderers#66, which falls back to `sampled_mask`).
- `\"all\"` → `role_to_mask=lambda m: True`.
- `LossMaskRolesConfig` → callable that looks up the per-role bool, same as before.

The chat-template fallback path (no renderer registered) has no `sampled_mask` signal so it falls back to `LossMaskRolesConfig()` defaults regardless of the configured mode.

## Backward compatibility

Existing TOML configs of the form:

```toml
[data.loss_mask]
assistant = true
tool = false
```

still parse: pydantic's union dispatch tries `Literal[\"sampled\", \"all\"]` first (fails for a dict), then `LossMaskRolesConfig` (succeeds). Per-role behavior is preserved verbatim.

Configs that don't set `loss_mask` at all pick up the new `\"sampled\"` default. This is a deliberate behavior change for SFT runs that relied on the implicit assistant-only filter — but the new default is what fixes the bug for renderers whose stop signal lives outside the assistant span. For configs that want to keep the old behavior, set `[data.loss_mask] assistant = true` explicitly.

## Test plan

- [x] Existing SFT unit tests (`tests/unit/train/sft`) and orchestrator SFT-trajectory tests pass under both old (per-role) and new (`\"sampled\"`) configs.
- [x] Type-adapter round-trip: `\"sampled\"`, `\"all\"`, and `{\"assistant\": true, \"tool\": true}` all parse to the expected variant.
- [x] Default `SFTDataConfig().loss_mask == \"sampled\"`.

## Renderer dependency

- [PrimeIntellect-ai/renderers#66](https://github.com/PrimeIntellect-ai/renderers/pull/66) — **merged**. Marks GLM `<|user|>` / `<|observation|>` openers as `is_sampled=True` when they close an assistant turn, and makes `role_to_mask` optional in `build_training_sample` (falls back to `sampled_mask`).
- This dependency is **required**, not cosmetic: the `"sampled"` path passes `role_to_mask=None`, which only resolves to the `sampled_mask` fallback as of renderers#66. The branch now pins `deps/renderers` at `35c2407` (via the `main` merge), a descendant of #66 (`1d69232`), so the feature works against its pinned renderer.

---

> [!NOTE]
> **Medium Risk**
> Changes default SFT supervision (implicit configs now train renderer-sampled tokens across roles) and token/EOS handling on the renderer path, which can shift training behavior for GLM-style templates and existing runs without explicit `[data.loss_mask]`.
> 
> **Overview**
> **SFT loss masking** now defaults to trusting the renderer’s `is_sampled` signal (`"sampled"`) instead of always AND-ing with assistant-only per-role flags. `LossMaskConfig` is a union of `"sampled"`, `"all"`, or `LossMaskRolesConfig` (the former per-role struct); explicit `[data.loss_mask]` tables still parse as `LossMaskRolesConfig`.
> 
> `SFTDataset` maps those modes to `role_to_mask` for `build_training_sample` / the chat-template fallback: `"sampled"` passes `None` on the renderer path; `"all"` trains every token; role config keeps the old boolean lookup. The legacy tokenizer path still uses assistant-only role defaults when `"sampled"`/`"all"` are set, since it has no `sampled_mask`.
> 
> **EOS handling** is renderer-aware: EOS is only auto-appended and required in `target_ids` on the chat-template fallback, not when a renderer owns the token stream (stop tokens may be turn markers like GLM’s `<|observation|>`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba47c3c1851c30035a0373e8a29702088f8624a8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/trainer/sft/data.py
```diff
@@ -27,6 +32,41 @@
 STACKING_DATASET_BUCKET_TIMEOUT = 10
 
 
+def _always_true_role_filter(message: dict) -> bool:
```
snimu, line 35: why private

src/prime_rl/trainer/sft/data.py
```diff
@@ -27,6 +32,41 @@
...
+    """Role filter that trains on every message regardless of role.
+
+    Used when :data:`LossMaskConfig` is ``"all"`` to bypass the
+    role-based ``role_to_mask`` gate in
+    :func:`renderers.base.build_training_sample`.
+    """
+    return True
+
+
+def _role_filter_for(cfg: LossMaskRolesConfig):
```
snimu, line 45: why private

src/prime_rl/trainer/sft/data.py
```diff
@@ -219,19 +259,23 @@ def resolve_messages(example: dict) -> list[dict]:
...
-                case _:
-                    raise ValueError(f"Invalid message role: {message['role']}")
+        # Dispatch the loss-mask config to a renderer-side role filter.
+        # The string sentinels map to no filter ("sampled" → trust
+        # sampled_mask) or an always-True filter ("all"); the
+        # :class:`LossMaskRolesConfig` instance maps to a per-role
+        # boolean lookup. The chat-template fallback below has no
+        # ``is_sampled`` signal so we treat sentinel modes as the
+        # ``LossMaskRolesConfig()`` defaults (assistant-only).
+        loss_mask_cfg = self.loss_mask_config
```
snimu, line 269: I feel like all this logic, plus the `always_true_role_filter`, should be in `role_filter_for` which just returns a tuple of the `renderer_role_filter` and `fallback_role_filter`. I found this split in logic between the role filter helper and this dataset quite confusing.

src/prime_rl/trainer/sft/data.py
```diff
@@ -246,15 +290,15 @@ def should_mask(message: dict) -> bool:
             input_ids, loss_mask = build_training_sample(
```
snimu, line 290: i dislike the extremely different names between `build_training_sample` and `build_incremental_token_mask` even though they do the same thing, just with a different backend. Not  necessarily important for this PR, but I didn't want to forget it

PR #1940, opened 2026-03-03
add dp-rank-routing

faresobeid:
> [!NOTE]
> **Medium Risk**
> Changes rollout scheduling and inference client selection, which can affect throughput and rollout consistency, especially in elastic pools. Risk is mitigated by a default `dp_rank_count=1` behavior preserving existing routing.
> 
> **Overview**
> Adds data-parallel rank–aware inference routing via new `client.dp_rank_count`: each `client.base_url` can be expanded into multiple logical clients with an `X-data-parallel-rank` header to keep multi-turn rollouts pinned to a consistent DP engine (improving KV cache reuse).
> 
> The RL config now auto-derives `orchestrator.client.dp_rank_count` from inference DP settings when not explicitly set, the scheduler pins a rollout group to a single selected client and tracks load by `(api_base_url, dp_rank)` to remain stable across elastic pool refreshes, and both static + elastic inference pool client construction is updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 41c28e2024f3ffaa7e9dd7a9e5656dbf251a8784. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #2708, opened 2026-06-04
fix(configs): remove vestigial client dp_rank_count / X-data-parallel-rank

S1ro1:
## Summary

Removes the vestigial client-side `dp_rank_count` / `X-data-parallel-rank` mechanism.

## Background

`dp_rank_count` + the `X-data-parallel-rank` header come from #1940 ("add dp-rank-routing"): with a single hybrid-LB server fronting multiple DP engines, the client pinned each rollout to a DP shard for KV reuse. **#2696 ("use external-LB everywhere, drop hybrid-LB")** replaced that with one endpoint per DP rank, router-balanced — so the header no longer has a job.

It isn't just dead, it's harmful. The regular `vllm-router` forwards client headers to the worker, and vLLM validates the rank:

```
data_parallel_rank N is out of range [0, num_ranks).
```

Single-DP backends (`--data-parallel-size 1`, how dense external-LB launches) have `num_ranks = 1`, so any request carrying rank > 0 is rejected. Affinity / KV reuse is now the router's job (prefix-cache scoring), not the client's.

## Change

- Remove `client.dp_rank_count` from `ClientConfig`.
- `setup_clients`: one client per `base_url`, no `X-data-parallel-rank` header.
- Drop the `dp_rank_count` auto-set in the RL resolver and the field pass-through in the elastic pool.
- `client_identity` keys on `api_base_url` alone (each DP rank is its own endpoint).
- CHANGELOG + tests updated.

No checked-in config sets `dp_rank_count`, and the server-side header reader already handles its absence. Breaking only for external configs that set it (`extra="forbid"` rejects it); nothing to migrate.

Related: #2707 (NCCL world-size) — together they fully fix dense multi-node external-LB.

---

> [!NOTE]
> **Medium Risk**
> Breaking config removal (`extra="forbid"`) and changed inference routing identity; fixes harmful rank headers on single-DP vLLM but requires one base URL per DP rank in user configs.
> 
> **Overview**
> Removes client-side **`dp_rank_count`** and the **`X-data-parallel-rank`** header path that expanded each `base_url` into multiple pinned DP clients. With external-LB, each DP rank is already a separate endpoint in **`client.base_url`**, so the orchestrator now builds **one inference client per URL** and load-balances on **`api_base_url`** only.
> 
> **`ClientConfig`** drops the field (breaking for configs that still set it). **`setup_clients`** no longer loops ranks or sets rank headers. The RL **`auto_setup_inference_client`** hook stops copying inference DP size into the client. **Elastic** client rebuilds no longer forward **`dp_rank_count`**. CHANGELOG documents the migration (delete the key; no replacement).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 55d3fa853185b8eb595491e96c527881a9bc2012. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


samsja (approved):

samsja (changes requested):
ah wait actually do we have router on single node ? if not we cant remove

PR #2721, opened 2026-06-05
feat(orchestrator): per-env advantage strategy

hallerite:
## What

Make advantage computation configurable **per training environment** instead of a single global setting.

## Why

Mixed-env runs currently share one advantage function. Different envs often want different advantage computation — plain GRPO for one, a custom or length-penalized advantage for another. This lets each env carry its own.

## Changes

- `TrainEnvConfig` gains an `advantage` field (`default` / `custom`), inheriting the top-level `[orchestrator.advantage]` when unset — the same inheritance pattern already used for `group_size`.
- Each `TrainEnv` builds its own advantage fn from `config.advantage` (mirroring how it derives `sampling_args`); `TrainSink.process_group` applies it via `self.train_envs.get(env_name).advantage_fn`. The global `advantage_config` param and its call-site arg are removed.
- Moved the advantage / length-penalty config classes above `EnvConfig` so `TrainEnvConfig` can reference `AdvantageConfig` (the module doesn't use `from __future__ import annotations`, so annotations evaluate eagerly).
- Doc note in `docs/algorithms.md`.

## Behavior

Behavior-preserving: with a default config, every env inherits the global default → identical advantages to before.

```toml
[orchestrator.advantage]
type = "default"            # inherited by every env unless overridden

[[orchestrator.train.env]]
id = "math-env"             # inherits the default

[[orchestrator.train.env]]
id = "agent-env"
advantage = { type = "custom", import_path = "my_module.normalized_advantage" }
```

## Testing

- `ruff check` + `ruff format --check` clean on edited files.
- `pytest tests/unit/orchestrator/test_advantage.py` → 17 passed; `tests/unit/test_configs.py` → 106 passed.
- Verified per-env inheritance, per-env override, and the `rollouts_per_example` alias via a config-construction check.
- Validated end-to-end on 2× RTX PRO 6000 (Blackwell): a **single-env** run (one env with a length-penalized advantage) and a **multi-env** run (two `reverse-text` envs with *separate* advantages — plain GRPO vs length-penalized) each completed 3 steps cleanly (exit 0, `Error 0.0%`, both envs trained, GPUs released).

---

> [!NOTE]
> **Medium Risk**
> Changes how advantages are computed in multi-env runs when overrides are used; single-env default inheritance should match prior behavior but mixed-env training logic is now env-scoped.
> 
> **Overview**
> **Per-env advantage** replaces a single orchestrator-wide advantage function with optional `advantage` on each `[[orchestrator.train.env]]`, inheriting `[orchestrator.advantage]` when unset (same pattern as `group_size`).
> 
> Each `TrainEnv` now builds its `advantage_fn` once at init; `TrainSink.process_group` calls `assign_advantages` with `train_envs.get(env_name).advantage_fn` instead of a shared fn on the sink. Config validation copies the top-level `advantage` into envs that did not override it; advantage/length-penalty Pydantic types were moved above `EnvConfig` so `TrainEnvConfig` can reference them. `docs/algorithms.md` documents mixed-env TOML overrides.
> 
> Default configs behave as before when every env inherits the global advantage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c411dab2a4ba1c49bb2cae81dc3a1ca69ad0af63. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
@@ -143,6 +143,49 @@ class EvalSamplingConfig(BaseConfig):
         return data
 
 
+class TokensLengthPenaltyConfig(BaseConfig):
+    type: Literal["tokens"] = "tokens"
+
+    completion_weight: float = Field(1.0, ge=0, allow_inf_nan=False)
+    """Weight on model completion tokens. Finite and non-negative."""
+
+    tool_response_weight: float = Field(1.0, ge=0, allow_inf_nan=False)
+    """Weight on tool-response tokens (read from the rollout's ``*_total_tool_response_tokens`` harness metric; 0 if absent). Finite and non-negative."""
+
+
+class TurnsLengthPenaltyConfig(BaseConfig):
+    type: Literal["turns"] = "turns"
+
+
+LengthPenaltyConfig: TypeAlias = Annotated[
+    TokensLengthPenaltyConfig | TurnsLengthPenaltyConfig,
+    Field(discriminator="type"),
+]
+
+
+class DefaultAdvantageConfig(BaseConfig):
+    type: Literal["default"] = "default"
+
+    length_penalty: LengthPenaltyConfig | None = None
+    """Correctness-gated length penalty. ``tokens`` shapes by weighted token cost; ``turns`` shapes by trajectory turn count; None disables shaping. In mixed groups, lower-cost correct rollouts get amplified advantage (up to 2x), higher-cost correct rollouts are unchanged, incorrect untouched. In all-correct groups, below-average-cost rollouts get advantage in [0, 1], others get 0."""
+
+
+class CustomAdvantageConfig(BaseConfig):
+    type: Literal["custom"] = "custom"
+
+    import_path: str
+    """Import path to the advantage function (e.g. ``my_module.my_advantage``)."""
+
+    kwargs: dict[str, Any] = Field(default_factory=dict)
+    """Kwargs forwarded to the advantage function."""
+
+
+AdvantageConfig: TypeAlias = Annotated[
+    DefaultAdvantageConfig | CustomAdvantageConfig,
+    Field(discriminator="type"),
+]
+
+
 class EnvConfig(BaseConfig):
     id: str = "reverse-text"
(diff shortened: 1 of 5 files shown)


packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -143,6 +143,49 @@ def _deprecate_max_tokens(cls, data: Any) -> Any:
...
+
+    tool_response_weight: float = Field(1.0, ge=0, allow_inf_nan=False)
+    """Weight on tool-response tokens (read from the rollout's ``*_total_tool_response_tokens`` harness metric; 0 if absent). Finite and non-negative."""
+
+
+class TurnsLengthPenaltyConfig(BaseConfig):
+    type: Literal["turns"] = "turns"
+
+
+LengthPenaltyConfig: TypeAlias = Annotated[
```
mikasenghaas, line 160: wondering if default should not just be plain reward - mean(reward). and the length penalty addition is a custom function

src/prime_rl/orchestrator/train_sink.py
```diff
@@ -58,9 +57,13 @@ def __init__(
...
-        # Built once — custom advantage funcs do an ``import_object`` and
-        # we don't want to pay that per group. ``None`` = reward-only path
-        self.advantage_fn = setup_advantage_fn(advantage_config) if advantage_config is not None else None
+        # Built once per env — custom advantage funcs do an ``import_object`` and
+        # we don't want to pay that per group. Each env carries its own advantage
+        # config (inheriting the top-level default when unset). ``None`` = reward-only path.
+        self.advantage_fns: dict[str, AdvantageFn | None] = {
+            env.name: setup_advantage_fn(env.config.advantage) if env.config.advantage is not None else None
+            for env in train_envs
+        }
```
mikasenghaas: should we init them in the training envs themselves, might make sense if they are tighlty coupled no?

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
hallerite: I can see why, but I think it may just be more convenient for now to keep it as-is, as we would need to find a nice and general way to combine different advantage functions or have to implement every combination.

mikasenghaas (approved):

PR #2723, opened 2026-06-07
improve sequence packing

faresobeid:
remove get_response_lengths and replace with explicit sequence lengths

Use flops aware micro batch scheduling and some other improvements

Mainly from https://x.com/slime_framework/status/2063287832182182249?s=20

---

> [!NOTE]
> **Medium Risk**
> Changes the distributed batching path and how RL loss is segmented; incorrect `sequence_lengths` or partitioning could skew gradients or cause straggers, though behavior is heavily covered by new unit tests.
> 
> **Overview**
> Improves **RL micro-batch packing and multi-GPU scheduling** by carrying **`sequence_lengths`** on every `MicroBatch` (set at pack time, updated on pad) and using them for **per-sequence loss splits**, token export, and fake/real dataloaders—**replacing `get_response_lengths(position_ids)`**.
> 
> Packing is refactored around **`_MicroBatchBin`** (FFD pack → **`_materialize_bin`**). When there are multiple train workers, bins can be **split by estimated workload** (`_expand_bins_by_splitting`) so micro-batch count aligns with DP without relying only on zero-loss dummy batches.
> 
> **`build_bin_cost(model.config)`** scores bins from estimated forward FLOPs (linear + quadratic in seqlen); **`balanced_partition`** (Karmarkar–Karp + swap refinement) assigns micro-batches within each text/multimodal group so ranks get **near-equal compute**, not just strided round-robin. **`bin_cost`** is threaded through packer, `prepare_batch`, and `DataLoader`.
> 
> Tests add randomized packing invariants and cases for token vs FLOP balancing and split-to-align behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbe46c92f873da9690744f4cd12a066771efb384. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -450,6 +450,9 @@ class DataLoaderConfig(BaseConfig):
     fake: FakeDataLoaderConfig | None = None
     """Use a fake data loader sampling random micro-batches (for debugging)."""
 
+    balance_by_flops: bool = True
+    """Balance packed micro-batches across data-parallel ranks using model-estimated FLOPs."""
```
samsja: would make this always true tbh

faresobeid: Fair, only made it an option in case for some new models we don't have an accurate flops calculation instantly but does seem useless

samsja (commented):
can we merge all the utils file together, dont want an parittion file

src/prime_rl/trainer/rl/packer.py
```diff
@@ -32,6 +34,7 @@ def __init__(
         pad_to_multiple_of: int,
         tokenizer: PreTrainedTokenizer,
         config: TransportConfig,
+        bin_cost: Callable[[Sequence[int]], int] | None = None,
```
samsja: why is this configurable ?

src/prime_rl/trainer/batch.py
```diff
@@ -141,8 +134,149 @@ def _is_multimodal_sample(sample: MicroBatch) -> bool:
...
+        teacher_logprobs=teacher_logprobs,
+        temperatures=temperatures,
+        rewards=rewards,
+        lora_num_tokens=lora_num_tokens,
+        routed_experts=routed_experts,
+        mm_token_type_ids=mm_token_type_ids,
+        env_names=env_names,
+        mm_kwargs=first_sample.mm_kwargs if _is_multimodal_sample(first_sample) else None,
+        training_mode=first_sample.training_mode,
+    )
```
samsja, line 254: is it time to have a dataclass here instead of all the param ?

faresobeid: hmm idk seemed nicer, otherwise u would duplicate most of MicroBatch

samsja (dismissed):

samsja (approved):

PR #2725, opened 2026-06-07
orch improvements

faresobeid:
If trainer goes too far ahead inference but inference started evals, don't pause evals

Clean up evaluate function that isn't used anymore 

Make finalize_eval and ZMQ sends async

Surface orchestrator component task failures 

---

> [!NOTE]
> **Medium Risk**
> Changes dispatch gating and off-policy cancellation semantics for train vs eval, which affects RL pipeline timing and metric correctness; async ZMQ and threaded saves alter failure/backpressure behavior on the orchestrator event loop.
> 
> **Overview**
> **Eval scheduling and off-policy behavior** change so in-flight evals are version-pinned measurements: `on_new_version` only ages/cancels **train** rollouts, and in `PREFER_EVAL` mode eval work is scheduled even when the orchestrator has cleared `dispatch_allowed` (train still respects the gate when the trainer is ahead).
> 
> **Async I/O on the event loop**: `finalize_eval_batch` is awaited from the main loop, writes eval JSONL via `asyncio.to_thread`, and logs **policy version** (with a wandb metric and mixed-version warning) instead of max off-policy on eval success lines. **ZMQ** training-batch sends use `zmq.asyncio` and `await send_multipart` so SNDHWM backpressure does not block the loop.
> 
> **Cleanup**: removes the unused `EvalEnv.evaluate` bulk path from `envs.py`. Adds **`orjson`** for faster rollout JSONL serialization (used with threaded saves).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4f5a5af4abc853d85936381938ac4229950b260. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/orchestrator/envs.py
```diff
@@ -183,147 +177,6 @@ def __init__(self, config: EvalEnvConfig):
         self.sampling_args = config.sampling.to_sampling_args()
         self.examples = self.env.get_eval_dataset(n=config.num_examples).to_list()
 
-    async def evaluate(
```
mikasenghaas, line 186: oh lol wtf, how is this still in

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -509,6 +510,17 @@ async def main_loop(self) -> None:
             if train_batch is not None and not self.draining and not self.stopped.is_set():
                 await self.finalize_train_batch(train_batch)
 
+    def check_component_tasks(self) -> None:
```
mikasenghaas: hm, have we seen a failure mode for this?

src/prime_rl/orchestrator/dispatcher.py
```diff
@@ -12,7 +12,9 @@
   source's emptiness), so in-flight rollouts of the opposite kind drain
   naturally on either side of an eval boundary.
 - ``on_new_version`` (called by the watcher) bumps ``off_policy_steps`` on
-  every in-flight rollout and drops groups past ``max_off_policy_steps``.
+  in-flight train rollouts and drops groups past ``max_off_policy_steps``.
+  Eval rollouts are measurements for the policy version they started with,
+  so they are allowed to finish even if training advances.
```
mikasenghaas, line 17: im not sure they should? i agree if we are at async barrier we shoudn't block evals. but i still think we should prevent infinite eval staleness? at least this is how the configs intend it rn

faresobeid: Hmm I mostly thought of it as a special case when the barrier is hit right before switching to evals
In any other setting we would pause training no?
So it's like to let evals go then pause the inference after that

src/prime_rl/orchestrator/orchestrator.py
faresobeid: Idk this part was a codex recommendation that seemed reasonable

mikasenghaas: hm i dont like codex, can we remove this until we have an actual failure somewhere

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: hm but the code is not a special case, no? like it allows an eval rollout to run across infinite training steps or am i reading this wrong?

faresobeid: no its basically if our inference went really fast and produced 3 batches of data extra, when evals start, our trainer keeps going on these 3 batches. If our max off policy steps was 2, then when the second trainer step happens we would cancel our evals no?

mikasenghaas (approved):

PR #2733, opened 2026-06-08
fix: add prime-pydantic-config as direct dep so uv uses editable path…

samsja:
… source

The [tool.uv.sources] override for prime-pydantic-config was being ignored because it was only a transitive dependency (via prime-rl-configs). uv only applies source overrides for packages that appear in project.dependencies. Adding it as a direct dependency makes uv resolve from the local editable path (deps/pydantic-config) instead of PyPI.

---

> [!NOTE]
> **Low Risk**
> Config loading and dependency resolution only; behavior is additive (env overrides) with tests and no changes to training or inference runtime logic.
> 
> **Overview**
> Adds **environment-variable overrides** for top-level run configs by declaring per-entrypoint `env_prefix` values (e.g. `PRIME_RL_ORCH_`, `PRIME_RL_TRAINER_`, `PRIME_RL_INFER_`) on `RLConfig`, `TrainerConfig`, `OrchestratorConfig`, `InferenceConfig`, `SFTConfig`, and `EnvServerConfig`, using nested `__` field paths.
> 
> **`prime-pydantic-config`** is listed as a **direct** `prime-rl` dependency so uv honors the editable `deps/pydantic-config` source override (it was previously only transitive via `prime-rl-configs`). The lockfile moves from a PyPI pre-release to the local editable **0.3.0** package.
> 
> `InferenceConfig` now imports **`BaseConfig` from `prime_rl.utils.config`** (re-export of pydantic-config) instead of importing it straight from `pydantic_config`.
> 
> New unit tests cover env injection, nested keys, CLI winning over env, `RLConfig` sub-config propagation, and **no cross-prefix leakage** between orchestrator and trainer.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 66d1aee406900b09c90d2f03ac793ad69a5770be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,7 @@ readme = "README.md"
 requires-python = "~=3.12.0"
 dependencies = [
     "prime-rl-configs",
+    "prime-pydantic-config",
     "beartype>=0.21.0",
     "datasets>=4.0.0",
     "jaxtyping>=0.3.2",
--- a/uv.lock
+++ b/uv.lock
@@ -3476,19 +3476,33 @@ wheels = [
 
 [[package]]
 name = "prime-pydantic-config"
-version = "0.3.0.dev86"
-source = { registry = "https://pypi.org/simple" }
+version = "0.3.0"
+source = { editable = "deps/pydantic-config" }
 dependencies = [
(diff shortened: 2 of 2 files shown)


Jackmin801 (dismissed):
ook lgtm

PR #2742, opened 2026-06-09
feat: vf v1 <> nano bridge

mikasenghaas:
Companion PR to PrimeIntellect-ai/verifiers#1576 for verifiers v1 training integration.

## Summary

Integrates **verifiers v1** into prime-rl. Training rolls out against a v1 **env server** that returns `vf.Trace` rollouts, with **renderer-only** (token-in/out) training and a unified **`token_ids`** sample schema.

- **v1 env server.** The orchestrator spawns a `vf` `EnvServer` per environment (subprocess runtime, ZMQ) and drives it purely by **task index** — it never loads the env. v0 envs run through the verifiers legacy bridge; native v1 tasksets serve directly; both return `vf.Trace` over one protocol.
- **`vf.Trace` rollouts.** `Rollout` *is* a `vf.Trace[TaskT]`; prime-rl's orchestration metadata lives on it as `exclude=True` fields, so a dumped rollout is a plain trace. The wire trace is re-typed into `Rollout` via `model_construct` (no serialize round-trip of the node tensors).
- **Unified `token_ids` schema.** Rollouts are tokenized renderer-side into one `TrainingSample` / `MicroBatch` representation (replaces the prompt/completion split); main's `bin_cost` packing and `MicroBatch.sequence_lengths` are grafted onto it.
- **Renderer-only training.** RL/OPD student and SFT teacher all roll out through the renderer (token) client; SFT re-renders to backfill tokens.
- **Configs consume vf v1.** `EnvConfig` extends `vf.EnvServerConfig` (`taskset` / `harness` / `runtime` / `pool`); `EvalClientConfig` / `TrainClientConfig` replace the prior client config.
- **verifiers** submodule bumped to the v1 line; **`main` (v0.6.0)** merged in.

## Verification

Scaleswe v1 ([W&B](https://wandb.ai/primeintellect/prime-rl/workspace?nw=xaxh64baiy))

Screenshot 2026-06-22 at 9 52 18 PM

Hendrycks v0/v1 ([W&B](https://wandb.ai/primeintellect/prime-rl/workspace?nw=l4hzuis9mv))

Screenshot 2026-06-22 at 9 50 42 PM

General Agent v1 on Modal ([W&B](https://wandb.ai/primeintellect/prime-rl/workspace?nw=xbxj6su68z))

Screenshot 2026-06-22 at 9 51 04 PM

Wiki Search v0 ([W&B](https://wandb.ai/primeintellect/prime-rl/workspace?nw=e8cxkkl1jgl))

[...]

PR #2720, opened 2026-06-04
[codex] Allow SFT without a teacher

tim0120:
## Summary

- allow `orchestrator.training_mode = "sft"` without a configured teacher
- route teacher-backed SFT train rollouts through the teacher pool, and teacherless SFT through the existing student rollout pool
- keep renderer forcing scoped to teacher-backed SFT; teacherless SFT keeps the normal student renderer path
- keep prefix-cache salting for student-backed train rollouts while preserving unsalted cache reuse for frozen teacher-backed SFT

## Motivation

This is needed for replay-backed SFT workflows where the train env can provide the supervision without a teacher model. PrimeRL should only need the generic capability: SFT can run without `orchestrator.teacher`; replay-specific taskset/env behavior stays outside this PR.

## Validation

- `uvx ruff==0.13.0 check --config=pyproject.toml packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py src/prime_rl/orchestrator/orchestrator.py src/prime_rl/orchestrator/dispatcher.py`
- `python3 -m py_compile packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py src/prime_rl/orchestrator/orchestrator.py src/prime_rl/orchestrator/dispatcher.py`
- `git diff --check`

## Notes

The earlier replay-specific package wiring, debug config, docs, tests, Verifiers submodule pointer, token-usage backfill, and zero-adv filter changes have been removed from the effective diff. The PR is now limited to the teacher-optional SFT routing/cache consequences.

---

> [!NOTE]
> **Medium Risk**
> Changes SFT train inference routing and prefix-cache salting when no teacher is set; misconfiguration could affect rollout correctness or cache reuse, but scope is limited to orchestrator config and dispatch.
> 
> **Overview**
> **SFT can run without `orchestrator.teacher`.** Config validation no longer requires a teacher for `training_mode = "sft"` (only **OPD** still mandates one). Forcing `renderer = None` is limited to **teacher-backed SFT**; teacherless SFT keeps the normal student renderer path.
> 
> **Train rollout routing** is split: SFT with a teacher still uses the frozen teacher inference pool and **disables** prefix-cache salting on train rollouts; SFT without a teacher uses the **student** rollout pool with **policy-version cache salt**, same as RL/OPD train paths. The dispatcher takes an explicit `use_cache_salt` flag instead of inferring this from `training_mode == "sft"` alone.
> 
[...]

PR #2641, opened 2026-05-26
[feat] multiplex trainer token export and agg to prime monitor on orch side

Jackmin801:
On trainer side, each rank appends the token export metrics to its own `output_dir/<run_id>/token_exports/step_<run_step>/rank_<rank>.jsonl` at every batch. Before we are about to take an optimizer step for a run, we write to that steps `token_exports/step_<run_step>`  stable file. This is required because of the round robin batching behavior. So each microbatch for the run, it accumulates more jsonls. We know we have all the jsonls for that run at that step when it tries to take optimizer step.

On orchestrator side, everytime we are about to log to monitor, we check for the last token exported step at `output_dir/<run_id>/token_exports/step_<run_step>/` for a stable file. If we find it, then we log the mean and max of `entropy`, `kl_mismatch` and `trainer/step`. We need to do it this way because of the async. So the trainers metrics will lag the orchestrator and the lag can be corrected by using `trainer/step` as x-axis.

Because of the lag, the orchestrator's ending now needs a trainer metric drain. So when the orchestrator has finished doing all its inference and writing all its rollouts. It needs to stay alive in a loop that waits for all the trainer metrics to come in. This way we collect the trainer metrics for every step despite the trainer metrics lagging the orchestrator metrics.

---

> [!NOTE]
> **Medium Risk**
> Breaking trainer config migration and new shutdown drain behavior; orchestrator disk I/O on the training hot path when export is enabled, but gated on STABLE files and best-effort parsing.
> 
> **Overview**
> Replaces **`trainer.experimental.token_export`** with a top-level **`trainer.enable_token_export`** flag (CLI **`--enable-token-export`**), documented as a breaking change with no alias.
> 
> On the **trainer**, token exports are reworked for **multi-run** packing: microbatches carry **`run_id`** / **`run_step`**, JSONL is appended per rank under run-local **`token_exports/step_<run_step>/`**, and a **`STABLE`** marker is written only after a distributed barrier when that run is **`ready_to_update`** (so a run step split across trainer steps is complete before orchestrator reads it).
> 
> On the **orchestrator**, a new **`token_export_metrics`** collector scans stable export steps, aggregates loss-masked **entropy** and **mismatch_kl** (mean/max), and **`MetricsBuilder`** merges them into monitor logs as **`trainer/*`** plus **`trainer/step`** for lag-correct plotting. After **`max_steps`**, **`_drain_token_export_metrics`** polls for any trailing stable steps so late trainer exports still get logged.
> 
[...]

PR #2746, opened 2026-06-09
feat: algorithm abstraction — named algorithm classes + inline frozen-model references (grpo, opd, sft_distill, self_distill, echo)

hallerite:
> [!NOTE]
> **On verifiers-v1.** Targets `main` after the vf-v1 ⇆ nano bridge ([#2742](https://github.com/PrimeIntellect-ai/prime-rl/pull/2742)) merged. `Rollout` is a `vf.Trace` subclass, samples are flat-token `TrainingSample`, scoring runs off `vf.Trace` branches/nodes, and the renderer is required. `origin/main` is merged in (latest: #2879 internal cleanup, #2880 eval-metric fix, #2881 dead-`processor=` removal, verifiers bumped to `62dc38a5`); the branch is even with `main` and CI is green.
>
> **The two prior reach-arounds are now resolved on this branch:**
> 1. **OPD/OPSD prefill scoring** — the hand-rolled `compute_prefill_logprobs` is gone; prefill logprobs go through `InferencePool.score()` (`PrefillScorer`).
> 2. **echo granularity** — upstream `MessageNode.is_content` landed, so echo weights only message-body tokens (scaffold excluded), with a whole-span fallback when attribution is absent. Pre-v1 parity restored.
>
> **Verification:** verifiers pinned at submodule `62dc38a5` (has `is_content`, `Trace.to_record()`, `Branch.input_len`/`output_len`), no merge conflicts, `ruff` clean, imports resolve against the pinned verifiers, all CI green (unit + every integration suite). GPU smokes in a prior session confirmed all three loss components (`rl` / `ce` / `ref_kl`) train and learn.

---

## What

Makes prime-rl's training algorithm a first-class, hackable abstraction — removes model *roles* from the pipeline, unifies every training signal under one concept (the **advantage**: credit assignment and loss routing, fused), and makes each algorithm a **named runtime class that owns its methods**. There is no registry and no separate "token scorer": there is the live **policy** — the only model prime-rl ever hosts — and per-env algorithms whose model references are either `"policy"` or an inline externally-hosted frozen endpoint.

An algorithm is a bundle of **two components**, configured under `[orchestrator.algo]` (per-env: `algo = {...}`) and resolved **per env**. **`algo.type` names the algorithm** — `grpo`, `max_rl`, `opd`, `opsd`, `sft`, `echo` — and each type's class defaults are its vetted setting; there is no separate preset layer:

1. **Sampling** — how train rollouts are produced: which model generates them, `"policy"` or an inline frozen endpoint. At runtime this builds the env's `Sampler` (`orchestrator/sampler.py`) — the pool rollouts come from, the liveness consequences (logprobs, prefix-cache salting, staleness), and the home of future sampling strategies (replay buffers, branching).
2. **Advantage** — the per-token training signal: one mapping from a finalized rollout to per-token *(loss component, weight)* pairs. Credit assignment computes the magnitude, loss routing picks the component — two coordinates of the same output, so they are one config component and one runtime object. Group-relative strategies compute scalars on the orchestrator and ship numbers; reference-KL strategies query a reference model as each rollout arrives and ship its prefill logprobs for the trainer to evaluate against the live policy. The strategy determines the action-token loss component (`rl` / `ce` / `ref_kl`) and what happens to env-provided observation tokens (masked out by default; `echo` trains on them with weighted CE — selected **by message role** via the renderer's per-token attribution, each role at its own `alpha`, tool-response bodies at λ=0.1 by default, optionally narrowed by a user-supplied per-rollout token filter).

The training loss is a **sum of three components** — `rl` (DPPO+KL or custom), `ce` (masked NLL), `ref_kl` (reverse KL to a reference as the PG signal) — **each normalized by its own global token count**:

$$\mathcal{L} = \tfrac{\sum \mathcal{L}_{rl}}{N_{rl}} + \tfrac{\sum \mathcal{L}_{ce}}{N_{ce}} + \tfrac{\sum \mathcal{L}_{ref\_kl}}{N_{ref\_kl}}$$

The orchestrator stamps per-token component **weight streams** (`rl_weights` / `ce_weights` / `ref_kl_weights`) plus the per-token **advantage stream** (`advantages` — there is no scalar advantage anywhere; uniform group credit is broadcast over completion tokens at assignment); a weight scales that component's per-token loss, `0.0` removes the token from the component's mask *and* denominator, and components may **overlap** on the same token (gradients sum). Per-component normalization means components never dilute each other: echo's observation tokens no longer shrink the rl term's effective per-token learning rate (previously both shared one global denominator, so the rl gradient scaled with the batch's obs/action ratio), and a supervised env packed next to a GRPO env doesn't soften its gradient.

```toml
[orchestrator.algo]
type = "opd"

[orchestrator.algo.teacher]   # opd's own field: the frozen model it scores against
name = "Qwen/Qwen3-32B"
base_url = ["http://localhost:8001/v1"]
```

The trainer is **algorithm-blind** (component weight streams ship per token on the wire; the trainer executes the three fixed components), and the orchestrator pipeline is too: dispatcher, train sink, and orchestrator call hooks on each env's `Sampler` + `Algorithm` objects and never branch on algorithm config or model roles.

### The API — two seams to override

Writing an algorithm is subclassing `Algorithm` (`orchestrator/algo/base.py`) and overriding the hook(s) your signal needs. The whole override surface is small and identical for every algorithm — two declarations, a lifecycle hook, and **two scoring hooks on one scope/timing ladder**:

```python
class Algorithm:
    # declaration — read by the pipeline, never branched on
    action_loss_type: ClassVar[ActionLossType] = "rl"   # which component action tokens feed: rl | ce | ref_kl

    async def setup(self) -> None: ...                   # connect pools to frozen models via self.connect(ref)

    async def score_rollout(self, rollout: Rollout)      -> None: ...   # one rollout, on arrival (model I/O here)
    async def score_group  (self, group:  list[Rollout]) -> None: ...   # the cohort, before filtering
```

The two hooks are **one ladder of scope, the wider scope unlocked by a later barrier** — so scope and timing coincide:

| hook | input | fires | scope | model access |
|---|---|---|---|---|
| `score_rollout` | `Rollout` | as each rollout is tokenized | one rollout, no siblings | **yes** (`self.policy_pool` / `self.teacher_pool`) |
| `score_group` | `list[Rollout]` | on group completion, **before** filtering | the whole cohort | — |

- **`score_rollout`** — rollout-local signals (raw reward, process rewards, echo's observation weighting) *and* the model-access seam: query a reference (OPD's frozen `teacher`, OPSD's live policy) and attach per-token logprobs as each rollout arrives. It runs *before* the pre-batch filters, so it pays reference compute on rollouts that may then be filtered — OPD/OPSD assign no advantage, so advantage filters never drop them anyway, and the serial main loop bounds in-flight scoring without an explicit cap.
- **`score_group`** — the group-relative seam: GRPO / MaxRL baselines, std-normalization. It runs *before* filtering on purpose, because advantage-based filters read the streams it writes.

Both hooks are `async`, so either may do I/O — a teacher or process-reward model at arrival, or a judge at group time whose verdict a pre-batch filter then reads; a hook that only does advantage math never awaits.

**The input type is always `Rollout`** (`orchestrator/types.py`) — which *is* the env's typed `vf.Trace` subclass, with prime-rl's training fields bolted on. A hook reads the trace directly and writes credit through one method:

```python
rollout.reward       # the rollout's scalar reward (vf.Trace property)
rollout.nodes        # the message graph; rollout.num_turns / .task / .info / ... read straight off the trace
rollout.samples      # the TrainingSamples the renderer built (per branch)
rollout.env_name     # the producing env
rollout.assign_advantages(scalar | list[float])   # write the rl credit stream
```

`assign_advantages` takes either a **scalar** (broadcast over the rollout's trainable mask-True tokens — the common, uniform-credit case) or an explicit **per-token list** (aligned full-length to the samples' tokens). It writes the per-token `advantages` stream; advantages are per-token everywhere stored or shipped, the scalar is only a write-time convenience. For non-rl components an algorithm writes the sample stream directly — echo writes `sample.ce_weights` over `is_content` tokens at `score_rollout` time — at the stage where that data is valid. There's no read-only facade: this is a power-user extension surface, so a hook gets the whole `Rollout` (an earlier `RolloutView` handle was collapsed away — everyone reached through its `.raw` escape hatch anyway). Stage discipline (don't read `is_filtered` before filtering, don't read `advantages` before assigning) is a convention, not an enforced wall.

The pipeline drives the hooks through two non-virtual `Algorithm` methods named by the barrier they sit on — `finalize_rollout` / `finalize_group` — so the **hooks** answer *"what do I override"* and the **finalize methods** answer *"when does the pipeline call me"*. Default hook bodies are no-ops: an algorithm that only needs a group baseline overrides `score_group` alone and inherits the rest.

### The algorithm classes

Each algorithm is a named runtime class — the algorithm object *is* the algorithm. Dispatch is keyed on `algo.type` — it names the algorithm, and each config class's defaults are its vetted parameterization:

Each class overrides the scoring hook(s) its signal needs, on the scope/timing ladder `score_rollout` (one rollout, on arrival; async, model access) → `score_group` (the cohort, before filtering):

| `algo.type` | Class | Component | hook(s) |
|---|---|---|---|
| `grpo` | `GRPOAlgorithm` | `rl` | `score_group`: group-norm scalars (optional length penalty) |
| `max_rl` | `MaxRLAlgorithm` | `rl` | `score_group`: mean-normalized group scalars ([arXiv:2602.02710](https://arxiv.org/abs/2602.02710): unbiased for the order-`group_size` truncation of the max-likelihood objective) |
| `echo` | `EchoAlgorithm` | `rl` + `ce` | `score_rollout`: env-provided tokens selected by message role (`roles.<role>.alpha`, tool bodies @ 0.1 default), optional user token filter; `score_group`: group-norm scalars (inherited) |
| `opd` | `OPDAlgorithm` | `ref_kl` | `score_rollout`: own-context prefill under the **teacher** (a required frozen model; no credit — `advantages=None`, filters skip) |
| `opsd` | `OPSDAlgorithm` | `ref_kl` | `score_rollout`: demo-conditioned prefill under the **live policy** (self-distillation, no model to declare; no credit) |
| `sft` | `SFTDistillAlgorithm` | `ce` | `score_group`: group-norm (feeds reward-based filtering) |

Reading a class top to bottom reads the algorithm — one module per algorithm under `orchestrator/algo/` (`grpo.py`, `opd.py`, …, with the base class and its non-virtual `finalize_*` drivers in `base.py` and dispatch in the package `__init__`); writing your own is subclassing `Algorithm` and overriding the seam(s) your signal needs (see [The API — two seams to override](#the-api--two-seams-to-override) above). Duplication of orchestration between similar algorithms (OPD vs OPSD) is accepted so each class stays self-contained; shared *math* (group normalization, prefill alignment, length penalties) lives as plain functions in `algo/advantage.py`. Algorithms take exactly one host-owned resource beyond their config — the live policy pool (`Algorithm(config, policy_pool)`, never closed by the algorithm). Anything else an algorithm needs it builds from its own config in `setup()`: `opd` connects its frozen `teacher`; `opsd` builds the renderer for its demonstration hint (the tokenizer is always the live policy's — self-distillation has no separate model). Text → token ids always goes through the renderer, the same path the policy's own prompts take.

One deliberate expressiveness trade: loss routing is not a free config axis (there is no `algo.loss`; you can't write `opd` + observation-CE in TOML) — routing variation *is* algorithm variation, expressed as a class. ECHO is a proper advantage type, not a flag on GRPO. Its selection surface is configurable where it matters: per-role `alpha` (`roles.system/user/assistant/tool` — setting any role replaces the whole table) and an optional `filter` hook (`import_path` + `kwargs`, called once per rollout as `filter_fn(rollout, **kwargs) -> list[list[bool]]`, one keep-mask per trainable branch spanning that branch's `token_ids`; it can only *narrow* the role selection, never widen it) — the raw rollout exposes message text and sampling logprobs, so warning filters and low-probability filters are user code, no framework surface.

### Extending it: code, not config-as-code

**The loss is fixed; the *signal* is open — and you open it in code.** There are exactly three loss components (`rl` / `ce` / `ref_kl`), each globally normalized, and the trainer is algorithm-blind — it runs those three and nothing else. `action_loss_type` is **not** a config field and has no `"custom"` value: it is a `ClassVar` *declaration* on each algorithm class (`Literal["rl", "ce", "ref_kl"]`) routing the action tokens into one of the three. You cannot configure a fourth loss component into existence.

A new credit-assignment scheme is a **new named algorithm in the repo**, not a config that points at an import path. Subclass `Algorithm`, assign credit in the scoring hook whose timing fits (`score_rollout` / `score_group`) via `rollout.assign_advantages(...)`, add a typed `*AlgoConfig` to the discriminated union, and register the class in `ALGORITHM_CLASSES`. There is **no `custom` advantage type and no advantage `import_path`** — the config surface is a closed menu of vetted algorithms with typed, validated fields; extension is forking, and the authoring path is deliberately small (one `score_group` that builds rewards and assigns; shared math like `efficiency_shaping` lives in `algo/advantage.py`).

The **one** remaining import-path escape hatch is echo's optional token `filter` — and it is deliberately the *weakest* kind: it does not define an algorithm or touch the loss, it is a per-rollout boolean predicate (`filter_fn(rollout, **kwargs) -> list[list[bool]]`, one keep-mask per trainable branch) that can only **narrow** the role-selected echo tokens, never widen them or assign credit. It parameterizes a vetted algorithm's data selection (drop tool-output warning lines, drop low-probability tokens), the same category as an env's reward function — env-specific data, not algorithm structure. The flip side is the hosted story: because every shipped algorithm is a named class plus per-token weight streams *on the wire* (data, not code), a hosted product exposes the vetted algorithm names and their knobs, disables that lone `filter` import-path, and still gets the full combinatorial reach over which streams get written **without** arbitrary code execution.

### The algorithms

| `algo.type` | Sampling | Loss |
|---|---|---|
| `grpo` (default) | policy | `rl` |
| `max_rl` ([arXiv:2602.02710](https://arxiv.org/abs/2602.02710)) | policy | `rl` |
| `opd` | policy | `ref_kl` (needs a frozen `teacher`) |
| `sft` | frozen model (`sampling.source`) | `ce` |
| `opsd` (SDFT, [arXiv:2601.19897](https://arxiv.org/abs/2601.19897)) | policy | `ref_kl` vs the live policy (self-distillation; no model to declare) |
| `echo` (ECHO) | policy | `rl` on actions + per-role α·`ce` on env-provided tokens |

There is **no preset layer**: the type IS the algorithm, its class defaults ARE the vetted setting, and every key beyond `type` is visibly the user's own assembly (an earlier iteration had named atomic presets; with type-plus-defaults equal to the preset for every algorithm, the layer had nothing left to do and was deleted). Per-env algorithms compose in one run. Because a reference can be the literal `"policy"`, `opsd` runs the SDFT paper's setting with zero extra deployments — that's its default.

### Model references — no registry; roles are algorithm-local

prime-rl assumes it is **never responsible for hosting any model other than the trainable policy**. Everything else is an external OpenAI-compatible endpoint, declared inline where it's used:

- `ModelReference = "policy" | FrozenModelConfig`, where `FrozenModelConfig` is just the existing `ClientConfig` plus the served model's `name` — no new declaration scheme. `base_url` (or an elastic deployment) is required: a frozen reference with no endpoint fails at parse time.
- **Each algorithm names its reference on the field where the model is actually used — there is no shared `teacher` slot and no `model_role` indirection.** `opd` declares a required frozen `teacher` (typed `FrozenModelConfig`); `sft` samples from its `sampling.source` (a frozen model); `opsd` self-distills against the live policy and declares no model at all. The base config knows nothing about teachers. Validation errors name the field directly ("algorithm 'opd' ... teacher: Field required").
- **Liveness is a property of the reference, not a role**: policy-sourced rollouts get version-salted prefix caches, carry sampling logprobs, and age off-policy; frozen-sourced rollouts get a stable prefix cache, skip the logprobs knob, and are never off-policy-cancelled. The pipeline branches on liveness alone.
- Config-time validation: `opd`'s teacher is typed `FrozenModelConfig`, so `"policy"` isn't even representable (the KL would be ≡ 0); frozen sampling can't feed an `rl`/`ref_kl`-type algorithm (no policy sampling logprobs for the importance ratio); `sft` without a frozen source is rejected (CE on the policy's own tokens is not a distillation target); and group-relative advantage with `group_size=1` warns loudly.

## Transport (orchestrator → trainer wire)

Everything an algorithm decides reaches the trainer as **per-token streams** on `TrainingSample` (`src/prime_rl/transport/types.py`) — there is no scalar advantage and no algorithm tag on the wire. The trainer reads the streams, not the algorithm. Five optional per-token arrays carry the whole training signal, each the full `prompt + completion` length:

| field | what it carries | `None` means |
|---|---|---|
| `advantages` | the rl credit stream (one number per token) | no rl credit assigned — `opd`/`opsd`; legal only for samples with no live rl tokens, else the trainer raises |
| `rl_weights` | scales the `rl` (DPPO + KL) component per token | rl weight `1.0` on every trainable token |
| `ce_weights` | scales the `ce` (masked NLL) component per token | no `ce` component |
| `ref_kl_weights` | scales the `ref_kl` (reverse-KL-to-reference) component per token | no `ref_kl` component |
| `ref_logprobs` | the reference model's prefill logprobs, against which the trainer takes the KL | no reference |

Three rules make the wire legible:

1. **Per-token from end to end.** Uniform group credit (GRPO's reward-minus-mean) is *broadcast* over the completion tokens at assignment time — the scalar never reaches the wire. A weight of `0.0` removes a token from its component's mask *and* its global denominator; components may overlap on a token (their gradients sum).
2. **`None` = absent, and absent is the hot path.** Plain GRPO ships exactly **one** extra stream (`advantages`, same order as `completion_logprobs`) and nothing else — the wire is as small as before this PR. `ref_kl` ships *logprobs*, not precomputed advantages, because the trainer evaluates the KL against the live policy each microbatch (a scalar couldn't capture that).
3. **Streams materialize on demand at packing.** The packer (`trainer/batch.py`) concatenates samples into a `MicroBatch`; when one packed sample carries a stream a neighbor lacks, the neighbor's span is backfilled with the component's identity fill (`STREAM_FILL = {rl_weights: 1.0, ce_weights: 0.0, ref_kl_weights: 0.0}`), so every per-token array stays aligned to `input_ids`. This is exactly what lets samples from **different algorithms** (e.g. a GRPO env packed next to an OPD env) share one micro-batch — membership is per token, not per bin. A post-pack assert pins every stream's length to `len(input_ids)`.

The trainer then runs each component over its stream (rl mask `= loss_mask & rl_weights != 0`; ce/ref_kl mask `= weights != 0`) and sums the per-component means; the three global denominators `[N_rl, N_ce, N_ref_kl]` come from one batched all-reduce, so every rank issues the same collective and the no-stream path is identical to the old single-`loss_scale` math.

One field on `TrainingSample` is deliberately *not* on the wire: `obs_spans`, interleaving's provenance record for env-provided observation tokens (`[completion_start, step_idx, step_prompt_start, length]` per later-turn extension span). Algorithms that train on observations (`echo`) consume it at group time and write the `ce_weights` stream directly; it is **cleared before transport**.

## How

- **Configs** (`prime_rl.configs.algorithm`): one algorithm discriminated union absorbs the former token scorers (`logprobs`/`demo_logprobs` → `opd`/`opsd` with a `ModelReference`) *and* loss routing (`EchoAlgoConfig(GRPOAlgoConfig)` carries the role table + filter); the action-token loss component is an `action_loss_type` class property of the algorithm, never configured; each variant declares its own reference field where the model is used (`opd.teacher` as a required `FrozenModelConfig`, `sft.sampling.source`; opsd none) — no `algo.model` shorthand and no `model_role` indirection.
- **No preset machinery at all**: env algorithm inheritance is a one-loop after-validator (envs without their own `algo` inherit the top-level one). Every `AlgoConfig` is built exactly once with everything in place: no preset tables, no name field, no merge machinery, no `__pydantic_fields_set__` surgery.
- **Runtime** (the `orchestrator/algo/` package + `orchestrator/sampler.py`):
  - `Algorithm` — base class with the two scoring hooks (`score_rollout` / `score_group`, detailed under [The API](#the-api--two-seams-to-override) above) plus `setup`/`connect` for frozen-model pools. Two non-virtual methods drive them and stamp the wire fields: `finalize_rollout` runs `score_rollout` as each rollout is tokenized (rollout-local credit *and* reference prefill — OPD/OPSD attach `ref_logprobs` here); `finalize_group` runs `score_group` then stamps each sample (the advantage stream — prompt-padded and sliced across samples — plus per-token loss routing into the declared component). `setup()` connects an `InferencePool` to the algorithm's inline frozen reference (`connect_frozen_pool` — client-side only: prime-rl connects and waits, never launches) and tracks `connected_pools` for shutdown. The live policy pool, by contrast, is handed to every algorithm as `self.policy_pool` (host-owned, never closed by the algorithm — `opsd` self-distills against it, others may judge against it or ignore it); `connect()` only ever opens *frozen* endpoints, so `connected_pools` is exactly the set the algorithm owns and must close. The named classes own their hook bodies outright; `build_algorithm` dispatches on `algo.type`.
  - `Sampler` — one per env, owns the rollout source: the generating pool (policy, or a frozen pool it connects in its own `setup()`), `samples_from_live_policy`, and `sampling_args()` (strips the logprobs knob for frozen endpoints). The dispatcher reads `env.sampler` for pool/liveness; the algorithm never sees sampling.
  - The pipeline has zero algorithm conditionals: the dispatcher derives cache-salt/aging from the sampler's liveness; the sink calls `finalize_rollout` per rollout (reference scoring runs inline here) and `finalize_group` per group; orchestrator setup is one `asyncio.gather` over both objects' `setup()`.
- **Wire** (`transport/types.py`): the per-token streams laid out under [Transport](#transport-orchestrator--trainer-wire) above. `TrainingSample`/`MicroBatch` carry the optional `rl_weights` / `ce_weights` / `ref_kl_weights` streams, the `advantages` stream, and `ref_logprobs`; absent streams keep the plain-GRPO wire minimal, and the packer backfills streams across heterogeneous bins (`STREAM_FILL`) so samples of different algorithms pack freely. The historical collapse of the scalar `advantage` + `token_advantages` into one per-token `advantages` stream, and `teacher_logprobs` → `ref_logprobs`, are in [Breaking changes](#breaking-changes-intentionally-no-deprecation-aliases) below.
- **Trainer**: `compute_loss` runs each component over its weight stream (rl mask = `loss_mask & rl_weights != 0`; ce/ref_kl mask = `weights != 0`) and sums the per-component means; the three global denominators come from one batched all-reduce (`[N_rl, N_ce, N_ref_kl]`), so every rank issues the same collective. The no-stream hot path is byte-identical to the old single-`loss_scale` math with zero extra device syncs. Mixed-algorithm packing fix: bins mixing ref-bearing (e.g. OPD) and ref-less (e.g. GRPO) samples now keep `ref_logprobs` position-aligned with `input_ids` (0.0 placeholders, backfill-or-pad like the sibling per-token arrays), with a regression test and a post-pack invariant assert that every per-token array on every micro batch matches `len(input_ids)`.

## Breaking changes (intentionally, no deprecation aliases)

- `orchestrator.training_mode` is **deleted** (was `rl`/`opd`/`sft`) → `[orchestrator.algo] type = ...`.
- **`advantage.type` → `algo.type`**: the `algo` bundle is a discriminated union keyed on `type` directly — `[orchestrator.algo] type = "grpo"` (was `[orchestrator.algo.advantage]`). The unused `advantage = {}` env/top-level shorthand is **deleted**; `opd` declares a native `algo.teacher` field.
- **The advantage `type` names the algorithm**: `group_norm` → `grpo`, `ref_kl` → `opd`, `demo_ref_kl` → `opsd`, `supervised` → `sft` (config classes renamed to match). The preset `name` field is **deleted** — type-plus-defaults is the algorithm.
- Echo selection is a **role table**: `observations` / `observation_weight` are replaced by `roles.<role>.alpha` (tool bodies @ 0.1 default; setting any role replaces the table) plus an optional `filter` hook. Echo always requires `orchestrator.renderer` (the no-attribution `"all"` mode is gone).
- `sft` requires a frozen sampling source — CE on the policy's own tokens is rejected at validation (previously silently assemblable).
- `[orchestrator.teacher]` is **deleted** → each algorithm declares its frozen model inline where used: `[orchestrator.algo.teacher]` for `opd`, `[orchestrator.algo.sampling.source]` for `sft` (`opsd` needs none), with `name` + `base_url`.
- The policy sub-config is `[orchestrator.model]`, **flat**: `name` / `lora` / `vlm` / `trust_remote_code` directly on it, the deployment under `[orchestrator.model.client]` — no more `model.model` stutter, in configs or resolved dumps (`HostedModelConfig` is now `ModelConfig` + a `client` field, mirroring `FrozenModelConfig`). The `[orchestrator.policy]` / `[orchestrator.student]` aliases, the `[orchestrator.client]` shorthand, and the flat-key re-nesting shim are **deleted**.
- `algorithm.token_scorer` is **deleted** — scorers are advantage strategies now (`logprobs` → `ref_kl`, `demo_logprobs` → `demo_ref_kl`).
- `algo.loss` does not exist — the action-token loss component derives from the advantage strategy, and observation-token routing is the `echo` advantage type (`observation_weight`).
- Advantage types: `default` → `group_norm`; `none` is **deleted** (the ref_kl-family strategies carry its no-scalar semantics; nothing else used it).
- Wire/trainer: `TrainingSample.advantage` (scalar) + `token_advantages` → one per-token `advantages` stream; `AdvantageOutputs` is deleted — algorithms assign credit in their scoring hooks via `rollout.assign_advantages(scalar | list[float])` (scalar broadcast for uniform credit), and advantage-based filters/metrics derive from the streams (zero-advantage filter = all-zero stream; logged distributions = per-rollout means). `TrainingSample.teacher_logprobs` → `ref_logprobs`; the loss-type *partition* (`loss_type` / `token_loss_types` / `token_loss_weights` / `loss_type_ids`, `LOSS_CORE_*`) is **replaced by the component weight streams** `rl_weights` / `ce_weights` / `ref_kl_weights` (the partition was the degenerate disjoint case); the `time/teacher_logprobs` metric is removed (reference scoring now runs inline at rollout arrival, not as a distinct batch phase).
- `configs/debug/training_modes/` → `configs/debug/algorithms/` (incl. new `self_distill.toml` running SDFT against the live policy); CI integration configs migrated.

Configs are `extra="forbid"`, so stale configs fail loudly at parse time with the exact unknown key.

Merged with `main` including #2720 (teacherless SFT). Its spelling is **deliberately not carried forward**: `sft` means distillation from a frozen teacher, and pointing it at `"policy"` is rejected at validation — CE on the policy's own rollouts is a different algorithm, and if wanted it gets its own advantage type/class rather than a degenerate spelling of `sft`.

Also merged `main` including #2641 (multiplexed trainer token export): the export machinery composes with the component weight streams — each exported record carries the per-token rl/ce/ref_kl arrays, and micro-batches carry `run_id`/`run_step` alongside the streams (`training_mode` stays deleted).

> **Note:** this PR absorbed #2764 (named algorithm classes → two-component model → `Sampler` split), #2778 (MaxRL), and #2782 (echo per-role selection + filter hook); those branches were merged into this one and GitHub marked their PRs merged.

## Deferred (next PRs)

The two-component model is built so each of these lands as its own PR without re-touching the wire:

- **Sampling strategies behind the `Sampler`** — today the Sampler answers one question (which pool, and is it live); next it absorbs within-env example iteration and group production, plus a sink→sampler `observe()` feedback edge at group finalization. In increasing order of machinery that unlocks: difficulty-pool curricula (per-example selection state fed back from group rewards), static dataset sources (supervised training from demonstrations instead of a frozen endpoint), and replay buffers / offline experience (a store between stamping and batch assembly — advantages and weight streams are already stamped-then-frozen at group finalization, which is the prerequisite).
- **Trainer collapse** — the three fixed components reduce to two cores (policy-gradient × supervised) composed with per-token factor streams. Dissolves the rl-slot-only custom-loss wart and lets the stability guards (importance correction, trust region) compose *around* custom losses instead of being replaced by them.
- **Remaining ECHO surface** — per-role α, arbitrary roles, and content/sampling-logprob filters shipped in this PR; still open: `tool_names` filtering (non-breaking optional field on the tool role; `message_tool_names` already rides in the attribution) and θ-dependent low-probability filters, which need a trainer-side per-component knob since the denominator collective happens pre-forward.
- Smaller: config knobs for component sums (per-env α/β/γ folded into the streams orchestrator-side — the wire and trainer already support overlap); opt-in weight-sum normalization (`Σw` denominators — a globally-folded λ cancels under it, so it pairs with a trainer-side coefficient); per-slot loss-fn configs for ce/ref_kl (the ref_kl trust region — one-sided today — becomes a deliberate choice there); barrier-blind sink (the group wait becomes a dependency declared by the algorithm; `finalize_group` → `finalize`); EMA/lagged reference endpoints; exposing frozen endpoints to envs as judge clients.

`87c38c704` renames the config surface from `advantage.type` to **`algo.type`** per review: the `algo` bundle flattens into a discriminated union (each `*AlgoConfig` variant carries `type` + its params + the shared `sampling`/`teacher` on `BaseAlgoConfig`), the unused `advantage = {}` shorthand + its fold validator are dropped, and each algorithm declares its reference inline (`opd.teacher`, `sft.sampling.source`). Runtime `Algorithm` takes the algo config; `build_algorithm` dispatches on `algo.type`. 12 debug/CI TOMLs, `docs/algorithms.md` + `training.md`, and the unit tests migrated. Orchestrator+config unit tests pass (incl. the full TOML-load sweep) + a `build_algorithm` dispatch smoke over every algorithm type; ruff clean.

`cf9a21a3b` / `f13119938` collapse the override surface from three scoring hooks to two: `score_batch` is removed and OPD/OPSD do their reference prefill in `score_rollout` (at arrival); `finalize_rollout` / `finalize_group` become non-virtual `Algorithm` methods (the cross-env `finalize_batch` is gone), and `Algorithm` is now constructed with just `(config, policy_pool)` — `opsd` builds its own hint renderer in `setup()` with the live policy's tokenizer (the `model` knob is gone; self-distillation has no separate model). The per-rollout `max_concurrent` scoring semaphore is dropped — the serial main loop already bounds in-flight reference calls — and with it the now-orphaned `time/scoring` metric. Docs + tests updated; orchestrator/config unit tests green.

---

> [!NOTE]
> **High Risk**
> Large breaking config and orchestrator/trainer contract change (per-token streams, deleted keys); mistakes in heterogeneous batches or frozen-model wiring could silently skew gradients until caught by tests.
> 
> **Overview**
> **Replaces the `rl` / `opd` / `sft` training-mode switch** with `[orchestrator.algo] type = …` (`grpo`, `max_rl`, `opd`, `opsd`, `sft`, `echo`). Frozen teachers move inline (`algo.teacher` for OPD, `algo.sampling.source` for SFT); the trainable model is a flat `[orchestrator.model]` with `[orchestrator.model.client]` (no `student` / `orchestrator.teacher`).
> 
> **Orchestrator runtime** gains per-env `Sampler` (which pool generates rollouts) and `Algorithm` subclasses with `score_rollout` / `score_group` / `score_batch`, plus wire stamping of `advantages`, `rl_weights` / `ce_weights` / `ref_kl_weights`, and `ref_logprobs`. The dispatcher keys off policy vs frozen sampling for cache salting and off-policy cancellation; custom `orchestrator.advantage` and `advantage.py` are removed in favor of algorithm classes.
> 
> **Trainer/docs/configs**: loss is documented and wired as three globally normalized components; CI and `configs/debug/training_modes/` become `configs/debug/algorithms/` (new echo, max_rl, self_distill, mixed GRPO+OPD). Metrics rename `time/teacher_logprobs` → `time/scoring`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3268685ed4618da44df82c8ce146dc170f798c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
mikasenghaas, line 1: can put into orch configs? i think rn we have one config per entrypoint

examples/glm5_pd_disag/rl.toml
```diff
@@ -66,7 +66,7 @@ group_size = 16
 oversampling_factor = 3
 max_off_policy_steps = 16
 
-[orchestrator.student.model]
+[orchestrator.policy.model]
```
mikasenghaas: i think we should fall back to plain `model` (and have alias for `student.` and `policy.`

configs/debug/algorithms/echo.toml
```diff
@@ -0,0 +1,48 @@
...
+
+[wandb]
+project = "algorithms-debug"
+name = "debug-echo"
+
+[orchestrator]
+batch_size = 32
+group_size = 4
+
+[orchestrator.algorithm]
```
mikasenghaas: would like `algo` as well lol but personal preference, could maybe alias

docs/algorithms.md
```diff
@@ -20,6 +27,85 @@ This page covers the math and the configurable algorithmic components: how off-p
...
+2. **Advantage** (`algorithm.advantage`) — the per-token training signal, one concept at different granularities and evaluation sites. Group-relative strategies compute scalars on the orchestrator and ship numbers; reference-KL strategies query a reference model at batch-ship time (bounded concurrency) and ship its prefill logprobs for the trainer to evaluate against the live policy. The strategy determines which loss core consumes the action tokens (`rl` / `ce` / `ref_kl`).
+3. **Loss routing** (`algorithm.loss`) — what happens to env-provided observation tokens in multi-turn rollouts (tool output, terminal responses): `observation = "none"` masks them out (the default), `"ce"` trains on them with weight `observation_weight`.
+
+The trainer is algorithm-blind: routing ships per token on the wire (`loss_core` / `token_loss_cores` / `token_loss_weights` on each training sample) and the trainer just executes loss cores. Adding an algorithm never touches the dispatcher, packer, or trainer hot path.
```
mikasenghaas: i think `loss_core` reads weird, whats that?

docs/algorithms.md
```diff
@@ -20,6 +27,85 @@ This page covers the math and the configurable algorithmic components: how off-p
...
+
+### The Model Registry
+
+There are no model *roles* in `prime-rl` — no teacher slot, no judge slot. There is the live policy (registered under the reserved name `"policy"`) and a registry of named frozen hosted models under `[orchestrator.models]`. Algorithm components hold *references* into that registry, so the same deployment can serve different algorithms in the same run:
+
+```toml
+[orchestrator.models.qwen-32b]
+model.name = "Qwen/Qwen3-32B"
+client.base_url = ["http://localhost:8001/v1"]
```
mikasenghaas: i wanted to do the same change but i think @samsja is right that its cleaner to have the invariaant that prime-rl orch only holds ref to trainable policy. any other model is algo-specific and as such should live on the algo config. eg.

```

[orchestrator.algorithm]
name = "opd"

[orchestrator.algorithm.teacher]
model = "qwen-32b"
```

docs/algorithms.md
```diff
@@ -69,17 +163,11 @@ The knobs (under `[trainer.loss]` with `type = "default"`):
...
-The trainer dispatches automatically based on the batch's training mode (set by the orchestrator via `orchestrator.training_mode`):
-
-- `rl` mode → DPPO + KL with the advantage signal.
-- `opd` mode → KL distillation against the teacher's per-token logprobs. The teacher must be a vLLM server (it's the only one that exposes `prompt_logprobs`).
-- `sft` mode → standard token-level NLL on teacher-generated rollouts.
-
-Set `[trainer.loss] type = "default"` and configure via the knobs above. SFT and OPD modes ignore the policy-gradient–specific fields.
+Set `[trainer.loss] type = "default"` and configure via the knobs above. The `ce` and `ref_kl` cores are fixed and unaffected by `[trainer.loss]`.
 
 ### Custom Loss
```
mikasenghaas, line 214: isnt this obsolete?

configs/debug/algorithms/echo.toml
hallerite: agreed, has more aura

docs/algorithms.md
hallerite: we could make it obsolete by making the RL loss more configurable in the algo config I guess

docs/algorithms.md
hallerite: the objective, renamed it

configs/ci/integration/reverse_text_rl_opd/start.toml
```diff
@@ -15,7 +15,7 @@ project = "reverse-text-ci"
 name = "ci-rl-opd"
 
 [orchestrator]
-training_mode = "opd"
+algo = { name = "opd", model = "reverse-text-rl" }
```
samsja: lets nmot use dicti n toml, lets just do sun section like toml god intented

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
```diff
@@ -0,0 +1,398 @@
...
+from pydantic import Field, model_validator
+
+from prime_rl.utils.config import BaseConfig
+
+AlgorithmName: TypeAlias = Literal["grpo", "opd", "sft_distill", "self_distill", "echo"]
+
+# Reserved model-registry key for the live policy (weight-updated, prefix
+# caches salted per version). Every other key names a frozen hosted model
+# from ``[orchestrator.models]``.
+POLICY_MODEL: str = "policy"
```
samsja: why do we need this ?

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
```diff
@@ -0,0 +1,398 @@
...
+
+# Preset component tables. Fields the user sets explicitly always win. Model
+# references (``sampling.source`` of sft_distill, ``advantage.model`` of opd /
+# self_distill) have no sensible default — presets leave them ``None`` and
+# validation demands ``algo.model`` (or the component field).
+_PRESETS: dict[str, dict[str, Any]] = {
+    "grpo": dict(
+        sampling=lambda: SamplingConfig(),
+        advantage=lambda: GroupNormAdvantageConfig(),
+        loss=lambda: LossRoutingConfig(),
```
samsja: i dont like lambda

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
```diff
@@ -0,0 +1,398 @@
...
+                setattr(self, component, preset_value)
+                self.__pydantic_fields_set__.discard(component)
+            else:
+                for field_name in type(current).model_fields:
+                    if field_name not in current.model_fields_set:
+                        setattr(current, field_name, getattr(preset_value, field_name))
+        if self.advantage is None:
+            self.advantage = preset["advantage"]()
+            self.__pydantic_fields_set__.discard("advantage")
+        return self
```
samsja: this seems brittle also dont like the pydantic set thingy

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
```diff
@@ -0,0 +1,398 @@
+"""Algorithm abstraction: sampling, advantage, and loss routing.
+
+An algorithm is a preset of three pieces:
+
+1. **Sampling** — which model generates train rollouts. ``source`` is a model
```
samsja: Not the bigget fan of this section, cofnig and preset seems off like it is rn, maybe need more explanation why we need the complexity

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -637,19 +624,17 @@ def _preserve_mito_renderer(self, handler: SerializerFunctionWrapHandler) -> dic
...
     @classmethod
-    def fold_student_shortcuts(cls, data: Any) -> Any:
-        """Accept top-level ``[orchestrator.model]`` / ``[orchestrator.client]``
-        as shorthand for the student sub-config. Useful for ergonomic rl configs
-        where ``[orchestrator.student.*]`` is overkill, and required for
-        pre-refactor configs that used the flat layout to keep parsing:
-
-        - [orchestrator.client.*]     -> [orchestrator.student.client.*]
-        - [orchestrator.model.<k>]    -> [orchestrator.student.model.<k>]
+    def fold_policy_shortcuts(cls, data: Any) -> Any:
```
samsja: ehh i feel this function is going to cause us issue in the future, but lets keep as is for now

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -726,15 +714,15 @@ def _env_to_train(cls, data: Any) -> Any:
     @model_validator(mode="after")
     def auto_setup_tokenizer(self):
         if self.tokenizer.name is None:
-            self.tokenizer.name = self.student.model.name
+            self.tokenizer.name = self.model.model.name
```
samsja: lol i remenber we change this exact line 2 weeks ago

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -758,22 +746,89 @@ def validate_unique_filter_types(self):
...
     @model_validator(mode="after")
-    def _force_no_renderer_for_sft(self):
-        """Teacher-backed SFT rolls out via the teacher's plain chat-completions
-        endpoint; the renderer client doesn't apply. When no teacher is
-        configured, SFT uses the student rollout path and keeps the renderer."""
-        if self.training_mode == "sft" and self.teacher is not None:
+    def resolve_algorithm(self):
+        """Fold the ``advantage`` shorthands and propagate the resolved
+        algorithm into every train env. Declared before any validator that
+        reads ``algo``."""
```
samsja: hmmm i dont like this function

src/prime_rl/transport/types.py
```diff
@@ -1,8 +1,10 @@
-from typing import Literal
-
 import msgspec
 
-TrainingMode = Literal["rl", "opd", "sft"]
+# Per-token loss types. The orchestrator routes each token to a loss type (the
+# algorithm's loss routing); the trainer just executes them.
+LOSS_TYPE_RL = 0  # configured RL loss (importance-weighted PG + KL)
+LOSS_TYPE_CE = 1  # masked NLL (SFT / observation prediction)
+LOSS_TYPE_REF_KL = 2  # per-token reverse KL to a reference model as the PG signal (OPD / SDFT)
```
samsja: nah no lest use problem enum here instead ? or like why not just use strig / literal

configs/ci/integration/reverse_text_rl_opd/start.toml
mikasenghaas: i acc think sometimes (like in this case) this syntax is more concise and easier to read

mikasenghaas: but whatever no strong feelings

configs/ci/integration/reverse_text_rl_opd/start.toml
```diff
@@ -38,10 +38,10 @@ max_completion_tokens = 128
 [[orchestrator.eval.env]]
 id = "reverse-text"
 
-[orchestrator.teacher.model]
+[orchestrator.models.reverse-text-rl.model]
```
mikasenghaas: ah yae this is defo weird (`models.reverse-text-rl`). should just be `orchestrator.model` (and mean trainable policy)

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
hallerite: I feel like if teacher is not priviliged as a concept in the main loop, then neither should the model we train be called student

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
hallerite: then let's remove. we don't need aliases imo

cc: @mikasenghaas

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
```diff
@@ -0,0 +1,428 @@
...
+
+    @model_validator(mode="after")
+    def validate_component_compatibility(self):
+        source_role = getattr(self.advantage, "source_role", None)
+        if source_role is not None and self.sampling.source == "policy":
+            raise ValueError(
+                f"advantage '{self.advantage.type}' needs a {source_role} to sample rollouts from — "
+                f"CE on the policy's own tokens is not a distillation target. Set '{source_role}' on "
+                "the algorithm (an inline hosted model: name + base_url), or sampling.source explicitly."
+            )
```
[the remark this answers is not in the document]
hallerite: By design, not a regression: `sft` means distillation from a frozen teacher — CE on the policy's own rollouts is a different algorithm, and the type system should say so rather than accept a degenerate spelling of it. The earlier `sft_distill + model="policy"` expressibility (the #2720 migration note) is deliberately dropped with the preset layer; if on-policy CE is ever wanted, it gets its own advantage type/class (one union member + one dispatch entry). PR body updated to match.

configs/ci/integration/reverse_text_rl_opd/start.toml
```diff
@@ -15,10 +15,16 @@ project = "reverse-text-ci"
...
 
 [orchestrator]
-training_mode = "opd"
 batch_size = 128
 group_size = 16
 
+[orchestrator.algo.advantage]
+type = "opd"
+
+[orchestrator.algo.model]
```
mikasenghaas: should we name this `teacher` or `teacher_model`?

configs/ci/integration/reverse_text_rl_sft/start.toml
```diff
@@ -15,10 +15,16 @@ project = "reverse-text-ci"
...
 [orchestrator]
-training_mode = "sft"
 batch_size = 128
 group_size = 16
 
+[orchestrator.algo.advantage]
+type = "sft"
+
+[orchestrator.algo.model]
+name = "PrimeIntellect/Qwen3-0.6B-Reverse-Text-RL"
```
mikasenghaas, line 26: same here

configs/elastic/rl.toml
```diff
@@ -35,7 +35,7 @@ group_size = 8
 [orchestrator.train.sampling]
 max_completion_tokens = 768
 
-[orchestrator.client.elastic]
+[orchestrator.model.client.elastic]
```
mikasenghaas, line 38: is this breaknig or auto-translated?

src/prime_rl/orchestrator/algo/algorithm.py
```diff
@@ -0,0 +1,385 @@
...
+    action_loss_type: ClassVar[ActionLossType] = "rl"
+    model_role: ClassVar[str | None] = None
+
+    def __init__(self, config: AlgorithmConfig, policy_pool: InferencePool, renderer: Renderer | None):
+        self.config = config
+        self.policy_pool = policy_pool
+        self.renderer = renderer
+        self.reference_pool: InferencePool | None = None  # resolved in setup() when the algorithm declares a model
+        self.connected_pools: list[InferencePool] = []  # client pools connected in setup(); closed at shutdown
+        # Echo selection: message role -> per-token ce weight for env-provided
```
mikasenghaas: this feels like it should not be on base

src/prime_rl/orchestrator/algo/algorithm.py
```diff
@@ -0,0 +1,385 @@
...
+    itself, conditioned on an expert demonstration — no extra deployment.
+
+    The scoring prefix is rebuilt from the rollout's first-turn prompt
+    messages with the demonstration woven into the last user message; the
+    returned completion logprobs are aligned back onto the sample (the
+    sample's prompt positions are 0.0 and stay outside the loss mask). No
+    scalar advantage is assigned."""
+
+    action_loss_type = "ref_kl"
+    model_role = "teacher"
```
mikasenghaas: what are these?

src/prime_rl/orchestrator/algo/algorithm.py
```diff
@@ -0,0 +1,385 @@
...
+
+    def __init__(self, config: AlgorithmConfig, policy_pool: InferencePool, renderer: Renderer | None):
+        super().__init__(config, policy_pool, renderer)
+        assert isinstance(config.advantage, OPSDAdvantageConfig)
+        assert renderer is not None, "opsd requires the renderer (validated at config time)"
+        self.demo_key = config.advantage.demo_key
```
mikasenghaas: is this the privileged informatioN? if so, dont like it

src/prime_rl/orchestrator/algo/algorithm.py
mikasenghaas, line 1: can we have one file per concrete algorithm, so its a bit easire to digest

src/prime_rl/orchestrator/algo/routing.py
```diff
@@ -0,0 +1,87 @@
...
+    if train_obs:
+        assert obs_weights is not None
+        ce_weights = sample.ce_weights if sample.ce_weights is not None else [0.0] * seq_len
+        for i, weight in enumerate(obs_weights):
+            if weight:
+                ce_weights[prompt_len + i] = weight
+        sample.ce_weights = ce_weights
+
+
+def spread_token_advantages(rollout: TrainRollout) -> None:
```
mikasenghaas: wondering if we should try land this after #2742, the prompt/ completion concept is gone then

src/prime_rl/orchestrator/dispatcher.py
```diff
@@ -273,6 +268,10 @@ async def on_new_version(self, step: int) -> None:
         for meta in self.inflight.values():
             if meta.kind != "train":
                 continue
+            # Frozen-sourced rollouts never go stale — their sampler doesn't
+            # change with policy updates.
+            if not self.train_envs.get(meta.env_name).sampler.samples_from_live_policy:
```
mikasenghaas, line 277: hmm

src/prime_rl/orchestrator/envs.py
```diff
@@ -162,14 +163,11 @@ def shutdown(self) -> None:
 class TrainEnv(Env):
     config: TrainEnvConfig
 
-    def __init__(self, config: TrainEnvConfig):
+    def __init__(self, config: TrainEnvConfig, sampler: Sampler, algorithm: Algorithm):
```
mikasenghaas, line 206: seems weird to have train env contain sampler + algorithm? maybe not.. have to think

hallerite: I think it actually makes sense. It should contain everything that is needed to return training samples and just be the env + all the stuff you need at train time, hence train env

configs/ci/integration/reverse_text_rl_opd/start.toml
hallerite: `teacher` parses already — the distillation algorithms declare their reference's role as `teacher`, so it's an accepted alias for the `model` shorthand. Flipped every teacher-meaning config to that spelling in 639b7b522.

configs/ci/integration/reverse_text_rl_sft/start.toml
hallerite: Flipped to `[orchestrator.algo.teacher]` in 639b7b522 (all the debug configs too).

configs/elastic/rl.toml
hallerite: Breaking, deliberately — this PR drops the old spellings without deprecation aliases (see Breaking changes in the description). Every shipped config is flipped in the PR and the config test suite parses them all.

src/prime_rl/orchestrator/algo/algorithm.py
hallerite: Agreed — fixed in 9e53a723c: the echo state is gone from the base. There's a third hook now, `observation_weights(output)` (default None = observations masked), and echo owns its whole selection (role table, attribution lookup, user filter + validation) in algo/echo.py. The base carries only the algorithm-neutral contract.

src/prime_rl/orchestrator/algo/algorithm.py
hallerite: Class-level declarations of what the algorithm needs: `action_loss_type` is which loss component its action tokens feed (what `stamp_loss_routing` writes on the wire), `model_role` is what this algorithm calls its reference model — it's why `[orchestrator.algo.teacher]` knows where to fold and why validation errors say "needs a teacher". Both documented on the base class; since 4c7d4f49b they live in algo/base.py with the per-algorithm modules overriding them.

src/prime_rl/orchestrator/algo/algorithm.py
hallerite: Done in 4c7d4f49b — one module per algorithm (algo/grpo.py, algo/opd.py, …) plus algo/base.py for the base class and pipeline hooks; dispatch lives in the package __init__.

src/prime_rl/orchestrator/dispatcher.py
hallerite: Staleness measures drift between the sampling policy and the trainer policy — a frozen sampler doesn't drift, so off-policy aging doesn't apply to its rollouts (and salting its prefix cache per policy version would just evict a perfectly valid cache every weight update). Frozen-sourced rollouts (sft) train as fixed-teacher data, like offline SFT; the importance-ratio machinery is off for them anyway (ce component, no sampling logprobs).

src/prime_rl/orchestrator/algo/algorithm.py
hallerite: why?

samsja (commented):
>>> prime-rl assumes it is never responsible for hosting any model other than the trainable policy. Everything else is an external OpenAI-compatible endpoint, declared inline where it's used:

should we have utility or docs tho to still help doing this ?

src/prime_rl/transport/types.py
```diff
@@ -52,9 +48,28 @@ class TrainingSample(msgspec.Struct, array_like=True, gc=False, omit_defaults=Tr
...
+    # (mask and denominator). ``None`` means absent: no ce/ref_kl component,
+    # and an rl weight of 1.0 on every trainable token — so the plain GRPO
+    # wire stays as small as before.
+    rl_weights: list[float] | None = None
+    ce_weights: list[float] | None = None
+    ref_kl_weights: list[float] | None = None
+
+    # Per-token advantages (full sequence length). ``None`` broadcasts the
+    # rollout-level ``advantage`` scalar over the sequence.
+    token_advantages: list[float] | None = None
```
mikasenghaas: just advantages?

src/prime_rl/trainer/rl/train.py
```diff
@@ -149,9 +149,9 @@ def train(config: TrainerConfig):
     logger.info(f"Initializing tokenizer ({config.tokenizer})")
     tokenizer = setup_tokenizer(config.tokenizer)
 
-    # Set up the loss function
+    # Set up the loss function for the RL loss type (ce / ref_kl are fixed)
     logger.info(f"Setting up loss function ({config.loss})")
-    loss_fns = setup_loss_fns(config.loss)
+    rl_loss_fn = setup_rl_loss_fn(config.loss)
```
mikasenghaas, line 154: just loss_fn?

mikasenghaas (commented):
i still feel the algo api is a bit complicated to understand

src/prime_rl/trainer/rl/train.py
hallerite: no because its just the custom loss fn for rl

src/prime_rl/transport/types.py
hallerite: yeah lets unify and always do token-level

Ziems:
It looks like I would have no trouble implementing Pedagogy based on the `opsd` implementation, so it looks good to me! I don't have much experience/taste for the direction of the codebase yet though so I don't think my opinions should have much weight. Regardless, it looks great for my current needs!

configs/ci/integration/reverse_text_rl_opd/start.toml
```diff
@@ -1,10 +1,10 @@
...
-# The student inference deployment is launched by the rl entrypoint on GPU 0;
-# the teacher inference server is started externally by the test fixture (see
+# Smoke test for the RL-entrypoint opd algorithm (on-policy distillation).
+# The policy inference deployment is launched by the rl entrypoint on GPU 0;
+# the frozen reference server is started externally by the test fixture (see
 # tests/integration/test_reverse_text_rl_opd.py) on the same GPU. Trainer
 # takes GPU 1. Training config mirrors `reverse_text/start.toml`.
 
-max_steps = 5
+max_steps = 20
```
mikasenghaas: hm?

configs/ci/integration/reverse_text_rl_sft/start.toml
```diff
@@ -1,10 +1,10 @@
...
-# fixture (see tests/integration/test_reverse_text_rl_sft.py) on the same GPU.
-# Trainer takes GPU 1. Training config mirrors `reverse_text/start.toml`.
+# Smoke test for the RL-entrypoint sft algorithm (hard distillation).
+# The policy inference deployment is launched by the rl entrypoint on GPU 0;
+# the frozen sampling server is started externally by the test fixture (see
+# tests/integration/test_reverse_text_rl_sft.py) on the same GPU. Trainer
+# takes GPU 1. Training config mirrors `reverse_text/start.toml`.
 
-max_steps = 5
+max_steps = 20
```
mikasenghaas: also here, why change

configs/ci/integration/reverse_text_rl_sft/start.toml
```diff
@@ -15,13 +15,24 @@ project = "reverse-text-ci"
...
+
+# The student's renderer tokenizes the teacher's messages into the student's
+# token space (backfill). Use `default` (wraps the student tokenizer's
+# apply_chat_template) so the CE target is faithful to the student's actual
+# chat template. The stock `qwen3` renderer reimplements the Qwen format and
+# injects an empty `<think></think>` block, which this model's custom template
+# does not emit — corrupting the distillation target.
 [orchestrator.renderer]
-name = "qwen3"
+name = "default"
```
mikasenghaas, line 36: hm im not 100% sure im following here, does `orchestrator.renderer` set the teacher's renderer here?

configs/elastic/rl.toml
mikasenghaas: have to let @eexwhyzee know then

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -473,21 +443,25 @@ class OrchestratorExperimentalConfig(BaseConfig):
...
-
-    teacher: RolloutModelConfig | None = Field(None, validation_alias=AliasChoices("teacher", "teacher_model"))
-    """Teacher rollout participant (model + client). Role depends on ``training_mode``: ``opd`` — teacher computes logprobs; ``sft`` — teacher generates rollouts."""
+    algo: AlgorithmConfig = GRPOAlgorithmConfig()
+    """Training algorithm: sampling plus the per-token training signal (credit
+    assignment and loss routing, fused — its ``type`` names the algorithm).
+    Defaults to ``grpo``. Override per env via ``[[orchestrator.train.env]]``'s
+    ``algo``."""
+
+    model: HostedModelConfig = HostedModelConfig()
```
mikasenghaas: not sure i love `HostedModelConfig`, whats stopping us from just `ModelConfig`?

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
```diff
@@ -0,0 +1,401 @@
...
+    def require_explicit_endpoint(self):
+        if "base_url" not in self.model_fields_set and not self.is_elastic:
+            raise ValueError(
+                "a frozen model reference needs base_url — frozen models are externally "
+                "hosted; prime-rl only ever hosts the trainable policy."
+            )
+        return self
+
+
+ModelReference: TypeAlias = Literal["policy"] | FrozenModelConfig
```
mikasenghaas, line 68: hmm, not sure we need this. why treat these different? imo both the trainable policy and any algo can reference `ModelConfig` which holds name and `ClientConfig`

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
```diff
@@ -0,0 +1,401 @@
...
+"""``"policy"`` (the live policy — weight-updated: prefix caches salted per
+version, sampling logprobs carried, rollouts age off-policy) or an inline
+externally-hosted frozen model."""
+
+ActionLossType: TypeAlias = Literal["rl", "ce", "ref_kl"]
```
mikasenghaas, line 73: just `LossType`?

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
mikasenghaas, line 1: i acc wonder if we should colocate the algo configs with the actual code. would make it more readable imo

mikasenghaas: at least the concrete algo configs, base can stay here

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
```diff
@@ -0,0 +1,401 @@
...
+    kwargs: dict[str, Any] = Field(default_factory=dict)
+    """Kwargs forwarded to the filter."""
+
+
+# ---------------------------------------------------------------------------
+# The algorithms (a discriminated union keyed on ``type``)
+# ---------------------------------------------------------------------------
+
+
+class BaseAlgorithmConfig(BaseConfig):
```
mikasenghaas: `AlgoConfig`?

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
```diff
@@ -0,0 +1,401 @@
...
+    teacher: ModelReference | None = Field(None, exclude=True)
+    """Reference-model shorthand: an inline frozen hosted model (``name`` +
+    ``base_url``). Folds into the slot the algorithm declares for it —
+    ``model`` for the distillation algorithms (opd/opsd), ``sampling.source``
+    for sft. ``grpo`` / ``max_rl`` / ``custom`` take no teacher.
+    Write-only input sugar — folded by validation and excluded from dumps so
+    resolved configs round-trip."""
+
+    @model_validator(mode="after")
+    def fold_teacher(self):
```
mikasenghaas: this seems a bit off

src/prime_rl/entrypoints/rl.py
```diff
@@ -181,19 +182,27 @@ def sigterm_handler(signum, frame):
...
+                "make sure one is running at orchestrator.model.client.base_url "
+                f"({', '.join(config.orchestrator.model.client.base_url)}), otherwise the orchestrator "
                 "will hang waiting for it."
             )
 
-        if config.orchestrator.teacher:
+        frozen_endpoints: list[str] = []
+        for env in config.orchestrator.train.env:
+            algo = env.algo
+            if algo is None:
```
mikasenghaas: when is this possible?

src/prime_rl/orchestrator/algo/advantage.py
```diff
@@ -0,0 +1,126 @@
...
+    from prime_rl.orchestrator.types import Rollout
+
+from prime_rl.configs.algorithm import (
+    LengthPenaltyConfig,
+    TokensLengthPenaltyConfig,
+    TurnsLengthPenaltyConfig,
+)
+from prime_rl.orchestrator.utils import get_tool_response_len
+
+AdvantageFn = Callable[..., list[float | list[float]]]
```
mikasenghaas: imo could be strict here and always make token based

src/prime_rl/orchestrator/algo/base.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Constructed with the algorithm config it interprets plus the two
+    host-owned resources: the policy pool and the policy's renderer (the
+    canonical messages → token ids path; ``None`` under MITO)."""
+
+    action_loss_type: ClassVar[ActionLossType] = "rl"
+
+    def __init__(self, config: AlgorithmConfig, policy_pool: InferencePool, renderer: Renderer | None):
+        self.config = config
+        self.policy_pool = policy_pool
```
mikasenghaas, line 123: wondering if we can avoid passing the policy pool through. would make algo even more self-contained/ disentangled from general orch. for opsd where orch (for rollout) and teacher (for scoring) are the same model, both have ind. clients which point to the same url

src/prime_rl/orchestrator/algo/base.py
```diff
@@ -0,0 +1,191 @@
...
+    async def score_rollout(self, rollout: Rollout) -> None:
+        """Arrival phase, one rollout, before its group is complete: write
+        rollout-local credit (``rollout.assign_advantages``) or observation ce
+        weights (echo). No siblings, no group stats."""
+
+    async def score_group(self, group: list[Rollout]) -> None:
+        """Group phase, the finalized cohort, before filtering: write
+        group-relative credit."""
+
+    async def score_batch(self, batch: list[Rollout]) -> None:
```
mikasenghaas: do we have an example where we need this hook?

src/prime_rl/orchestrator/algo/base.py
```diff
@@ -0,0 +1,191 @@
...
+        """Ship phase, survivors only, after filtering, async: query the
+        algorithm's reference models and attach per-token results (e.g. teacher
+        logprobs)."""
+
+
+async def finalize_rollout(algorithm: Algorithm, rollout: Rollout) -> None:
+    """Arrival phase: rollout-local scoring as each rollout is tokenized."""
+    if rollout.samples:
+        await algorithm.score_rollout(rollout)
```
mikasenghaas: any reason for these to not be methods on the class?

src/prime_rl/orchestrator/algo/grpo.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class GRPOAlgorithm(Algorithm):
+    """Group Relative Policy Optimization: sample a group of rollouts from the
+    policy per example; credit = reward minus the group mean (optionally
+    length-shaped); action tokens feed the ``rl`` loss."""
+
+    def __init__(self, config: AlgorithmConfig, policy_pool: InferencePool, renderer: Renderer | None):
+        super().__init__(config, policy_pool, renderer)
+        assert isinstance(config, GRPOAlgorithmConfig)
```
mikasenghaas: lets use generic typing here

`GRPOAlgorithm(Algorithm[GRPOConfig])` 

then you get tight types

src/prime_rl/orchestrator/algo/max_rl.py
```diff
@@ -0,0 +1,21 @@
...
+class MaxRLAlgorithm(Algorithm):
+    """Maximum-likelihood RL (arXiv:2602.02710): the GRPO pipeline with
+    mean-normalized advantages — ``(reward − group mean) / group mean``
+    instead of plain centering. The mean normalization upweights low-pass-rate
+    examples like the maximum-likelihood gradient does, and ``group_size``
+    doubles as the truncation order of the likelihood expansion the gradient
+    is unbiased for (REINFORCE at 1 → exact maximum likelihood as it grows)."""
+
+    async def score_group(self, group: list[Rollout]) -> None:
+        apply_advantage_fn(group, max_rl_advantage_fn)
```
mikasenghaas: this seems off, should we not inline the max rl adv fn here?

src/prime_rl/orchestrator/algo/grpo.py
```diff
@@ -0,0 +1,27 @@
...
+        self.length_penalty = config.length_penalty
+
+    async def score_group(self, group: list[Rollout]) -> None:
+        assign_group_norm(group, self.length_penalty)
```
mikasenghaas: same here, shouldnt we inline this function here

src/prime_rl/orchestrator/algo/opd.py
```diff
@@ -0,0 +1,54 @@
...
+        self.teacher = config.model
+        self.teacher_pool: StaticInferencePool | None = None  # static teacher endpoint, connected in setup()
+
+    async def setup(self) -> None:
+        pool = await self.connect(self.teacher)
+        if not isinstance(pool, StaticInferencePool):
+            raise TypeError("opd teacher must be a static endpoint — prefill scoring needs fixed endpoints")
+        self.teacher_pool = pool
+
+    async def score_batch(self, batch: list[Rollout]) -> None:
```
mikasenghaas: what is this doing?

configs/ci/integration/reverse_text_rl_opd/start.toml
hallerite: it's because we use a renderer and that has a bit of a mismatch in CI with the custom chat template

configs/ci/integration/reverse_text_rl_sft/start.toml
hallerite: same reason as above

src/prime_rl/orchestrator/algo/base.py
hallerite: the idea is that we do computationally heavy ops like teacher-prefill here, because these are the rollouts that have made it into the batch. this shape will likely change once we do proper env samplers, I think.

src/prime_rl/orchestrator/algo/base.py
hallerite: I don't think so. It will be worth it for when adv computation involves judging and the model that we are training is used as judge, for example.

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
hallerite: nah, because this is the loss type for the action tokens, i.e. tokens that were sampled by the policy.

src/prime_rl/orchestrator/algo/base.py
hallerite: I guess finalize_rollout and `finalize_group` could be methods, but `finalize_batch` couldn't, because it does it across different envs (as rollouts from a batch can come from different envs). imo keeping symmetry here is better than slightly improved call-site ergonomics, but an asymmetry on where these methods live, but happy to debate this

packages/prime-rl-configs/src/prime_rl/configs/algorithm.py
hallerite: this goes against the slim-config

mikasenghaas (approved):
LFGTM

PR #2764, opened 2026-06-10
feat(orchestrator): named algorithm classes own assign/score; declared model roles

hallerite:
## What

Two changes that are really one insight: **an algorithm is credit assignment and loss routing, fused — and sampling is the other, separate thing.**

### 1. Named algorithm classes own their methods

Each training algorithm is a named runtime class — the algorithm object *is* the algorithm:

| `advantage.type` | Class | `assign` (group time) | `score` (ship time) |
|---|---|---|---|
| `group_norm` | `GRPOAlgorithm` | group-norm credit (optional length penalty) | — |
| `echo` | `EchoAlgorithm` | group-norm credit + weighted ce on observation tokens | — |
| `ref_kl` | `OPDAlgorithm` | group-norm credit (DPPO sign steering) | own-context prefill under the **teacher** |
| `demo_ref_kl` | `OPSDAlgorithm` | — | demo-conditioned prefill under the **teacher** |
| `supervised` | `SFTDistillAlgorithm` | group-norm credit (feeds filters) | — |
| `reward` | `RewardAlgorithm` | raw reward | — |
| `custom` | `CustomAlgorithm` | user function (`import_path`) | — |

Reading a class top to bottom reads the algorithm; writing your own is subclassing `Algorithm` and overriding the same two methods. The intermediate `AdvantageStrategy` layer is deleted — the algorithm object owns `assign`/`score` directly, duplication of orchestration between similar algorithms (OPD vs OPSD) is accepted so each class stays self-contained, and shared *math* (group normalization, prefill alignment, length penalties) stays as plain functions in `algo/advantage.py`.

Algorithms also **declare what they need**:

- `action_loss_type` — which loss component their action tokens feed.
- `model_role` — what the algorithm calls its reference model. The distillation algorithms declare `"teacher"`, which makes `[orchestrator.algo.teacher]` a parse-time alias for the `model` shorthand and puts the same word in validation errors ("algorithm 'opd' needs a teacher — set 'teacher' on the algorithm ...").

Roles stay strictly algorithm-local: the dispatcher, sink, and trainer keep branching on liveness alone — no role ever reaches flow code or the wire. Algorithms take exactly two runtime resources — `Algorithm(config, policy_pool, renderer)`; text → token ids always goes through the renderer (`demo_ref_kl` requires one, validated at config time).

### 2. Three components become two: `loss` merges into `advantage`, sampling becomes the `Sampler`

The old `algo.loss` component was a fiction. Credit assignment computes a token's weight; loss routing picks which component stream it lands in — but those are two coordinates of the *same* output, the per-token weight streams the trainer executes. The code already agreed: `action_loss_type` had migrated onto the algorithm class (with an assert that the config copy matched), and `LossRoutingConfig` was down to echo's two knobs. So:

- **`algo.loss` is deleted.** ECHO becomes a proper advantage type: `EchoAdvantageConfig(GroupNormAdvantageConfig)` with `observation_weight`, built into `EchoAlgorithm(GRPOAlgorithm)`. Every preset delta is now uniformly a component-type swap (`echo` → `{advantage: {type: "echo"}}`).
- **The algorithm bundle is two components**: `sampling` (how rollouts are produced) and `advantage` (the per-token training signal — credit assignment + loss routing, fused).
- **The runtime splits along the same seam.** A new `Sampler` (`orchestrator/sampler.py`) owns the rollout source: the generating pool, frozen-source connection, `sampling_args` (logprob stripping), and `samples_from_live_policy` (prefix-cache salting / staleness). Future sampling strategies — replay buffers, branching — extend here, not the algorithm. The `Algorithm` keeps only the signal half: `assign`, `score`, the reference pool, and wire stamping. Each `TrainEnv` carries both (`env.sampler`, `env.algorithm`).

One deliberate expressiveness trade: routing is no longer a free config axis (you can't write `opd` + observation-CE in TOML anymore) — routing variation *is* algorithm variation, expressed as a class. That's the point.

## Breaking changes

- **Config**: `[orchestrator.algo.loss]` is gone. `name = "echo"` works unchanged; an explicit λ moves from `algo.loss.observation_weight` to `algo.advantage.observation_weight`. A per-env `advantage` shorthand now replaces the inherited algorithm's routing along with its credit (advantage swaps are wholesale).
- **Orchestrator-internal Python API**: `AdvantageStrategy` / `setup_advantage_strategy` gone (named classes exported instead); `Algorithm` no longer has `sampling_pool` / `sampling_args` / `samples_from_live_policy` (use `env.sampler`); `stamp_loss_routing` takes a plain `observation_weight: float | None`. The custom advantage *function* interface (`AdvantageInputs` / `AdvantageOutputs`, `import_path`) is unchanged.
- **Wire/trainer**: untouched — same three weight streams.

## Validation

- Full CPU unit suite green (526 passed; pre-existing `test_qwen3_5_moe*` tilelang failures on this box excluded), ruff clean. Updated tests: preset expansion (echo → advantage type `echo`), echo λ override on `advantage`, `stamp_loss_routing` weight signature, env-inheritance semantics, `teacher`-alias, role-aware error regexes.
- 5-step GPU smokes (2-GPU, all exit 0): **grpo** (GRPOAlgorithm hot path), **self_distill** (OPSDAlgorithm `score()` against the live policy pool, demo prefix via `renderer.render_ids`), and **echo** on the multi-turn alphabet-sort env (EchoAlgorithm + `Sampler` split: observation tokens tagged, ce-weighted, and trained).

Stacked on #2746 (`feat/algorithm-abstraction`).


hallerite:
Folded into #2746 — the stack is one PR now (this branch was fast-forwarded into `feat/algorithm-abstraction`, so GitHub shows it as merged).

Issue #1166, opened 2025-10-27
Enable Nvidia Dynamo Inference backend


AmeenP:

plugyawn:
@AmeenP is this still planned?

AmeenP:
@plugyawn Yes it's being worked on internally in collaboration with Nvidia! We plan to share results to this repo.

plugyawn:
Oh fun! Would've loved to contribute, can't wait to see faster inference. 
Is #737 still planned after this (the comment says SGLang inference is not planned)?

AmeenP:
@plugyawn Our focus is on dynamo+vLLM at the moment, but SGLang will also be supported through the dynamo path as a followup. Easily being able to swap engines is one of the advantages of supporting dynamo, along with various optimizations for scalable and fast production deployments.

Reach out if you're interested in getting involved.

plugyawn:
Would love to help out and contribute, yep!

PR #2777, opened 2026-06-11
Add inference backend A/B benchmark

RitwijParmar:
## Summary
- add an endpoint-level benchmark framework for OpenAI-compatible inference backends such as vLLM, routers, and Dynamo experiments
- support both one-off comparisons and JSON scenario-suite runs
- include a ready-to-run four-profile suite for short-rollout latency, long-context prefill, high-concurrency decode, and session-cache reuse
- measure warmup-excluded request throughput, output throughput, streaming TTFT, p50/p95/p99 latency, error rate, and per-request failures
- snapshot /metrics before and after each backend with a self-contained vLLM counter parser for token deltas, prefix-cache hit rate, and NIXL failure counters
- add optional per-scenario regression gates so backend experiments can fail on throughput, latency, or error-rate regressions before being wired into full RL runs
- write aggregate Markdown reports and JSON samples so backend comparisons can be reviewed and debugged after the run

Related to #1166. This does not duplicate the Dynamo backend implementation. It gives the project a repeatable way to compare Dynamo, vLLM, or router endpoints under rollout-like traffic before moving a backend into training.

## Checks
- uv run --no-sync ruff check benchmarks/scripts/inference_backend_benchmark.py tests/unit/test_inference_backend_benchmark.py
- uv run --no-sync python -m py_compile benchmarks/scripts/inference_backend_benchmark.py tests/unit/test_inference_backend_benchmark.py
- PYTHONPATH=src uv run --no-sync pytest --confcutdir=tests/unit tests/unit/test_inference_backend_benchmark.py -q
- git diff --check

Note: regular uv run sync is blocked locally on macOS because this checkout's lockfile only supports Linux platforms. The repo-level pytest conftest also pulls in heavier runtime setup that is unrelated to this pure benchmark module, so I ran the focused tests with --confcutdir=tests/unit after populating submodules and installing the minimal local test tooling.

### DIFF
--- a/benchmarks/configs/inference_backend_suite.json
+++ b/benchmarks/configs/inference_backend_suite.json
@@ -1,0 +1,62 @@
+{
+  "model": "Qwen/Qwen3-4B-Instruct-2507",
+  "scenarios": [
+    {
+      "name": "short_rollout_latency",
+      "requests": 256,
+      "warmup_requests": 32,
+      "concurrency": 16,
+      "sessions": 64,
+      "prompt_words": 256,
+      "max_tokens": 64,
+      "gate": {
+        "min_request_throughput_ratio": 1.05,
+        "max_latency_p95_ratio": 1.1,
+        "max_error_rate": 0.01
+      }
+    },
+    {
+      "name": "long_context_prefill",
+      "requests": 128,
+      "warmup_requests": 16,
+      "concurrency": 16,
+      "sessions": 32,
+      "prompt_words": 4096,
+      "max_tokens": 128,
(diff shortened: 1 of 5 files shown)


RitwijParmar:
I opened this as a draft because the next useful step is a real vLLM vs Dynamo run.

If there is a preferred Dynamo branch or launch command, I can run the suite against it and add the result artifact here.

The suite covers short rollout latency, long-context prefill, high-concurrency decode, and session-cache reuse.

samsja:

PR #2786, opened 2026-06-12
fix(v1): multimodal training — pick up the mm wire fix + keep tensors off disk

mikasenghaas:
## Summary

- Companion to PrimeIntellect-ai/verifiers#1653 — the core fix: multimodal image tensors were dropped at the env-server → orchestrator wire (`MessageNode.multi_modal_data` was `exclude=True`), so the trainer forwarded `<|image_pad|>` tokens with no pixels → multi-OOM train/inference logprob mismatch on VLM RL.
- Bumps `deps/verifiers` to the wire-survive fix.
- `orchestrator.py`: dump train/eval rollouts with `exclude={"nodes": {"__all__": {"multi_modal_data"}}}` so the now-wire-carried tensors stay out of `results.jsonl` / wandb sample tables.

## Verification

- `color-codeword-v1` (Qwen3-VL-4B): completion-token `masked_mismatch_kl` dropped from **~5.24** to **~0.001** (matching the v0/main path). Fixed v1 eval reward **climbs and holds** (step 0/5/10/15 = 0.63 / 0.94 / 0.94 / 0.94) vs the buggy run that *declined* (0.50 → 0.47 → 0.38) — matches main's healthy curve.

## Note

verifiers #1653 has merged; this is pinned at verifiers `feat/nano-as-v1` (includes the fix).

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -101,6 +101,11 @@ MAX_CONSECUTIVE_EMPTY_BATCHES = 10
 # resumed when the watcher advances ``policy.version``.
 TARGET_LAG = 1
 
+# Drop the per-node multimodal tensors when dumping a Trace to disk (rollout jsonl / wandb
+# tables): they're the training mm_kwargs carrier, not part of the rollout record, and base64
+# pixel tensors would bloat every line. `__all__` applies the exclude to every node in the list.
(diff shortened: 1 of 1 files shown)


mikasenghaas:
Superseded by https://github.com/PrimeIntellect-ai/prime-rl/pull/2794 — a clean re-do of this PR's only remaining unique content (the rollout-dump mm-exclude) off current `feat/nano-as-v1`. The pin / uv.lock / hendrycks-config / client.py bits here are now stale or already landed.

PR #2787, opened 2026-06-12
Fix/readme trainer torchrun

anravich13-cloud:
> [!NOTE]
> **Low Risk**
> Default orchestrator behavior is unchanged; changes are README/debug-config plus an opt-out flag for standalone validation runs.
> 
> **Overview**
> Fixes **environment validation** docs so the RL trainer step matches how `prime_rl.entrypoints.trainer` must be started (`torchrun`), and swaps the eval smoke test to **`vf-eval`** against a running inference server.
> 
> Adds **`wait_for_trainer`** on the orchestrator (default **true**). When **false**, `update_dispatch_gate` keeps dispatch open so **`configs/debug/orch.toml`** can finish **`max_steps`** without a trainer consuming weight updates. That debug config also sets **`zero_advantage`** post-batch filter to **monitor** mode so short completions that yield all-zero advantages do not abort the run.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08a169a6643365d8fc82566803a9db62e1f13e0c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,8 +157,10 @@ uv run sft @ configs/debug/sft/train.toml
 4. Check that you can run the RL trainer (*this requires 1 GPU*)
 
 ```bash
-uv run trainer @ configs/debug/rl/train.toml
+uv run torchrun --nproc-per-node 1 -m prime_rl.entrypoints.trainer @ configs/debug/rl/train.toml
 ```
+
+*The `trainer` entrypoint expects to be launched under `torchrun` (normally handled automatically by the `rl` launcher), so it cannot be invoked directly.*
 
 5. Check that you can run the inference server (*this requires 1 GPU*)
 
--- a/README.md
+++ b/README.md
@@ -176,8 +176,6 @@ uv run inference @ configs/debug/infer.toml
 uv run orchestrator @ configs/debug/orch.toml
 ```
 
-*This runs the orchestrator standalone (no trainer); the debug config sets `wait_for_trainer = false` so it completes 5 steps and exits cleanly on its own.*
-
 5.2. Check that you can run evals against the inference server
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ uv run orchestrator @ configs/debug/orch.toml
 5.2. Check that you can run evals against the inference server
 
 ```bash
-uv run eval @ configs/debug/eval.toml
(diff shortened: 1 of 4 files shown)


samsja (commented):
not clear to me why the wait_for_trainer is needed, I do understand what it does but where we need to use it

README.md
```diff
@@ -157,9 +157,11 @@ uv run sft @ configs/debug/sft/train.toml
 4. Check that you can run the RL trainer (*this requires 1 GPU*)
 
 ```bash
-uv run trainer @ configs/debug/rl/train.toml
```
mikasenghaas, line 160: this should work for single gpu no?

README.md
```diff
@@ -174,10 +176,12 @@ uv run inference @ configs/debug/infer.toml
 uv run orchestrator @ configs/debug/orch.toml
 ```
 
+*This runs the orchestrator standalone (no trainer); the debug config sets `wait_for_trainer = false` so it completes 5 steps and exits cleanly on its own.*
```
mikasenghaas: remove

configs/debug/orch.toml
```diff
@@ -1,6 +1,22 @@
...
 max_completion_tokens = 16
 
+# With 16 completion tokens, all rollouts in a group score 0 reward, so every
+# advantage is zero. Keep the zero_advantage filter in monitor mode here —
+# the default (enforce) would drop 100% of rollouts and abort the run.
+[[post_batch_filters]]
+type = "gibberish"
+
+[[post_batch_filters]]
+type = "repetition"
```
mikasenghaas: can remove

README.md
samsja: it used to but maybe it failed now ?

aniravich:
> not clear to me why the wait_for_trainer is needed, I do understand what it does but where we need to use it
@samsja 

This is only needed for the readme step 5.1, which otherwise just hangs when there's no trainer -  the dispatch gate pauses at TARGET_LAG waiting for a weight broadcast that never comes without a trainer — exactly what happens in README "Validate your environment setup" step 5.1 (uv run orchestrator @ configs/debug/orch.toml), which hangs after ~2 steps instead of finishing. wait_for_trainer = false keeps the gate open so the run ships all max_steps batches and exits 0.

PR #2793, opened 2026-06-12
Validate requested attention implementation per model at load time

hallerite:
Adds a declarative, per-model constraint on which attention implementations a custom model supports, validated once in `get_model` before construction.

## Why

The constraint information existed but was scattered across three half-mechanisms with bad failure modes:

- Every modeling file dispatches via `ATTN_IMPL2CLASS[config._attn_implementation]` — an unsupported impl died with a bare `KeyError: 'eager'` deep inside layer construction.
- Requesting `flash_attention_3`/`fa4` without the kernel installed passed load and only failed with a confusing TypeError at the **first forward pass** (`self.func = None`).
- The one clean guard (GPT-OSS eager-only off Hopper) was hardcoded in `get_model`, far from the model it describes.
- transformers' `_supports_flash_attn`/`_supports_sdpa` flags can't distinguish fa2/fa3/fa4 or express hardware conditions, and prime's `fa4` string is invisible to them entirely.

## What

- `PreTrainedModelPrimeRL.supported_attn_impls` — class-level set, defaulting to the shared `layers.attn` dispatch (`sdpa`, `flash_attention_2`, `flash_attention_3`, `fa4`).
- `PreTrainedModelPrimeRL.validate_attn_impl(config, attn)` — membership check with an actionable error (model name + supported list + the config knob to set), plus a fail-fast kernel-availability check (`require_flash_attn_kernels` in `layers/attn.py`). Models override it for conditional constraints.
- `get_model` calls it once for custom-impl models (text and VLM paths) right after impl resolution. HF-impl models keep transformers' own validation.
- Migrations: **GPT-OSS** Hopper rule moves from `get_model` into the class (same error message); **Laguna**, **AFMoE**, and **Qwen3.5-MoE** declare `eager` (their factories alias it to SDPA); **GLM DSA** overrides to a no-op (its sparse MLA kernels ignore the knob entirely).
- One-paragraph note in `docs/development.md` under "Implement the Modeling Code".

## Validation

All paths exercised on this box (RTX PRO 6000, SM 12.0):

```
glm4 + flash_attention_2          → OK
glm4 + eager                      → ValueError: Glm4MoeForCausalLM does not support attn='eager'. Supported: [...]
gpt-oss + flash_attention_2       → ValueError: GPT-OSS requires 'attn = "eager"' on non-Hopper GPUs (detected SM 120)
gpt-oss + eager                   → OK
laguna + eager                    → OK
glm_moe_dsa + flash_attention_3   → OK (knob ignored, no kernel requirement)
afmoe / qwen3_5_moe + eager       → OK (aliased to SDPA, caught by Bugbot — full sweep below)
get_model(impl=custom, attn=eager, samsja/mini-glm-moe)  → clean ValueError from the new hook
get_model(impl=custom, attn=sdpa,  samsja/mini-glm-moe)  → loads
```

Full sweep of every registered family against every `attn` value, asserted against each modeling file's actual dispatch:

| Model | Accepted |
|---|---|
| llama, qwen3, qwen3_moe, glm4_moe, minimax_m2, nemotron_h | sdpa, fa2, fa3, fa4 |
| afmoe, qwen3_5_moe, laguna | eager (aliased to sdpa), sdpa, fa2, fa3, fa4 |
| glm_moe_dsa | all (own sparse MLA kernels; knob unused) |
| gpt_oss | eager (+ flash on Hopper) |

`pytest tests/unit -m "not gpu" -k "conversion or model or attn"`: 32 passed.

The in-flight Hy3 (#2789) and MiMo-V2.5 (#2791) models are covered by the base default and will pick this up on rebase (MiMo's attention also dispatches `eager`, so it will declare its set then).

### DIFF
--- a/src/prime_rl/trainer/models/gpt_oss/modeling_gpt_oss.py
+++ b/src/prime_rl/trainer/models/gpt_oss/modeling_gpt_oss.py
@@ -185,6 +185,21 @@ class GptOssPreTrainedModel(PreTrainedModelPrimeRL):
     _can_record_outputs = {"hidden_states": GptOssDecoderLayer}
 
     @classmethod
+    def validate_attn_impl(cls, config, attn_impl: str) -> None:
+        # GPT-OSS only supports FlashAttention via kernels-community/vllm-flash-attn3, which
+        # requires Hopper (SM 90). On other architectures (e.g. Blackwell), users must fall
+        # back to eager attention.
+        HOPPER_MAJOR = 9
+        if attn_impl != "eager":
+            major, minor = torch.cuda.get_device_capability()
+            if major != HOPPER_MAJOR:
+                raise ValueError(
+                    f"GPT-OSS requires 'attn = \"eager\"' on non-Hopper GPUs (detected SM {major}{minor}). "
+                    f"The only flash attention kernel supported by GPT-OSS (kernels-community/vllm-flash-attn3) is Hopper-only. "
+                    f'Set [trainer.model] attn = "eager" in your config.'
+                )
+
+    @classmethod
     def is_hf_state_dict(cls, state_dict: dict[str, Tensor]) -> bool:
         return is_hf_state_dict(state_dict)
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -69,6 +69,8 @@ Bringing up a new model family is three steps: implement the modeling code, regi
 
 Drop the modeling code under `src/prime_rl/trainer/models/<arch>/` (HF-compatible config, modeling, and weight conversion). Mirror the layout of an existing family — `glm4_moe/` or `qwen3_moe/` are good starting points.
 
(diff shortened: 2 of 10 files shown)


src/prime_rl/trainer/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -184,6 +184,21 @@ class GptOssPreTrainedModel(PreTrainedModelPrimeRL):
...
+        # back to eager attention.
+        HOPPER_MAJOR = 9
+        if attn_impl != "eager":
+            major, minor = torch.cuda.get_device_capability()
+            if major != HOPPER_MAJOR:
+                raise ValueError(
+                    f"GPT-OSS requires 'attn = \"eager\"' on non-Hopper GPUs (detected SM {major}{minor}). "
+                    f"The only flash attention kernel supported by GPT-OSS (kernels-community/vllm-flash-attn3) is Hopper-only. "
+                    f'Set [trainer.model] attn = "eager" in your config.'
+                )
```
samsja, line 200: huh no blackwell ?

PR #2801, opened 2026-06-13
[codex] add no-inference orchestrator debug mode

samsja:
## Summary

Adds orchestrator-only debug mode for no-inference/no-trainer memory and scheduling runs.

- `debug.no_inference`: uses a no-op inference pool and debug tokenizer.
- `debug.no_trainer`: writes batches through filesystem transport and advances fake policy versions.
- Adds `no-inference-env` for deterministic heavy trajectories, optional routed-expert payloads, and mean/std rollout delays.

## Validation

- `PYTEST_OUTPUT_DIR=/tmp/outputs uv run pytest tests/unit -m "not gpu"`
- `uv run ruff check tests/debug_envs/no_inference_env/no_inference_env.py tests/unit/orchestrator/test_orchestrator_setup.py`
- `git diff --check`

---

> [!NOTE]
> **Medium Risk**
> Touches core orchestrator setup, weight watching, and dispatch gating; production paths are gated on debug flags, but mis-enabled debug in a real run would skip inference/trainer integration.
> 
> **Overview**
> Adds **`OrchestratorDebugConfig`** (`debug.no_inference`, `debug.no_trainer`) on `OrchestratorConfig`, with **`resolve_debug`** clearing the renderer and forcing filesystem rollout transport when those flags are set, and validators skipping renderer/VLM checks in no-inference mode.
> 
> **`debug.no_inference`** wires a **`NoOpInferencePool`** and drops orchestrator-level tokenizer/renderer setup; batch filters are disabled when there is no renderer (vocab comes from `AutoConfig` only when a renderer exists). **`debug.no_trainer`** uses **`DebugWeightWatcher`** to bump `policy.version` locally (including after each shipped batch via async **`update_dispatch_gate`**) so dispatch gating and off-policy cancellation still behave without a trainer.
> 
> Adds **`no-inference-env`**: synthetic multi-turn trajectories (optional routed-expert payloads, configurable rollout delays) for stress-testing the pipeline. **`psutil`** RSS logging runs per step when debug is enabled; docs cover standalone **`uv run orchestrator`** with `--debug.no-inference --debug.no-trainer`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28659c5069b6900a34b7fa395585c1f8f15c7dfd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2807, opened 2026-06-14
[codex] reduce retained train trace memory

samsja:
## Summary

Stacked on #2801. This PR contains only the train-trace memory reductions built on top of the fake-R3 orchestrator debug setup.

### What changed

- Prunes heavy raw train trajectory token/R3 payloads after samples, advantages, and pre-batch filters have run.
- Keeps lightweight raw trajectory token/R3 summaries for logging and monitors.
- Preserves trainer-bound `TrainingSample` payloads needed for training.
- Keeps token length penalty compatible with pruned raw payloads by reading compacted `completion_ids_len` summaries.
- Adds sample-based filter fallback so filters still work after raw payload pruning.
- Centralizes pruned-token sample logging helpers for Prime/W&B monitor compatibility.
- Moves interleave-time raw pruning through the shared pruner after all steps are prepared, avoiding partial compaction on malformed rollouts.
- Adds compact scalar `completion_temperature` support so the orchestrator does not fan out identical per-token temperature lists.
- Releases sent train-batch sample references after `TrainingBatch` send.
- Releases finalized rollout references after metrics/logging consume them.
- Always calls shared `trim_process_memory()` after each finalized train batch so freed glibc heap pages can return to the OS.

### Validation

- `uv run pytest tests/unit/orchestrator` -> 86 passed
- `uv run pytest tests/unit/orchestrator/test_advantage.py tests/unit/orchestrator/test_train_sink_memory.py` -> 22 passed
- `uv run pytest tests/unit/orchestrator/test_train_sink_memory.py tests/unit/orchestrator/test_filters.py tests/unit/utils/test_prime_monitor.py tests/unit/utils/test_monitor_samples.py` -> 35 passed
- `uv run ruff check src/prime_rl/orchestrator/utils.py tests/unit/orchestrator/test_advantage.py` -> passed
- `uv run ruff format --check src/prime_rl/orchestrator/utils.py tests/unit/orchestrator/test_advantage.py` -> passed
- `git diff --check` -> passed

### Fake-R3 measurements

Shape: `batch_size=512`, `max_inflight_rollouts=1024`, `seq_len=30000`, `turns=30`, GLM5 R3 defaults (`78` routed layers, top-k `8`, `256` experts), fake env, no inference, no trainer.

Earlier baselines:

- Before raw pruning, R3 logged peak: `137.190 GiB`
- After raw trajectory pruning, R3 logged peak: `66.002 GiB`

Additional lifecycle cleanup run, before final `batch.rollouts.clear()` was included in the loaded process:

- Before cleanup/scalar-temp run: step 6 logged `76.301 GiB`, final logged peak `107.765 GiB`, `/usr/bin/time` max RSS `109,208,896 KB`
- With sample cleanup + `malloc_trim`: step 6 logged `37.751 GiB`, final logged peak `79.611 GiB`, `/usr/bin/time` max RSS `109,090,388 KB`

Interpretation: this reduces retained after-step RSS materially. Peak RSS during batch assembly/send is still high and should be handled by a follow-up compact `TrainingSample` transport change for the remaining `list[int]`, `list[bool]`, and `list[float]` payloads.

---

> [!NOTE]
> **High Risk**
> Large cross-cutting surface (orchestrator memory, weight-update timing, vLLM patches, breaking advantage config, trainer batching) with correctness and stability impact on multi-node disaggregated runs.
> 
> **Overview**
> **Train-trace memory (stated PR focus):** After interleaving and pre-batch filters, raw train trajectories are **compacted** to length/metadata summaries instead of retaining duplicate token and routed-expert arrays; `TrainingSample` payloads stay intact for the trainer. Filters and length penalties read from **samples** or `*_len` fields when raw arrays are gone; Prime/W&B logging gets helpers for compacted tokens. The orchestrator uses **scalar `completion_temperature`**, clears sample/rollout references after send/logging, and calls **`trim_process_memory()`** each train step.
> 
> **GRPO advantage (breaking config):** Replaces `tokens`/`turns` length-penalty variants with a **linear** `coef * pass_rate * (completion_tokens / seq_len)` penalty and optional **`length_weighted_baseline`**; `max_seq_len` is threaded through env setup.
> 
> **Trainer & transport:** Adds **`IPO` loss**, explicit **`sequence_lengths`** on micro-batches (replacing inferring lengths from `position_ids`), and **FLOP-aware** micro-batch packing plus Karmarkar–Karp balancing across data-parallel workers.
> 
> **Inference & weight updates:** Strips **`routed_experts`** from chat-completions (eval/PD router fix), raises **DP coordinator startup timeout**; off-policy cancellation moves to **`on_version_pending`** (before engine pause) to avoid NIXL/KV abort races.
> 
> **Deployment & config:** **`orchestrator_on_inference`** relocates the orchestrator to the last inference node; multi-node runs pin **`dp_rank_count = 1`** for router URLs. Changelog documents other breaking orchestrator/inference changes; adds **`rlm-uuid-ctf`** env and Gemma4 VLM registry / softcapping fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10ed89da308eda44688773ac1bc7b9918962cc9a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/utils/monitor/prime.py
```diff
@@ -31,6 +31,19 @@ def _json(val: Any) -> str:
     return json.dumps(val)
 
 
+def _token_len(tokens: dict[str, Any] | None, key: str) -> int | None:
```
samsja: not a fan of this function

src/prime_rl/utils/monitor/wandb.py
```diff
@@ -18,6 +18,24 @@
 from prime_rl.utils.monitor.base import Monitor, sample_items_for_logging
 
 
+def _messages_to_text(tokenizer: PreTrainedTokenizer, prompt: Any, completion: Any) -> str:
```
samsja: not a fan of this function

src/prime_rl/orchestrator/trajectories.py
```diff
@@ -276,6 +339,11 @@ def prepare_step_tokens(step: vf.TrajectoryStep, step_idx: int) -> dict[str, Any
         if prepared is None:
             return None
         prepared_steps.append(prepared)
+        if prune_raw_payload and isinstance(step, dict) and isinstance(step.get("tokens"), dict):
+            step["tokens"] = _compact_token_payload(step["tokens"])
+            step.pop("response", None)
+    if prune_raw_payload:
+        output["trajectory_payload_pruned"] = True
```
samsja: eh why do we need this here ?

src/prime_rl/orchestrator/filters.py
```diff
@@ -19,6 +19,37 @@
     from prime_rl.orchestrator.types import TrainRollout
 
 
+def _generated_completion_tokens(rollout: "TrainRollout"):
```
samsja: not clear how this is woring and why we need it

PR #2844, opened 2026-06-20
fix(vlm): dispatch Gemma 4 as a VLM and read softcapping from the text config

### LINKED ISSUE #2362: Gemma 4 31B SFT + LoRA: pre-clip grad norms in the millions, several silent bugs along the way
## TL;DR

Trying to fine-tune **Gemma 4 31B (`google/gemma-4-31B-it`)** with the SFT trainer + LoRA (r=16, α=32) on `main`. Hit a chain of issues:

1. **`gemma4` is not in `VLM_REGISTRY`** → model dispatch silently picks `AutoModelForCausalLM` instead of `AutoModelForImageTextToText`, then `language_model.layers` lookup fails. Quick fix needed just to load the model. (PR #2347 partially addresses but is unmerged.)
2. **Gemma 4's global-attention layers use `head_dim=512`** which Flash-Attention does not support. With `attn = "flash_attention_2"` the model crashes inside attention. Need to fall back to SDPA for those layers only.
3. **`final_logit_softcapping` is silently ignored**. `inject_prime_lm_head` reads `getattr(model.config, "final_logit_softcapping", None)` which returns `None` for Gemma 4 because the attribute lives on `model.config.text_config.final_logit_softcapping` (HF reads it via `self.config.get_text_config().final_logit_softcapping`). prime-rl falls through to vanilla LM head with no cap. **Logits run to ±120+** instead of the trained ±30 bound. Loss/grad pathology follows.
4. **Even after fixing all the above**, pre-clip grad norm during the first ~30 steps of SFT is **1k–1M**, with sample-dependent spikes up to ~1.7M. A vanilla HF + PEFT setup on the same data produces 15k–80k for the same samples (1–2 orders of magnitude lower). Something else amplifies grad in prime-rl that we couldn't pin down.

The first three are concrete bugs. The fourth is the still-unsolved one — extra eyes welcome.

This appears to be the same family of "Gemma 4 + sharded training" issues already reported elsewhere:

- **verl** [issues/5999](https://github.com/verl-project/verl/issues/5999) — *"Gemma-4-31B: The model fails to load entirely … FQN (Fully Qualified Name) mismatch between the model and the keys in the checkpoint"* and *"Gemma-4-E4B: peaking at 50+, leading to training instability"*.
- **Unsloth** [issues/4982](https://github.com/unslothai/unsloth/issues/4982) — *"unusually high gradient norms … initial loss around 185.9"* (for us the initial loss is ~1-5, so we are not hitting the multimodal-collator variant of this; we are hitting the gradient-norm variant.)
- HF transformers [issues/43899](https://github.com/huggingface/transformers/issues/43899)
- google-deepmind/gemma [issues/270](https://github.com/google-deepmind/gemma/issues/270)
- HF model card [discussions/3](https://huggingface.co/google/gemma-4-31B/discussions/3)

So this seems to be a known "Gemma 4 day-zero" cluster of issues; filing here so the prime-rl-side surface area is on record.

---

## Environment

- prime-rl `main` (commit `24a2af00546394d836b8e256593f6b649ac16dfc`)
- Python 3.12, PyTorch 2.x (whatever the lockfile pins)
- 8× H200 (143 GB)
- transformers from prime-rl's pinned lockfile
- Gemma 4 31B base model (`google/gemma-4-31B-it`) — VLM, but using only the text path
- Custom Jinja chat template using Gemma 4's native special tokens (`<|turn>`, `<|channel>`, `<|tool_call>`, etc.) — these are real Gemma 4 vocab tokens, not OOD.

## Config

```toml
output_dir = "outputs/sft_v9"
max_steps = 400
loss_impl = "liger_fused"

[deployment]
type = "single_node"
num_gpus = 8

[model]
name = "google/gemma-4-31B-it"
trust_remote_code = true
attn = "flash_attention_2"
impl = "hf"
seq_len = 16384
optimization_dtype = "bfloat16"
reduce_dtype = "bfloat16"

[model.ac]
mode = "full"

[model.vlm]
vision_encoder_attr = "model.vision_tower"
language_model_attr = "model.language_model"
freeze_vision_encoder = true

[model.lora]
rank = 16
alpha = 32
dropout = 0.0

[data]
name = "<our private SFT dataset; see notes below>"
seq_len = 16384
batch_size = 32
micro_batch_size = 1
pack_function = "stack"

[data.loss_mask]
assistant = true

[optim]
type = "adamw"
lr = 1e-4
weight_decay = 0.0
max_norm = 1.0

[scheduler]
type = "cosine"
warmup_steps = 20
```

---

## Issue 1: `gemma4` missing from `VLM_REGISTRY`

**Symptom on `main`:** the model fails to load with

```
AttributeError: 'Gemma4Model' object has no attribute 'layers'
```

raised from `configure_moe_ep_backend(model, config)` → `for transformer_block in language_model.layers:`.

**Why:** `is_vlm_architecture(model_config)` checks `model_type` in `VLM_REGISTRY`. Gemma 4's `model_type` is `"gemma4"`. That key is not in the registry. So `get_model()` ends up loading via `AutoModelForCausalLM` (wrong class for a composite VLM), and `get_language_model(model)` falls back to `model.model` (which is `Gemma4Model`, not `Gemma4Model.language_model` where the layers actually live).

**Minimal fix** — `src/prime_rl/utils/vlm.py`:

```python
VLM_REGISTRY: dict[str, VLMModelInfo] = {
    "qwen3_vl":      VLMModelInfo("model.visual",       "model.language_model"),
    "qwen3_5":       VLMModelInfo("model.visual",       "model.language_model"),
    "qwen3_5_moe":   VLMModelInfo("model.visual",       "model.language_model"),
    "qwen3_vl_moe":  VLMModelInfo("model.visual",       "model.language_model"),
    "gemma4":        VLMModelInfo("model.vision_tower", "model.language_model"),  # <-- add this
}
```

This is also what the user's config implies via `[model.vlm]` overrides — but `is_vlm_architecture()` uses the registry, not the override, so without the registry entry the wrong load class is selected before the override is consulted.

PR #2347 partially adds this but is still open.

---

## Issue 2: Flash-attention crashes on Gemma 4 global layers

`Gemma4TextConfig.layer_types` interleaves `sliding_attention` (head_dim=256) and `full_attention` / global (head_dim=**512**, see `config.text_config.global_head_dim`). Flash-Attention 2/3 max out at head_dim 256. With `attn = "flash_attention_2"` the global layers crash inside the kernel.

**Workaround we used** (in `setup_model`, after the model is built):

```python
def _patch_gemma4_global_attention_to_sdpa_if_needed(model, config, logger):
    if config.attn not in ("flash_attention_2", "flash_attention_3"):
        return
    if getattr(getattr(model, "config", None), "model_type", None) != "gemma4":
        return
    text_model = getattr(getattr(model, "model", None), "language_model", None)
    if text_model is None or not hasattr(text_model, "layers"):
        return
    patched = 0
    for layer in text_model.layers:
        attn = getattr(layer, "self_attn", None)
        if attn is None or getattr(attn, "is_sliding", True):
            continue
        new_cfg = copy.copy(attn.config)
        new_cfg._attn_implementation = "sdpa"
        attn.config = new_cfg
        patched += 1
    if patched:
        torch.backends.cuda.enable_math_sdp(False)
        # mem-efficient SDPA (cutlass) supports head_dim=512 but rejects enable_gqa=True
        # with broadcast K/V — force the repeat_kv path:
        import transformers.integrations.sdpa_attention as _sdpa_mod
        _sdpa_mod.use_gqa_in_sdpa = lambda *a, **kw: False
```

Sliding layers stay on FA2; global layers get SDPA (mem-efficient backend). PR #2347 does something similar but is unmerged.

It would be good to have first-class support for hybrid kernel dispatch so users don't have to monkey-patch.

---

### What happens

`src/prime_rl/trainer/models/layers/lm_head.py:303`:

```python
final_logit_softcapping = getattr(model.config, "final_logit_softcapping", None)
if final_logit_softcapping:
    ...inject GemmaVanillaOutputLinear with softcap...
    return

# fallthrough — vanilla, no softcap
```

For Gemma 4's `Gemma4Config`, the attribute is at `config.text_config.final_logit_softcapping` (= 30.0), **not** at the top level. `getattr(model.config, "final_logit_softcapping", None)` returns `None`, so prime-rl falls through and injects `VanillaOutputLinear` with no softcap. Log line:

```
Injecting vanilla LM head
```

(should have been `Injecting Gemma LM head with softcap=30.0`).

### Why this matters

Gemma 4's training applied `tanh(logits/30) * 30` so logits are bounded to ±30. Without the cap, prime-rl produces logits up to **±122** in our runs. CE on such logits saturates very hard on confidently-wrong tokens, producing massive upstream gradients.

For comparison HF's own `Gemma4ForConditionalGeneration.forward` reads softcap correctly:

```python
if (final_logit_softcapping := self.config.get_text_config().final_logit_softcapping) is not None:
    logits = logits / final_logit_softcapping
    logits = torch.tanh(logits)
    logits = logits * final_logit_softcapping
```

### One-line fix

```python
# src/prime_rl/trainer/models/layers/lm_head.py
final_logit_softcapping = getattr(model.config, "final_logit_softcapping", None)
if final_logit_softcapping is None:                              # NEW
    text_config = getattr(model.config, "text_config", None)     # NEW
    if text_config is not None:                                  # NEW
        final_logit_softcapping = getattr(text_config, "final_logit_softcapping", None)  # NEW
if final_logit_softcapping:
    ...
```

### Effect

On a representative spike step we measured (8 GPU FSDP, seq_len=16384, batch_size=32):

| Metric | main (broken) | + softcap fix |
|--------|--------------:|--------------:|
| `max_logit` observed | **122.0** | **30.0** ✓ |
| Pre-clip grad norm (step 0, micro=1) | 412,250 | ~234,000 |
| Pre-clip grad norm (step 1, micro=1) | 1,690,696 | ~934,000 |

End-to-end repro on `main` (commit above), `world_size=8`, `seq_len=16384`, `batch_size=32`, LoRA r=16 α=32:

```
Step 0 | Time: 135.84s | Loss: 4.6154 | Grad. Norm: 326.0000 | LR: 1.00e-12 ...
```

Loss=4.6 is normal for cold-start LoRA on this corpus. **Grad. Norm: 326** at step 0 (post-FSDP-averaging, post-token-normalization, so the underlying per-sample raw grad is in the hundreds-of-thousands range — see Issue 4 detail below). For comparison, Qwen3-32B + LoRA on the same dataset shows pre-clip grad norm `< 1` from step 0.

(Will append more `Step` lines once the 15-step run finishes.)

Softcap matters but only halves the worst case — there is more going on (Issue 4).

---

### Setup

To rule out architecture/data noise, we ran **the exact same input** (`input_ids`, `target_ids`, `loss_mask` from one micro-batch) through:

1. **prime-rl** SFT path (after applying fixes 1–3 above), `world_size=1`, AC on, seq=12288 (truncated for memory)
2. **Vanilla HF + PEFT** (`AutoModelForImageTextToText.from_pretrained` + `get_peft_model(LoraConfig(r=16, alpha=32, dropout=0))` + `gradient_checkpointing_enable()`), single GPU

Same model weights, same LoRA target modules, same softcap, same loss formulation (sum over masked tokens, then `loss.backward()`).

### Result

| Sample | HF + PEFT grad norm | prime-rl grad norm | Ratio |
|--------|--------------------:|-------------------:|------:|
| Sample A (596 trained tokens) | 15,262 | 16,214 | 1.06× ✓ |
| Sample B (596 trained tokens, different conv.) | 78,744 | **7,193,135** | **91× ❌** |

Both samples are normal multi-turn assistant conversations from the same dataset. Sample A reproduces faithfully. Sample B explodes only in prime-rl.

For Sample B, breaking down by layer:

| Layer | prime-rl `lora_B` grad | HF + PEFT `lora_B` grad | Ratio |
|------:|-----------------------:|------------------------:|------:|
| 7     | 1,878,692              | 15,415                  | **122×** |
| 11    | 1,825,194              | 15,493                  | **118×** |
| 12    | 1,794,482              | 15,358                  | **117×** |
| 15    | 2,057,758              | 17,011                  | **121×** |
| 16    | 1,733,073              | 16,830                  | **103×** |
| 17    | 1,503,309              | 20,721                  | **72×**  |
| 20+   | (similar to HF)        | —                       | ~1×    |

Amplification is concentrated in **early-to-middle decoder layers (5–17)**; late layers (20+) match HF. This is the classic shape of backward error compounding through a long layer chain.

### Things this is NOT

- **Not the multimodal-data-collator bug** documented in unslothai/unsloth#4982. That manifests as initial loss ≈ 185 due to missing `mm_token_type_ids`/`token_type_ids`. Our initial loss is normal (~1-5 per token, mean=4.6 at step 0). Our training is text-only and doesn't pass multimodal IDs because there are no image/video/audio placeholder tokens in the input — the input_id range is well below `image_token_id=258880`.
- **Not the FSDP `no_sync()` issue** (we don't use grad accumulation skip).
- **Not data quality** — we manually decoded the offending sample (`sample01_rank7.pt`); it's a perfectly normal multi-turn assistant conversation in the same dataset.

### Possible cause (FQN / FSDP wrap mismatch on Gemma 4 specific layer types)

The verl issue (verl-project/verl#5999) traces the Gemma-4 + FSDP grad-norm explosion to a Fully Qualified Name mismatch and to FSDP not picking up Gemma 4's custom layer classes (e.g. `Gemma4ClippableLinear`, used in the vision/audio path; doesn't subclass `nn.Linear`). prime-rl uses FSDP-2 (`fully_shard`) per transformer block, not auto-wrap-by-class, so the verl-specific cause may not apply directly — but the *symptom* matches very closely (early-middle layers amplified, late layers fine, sample-dependent).

### Things ruled out

We bisected fairly aggressively:

- ✅ Switched `checkpoint_wrapper` (NO_REENTRANT) → `torch.utils.checkpoint.checkpoint` (HF-style direct wrap): **still 7.1M**. AC mechanism is not the cause.
- ✅ `preserve_rng_state=False` → `True`: **still 7.1M**. Not RNG-state-related.
- ✅ FSDP `reshard_after_forward=False`: **still 7.1M**.
- ✅ Verified FSDP-2 averaging is correct: world=8 with the same sample broadcast to all ranks gives the same grad norm as world=1. ✓
- ✅ The trigger is a specific sample, not the framework alone — Sample A goes through prime-rl cleanly (matches HF within 6%). Only Sample B explodes.

### What's special about the offending sample

Looking at Sample B vs Sample A:

| | A (works) | B (explodes 91×) |
|---|---:|---:|
| Trained tokens | 596 | 814 |
| Trained range (positions) | 8179–9530 | 9754–**12287** |
| `<\|"\|>` (id 52) count | 132 | **396** |
| Tool-call markers count | 4 | 5 |
| Last trained pos vs end of seq (12288) | 758 tokens before end | **at the end** |

Sample B has trained tokens extending to the very last position of the truncated input, and ~3× more `<|"|>` (JSON-quote) tokens. We suspect one of these properties triggers a numerical issue specific to prime-rl's stack (LoRA layer impl, FSDP wrap, patched `model.forward`, custom LM head, or some interaction). We were not able to localize it further without more invasive surgery.

### Reproduction

We can attach `sample01_rank7.pt` (the 91× sample's `input_ids`/`target_ids`/`loss_mask`) and a minimal replay script. The script runs both prime-rl forward+backward and HF+PEFT forward+backward on the same tensors and prints the grad norms side-by-side. Happy to share if a maintainer wants to dig in.

---

## What we'd love help with

- **Confirm and merge fixes 1–3** (or close as duplicates of the relevant PRs).
- **Suggest where to look for issue 4**. The 91× amplification reproduces deterministically on at least one sample; we'd like to catch this before it bites others.

Thanks 🙏

EazyReal:
## What

Gemma 4 vision-language models fail to train, two ways:

1. **Wrong model dispatch.** `gemma4` is missing from `VLM_REGISTRY`, so `is_vlm_architecture()` returns `False` and the trainer loads the checkpoint with `AutoModelForCausalLM` instead of `AutoModelForImageTextToText`, then can't locate the language-model submodule (`AttributeError`).
2. **Softcapping silently dropped.** `final_logit_softcapping` is read from the top-level config, but on composite VLM configs the value lives under `text_config` — so it reads `None` and Gemma's logit softcapping is never applied.

## Fix

- Register `gemma4` in `VLM_REGISTRY` (`model.vision_tower` / `model.language_model`).
- Read softcapping via `config.get_text_config()`, which returns the nested text config for VLMs and the config itself for text-only models (so the text-only path is unchanged).

Fixes #2362

---

> [!NOTE]
> **Medium Risk**
> Changes model-class dispatch and LM-head softcapping for a new VLM family; behavior for existing registered VLMs and text-only models should stay the same if `get_text_config()` matches prior top-level reads.
> 
> **Overview**
> **Gemma 4** can train as a VLM again: `gemma4` is added to `VLM_REGISTRY` with `model.vision_tower` / `model.language_model`, so `is_vlm_architecture()` routes loading and submodule access like other VLMs instead of treating the checkpoint as text-only.
> 
> **Gemma-style final logit softcapping** is no longer skipped on composite configs. LM head injection now uses `get_final_logit_softcapping()`, which reads `final_logit_softcapping` from `config.get_text_config()` (nested `text_config` for VLMs, same as before for text-only models).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fe10791d99ab629a57a4a6a673bf0822fafc1dc6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


S1ro1:
@hallerite or @eligotts for VLM

hallerite (approved):
lgtm! thanks for the contribution :)

PR #2845, opened 2026-06-20
feat(orchestrator): make target_lag a configurable field

### LINKED ISSUE #2812: `TARGET_LAG = 1` is a silent, hardcoded config choice that I think should be surfaced?
Hi and thank you for the great framework!

I recently learned of `TARGET_LAG = 1` in `src/prime_rl/orchestrator/orchestrator.py:109` which determines: 
```python
# Maximum batches the orchestrator may run ahead of the trainer. The
# dispatcher is paused via ``update_dispatch_gate`` once this is exceeded;
# resumed when the watcher advances ``policy.version``.
```
May I ask why you hardcoded `TARGET_LAG = 1` instead of exposing it in the config? Why did you pick 1 instead of a larger value? 

I think the (silent) existence of `TARGET_LAG` makes `max_off_policy_steps` misleading. 
I thought that by setting `max_off_policy_steps = 8` (as set in some examples I followed) the rollout engine will continue generating rollout indefinitely and rollouts with `off_policy > 8` will discarded. 
My understanding is that this is *not* what happens, and for `max_off_policy_steps = 8` to really be hit, one needs to have `max_inflight_rollouts = 8 * batch_size` (examples and I set this to a much smaller value).

p.s. , I think `prime-rl` will strongly benefit from page of concentrated all (or most) config knobs like [what verl has](https://verl.readthedocs.io/en/latest/examples/config.html).

EazyReal:
## What

`TARGET_LAG` — the maximum number of batches the orchestrator may run ahead of the trainer — was a hardcoded module constant (`= 1`) in `orchestrator.py`. As noted in #2812, this is a meaningful async/off-policy knob that operators can't tune (and it interacts with `max_off_policy_steps`) without editing source.

## Fix

Promote it to `OrchestratorConfig.target_lag` (`Field(1, ge=1)`, documented next to `max_off_policy_steps`), delete the constant, and read `self.config.target_lag` at the dispatch gate.

Default `1` preserves the current one-step-ahead pipelining behavior exactly. Larger values allow deeper look-ahead at the cost of staleness; `0` is rejected because it would keep the dispatcher paused in steady state.

Focused config tests cover the default and the lower bound.

Closes #2812

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
@@ -557,6 +557,9 @@ class OrchestratorConfig(BaseConfig):
     max_off_policy_steps: int = Field(8, ge=0)
     """Maximum policies allowed to generate a single rollout. Rollouts generated more than ``max_off_policy_steps`` ahead of training are discarded. Higher values yield better throughput at the cost of off-policy noise."""
 
+    target_lag: int = Field(1, ge=1)
+    """Maximum number of batches the orchestrator may run ahead of the trainer. The dispatcher is paused once the orchestrator's next batch would lead ``policy.version`` by more than this; it resumes when the trainer advances the policy version. Default ``1`` is the standard one-step-ahead pipelining; larger values allow deeper look-ahead at the cost of staleness."""
+
     bench: bool = False
     """Benchmark mode. Sets ``max_steps`` to 5 and disables W&B."""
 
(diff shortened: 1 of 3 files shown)


EazyReal:
Good catch — you're right that `target_lag=0` deadlocks (in steady state `lead = (step+1) - version` is always ≥ 1, so `lead > 0` never reopens the gate). Since `0` isn't a usable value, I bounded the field `ge=1` (default unchanged at 1) and dropped the "0 = synchronous" claim from the docs/tests.

EazyReal:
@mikasenghaas Could I get a quick review on this config PR? It surfaces `target_lag` with default `1` and keeps `0` rejected to avoid dispatcher deadlock.

PR #2856, opened 2026-06-23
feat: adapt orchestrator to index-addressed env server

mikasenghaas:
## Summary

Companion to verifiers PR [#1829](https://github.com/PrimeIntellect-ai/verifiers/pull/1829), which makes the v1 env-server index-addressed with **no required task count** so it can serve lazy/unbounded tasksets — `InfoResponse.num_tasks` is now `int | None` (`None` = served lazily; the caller drives `task_idx`). This adapts the orchestrator to that API.

- `Env.num_tasks` is now `int | None` (`None` when the taskset is served lazily).
- `EvalEnv.start` requires `num_examples` to bound a lazily-served eval — an unknown/infinite count can't be enumerated (raises a clear error if `num_examples < 0`).
- `TrainSource`:
  - **Finite path unchanged** — enumerate the env's index range, shuffle, reshuffle each epoch, weight by dataset size.
  - **Unbounded path** (`num_tasks is None`) — `next_example` streams a monotonically increasing `task_idx` (no epoch/reshuffle; the taskset's own generator, e.g. RNG-seeded pairs, supplies the variety).
  - An unbounded env has no dataset size to weight by, so it requires an explicit `ratio` (clear error otherwise).

Backward-compatible: against a server that still reports a count (every current taskset that returns a list, and the v0 bridge), every env is bounded and behavior is identical. The change is safe to land before the verifiers pin bump.

## Verification

Unit-checked `TrainSource` against stub envs:
- finite env → epochs cover all indices, reshuffled (e.g. `[2,0,1,3, 1,2,0,3, ...]`);
- unbounded env (`num_tasks=None`, with a `ratio`) → streams `[0,1,2,3,4,5]`;
- unbounded without a `ratio` alongside a finite env → `ValueError` (no size to weight by);
- mixed finite + unbounded with ratios → both envs sampled.

`ruff check` / `ruff format` clean (pre-commit).

## Follow-up

The verifiers pin bump (to the merged #1829 commit) lands separately once that PR merges; only then does an actual unbounded taskset report `num_tasks=None` end-to-end.

### DIFF
--- a/src/prime_rl/orchestrator/envs.py
+++ b/src/prime_rl/orchestrator/envs.py
@@ -3,8 +3,8 @@
 Each ``Env`` owns a v1 ``EnvServer`` (spawned as a child process, or an
 external one given by ``config.address``) and an ``EnvClient`` to drive it. The
 orchestrator never *runs* an environment: it asks the server for ``info``
-(``num_tasks`` + whether group scoring is needed), then runs rollouts purely by
-**task index**. The server returns a ``Trace`` (a plain ``model_dump`` — derived values are
+(``num_tasks`` — ``None`` when the taskset is served lazily — + whether group scoring is
+needed), then runs rollouts purely by **task index**. The server returns a ``Trace`` (a plain ``model_dump`` — derived values are
 properties, not serialized) which we validate into a ``Trace[WireTask]`` — a real ``vf.Trace``
 (never a loose dict) whose task keeps the env's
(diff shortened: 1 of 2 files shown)


mikasenghaas:
Superseded by #2882, which reimplements the same orchestrator change off `main` (this branch was based on `feat/nano-as-v1`).

PR #2868, opened 2026-06-24
feat: default ep to auto for MoE models

samsja:
## Summary

Two sets of sane defaults so users get memory-efficient + EP-enabled training out of the box without manual tuning.

### 1. Memory-saving defaults

| Feature | Old default | New default |
|---|---|---|
| `model.ac` | `None` (disabled) | `ActivationCheckpointConfig()` (full mode) |
| `model.ac_offloading` | `None` (disabled) | `ActivationOffloadingConfig()` |
| `model.compile` | `None` (disabled) | `CompileConfig()` (`fullgraph=false`) |
| `model.optim_cpu_offload` | `False` | `True` |
| `model.fused_lm_head_token_chunk_size` | `"disabled"` | `1024` |

- SFT trainer silently disables `fused_lm_head_token_chunk_size` (not supported yet)
- Multi-run CI config (`max_concurrent_runs=2`) explicitly sets `optim_cpu_offload = false`
- Debug configs explicitly disable compile

### 2. EP "auto"

- `model.ep` type changed from `int` to `int | Literal["auto"]`, default `"auto"`
- At trainer startup, `resolve_ep()` loads `AutoConfig` to read `num_experts`, then picks the largest EP degree up to 8 that divides `num_experts`, divides the FSDP island size (`world_size // dp_replicate`), and is a multiple of `cp`
- Non-MoE models resolve to 1 (no-op)
- Called before `get_parallel_dims()` in both RL and SFT trainers

### Files changed

- `trainer.py`: defaults + EP type + validator fixes
- `sft.py`: SFT chunked loss silent disable + EP validator fix
- `parallel_dims.py`: `resolve_ep()` + `_get_num_experts()`
- `rl/train.py` + `sft/train.py`: import + call `resolve_ep`
- Debug/CI tomls: override compile/optim_cpu_offload where needed
- `docs/scaling.md` + `CHANGELOG.md`

---

> [!NOTE]
> **High Risk**
> Breaking default trainer behavior changes memory, compile, and offload for every run without config overrides; `ep="auto"` loads HuggingFace `AutoConfig` at startup and can enable EP/DeepEP interactions users did not set explicitly.
> 
> **Overview**
> **Breaking trainer defaults** turn on memory-saving paths out of the box: activation checkpointing, activation offloading, `torch.compile`, optimizer CPU offload, and fused LM-head chunking (`1024` for RL). Existing configs that assumed the old “all off” behavior need explicit opt-outs (`None` / `"disabled"` / `false` as documented in `CHANGELOG.md`). Multi-run (`max_concurrent_runs > 1`) must set `optim_cpu_offload = false`; SFT **always** forces `fused_lm_head_token_chunk_size` to `"disabled"` instead of erroring.
> 
> **`model.ep` defaults to `"auto"`** (`int | "auto"`). RL and SFT trainers call new `resolve_ep()` before `get_parallel_dims()`: non-MoE or HF-only impl → `1`; MoE with custom/auto impl → `min(world_size // dp_replicate, 8)`. Validators now treat `ep="auto"` like disabled EP for custom-impl checks.
> 
> CI/debug TOMLs opt out of compile (and multi-run opts out of optim CPU offload). `docs/scaling.md` and benchmark JSON baselines are refreshed for the new defaults. Inference docs stress **required** NIXL built against UCX 1.19 for disaggregated P/D.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0e6ad0c79b44d65b8bc028db339e977530f5319. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/trainer/sft/train.py
```diff
@@ -94,6 +94,9 @@ def train(config: SFTConfig):
     if config.model.lora is not None:
         setup_multi_run_manager(config.output_dir, 1, torch.device("cuda", world.local_rank), config.model.lora)
 
+    # Resolve ep="auto" to a concrete integer before creating parallel dims
+    resolve_ep(config.model)
```
samsja, line 98:
```suggestion
    resolve_auto_ep(config.model)
```

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -666,7 +666,7 @@ def auto_setup_fused_lm_head_token_chunk_size(self):
 
     @model_validator(mode="after")
     def ep_only_with_custom_impl(self):
-        if self.model.ep > 1 and self.model.impl not in ("custom", "auto"):
+        if self.model.ep != 1 and self.model.ep != "auto" and self.model.impl not in ("custom", "auto"):
             raise ValueError("EP is only supported with the custom implementation or auto mode")
```
samsja, line 670: camt we use greater than int nstead for the type ?

configs/debug/moe/sft/train.toml
```diff
@@ -5,8 +5,7 @@ name = "PrimeIntellect/GLM-0.5B"
 moe_use_grouped_mm = false
 impl = "custom"
 attn = "sdpa"
-
-[model.compile]
+compile = "None"
```
mikasenghaas: this is wrong

mikasenghaas (commented):
i like it

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -143,8 +143,11 @@ class ModelConfig(BaseModelConfig):
     dp_replicate: int = 1
     """Data parallel dim where model weights are replicated."""
 
+    auto_ep: bool = True
```
mikasenghaas: hm i liked the `ep: Literal["auto"] | int` better

samsja: so what I didn't like is that the ep is not an int only anymore so its a bit confusing but the whole code because technically one should always check if its not auto anymore

mikasenghaas: ya i get it but mutually exclusive configs are also a bit ugly. maybe we can find a good solution for type narrowing after "auto" has been resovled?

we have this pattern in a few places alr iirc

samsja: hmmmm I don't think that there is a good solution at the p ydantic level, let me revert to auto for now

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -347,12 +347,6 @@ def validate_opt_and_fsdp_offload(self):
 
     @model_validator(mode="after")
     def validate_and_disable_chunked_loss(self):
-        if isinstance(self.model.fused_lm_head_token_chunk_size, int):
-            raise ValueError(
-                "Chunked loss is not supported for SFT training yet, please set "
-                "`model.fused_lm_head_token_chunk_size` to 'disabled'"
-            )
-
```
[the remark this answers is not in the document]
mikasenghaas: seems valid?

samsja:
Thanks BugBot. Here's a summary of the findings and their status on the latest commit:

**Fixed:**
- "Auto EP ignores HF impl guard" / "Auto EP skips impl guard" — `resolve_ep` now checks `config.impl not in ("custom", "auto")` and sets `ep=1` before returning. Config validators also skip the custom-impl check for `"auto"` because `resolve_ep` enforces it at runtime.
- "SFT never resolves auto EP" / "SFT skips auto EP resolution" — both RL and SFT trainers now call `resolve_ep(config.model)` before `get_parallel_dims()`.
- "SFT validator breaks on auto ep" — both SFT and RL `ep_only_with_custom_impl` validators now use `self.model.ep != 1 and self.model.ep != "auto"` instead of `self.model.ep > 1`, avoiding the `TypeError` from comparing `"auto"` to an int.

**Known / by design:**
- "DeepEP check skipped for auto" / "DeepEP check skipped on early exit" — `validate_ep_comm_backend` intentionally skips the check for `ep="auto"` because the actual EP degree isn't known at config-validation time. `resolve_ep` checks `ep_comm_backend != "torch" and ep <= 1` on the MoE path where EP could be >1. On early-exit paths (non-MoE, HF impl), `ep` is set to 1, so DeepEP simply won't be activated — the mismatch is benign.
- "Expert count skips text_config" — `_is_moe_model` only checks the root `AutoConfig` for `num_experts` / `n_routed_experts`. VLM MoE configs that store these under `text_config` would resolve to `ep=1`. This is an edge case; users can set `ep` explicitly as a workaround.
- "Auto EP ignores mesh constraints" — `resolve_ep` picks `min(fsdp_island_size, 8)` without checking divisibility against `num_experts`. `ParallelDims._validate` still asserts `ep % cp == 0` and `(dp_shard * cp) % ep == 0`, so invalid degrees crash loudly rather than silently producing wrong results. The missing `num_experts` divisibility check could produce a runtime error later in model setup for edge-case GPU counts; this can be tightened if it becomes a real issue in practice.

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -188,8 +188,8 @@ class ModelConfig(BaseModelConfig):
     debug: DebugModelConfig = DebugModelConfig()
     """Debugging knobs for the model and distributed training."""
 
-    fused_lm_head_token_chunk_size: int | Literal["auto", "disabled"] = "disabled"
-    """Flattened token chunk size for the fused LM head. ``int >= 1`` sets the tokens per LM-head chunk explicitly; ``auto`` auto-enables (RL training picks 8192); ``disabled`` uses the vanilla LM head. Integer values aren't supported for SFT training."""
+    fused_lm_head_token_chunk_size: int | Literal["auto", "disabled"] = 1024
```
mikasenghaas, line 191: lol should we remove "auto" here? seems a bit silly

mikasenghaas (commented):
nice looks good!

mikasenghaas (approved):

PR #2883, opened 2026-06-25
feat: vLLM 0.23 bump + GLM-5.2 16-node P/D inference config (mooncake + llm-d)

S1ro1:
Rebases the vLLM **0.23.0** bump (from #2825) onto current `main` and adds a validated standalone disaggregated-P/D inference config for **GLM-5.2** (16 nodes, mooncake + llm-d).

### Commits
- `chore: bump vLLM to 0.23.0 and prune/adapt monkeypatches` (from #2825)
- `docs(inference): cite vllm#44814 …` (from #2825)
- `feat(configs): add GLM-5.2 16-node P/D mooncake + llm-d inference config`
- `fix(inference): keep Qwen3.5-MoE on 2D per-expert LoRA format`

### Monkeypatch verification (against installed 0.23.0 source)
The pruned/adapted patches were re-checked against the actual 0.23.0 wheel:
- ✅ `monkey_patch_deep_gemm_silu_mul_quant_int64` — removed. Upstream kernel (`fp8_utils.py`) already uses int64 pointer arithmetic; the patch was a no-op.
- ✅ `_patch_qwen35_lora` (dense GDN, vllm#36372) — removed. Fixed natively in 0.23.0 via the N-slice `MergedColumnParallelLinearWithLoRA` + `expand_packed_lora` (`lora/layers/column_parallel_linear.py`).
- ⚠️→✅ `Qwen3_5MoeForConditionalGeneration.is_3d_moe_weight = False` — this sub-part of `_patch_qwen35_lora` was **not** an upstream bug fix and was dropped by mistake in the initial prune. 0.23.0 defaults the flag to `True` (3D stacked-expert adapters), but our trainer emits the **2D per-expert** layout (`MultiLoRAGroupedExperts.state_dict_for_adapter`), which vLLM only reads when the flag is `False` (or `enable_mixed_moe_lora_format=True`). Without it, Qwen3.5-MoE LoRA adapters fail to load. Re-added as a minimal standalone patch in `fix(inference): keep Qwen3.5-MoE on 2D per-expert LoRA format`.
- ✅ `monkey_patch_minimax_m2_for_lora` — adapted to the new `GateLinear` gate. Verified `GateLinear(out_dtype=float32)` with no `params_dtype` yields a **bf16** weight (resolves to the model dtype under `set_default_torch_dtype`) so the LoRA Triton kernel's fp16/bf16 assertion passes, while router logits stay fp32.

### ⚠️ Required setup for disaggregated P/D (NIXL/UCX)
The pip-wheel NIXL's bundled UCX **segfaults on the prefill→decode KV transfer** (`signal 11: invalid permissions for mapped object` in `libucs.so`) — reproduced on vLLM 0.22 **and** 0.23, with/without mooncake, with/without llm-d. Fix is to build NIXL against **UCX 1.19.x** from source:
```bash
bash scripts/install_nixl_from_source.sh
uv pip install --reinstall --no-deps deps/nixl_cu12-*.whl
```
(Re-run after every `uv sync`, since the lock pins the wheel.) This should be promoted from "optional" to **required** for P/D in the install docs.

### Validation (real deployments, vLLM 0.23, H200)
- **GLM-5.2 16n** (8 prefill / 4 replicas + 8 decode), plus GLM-5.2 8n and GLM-5.1 8n — each a 1000-request / 500-concurrent burst → **0 errors, 0 faults, 0 RPC_FAIL** through llm-d + mooncake.
- Note: llm-d head-startup is a flaky race (`wait -n` + `--kill-on-bad-exit`); resubmit 1–3× to win it. vllm-router starts first-try.

Supersedes #2825 (same bump, rebased + GLM-5.2 config).

---

> [!NOTE]
> **High Risk**
> Core inference stack upgrade (vLLM 0.23) plus changes to env defaults, LoRA loading, and disaggregated P/D Docker paths affect production serving and RL weight sync; misconfiguration of NIXL/UCX or runner choice can cause crashes or NaN logits.
> 
> **Overview**
> Bumps **vLLM from 0.22.0 to 0.23.0** (`pyproject.toml`, pinned wheels, `uv.lock` transitive deps such as flashinfer 0.6.12 and compressed-tensors 0.17.0).
> 
> **Inference compatibility for 0.23:** `setup_vllm_env` sets `VLLM_USE_V2_MODEL_RUNNER=0` so padded-input scrub and NIXL routed-experts stay on the V1 runner. Disagg token serving imports `get_max_tokens` from vLLM’s new `api_utils` path. NemotronH reload-shim comments are updated for 0.23.
> 
> **Monkeypatches pruned/adapted:** Removes the custom DeepGEMM SiLU/mul FP8 Triton patch and the dense Qwen3.5 GDN LoRA shims (fixed upstream in 0.23). Keeps a minimal **Qwen3.5-MoE** override (`is_3d_moe_weight = False`) so trainer 2D per-expert LoRA adapters load. **MiniMax M2** gate fix switches to rebuilding `GateLinear` with bf16 weights and fp32 router logits.
> 
> **Docker:** Optional `INCLUDE_DISAGG_EXTRAS` (`--extra disagg` on `uv sync`) and `INCLUDE_NIXL_FROM_SOURCE` (build NIXL against baked UCX 1.19 to avoid P/D KV segfaults).
> 
> **New config:** `configs/glm5.2_16node_llmd/inference.toml` — GLM-5.2-FP8, 16-node disaggregated prefill/decode with mooncake offload and llm-d router.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d959d3c949e80f37d7ced6a0d35766b94b7a7c92. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/inference/patches.py
+++ b/src/prime_rl/inference/patches.py
@@ -15,6 +15,7 @@ def transformers_v5_compat():
         Qwen3VLMoeTextConfig.tie_word_embeddings = False
 
     _patch_lora_key_prefix()
+    _patch_qwen35_moe_lora_format()
     monkey_patch_nano_v3_reasoning_parser()
     monkey_patch_vllm_padded_input_scrub()
     monkey_patch_return_routed_experts_with_nixl_connector()
@@ -179,6 +180,29 @@ def monkey_patch_strip_routed_experts_from_chat():
     logger.info(
         "Stripped routed_experts from chat-completions responses (PD router merges only the /generate object form)."
     )
+
+
+def _patch_qwen35_moe_lora_format():
+    """Force Qwen3.5-MoE onto vLLM's 2D per-expert LoRA format.
+
+    vLLM 0.23.0 defaults ``Qwen3_5MoeForConditionalGeneration.is_3d_moe_weight = True``,
+    which makes the LoRA loader expect 3D stacked-expert adapters
+    (``base_layer.lora_{A,B}.weight`` / ``lora_{A,B}.weight``, experts folded into the
+    rank dim; see ``_stack_moe_lora_weights``). Our trainer instead emits the 2D
+    per-expert layout (``{expert_id}.gate_proj.lora_A.weight`` ...) from
+    ``MultiLoRAGroupedExperts.state_dict_for_adapter`` -- vLLM only consults that layout
+    when ``is_3d_moe_weight`` is False (or ``enable_mixed_moe_lora_format=True``).
+    Without this override the adapters fail to load with key/shape mismatches.
+
+    The rest of the old Qwen3.5 LoRA shim (the in_proj_qkvz packed-mapping fix and the
+    N-slice ``can_replace_layer`` / ``slice_lora_a`` generalizations for vllm#36372) is
+    handled natively by 0.23.0 and was dropped. Remove this too once we either adopt the
+    3D stacked save format (like gpt-oss) or start the engine with
+    ``enable_mixed_moe_lora_format=True``.
+    """
+    from vllm.model_executor.models.qwen3_5 import Qwen3_5MoeForConditionalGeneration
+
+    Qwen3_5MoeForConditionalGeneration.is_3d_moe_weight = False
 
 
 def _patch_lora_key_prefix():
--- a/src/prime_rl/inference/patches.py
+++ b/src/prime_rl/inference/patches.py
@@ -1,5 +1,4 @@
 import torch
-from vllm.triton_utils import tl, triton
 
 from prime_rl.inference.vllm.padded_input_scrub import monkey_patch_vllm_padded_input_scrub
 
@@ -15,10 +14,8 @@ def transformers_v5_compat():
     if not hasattr(Qwen3VLMoeTextConfig, "tie_word_embeddings"):
         Qwen3VLMoeTextConfig.tie_word_embeddings = False
 
-    _patch_qwen35_lora()
     _patch_lora_key_prefix()
     monkey_patch_nano_v3_reasoning_parser()
-    monkey_patch_deep_gemm_silu_mul_quant_int64()
     monkey_patch_vllm_padded_input_scrub()
     monkey_patch_return_routed_experts_with_nixl_connector()
     monkey_patch_kv_xfer_finished_tolerate_freed()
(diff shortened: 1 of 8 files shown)


hallerite (dismissed):

src/prime_rl/inference/patches.py
```diff
@@ -15,10 +14,8 @@ def transformers_v5_compat():
     if not hasattr(Qwen3VLMoeTextConfig, "tie_word_embeddings"):
         Qwen3VLMoeTextConfig.tie_word_embeddings = False
 
-    _patch_qwen35_lora()
     _patch_lora_key_prefix()
     monkey_patch_nano_v3_reasoning_parser()
-    monkey_patch_deep_gemm_silu_mul_quant_int64()
```
S1ro1, line 21: Why are we removing this patch, afaik it wasn't fixed?

configs/glm5.1_llmd_inference/inference.toml
```diff
@@ -0,0 +1,65 @@
+# GLM-5.1 inference-only deployment with llm-d P/D disaggregation.
```
S1ro1: What's the reason for 2 identical configs with just a separate model version

src/prime_rl/inference/patches.py
samsja: its is coming from @hallerite pr

src/prime_rl/inference/patches.py
```diff
@@ -805,13 +608,17 @@ def monkey_patch_minimax_m2_for_lora():
...
+        # vLLM 0.23.0 builds the gate as GateLinear with a float32 weight; rebuild it
+        # with a bf16 weight (model dtype) so the LoRA Triton kernel's float16/bfloat16
+        # assertion passes, keeping out_dtype=float32 so router logits stay fp32 (the
+        # GateLinear bf16xbf16->fp32 path).
+        self.gate = GateLinear(
             config.hidden_size,
             config.num_local_experts,
             bias=False,
-            quant_config=None,
+            out_dtype=torch.float32,
```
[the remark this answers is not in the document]
hallerite: unfounded

samsja (approved):

PR #2884, opened 2026-06-26
exp: temp-experimental K8s e2e for vLLM 0.23 P/D + mooncake + llm-d

JannikSt:
- Removes `job_id` / `jobId` from `HostedTrainingRunResponse` and the `prime rl` dispatch surface (JSON output and human print).
- `job_id` is a platform-internal scheduling id; the backend is dropping it from `POST /api/v1/training/runs` so clients can't infer the cluster from it.
- Beta-gated, no compat shims needed. `run_id` and `token_value` remain.


JannikSt:
Wrong repo — this belongs in pi-rft (hosted-rl). Moving.

PR #2895, opened 2026-06-28
feat(metrics): add effective_reward metric

mikasenghaas:
## What

Adds an **`effective_reward`** metric to the orchestrator's per-step W&B dict, logged as `effective_reward/{all,<env>}/{mean,max,min}`.

It is the same as the existing `reward` metric, but computed over **only the non-zero-advantage rollouts** — i.e. the reward as it would look if the zero-advantage filter were applied *before* the batch. A rollout counts when its advantage stream is non-zero (`r.advantages is not None and any(a != 0.0 ...)`), mirroring `ZeroAdvantageFilter`'s predicate. Per-example mean first (matching the plain `reward` metric), then aggregated across examples.

## Why

`reward/all/mean` includes whole groups that contribute no gradient (all rollouts earned the same reward → centered advantage collapses to zero). `effective_reward` shows the mean reward among the samples that actually drive learning, which tracks training signal more faithfully — useful for spotting when a run's reward is rising only on already-solved/already-failed groups.

## Notes

- Computed directly from the advantage stream, independent of filter config — so it reflects the would-be effective reward even when the zero-advantage filter is set to *monitoring* (not enforcing).
- Returns nothing (metric omitted) for a step where no rollout has a non-zero advantage.
- Self-contained: one file, `src/prime_rl/orchestrator/metrics.py` (+23 lines). Stable metric names, no change to existing keys.

---

> [!NOTE]
> **Low Risk**
> Observability-only change in metrics assembly; no training, filtering, or API behavior changes.
> 
> **Overview**
> Adds **`effective_reward`** to per-step W&B logging (`effective_reward/{all,<env>}/{mean,max,min}`), parallel to the existing `reward` stats but restricted to rollouts with a **non-zero advantage stream** (inverse of `ZeroAdvantageFilter`’s all-zero check).
> 
> Rollouts are tagged via a new `nonzero_advantage` column on the step results frame; `effective_reward_stats` averages reward per `group_id` over those rows, then aggregates mean/max/min across examples. Keys are **omitted** when every rollout has zero advantage. Existing metric names are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec83a51ec762f51087a1b74b7e31cb1e718883fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/metrics.py
+++ b/src/prime_rl/orchestrator/metrics.py
@@ -135,7 +135,11 @@ class MetricsBuilder:
 
         eff = effective_reward_stats(results_df)
         if eff is not None:
-            to_log["effective_reward/all/mean"], to_log["effective_reward/all/max"], to_log["effective_reward/all/min"] = eff
+            (
+                to_log["effective_reward/all/mean"],
+                to_log["effective_reward/all/max"],
+                to_log["effective_reward/all/min"],
+            ) = eff
 
         # Per-env metrics
         per_env_columns = [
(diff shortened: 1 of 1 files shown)


mikasenghaas:

PR #2906, opened 2026-06-29
[ENG-4306] fix(docker): repair CUDA image build broken by uv 0.11.25 centralized project envs

eexwhyzee:
## Summary

The CUDA image build (`Dockerfile.cuda`, `Docker Push` workflow) began failing on every push to `main` as of 2026-06-27, at the runtime-stage interpreter relink:

```
RUN rm /app/.venv/bin/python && ln -s /usr/local/bin/python /app/.venv/bin/python
rm: cannot remove '/app/.venv/bin/python': No such file or directory
```

It first appeared on #2746 (`0be5ac4`), but that commit is **not** the cause — its build was simply the first to run after `uv` auto-upgraded. This PR fixes the real root cause and hardens the build so it can't recur.

## Root cause

`Dockerfile.cuda` installs `uv` from an **unpinned** installer (`ADD https://astral.sh/uv/install.sh`), so every build pulls the latest `uv`. Image builds only run on push/release, so a single build can cross weeks of upstream drift.

- Last green build: `uv 0.11.24`
- First red build: `uv 0.11.25`

uv `0.11.25` shipped the **`centralized-project-envs`** preview feature. Combined with prime-rl's global `preview = true` in `[tool.uv]`, `uv sync` stopped materializing the environment at `<project>/.venv` and instead built it in the uv cache (`$UV_CACHE_DIR/environments-v2/<hash>`), leaving `.venv` as a **symlink into the cache**.

The runtime stage ships the env via `COPY --from=builder /app /app`. That copies the `.venv` *symlink*, but its target (the cache store) lives **outside `/app`** and isn't copied — so in the runtime image `/app/.venv` is a dangling symlink, `/app/.venv/bin/` resolves to nothing, and `rm` aborts with `ENOENT`. (It isn't only `bin/python` — the entire venv is absent from the runtime image.)

`preview = true` was originally added in #1772 solely to silence a `uv run` warning; it globally opts into uv's entire experimental surface, which is what silently activated `centralized-project-envs` when the version rolled over.

## Fix

1. **`UV_PROJECT_ENVIRONMENT=/app/.venv`** (builder stage) — forces uv to materialize a real in-project venv directory, overriding centralized/cached placement. This is the durable fix: it holds whether the feature stays preview-gated (today) or becomes the default (later).
2. **Drop `preview = true`** — removes the trigger and stops the project from auto-adopting future uv preview features. The `uv run` warning it originally silenced no longer fires on `uv >= 0.11.9`.

The `bin/python` relink lines are still required and left unchanged — they address a separate concern (the venv's interpreter symlinks point at the uv-managed Python, which is not shipped to the slim runtime image; the relink repoints them at the apt-installed `python3.12`).

## Verification

Reproduced the failure and validated the fix with a minimal multi-stage build mirroring `Dockerfile.cuda` (managed Python in builder, `COPY /app` to runtime):

| uv | `preview` | `UV_PROJECT_ENVIRONMENT` | `.venv` in builder | runtime relink |
|----|-----------|--------------------------|--------------------|----------------|
| 0.11.24 | true | unset | real dir | OK (matches last-green) |
| 0.11.25 | true | unset | symlink into cache | **fails** (reproduces the break) |
| 0.11.25 | true | `/app/.venv` | real dir | OK, `python` runs |
| 0.11.25 | false | unset | real dir | OK |

A full `Dockerfile.cuda` image build was also confirmed working with these changes.

---

> [!NOTE]
> **Medium Risk**
> Changes how Python deps are resolved and packaged in the CUDA image; a misconfigured venv would break runtime, but scope is build/Docker and uv config only.
> 
> **Overview**
> Repairs the **CUDA multi-stage Docker build** after `uv sync` stopped leaving a real `/app/.venv` in the image (symlink into the uv cache under `preview = true` + uv 0.11.25’s centralized env behavior), which broke runtime `COPY /app` and the old `bin/python` relink steps.
> 
> **`Dockerfile.cuda` (builder):** installs **system Python 3.12**, sets **`UV_PYTHON_PREFERENCE=only-system`** and **`UV_PROJECT_ENVIRONMENT=/app/.venv`** so the venv is built in-project against that interpreter; drops **`UV_PYTHON_INSTALL_DIR`**. **Runtime stage:** removes the **`rm`/`ln` hacks** for `.venv/bin/python*` because the venv should already point at system Python.
> 
> **`pyproject.toml`:** removes **`preview = true`** under `[tool.uv]` so the project no longer auto-enables uv preview features that moved the project env into the cache.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9fb258849dbca8775de393b2f66f2ce5630c20db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Dockerfile.cuda
+++ b/Dockerfile.cuda
@@ -62,6 +62,11 @@ RUN INSTALLER_NO_MODIFY_PATH=1 UV_INSTALL_DIR="/usr/local/bin" sh /uv-installer.
 ENV PATH="/usr/local/bin:$PATH"
 ENV UV_PYTHON_INSTALL_DIR="/usr/local/share/uv/python"
 ENV UV_CACHE_DIR="/usr/local/share/uv/cache"
+# Force a real in-project venv directory. uv's centralized project environments
+# (preview today, possibly default later) otherwise place the env in the uv cache
+# and make .venv a symlink into it, which the runtime stage's `COPY /app` can't
+# follow (the cache store lives outside /app), leaving the venv missing.
+ENV UV_PROJECT_ENVIRONMENT="/app/.venv"
 
 # Install Python dependencies (The gradual copies help with caching)
 WORKDIR /app
--- a/Dockerfile.cuda
+++ b/Dockerfile.cuda
@@ -64,15 +64,12 @@ ADD https://astral.sh/uv/install.sh /uv-installer.sh
 # Set install dir to location accessible to non-root users
 RUN INSTALLER_NO_MODIFY_PATH=1 UV_INSTALL_DIR="/usr/local/bin" sh /uv-installer.sh && rm /uv-installer.sh
 ENV PATH="/usr/local/bin:$PATH"
-# Build .venv against the system CPython (== the runtime stage's apt python3.12,
-# same Ubuntu 24.04 base => same 3.12.x), so the venv's recorded interpreter
+# Build .venv against the system CPython, so the venv's recorded interpreter
 # matches the runtime interpreter and no bin/python relink is needed at runtime.
 ENV UV_PYTHON_PREFERENCE=only-system
 ENV UV_CACHE_DIR="/usr/local/share/uv/cache"
-# Force a real in-project venv directory. uv's centralized project environments
-# (preview today, possibly default later) otherwise place the env in the uv cache
-# and make .venv a symlink into it, which the runtime stage's `COPY /app` can't
-# follow (the cache store lives outside /app), leaving the venv missing.
+# Explicity set venv dir inside /app so to ensure it gets copied over during
+# the runtime stage's `COPY /app`
 ENV UV_PROJECT_ENVIRONMENT="/app/.venv"
 
 # Install Python dependencies (The gradual copies help with caching)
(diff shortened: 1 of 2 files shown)


Dockerfile.cuda
```diff
@@ -62,6 +62,11 @@ RUN INSTALLER_NO_MODIFY_PATH=1 UV_INSTALL_DIR="/usr/local/bin" sh /uv-installer.
 ENV PATH="/usr/local/bin:$PATH"
 ENV UV_PYTHON_INSTALL_DIR="/usr/local/share/uv/python"
 ENV UV_CACHE_DIR="/usr/local/share/uv/cache"
+# Force a real in-project venv directory. uv's centralized project environments
```
mikasenghaas: afaiu this is not needed yet? only if the current preview feature is on stable release? if so, let's not add yet but only once it is in stbale release imo

eexwhyzee: i think it's still beneficial to include and [explicitly set the venv dir](https://docs.astral.sh/uv/reference/environment/#uv_project_environment) to ensure it makes it over during the runtime image `COPY /app` stage and avoid any other surprises

mikasenghaas (approved):

PR #2908, opened 2026-06-29
feat: adaptive concurrency control for rollout dispatch

### LINKED ISSUE #2902: feat: adaptive concurrency control for rollout dispatch
## Summary

The rollout dispatcher currently caps concurrency with a **fixed** `max_inflight_rollouts` limit. This static ceiling can't track the inference server's actual available compute, so we either under-utilize it (limit set too low → idle GPUs) or oversubscribe it (limit set too high → queue buildup, latency, off-policy staleness). We should make concurrency control **adaptive**: continuously adjust the in-flight rollout limit on the fly toward whatever the inference cluster can actually serve, so we always optimally use the available inference compute.

## Current behavior

Concurrency is governed by `max_inflight_rollouts` (orchestrator config):

- Defined at `packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py:541` — `max_inflight_rollouts: int | None = Field(None, ge=1)`. With `batch_size`, defaults to `batch_size * oversampling_factor`; required for token-based batching.
- Enforced in `RolloutDispatcher` (`src/prime_rl/orchestrator/dispatcher.py`) via a fixed permit counter:
  - `available_permits = self.max_inflight - self.inflight_permits` (`dispatcher.py:190-191`)
  - the dispatch loop stops scheduling when `available_permits <= 0` (`dispatcher.py:303`)
  - permits are acquired on dispatch and released on completion.

`self.max_inflight` is set once at construction and never changes for the lifetime of the run.

## Problem

A single static number can't be optimal across:

- **Heterogeneous / changing inference capacity** — number of inference engines, KV cache headroom, and per-engine throughput vary across runs and even within a run (P/D disaggregation, node failures, autoscaling).
- **Variable rollout cost** — sequence lengths, group sizes, and tool-call depth change effective load per rollout, so a count-based ceiling is a poor proxy for actual compute pressure.
- **Operator burden** — getting the value right today requires manual tuning per cluster/model; too low wastes GPUs, too high inflates queueing latency and off-policy staleness.

## Proposed solution

Replace (or wrap) the fixed permit ceiling with an **adaptive controller** that adjusts the effective in-flight limit at runtime based on observed inference performance — converging on the concurrency that saturates inference compute without oversubscribing it.

Sketch:

- **Signal**: derive a load/utilization signal from what the dispatcher and inference server already expose — e.g. rollout completion throughput, time-to-first-token / end-to-end rollout latency, queueing delay at the engine, or KV-cache utilization reported by vLLM. Rising latency / queue depth at flat throughput ⇒ oversubscribed; idle capacity ⇒ undersubscribed.
- **Controller**: a closed-loop policy that nudges the effective `max_inflight` up while throughput keeps improving and backs off when latency/queueing degrades without throughput gains. Classic options: AIMD (additive-increase/multiplicative-decrease), gradient/throughput-based congestion control (à la TCP Vegas / BBR), or a simple PID toward a target utilization.
- **Integration point**: make `available_permits` read from a dynamic `current_limit` instead of a constant `self.max_inflight`. The controller runs alongside the dispatch loop and updates `current_limit` periodically. Keep `max_inflight_rollouts` (and a new lower bound) as **clamps / safety rails** rather than the operating point.
- **Guardrails**: min/max bounds, smoothing to avoid oscillation, and interaction with off-policy constraints — note that raising concurrency increases `max_off_policy_steps` pressure (`dispatcher.py:260-291`), so the controller should account for staleness, not just inference throughput.
- **Observability**: log the effective limit, the load signal, and increase/decrease decisions so runs are debuggable and the controller is tunable.

## Open questions

- Which signal(s) are both cheap to collect and reliable across our inference backends (vLLM, P/D disaggregated, mooncake/llm-d)? Does the server already expose KV-cache utilization / queue depth we can consume directly?
- AIMD vs. PID vs. throughput-gradient — what converges fastest without oscillating under bursty rollout-length distributions?
- How should adaptive concurrency interact with token-based batching (where `max_inflight_rollouts` is currently *required* and load-bearing)?
- Should the controller optimize purely for inference utilization, or jointly for end-to-end training step time (balancing inference saturation against off-policy staleness)?

## Acceptance criteria

- Dispatcher concurrency adapts at runtime instead of using a fixed ceiling, keeping inference compute well-utilized without runaway queueing.
- `max_inflight_rollouts` continues to work as an upper clamp for backward compatibility.
- Effective limit and control decisions are logged/metricized.

mikasenghaas:
## Summary

Adaptive concurrency control for rollout dispatch (closes #2902). The in-flight rollout limit is no longer a fixed ceiling — an always-on closed-loop controller ramps it at runtime toward a target **GPU KV-cache utilization** (default 80%), so inference compute stays well-used without oversubscribing.

## Control law

KV usage is a **lagging signal**: an admitted rollout's KV footprint grows over its (possibly hour-long) lifetime, so committed load far exceeds the instantaneous reading, and ramping faster than it materializes overshoots. `decide_limit` (a gradient-gated AIMD) handles this:

- **Grow** only when KV is below target **and has settled** (small per-tick rise) **and** the limit is saturated — the settle gate paces the ramp to the rate at which in-flight KV materializes, converging from below. The step tapers to ~1 as it approaches target.
- **Back off** (multiplicative) on a KV high-water mark (default 95%) or rising preemptions (`vllm:num_preemptions`); lowers the ceiling and lets the in-flight set drain (never cancels).
- **Hold** otherwise (deadband avoids oscillation).

## Architecture

- `InferenceMetricsCollector` (`orchestrator/metrics/inference.py`) is the **single `/metrics` poller**. It runs always (decoupled from W&B), and each poll pushes a `ControlSignal` (GPU KV max + preemption rate on the decode engines) to the controller via `on_sample`, plus logs `inference/*` to W&B when enabled.
- `ConcurrencyController` (`orchestrator/concurrency.py`) is a pure consumer: it EWMA-smooths the signal, runs `decide_limit`, and calls `dispatcher.set_limit`. No poller, no separate task.
- Dispatcher: dynamic `current_limit` (operating point) within `[group_size, max_inflight_rollouts]`; ramps from `initial_inflight_rollouts` (default `group_size`).
- New `orchestrator/metrics/` subpackage: `metrics.inference` + `metrics.rollouts` (was `inference_metrics.py` / `metrics.py`).

## Config (`[orchestrator.concurrency]`, always on)

- `target_kv_usage` (0.8), `high_water_kv_usage` (0.95), `initial_inflight_rollouts` (ramp start, default `group_size`), `max_inflight_rollouts` (optional hard ceiling; **None = no clamp**, bounded only by KV), plus tick/EWMA/settle/growth/backoff/preemption knobs. The lower clamp is always `group_size` (not configurable) — a group-scoring group reserves that many permits at once.

## Breaking

- `orchestrator.max_inflight_rollouts` → **`orchestrator.concurrency.max_inflight_rollouts`** (optional ceiling, no longer required for token batching).
- **`orchestrator.oversampling_factor` removed** and the `batch_size * oversampling_factor` auto-derivation of `max_inflight_rollouts` is gone — the limit is controller-driven now, not derived from `batch_size`.
- Checked-in configs migrated; CHANGELOG updated.

## Observability

`dispatcher/current_limit` (the effective limit) + `inference/*` (`kv_cache_usage_mean/max`, `preemptions_per_second`, …); the console pipeline line shows `N/limit inflight rollouts`; controller decisions log at debug (`concurrency: grow A->B …`). No duplicated `concurrency/*` gauges.

## CPU KV offloading caveat

`target_kv_usage` tracks the **GPU** KV pool (`vllm:kv_cache_usage_perc` = `1 − free_gpu_blocks/total_gpu_blocks`). vLLM 0.23.0 exposes no occupancy metric for the CPU offload tier, so a GPU+CPU "total" can't be targeted directly — with CPU offloading enabled, raise `target_kv_usage` toward ~0.95 so the GPU runs hot enough to engage the tier, and rely on the preemption backoff. Documented on the config.

## Verification

- E2E (`reverse-text` RL, `--no-wandb`, driven by the collector): the limit ramped **16→24→35→52→77** off the live KV signal (starts at 16 = `group_size` floor), clean finish, no errors — confirming the collector polls and drives the controller without W&B.
- Full orchestrator unit suite passes (94); ruff/import/config-parse clean.


mikasenghaas:
Superseded by a design-first approach on a fresh branch of the same name: feedforward token-budget admission + binary-signal AIMD trim, instead of the KV-target gradient controller here. New draft PR will cross-link.

mikasenghaas:
Superseded by #3285 (design doc for the feedforward token-budget approach).

PR #2921, opened 2026-07-01
feat: bump vLLM 0.23.0 -> 0.24.0

mikasenghaas:
## Purpose

Bump the pinned vLLM runtime **0.23.0 → 0.24.0**. Base-wheel bump only; HiSparse fork work is tracked separately.

## Changes

**Dependencies** — `vllm>=0.24.0` + v0.24.0 wheel URLs (cu129, cp38-abi3, x86_64/aarch64). Lock regenerated: only 4 version moves (`vllm`, `humming-kernels 0.1.6`, `prometheus-fastapi-instrumentator 8.0.2`, `starlette 1.3.1`), `gguf` dropped, torch stays 2.11.0. The large lock diff is uv re-simplifying platform markers.

**Weight sync — 0.24 MoE refactor.** `FusedMoE` became a factory returning `MoERunner`, with the expert map on `.routed_experts` and EPLB state on `.router`; expert params are now nested (`…experts.routed_experts.w13_weight`) while the trainer broadcasts flat names.
- `build_expert_map` rewritten for `MoERunner`; `load_weights_kernel` wraps the incoming stream in upstream's `remap_moe_expert_weights` compat shim (reached only with `quantize_in_weight_transfer=true`, i.e. GLM-5.x llmd).
- NemotronH `mixer.D` reload shim deleted — vllm#44814 is in 0.24.0 (`reload/meta.py` caps `copied_numel` at `param.numel()`, fixing the `composed_weight_loader` double-count generically).

**LoRA reload — native `load_inplace`.** Dropped `monkey_patch_load_lora_adapter` + `monkey_patch_LRUCacheWorkerLoRAManager`: 0.24's `add_adapter(load_inplace=True)` is our old patch upstreamed (same-name reload reuses `lora_int_id`, workers re-read from disk). The `/load_lora_adapter` wrapper forces the flag and resets it on the stored `LoRARequest` afterwards — left `True` it would force a disk reload every scheduler step. Also deleted `monkey_patch_skip_lora_module_warnings` (the loop it skipped was removed upstream before 0.23; the stale copy was silently reverting newer `_load_adapter` behavior). `_patch_lora_key_prefix` re-derived from the 0.24 original (vllm#38522 closed unmerged → permanent); only the dual-form `.experts` name acceptance is grafted in.

**0.24 behavior guards** (from a full source audit of the release against every prime-rl touchpoint — these change RL semantics without any prime-rl code being "broken"):
- `monkey_patch_qwen3_coder_param_newline_trim`: 0.24's parser engine full-`.strip()`s tool parameter values; 0.23 trimmed exactly one newline each side. A full strip corrupts whitespace-significant args (str_replace `old_str`/`new_str`, indented file content) → silent reward corruption on agentic runs. Restores 0.23 semantics; covers `nemotron_v3` (reuses `qwen3_config`).
- `monkey_patch_minimax_m2_think_end_passthrough`: 0.24's minimax engine deletes the literal `</think>` from content on tool turns and strips content when tools are present, breaking `minimax_m2_append_think` round-tripping. Drops the transition and the strip.
- `VLLM_ENFORCE_STRICT_TOOL_CALLING=0` defaulted in `setup_vllm_env`: 0.24 flipped it to True — grammar-constrained decoding for required/named/strict tools samples a distribution the trainer never sees.
- `temperature ≤ 2.0` on sampling configs, mirroring 0.24's new request validation (fail at config load, not per-rollout 400s).

**Patch hygiene.** `monkey_patch_no_moe_lora` wraps upstream `__post_init__` instead of copying it; MiniMax-M2 `forward` patch deleted (double all-reduce at TP>1 — `MoERunner` reduces internally); dead transformers-v5 shim removed and plugin entry point renamed `transformers_v5_compat` → `apply_shared_vllm_patches`. All 10 remaining monkeypatches audited against 0.24 source: targets exist, copied code matches, still needed (the harmony stop-token patch is a verified no-op at both versions — removal candidate).

## Validation

- **Source audit**: 65-agent adversarially-verified audit against the v0.23.0/v0.24.0 trees. Every vLLM import/env-var/RPC-string/endpoint prime-rl uses resolves at 0.24 with matching semantics. The **V1 sampler + logprob path is byte-identical** across the bump (all `v1/sample` diffs are pin-memory/ROCm/spec-decode-only), and both vLLM tags pin the identical DeepGEMM commit (`891d57b`), keeping the side wheels valid.
- `uv sync --all-extras --locked` clean; ruff clean; `tests/unit` pass.
- Parser patches verified behaviorally against the installed 0.24 wheel: stock turns `'\n    indented line\n\n'` into `'indented line'`; patched yields `'    indented line\n'` — exact 0.23 semantics on full and streaming-partial paths. `SamplingParams` bound confirmed (`temperature must be in [0, 2]`).
- **LoRA RL smoke** (2× RTX PRO 6000, `reverse_text_lora`, 15 steps): reward 0.15→0.72, exactly one adapter load per step, zero errors.
- **MoE RL smoke** (`reverse_text_moe`, mini-glm-moe, 20 steps): 18 layerwise reloads, no NaNs, zero errors (`moe_backend=triton` — pre-existing SM 12.0 FlashInfer limitation).
- **H200 cluster validation (SM90, the prod arch)**:
  - deep-gemm/deep-ep import smoke: all 17 `*_impl` symbols (incl. the 4 new 0.24 ones) resolve from the pinned external wheels; `deep_ep` imports clean.
  - **Reload-invariance smokes** (server boot with CUDA graphs + V1 forced + plugin, greedy logprobs, `/update_weights` layerwise reload from the HF snapshot, logprobs re-checked): dense Qwen3-0.6B and mini-glm-moe both **bitwise-identical logprobs after reload** (max delta 0.0) — covers the MoERunner weight-sync rewrite end-to-end.
  - Boot matrix on H200: 0.24 V1+graphs, V2+graphs, eager, plugin on/off, and 0.23 control all boot and generate. (Early runs hit silent EngineCore deaths that turned out to be node-level job contention, reproducing on neither version once runs were isolated.)
  - `pytest tests/unit -m gpu`: 70/74 pass. The 4 failures (`test_qwen3_5_moe*` tilelang kernels, `tvm.error.InternalError: Get different layout for b_dq` at sm_90a) reproduce **identically on main/0.23 across multiple nodes** — pre-existing trainer-side tilelang issue on H100/H200, unrelated to this bump; needs its own issue.

## Post-merge TODO (not reachable on a 2-GPU box)

- [ ] **fp8 kernel weight transfer** (any `quantize_in_weight_transfer=true` deployment): ≥2 updates land with no skipped/shape-mismatch errors and sane post-update rewards; exercises `update_mla_absorbed_weights` too.
- [ ] **KL spot check for GLM-5.x and Qwen3.5/Qwen3-Next** (one weight sync + trainer-vs-inference KL): 0.24 changes inference numerics on these archs — GLM-5 DSA sparse-attention accuracy fix (#36616) and a default-on fused qk-rmsnorm-rope-gate kernel for Qwen3.5/Next (#44176, no env gate). Self-consistent, but expect a small KL shift vs the trainer.
- [ ] **EP + EPLB** run (once each with `enable_eplb` off/on): wrong mappings raise loudly or show as post-update reward collapse.
- [ ] **Multi-node NCCL broadcast** incl. NemotronH (shim deleted per the 0.24 fix): finite generation after ≥2 updates.
- [ ] **P/D disaggregated stack** (NIXL + mooncake + llm-d): eval rollouts, router replay, abort-storm KV-transfer tolerance, cold DP-coordinator startup.
- [ ] Optional: quantify accepted parser shifts offline (0.24 silently drops hallucinated tool names and emits partial tool calls on length truncation) by re-parsing saved 0.23 rollouts.

## Follow-ups (pre-existing on main, not blocking)

- `NativeKVCacheOffloadConfig` emits tier type `fs_python` — not registered in any vLLM version (should be `fs`); native cpu+disk offload fails at startup.
- `InferenceConfig.rope_scaling` matches no `EngineArgs` field — silently ignored.
- `VLLM_ENABLE_MOE_DP_CHUNK` (glm5.2 config + docs) exists in no vLLM version — no-op.
- `scripts/install_deep_gemm.sh` defaults to a stale DeepGEMM ref (`d30fc36` vs pinned `891d57b`).
- Consider engine-level `enable_mixed_moe_lora_format=True` over `_patch_qwen35_moe_lora_format`, and dropping the no-op harmony patch.

---

> [!NOTE]
> **High Risk**
> Touches inference runtime version, RL rollout semantics (tool parsers, strict tool calling, temperature bounds), MoE/NCCL weight transfer, and per-step LoRA reload—any regression affects training fidelity or large disaggregated deployments.
> 
> **Overview**
> Pins **vLLM 0.24.0** (`pyproject.toml`, GLM config comment) and renames the general plugin to **`apply_shared_vllm_patches`**, dropping the transformers-v5 shim while adding **0.24-specific parser patches** (Qwen3 coder single-newline trim, MiniMax-M2 `</think>` passthrough) plus defaults **`VLLM_ENFORCE_STRICT_TOOL_CALLING=0`** and **`temperature ≤ 2.0`** on orchestrator sampling configs.
> 
> **Weight sync** adapts to 0.24’s **`MoERunner`** layout: `build_expert_map` reads `routed_experts` / `router`, and kernel loads wrap the stream in **`remap_moe_expert_weights`**. The NemotronH **`mixer.D`** reload workaround is removed (fixed upstream in 0.24).
> 
> **LoRA**: removes several monkeypatches in favor of native **`load_inplace`** on `/load_lora_adapter` (set then reset on the stored request); **`_patch_lora_key_prefix`** is re-based on 0.24’s loader with dual-form expert name acceptance. Worker-side LRU/skip-warning/load-adapter patches are deleted.
> 
> **Hygiene**: MiniMax-M2 drops the redundant `forward` patch; `monkey_patch_no_moe_lora` delegates to upstream `__post_init__`; DP coordinator timeout comment updates to 120s baseline.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a2d08b31817445a5756a9b7f1f31f81cbc6d846. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


samsja (dismissed):

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -50,8 +50,8 @@ class ModelConfig(BaseModelConfig):
 
 
 class TrainSamplingConfig(BaseConfig):
-    temperature: float = Field(1.0, ge=0)
-    """Sampling temperature."""
+    temperature: float = Field(1.0, ge=0, le=2.0)
+    """Sampling temperature. vLLM 0.24 rejects requests with temperature above 2.0."""
```
S1ro1: can we remove this comment? its useless and very ai vibed

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -92,8 +92,9 @@ def _deprecate_max_tokens(cls, data: Any) -> Any:
 
 
 class EvalSamplingConfig(BaseConfig):
-    temperature: float | None = Field(None, ge=0)
-    """Sampling temperature. None defers to the inference server default."""
+    temperature: float | None = Field(None, ge=0, le=2.0)
+    """Sampling temperature. None defers to the inference server default. vLLM 0.24
+    rejects requests with temperature above 2.0."""
```
S1ro1: likewise

tests/unit/inference/test_padded_input_scrub.py
```diff
@@ -103,6 +103,25 @@ def test_zero_padded_model_inputs_zeroes_embeddings_and_mrope_positions():
     ]
 
 
+def test_general_plugin_entry_point_loads_and_applies():
```
S1ro1: lets remove

tests/integration/test_reverse_text_moe.py
```diff
@@ -48,3 +48,11 @@ def test_no_error(rl_process: ProcessResult, output_dir: Path):
 
 def test_moe_runs(rl_process: ProcessResult, test_no_error):
     """MoE RL with custom model impl completes without error."""
+
+
+def test_moe_mismatch_kl_in_range(rl_process: ProcessResult, test_no_error, output_dir: Path):
```
S1ro1: unrelated to this

tests/unit/inference/test_weight_transfer_kernel.py
```diff
@@ -0,0 +1,58 @@
+import pytest
```
S1ro1: also feel like these tests test nothing

src/prime_rl/inference/vllm/server.py
```diff
@@ -97,11 +93,26 @@ async def update_weights(request: Request):
...
+    every generation request; left ``True`` it would force a disk reload on each
+    scheduler step. The orchestrator awaits this endpoint before dispatching
+    rollouts for the new version, so the reset always lands before generation.
+    """
     handler = models(raw_request)
+    lora_request.load_inplace = True
     response = await handler.load_lora_adapter(lora_request)
     if isinstance(response, ErrorResponse):
         return JSONResponse(content=response.model_dump(), status_code=response.error.code)
+    handler.lora_requests[lora_request.lora_name].load_inplace = False
```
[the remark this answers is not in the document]
faresobeid: Stale — fixed in 7d8ddbd4e: the reset now runs unconditionally before the ErrorResponse check (see the current wrapper + docstring).

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
faresobeid: Removed in 0c0bc7bc7 (bound kept).

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
faresobeid: Removed in 0c0bc7bc7.

tests/unit/inference/test_padded_input_scrub.py
faresobeid: Removed in 0c0bc7bc7.

tests/integration/test_reverse_text_moe.py
faresobeid: Reverted in 0c0bc7bc7 — can send it as a separate PR if there's interest.

tests/unit/inference/test_weight_transfer_kernel.py
faresobeid: Dropped in 0c0bc7bc7. FWIW the remap/EP-slice path was also validated end-to-end on an H200 (server boot + /update_weights reload with bitwise-identical logprobs), so coverage doesn't hinge on these.

hallerite (approved):

PR #2922, opened 2026-07-01
fix: weight-sync hardening + dead-code / get_max_layer_num cleanup

mikasenghaas:
## Summary

Correctness fixes and low-risk cleanup, concentrated in the weight-sync path, plus two duplication consolidations. Commits:

### `fix:` weight-broadcast validation + prefix-cache docs
- **`validate_shared_weight_broadcast` bug**: the check was a Python chained comparison — `trainer.type != orchestrator.type != inference.type` — which only compares adjacent pairs. So `trainer=nccl, orchestrator=nccl, inference=filesystem` passed validation silently. Now checks trainer↔orchestrator, then inference↔both, with clearer messages.
- **Prefix-cache docs**: the `update_weights` / `load_lora_adapter` docstrings claimed the server resets the prefix cache on weight update. It doesn't — the orchestrator salts the prefix cache per weight version (`cache_salt`, see `orchestrator/envs.py`), so stale KV is never reused. Corrected the docs and removed the dead `reset_prefix_cache_after_update` state flag (set but never read).
- Heartbeat docstring said "2s timeout"; the request uses `timeout=1`.

### `chore:` remove dead code (grep-verified unused)
`load_balancing_loss_func` (qwen3_moe), `shift_logits` (rl/loss), `_count_trainable_params` (perf), `align()` (fp8_utils), a dead `_weight_scale` assignment (fp8_indexer), a commented-out `forward` variant (layers/moe), and `capitalize` / `to_row_format` / `mean_normalize` (utils). Also deduped `NCCL_READY_MARKER` (import from `utils.client` instead of redefining it in `broadcast/nccl.py`).

### `refactor:` consolidate `get_max_layer_num` into `trainer/weights`
Seven `converting_*.py` modules each defined their own copy — six as a fragile `int(key.split(".")[2])` one-liner, and `afmoe` as a regex variant returning the max **index** (callers compensated with `range(max_layer + 1)`). All replaced with the canonical, robust implementation already in `trainer/weights.py`. For afmoe the two callers switch to `range(num_layers)` — behavior-identical since the canonical returns `max+1`. Verified no import cycle.

### `refactor:` centralize flash-attn varlen import guard
`afmoe` and `qwen3_5_moe` each re-declared the identical three-way `try/except` import guard for `flash_attn_varlen_func` / `_3_` / `_4_` that already lives in `trainer/models/layers/attn.py`. They now import the (None-able) module-level names from there; each model's `_funcs` dispatch table is unchanged. (Their attention classes are genuinely model-specific — gated output, MRoPE — so they legitimately do *not* subclass the shared `FlashAttention`; only the optional-dependency guard was duplicated.)

## Verification
Every edited file compiles (`py_compile`) and passes `ruff check`; deletions leave no orphaned imports or bad spacing; the `NCCL_READY_MARKER` / `get_max_layer_num` / flash-attn import paths are cycle-free by static analysis. The full test suite was **not** run in the authoring environment (`uv sync` is blocked there by an unchecked-out submodule, unrelated to these changes). Reviewer should run the trainer/config unit tests before merge.

## Deliberately out of scope
- The broad `except Exception` in the multi-run broadcast backends was investigated and left as-is: it's per-run fault isolation, and the `STABLE`-marker handshake means a failed broadcast blocks the orchestrator rather than silently serving stale weights.
- Higher-risk refactors flagged in the audit (`converting_*` MoE weight-conversion dedup — blocked by 4 structural outliers: `minimax_m2`/`afmoe`/`nemotron_h`/`gpt_oss`; `PrimeMonitor`→`BackgroundAsync`) are left for follow-ups that warrant test coverage first.

---

> [!NOTE]
> **Medium Risk**
> The validation fix changes which configs fail at startup (desirable) but could break previously accepted misconfigured RL runs; weight-sync semantics are documented, not altered.
> 
> **Overview**
> Fixes **weight-broadcast config validation** so trainer, orchestrator, and inference must all agree on broadcast type—replacing a chained `!=` comparison that could let mismatched inference backends (e.g. NCCL vs filesystem) pass silently.
> 
> **Prefix-cache behavior** is documented accurately: weight update / LoRA load no longer claim the vLLM server resets the prefix cache; staleness is avoided via orchestrator `cache_salt` per weight version. Removes the unused `reset_prefix_cache_after_update` app state flag and corrects the heartbeat timeout doc (1s, not 2s).
> 
> **Refactors and cleanup:** `get_max_layer_num` lives in `trainer/conversion_utils.py` and replaces seven duplicate implementations in MoE `converting_*` modules; flash-attn varlen optional imports are centralized from `layers/attn` in afmoe and qwen3_5_moe; `NCCL_READY_MARKER` is imported from `utils.client` in NCCL broadcast. Removes grep-verified dead code (`load_balancing_loss_func`, `shift_logits`, unused utils helpers, etc.) with no behavior change to active paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7cacc21a9d4aae587d876dd973d548494f58ab1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/trainer/models/afmoe/converting_afmoe.py
```diff
@@ -122,26 +124,15 @@ def convert_tt_layer_to_hf(state_dict: dict[str, Tensor], layer_idx: int) -> Non
...
-            max_layer = max(max_layer, layer_idx)
-    return max_layer
-
-
 def convert_hf_to_tt_moe(state_dict: dict[str, Tensor]) -> None:
     """Convert all MoE layers from HF format to TT format in-place."""
-    max_layer = get_max_layer_num(state_dict)
-    for layer_idx in range(max_layer + 1):
+    num_layers = get_max_layer_num(state_dict)
+    for layer_idx in range(num_layers):
```
samsja, line 130: wait here we remove the + 1 isn't it wrong ?

mikasenghaas: the +1 is in `get_max_layer_num`

mikasenghaas (commented):
lgtm

faresobeid (approved):

PR #2923, opened 2026-07-01
refactor: flash-attn import guard + build_scope_metrics dedup (stacked on #2922)

mikasenghaas:
> **Stacked on #2922** (`fix/weight-sync-hardening`). Base branch is that PR, not `main`. **Merge #2922 first**, then this retargets to `main` and can merge.

## Summary

Follow-up refactors from the same audit as #2922. Scoped to changes that are **safe to verify by inspection**, since the test suite can't run in this environment (`uv sync` is blocked by the unchecked-out `deps/renderers` submodule). Two commits:

### `refactor:` centralize flash-attn varlen import guard
`afmoe` and `qwen3_5_moe` each re-declared the identical three-way `try/except` import guard for `flash_attn_varlen_func` / `_3_` / `_4_` that already lives in `trainer/models/layers/attn.py`. They now import the (None-able) module-level names from there; each model's `_funcs` dispatch table is unchanged.

> Note: the audit suggested these two could *subclass* the shared `FlashAttention`. On inspection they **can't** — both have genuinely model-specific attention (gated output via `gate_proj`, local/global split, MRoPE, GatedDeltaNet). Only the optional-dependency import guard was actually duplicated, so that's all this touches. The numerically-sensitive `_compute_attention`/`forward` are left alone.

### `refactor:` extract `_scope_value` helper in `build_scope_metrics`
The prefill/decode/aggregate branching was duplicated verbatim for the `throughput` and `avg_time_seconds` metrics (differing only in the aggregate — `sum` vs mean). Extracted `_scope_value(scope, prefill, decode, aggregate)`; both call sites collapse to three lines. Behavior-identical (verified case-by-case, including the empty-values guard).

## Verification
`ruff check` passes and all edited files compile. As above, the unit tests were **not** run here — a reviewer should run them (CI should be green on these files).

## Deliberately deferred (need a test run — not done here)
The remaining audit items turned out to be riskier or less applicable than first framed; I did **not** attempt them untested:

- **`PrimeMonitor` → `BackgroundAsync`**: not a clean mechanical swap — PrimeMonitor interleaves monitor-specific state (`enabled` gating, `_pending_sample_steps.clear()` on fork, direct `self._client` use in the sample-upload paths) with the generic async machinery. Doable, but wants test coverage.
- **`converting_*` MoE weight-conversion core dedup**: the per-family expert stack/unstack genuinely differs (e.g. afmoe uses per-expert regex keys; others use the fused gate_up format). This is correctness-critical weight math — unsafe to unify without a conversion round-trip test.
- **SFT/RL grad-rescale unify**: `sft/train.py` and `rl/train.py` compute *different* scalings on purpose; not a safe mechanical dedup.

Happy to pick up `PrimeMonitor` and the `converting_*` dedup as further commits here once there's a way to run the suite.

### DIFF
--- a/src/prime_rl/orchestrator/inference_metrics.py
+++ b/src/prime_rl/orchestrator/inference_metrics.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import time
 from collections import deque
+from collections.abc import Callable
 from dataclasses import dataclass, field
 
 import wandb
@@ -256,6 +257,22 @@ def mean(values: list[float]) -> float | None:
     return sum(values) / len(values)
 
 
+def _scope_value(
+    scope: str,
+    prefill_value: float | None,
+    decode_value: float | None,
+    aggregate: Callable[[list[float]], float],
+) -> float | None:
+    """Pick a scope's metric value: prefill/decode use their own value; the
+    aggregate scope combines both non-None values via ``aggregate`` (None if neither)."""
+    if scope == "prefill":
+        return prefill_value
+    if scope == "decode":
+        return decode_value
+    values = [v for v in (prefill_value, decode_value) if v is not None]
+    return aggregate(values) if values else None
+
+
 def build_scope_metrics(
     scope: str,
     samples: list[EndpointSample],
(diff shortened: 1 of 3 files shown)


mikasenghaas:
Folded the flash-attn import-guard commit into #2922; dropped the build_scope_metrics/_scope_value change per review. Consolidating into a single PR.

PR #2926, opened 2026-07-01
feat: optional per-node uv sync for multi-node RL

samsja:
Add a SlurmConfig.per_node_venv_sync flag (default False). When enabled, the multi-node RL sbatch runs uv sync on every node via srun (one task per node, blocking) instead of only the batch node.

This is needed when the venv is node-local (e.g. UV_PROJECT_ENVIRONMENT on /tmp): a single batch-node sync would leave the other nodes without a venv. The srun task re-sources .env so uv sees the node-local UV_PROJECT_ENVIRONMENT / UV_CACHE_DIR even if SLURM env propagation is disabled. With set -e, a failed sync on any node aborts the job. Default keeps the original head-only behavior for shared (NFS) venvs.

---

> [!NOTE]
> **Low Risk**
> Launcher-only SLURM bootstrap change with default preserving existing head-only sync; risk is mainly mis-setting shared_fs on clusters with local venvs.
> 
> **Overview**
> Multi-node SLURM jobs can now distinguish **shared** vs **node-local** Python environments via **`SlurmConfig.shared_fs`** (default **`True`**), exposed in **`template_vars`** for all bundled sbatch templates.
> 
> When **`shared_fs` is `False`**, **`inference`**, **`multi_node_rl`**, and **`multi_node_sft`** run **`srun --ntasks-per-node=1 … uv sync --all-extras`** so each node builds its own venv (e.g. **`UV_PROJECT_ENVIRONMENT`** on **`/tmp`**). The srun task **re-sources `.env`** so node-local uv settings are not lost across srun; with **`set -e`**, any failed sync aborts the job. When **`shared_fs` is `True`**, behavior is unchanged: a single **`uv sync`** on the batch node for NFS/shared venvs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 660de820c06e8a0f1c255d671ffdc92f9990aa54. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/shared.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/shared.py
@@ -58,8 +58,8 @@ class SlurmConfig(BaseConfig):
     cleanup_grace_period: int = Field(3600, ge=0)
     """Seconds to wait before tearing down a multi-node RL job that hit a non-zero exit, letting in-flight checkpoints flush. Set to 0 to tear down immediately."""
 
-    per_node_venv_sync: bool = False
-    """Run ``uv sync`` on every node via srun (one task per node) instead of only the batch node. Enable when the venv is node-local (e.g. ``UV_PROJECT_ENVIRONMENT`` on ``/tmp``); leave off for a shared (NFS) venv where a single sync suffices."""
+    shared_fs: bool = True
+    """Whether the project filesystem (including the venv) is shared across nodes (e.g. NFS). When True, a single ``uv sync`` on the batch node suffices. Set to False when the venv is node-local (e.g. ``UV_PROJECT_ENVIRONMENT`` on ``/tmp``) so ``uv sync`` runs on every node via srun."""
 
     @property
     def template_vars(self) -> dict:
@@ -74,7 +74,7 @@ class SlurmConfig(BaseConfig):
             "time": self.time,
             "pre_run_command": self.pre_run_command,
             "cleanup_grace_period": self.cleanup_grace_period,
-            "per_node_venv_sync": self.per_node_venv_sync,
+            "shared_fs": self.shared_fs,
         }
 
     @model_validator(mode="after")
--- a/packages/prime-rl-configs/src/prime_rl/configs/shared.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/shared.py
@@ -58,6 +58,9 @@ class SlurmConfig(BaseConfig):
     cleanup_grace_period: int = Field(3600, ge=0)
     """Seconds to wait before tearing down a multi-node RL job that hit a non-zero exit, letting in-flight checkpoints flush. Set to 0 to tear down immediately."""
 
+    per_node_venv_sync: bool = False
+    """Run ``uv sync`` on every node via srun (one task per node) instead of only the batch node. Enable when the venv is node-local (e.g. ``UV_PROJECT_ENVIRONMENT`` on ``/tmp``); leave off for a shared (NFS) venv where a single sync suffices."""
+
     @property
     def template_vars(self) -> dict:
         """Common template variables for all SLURM templates."""
@@ -71,6 +74,7 @@ class SlurmConfig(BaseConfig):
             "time": self.time,
             "pre_run_command": self.pre_run_command,
             "cleanup_grace_period": self.cleanup_grace_period,
+            "per_node_venv_sync": self.per_node_venv_sync,
         }
 
     @model_validator(mode="after")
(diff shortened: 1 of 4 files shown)


src/prime_rl/templates/multi_node_rl.sbatch.j2
```diff
@@ -144,7 +144,17 @@ echo "MASTER_PORT=${MASTER_PORT}"
 cd $PROJECT_DIR
 [ -f .env ] && source .env
 source .venv/bin/activate
+{% if per_node_venv_sync %}
+# Node-local venv (e.g. UV_PROJECT_ENVIRONMENT on /tmp): a bare `uv sync` would only
+# build the batch node, so srun runs it on every node (one task each) and blocks until
+# all finish. `set -e` above aborts the job if any node's sync fails. Re-source .env
+# inside the task so uv reads the /tmp paths even if SLURM env propagation is disabled;
+# otherwise uv would silently fall back to the shared .venv.
+srun --ntasks-per-node=1 bash -c 'cd $PROJECT_DIR && { [ -f .env ] && source .env; } && uv sync --all-extras'
```
hallerite: I think this fails if there is no `.env`, which is stricter than the rest of the script, no?

packages/prime-rl-configs/src/prime_rl/configs/shared.py
```diff
@@ -58,6 +58,9 @@ class SlurmConfig(BaseConfig):
     cleanup_grace_period: int = Field(3600, ge=0)
     """Seconds to wait before tearing down a multi-node RL job that hit a non-zero exit, letting in-flight checkpoints flush. Set to 0 to tear down immediately."""
 
+    per_node_venv_sync: bool = False
```
hallerite: if this is in shared, shouldn't the multi-node sft job also honor this?

src/prime_rl/templates/multi_node_rl.sbatch.j2
```diff
@@ -144,7 +144,17 @@ echo "MASTER_PORT=${MASTER_PORT}"
 cd $PROJECT_DIR
 [ -f .env ] && source .env
 source .venv/bin/activate
+{% if per_node_venv_sync %}
+# Node-local venv (e.g. UV_PROJECT_ENVIRONMENT on /tmp): a bare `uv sync` would only
+# build the batch node, so srun runs it on every node (one task each) and blocks until
+# all finish. `set -e` above aborts the job if any node's sync fails. Re-source .env
+# inside the task so uv reads the /tmp paths even if SLURM env propagation is disabled;
```
hallerite: isn't it rather because otherwise unexported vars in .env don't cross srun even with normal propagation instead of slurm propagation being disabled?

hallerite (commented):
should probably also update `docs/scaling.md` then?

packages/prime-rl-configs/src/prime_rl/configs/shared.py
mikasenghaas: also inference?

mikasenghaas: maybe `shared_fs: bool = True`. then we can reuse this as well for other stuff potentially?

src/prime_rl/templates/multi_node_rl.sbatch.j2
samsja: good point

packages/prime-rl-configs/src/prime_rl/configs/shared.py
samsja: true

src/prime_rl/templates/multi_node_rl.sbatch.j2
samsja: triue

samsja:
Addressed all review comments:

- **Renamed `per_node_venv_sync` to `shared_fs`** (default `True`) with inverted logic, per @mikasenghaas's suggestion. `shared_fs=True` means NFS/shared venv (single sync on batch node); `shared_fs=False` means node-local venv (srun sync on every node).
- **Fixed `.env` sourcing** in the srun command — added `|| true` guard so it doesn't fail when `.env` is absent, matching the rest of the script (@hallerite's point).
- **Fixed the comment** — re-sourcing `.env` inside the srun task is needed because unexported vars in `.env` don't cross `srun` even with normal SLURM env propagation, not because propagation is disabled (@hallerite's point).
- **Extended `shared_fs` to `multi_node_sft.sbatch.j2` and `inference.sbatch.j2`** since the flag lives in `shared.py` and should apply to all multi-node jobs, not just RL (@hallerite, @mikasenghaas).

mikasenghaas (approved):

hallerite (approved):

PR #2933, opened 2026-07-02
feat: auto-create curated W&B overview view per project

mikasenghaas:
## Summary

RL runs log a *lot* of W&B metrics: every rollout metric is emitted per subset (`all`/`effective`), per stat (`mean`/`max`/`min`/`p10`/`p90`), and per env plus a cross-env aggregate. A real ablation project logs **thousands** of series (see below), so the default auto-generated workspace buries the ~dozen metrics you actually watch — and today you hand-pick an "overview" section by hand in every new project. This PR auto-creates a curated `overview` saved view instead. See it live: [gsm8k-overview-demo](https://wandb.ai/primeintellect/gsm8k-overview-demo?nw=sv1peinym1v) (single env), [search-ablations](https://wandb.ai/primeintellect/search-ablations?nw=hfsan35zwvb) (5 train + 2 eval), [terminal-ablations](https://wandb.ai/primeintellect/terminal-ablations?nw=tvja0s187ij).

The `overview` view is created automatically on the first run into a project:

- **Sections**: `train` (per-env, plus `train/agg` when >1 env), `eval/<env>`, `stability`, `performance`. Train and eval show the **same metrics** (reward, has_error, is_truncated, num_total/turns/branches) — only the reward metric differs (`reward/mean` for train, `avg@k` for eval) — and all on the `all` subset.
- **Per-env breakdown** built from the orchestrator's resolved env names: a single train env collapses to `train/<env>`; multiple envs add `train/agg` on top followed by one section per env. Eval has no aggregate, so it's always per-env (`eval/<env>`).
- **x-axis**: everything on `step`, except `inference/*` panels which use wall time — those are logged against `_timestamp` (periodic logger), not the training step, so `step` would misalign them.
- **Env-change versioning**: the env set is encoded in the section names, so a run whose envs match an existing overview reuses it, and a run with a *new* env set gets a fresh versioned view (`overview` → `overview-v2` → …) instead of a stale one.
- **Panels** show raw metric names (no renaming); grid is dense (4/row) and sized so sections don't paginate.
- **Always on** when W&B logging is enabled — no config flag. Created by the run's primary process, online only, best-effort (a `wandb-workspaces`/API failure only logs a warning — never breaks training).

New dependency: `wandb-workspaces`. All of it lives in `src/prime_rl/utils/monitor/wandb.py`.

## Why — the volume of data we log

Counted from live ablation runs:

| project | train envs | eval envs | **metrics logged** | overview panels |
|---|---|---|---|---|
| `search-ablations` | 5 | 2 | **3,004** | 60 |
| `terminal-ablations` | 1 | 1 | **1,129** | 24 |

Each env contributes ~350 series (`{all,effective}` × `{mean,max,min,p10,p90}` × every rollout metric), so metric count grows with (train + eval) envs. The overview distills those thousands down to the few dozen panels that matter — the same layout in every project, no hand-picking.

Generated with [Claude Code](https://claude.com/claude-code)


packages/prime-rl-configs/src/prime_rl/configs/shared.py
```diff
@@ -211,6 +211,9 @@ class WandbConfig(BaseConfig):
     offline: bool = False
     """Run W&B in offline mode."""
 
+    create_overview: bool = True
```
samsja: would not even have a config and just alwys tryue

hallerite (approved):
lgtm

PR #2485, opened 2026-05-13
feat(sft): Enable VLM SFT 

eligotts:
## Summary

This PR makes SFT renderer-only and enables VLM SFT end-to-end, and builds out the Qwen3.5 (dense + MoE) custom model stack needed to train those models with context parallelism.

**Renderer-only SFT**
- Samples always go through `renderers.build_training_sample`; the legacy incremental tokenization path is removed. `[renderer]` defaults to `auto` (hand-coded renderer required; `DefaultRenderer` is rejected).
- `deps/renderers` advances to `99bedaf`, which surfaces the multimodal payload through `build_training_sample`.

**Multimodal SFT plumbing**
- `mm_kwargs`, `mm_token_type_ids`, and `seq_lens` flow through datasets, packing, collation, and trainer `forward()`.
- A single `CatDataset` packing path serves text-only and multimodal samples: text-only spans can share a multimodal pack with zero `mm_token_type_ids`; compatible multimodal kwargs concatenate along the leading item dimension. `pack_function = "cat"` is the SFT default again (`stack` remains opt-in).
- The trainer loads an `AutoProcessor`, attaches it to the renderer, saves it with VLM weight checkpoints (so `AutoProcessor.from_pretrained(step_dir)` works on exported dirs), and enforces VLM rules (custom PrimeRL VLM implementations only — the generic HF VLM path is rejected; image-only — video inputs are rejected at sample processing; `micro_batch_size = 1`, LoRA vs unfrozen vision, fail-fast when no multimodal processor can be loaded). New `examples/vlm_sft` and `examples/vlm_sft_moe` configs (data paths are placeholders).
- `data_files` cannot be combined with `subsets`/`splits` (config-rejected: subsets were silently ignored and non-train splits failed at load time).
- Local `.jsonl`/`.zst` `data_files` are normalized into a `$TMPDIR` cache before `load_dataset`; the cache (and zstd decompression output) is written to a `.part` sidecar and atomically renamed into place, so an interrupted run can't leave a truncated cache that later runs silently reuse.
- `pack_function = "stack"` is rejected at config for Qwen3.5 custom flash attention: its varlen path assumes a single flattened sequence (batch size 1), which `stack`'s genuine multi-row batches violate, crashing with a shape mismatch.
[...]

PR #2942, opened 2026-07-03
feat(sft): renderer-only SFT tokenization + local data files

hubert-marek:
## Summary

Part 1/7 of the split of #2485 (see that PR for review history). **Based on `main`; no dependencies.**

Makes SFT tokenization renderer-only and adds local dataset ingestion:

- **Renderer-only SFT**: `build_training_sample` with a hand-coded renderer replaces the incremental `apply_chat_template` masking path, which corrupted loss masks under position-dependent chat templates (e.g. Qwen3 stripping past `<think>` blocks across user turns). `DefaultRenderer` is rejected in `setup_renderer` (it falls back to incremental tokenization and doesn't fix the problem); fake data skips renderer construction so models without a hand-coded renderer can still benchmark step time. Non-assistant roles opt into the loss via the renderer's body-only path (`content_sft_roles`).
- **Local `data_files`**: glob patterns expand before materialization; `.zst` decompresses via the zstd CLI (HF handles gz/bz2/xz natively but not zstd); JSONL rows are stream-normalized for Arrow schema stability (`tool_calls` and `tools`/`tool_defs` stringified to JSON string columns — heterogeneous nested JSON otherwise hard-crashes `load_dataset` or injects phantom nulls; `deserialize_tool_calls` reverses it downstream), written atomically to a digest-keyed per-node cache with local-rank-0 gating. Normalization covers both the `messages` layout and the TRL `prompt`/`completion` layout (review finding: the latter previously passed through raw), and stringifies the *whole* `tool_calls` list with `""` for no-tool messages — per-batch Arrow inference types an all-null column chunk as `null`, which can't unify with later `list<struct>` (or even `string`) chunks, so files whose first ~10 MB contain no tool calls crashed with `DatasetGenerationError` (reproduced and fixed against a real 60k-row prompt/completion dataset). `data_files` cannot combine with `subsets`/`splits`.
- `_drop_null_fields` strips Arrow phantom nulls before tool-call deserialization so genuine nulls inside argument strings survive.
- Mixed-layout files resolve correctly (review finding): `messages` takes precedence only when non-null — Arrow schema union adds `messages: null` to prompt/completion rows whenever other rows have a `messages` column, and the previous key-presence check silently resolved those rows to empty message lists.
- Defaults: `seq_len` 256 (FakeDataConfig keeps 128), `loss_impl` `liger_fused`; CI + example reverse-text/wordle configs pin `[renderer] name = "qwen3"` (PrimeIntellect/* fine-tunes miss the auto-renderer's exact-match lookup); `deps/renderers` pin bumped to `5904fa24a` (upstream main, contains the multimodal `build_training_sample` change previously pinned as `99bedaf4b`).

## E2E (W&B project [`pr2485`](https://wandb.ai/primeintellect/pr2485))

- [x] **Baseline on `main`, same dataset**: cannot train at all — incremental tokenization rejects the Qwen3.5 chat template on effectively every example (**54,261 of 60,000 skipped, 0 training steps** before we killed it: `Mismatch in incremental tokenization at message 1 (role=assistant)`). This PR is the fix.
- [x] Text SFT at this tip (Qwen3.5-0.8B, hf impl, renderer path + `data_files`, 8×H100, seq 8192, batch 64, seed 0): 10/10 steps, loss 0.78–0.86 — [6zkfiy1r](https://wandb.ai/primeintellect/pr2485/runs/6zkfiy1r). All later stack runs reuse this dataset/seed for comparability.
- [x] Integration suite (SFT subset) at this tip: **10/10 passed**

---

> [!NOTE]
> **High Risk**
> SFT is a breaking change to tokenization and defaults (renderer required for real data, new loss/seq_len defaults), and adds substantial new model and data-loading paths that affect training correctness and distributed I/O.
> 
> **Overview**
> **SFT tokenization is now renderer-only.** The incremental `apply_chat_template` path and `build_incremental_token_mask` are removed; samples go through `build_training_sample` with a required hand-coded renderer (`setup_renderer` rejects `DefaultRenderer`). `[renderer]` defaults to auto-resolution; per-example `chat_template_kwargs` are ignored. Mixed Arrow layouts use null checks so `messages: null` rows still use `prompt`/`completion`.
> 
> **Local `data.data_files`** adds glob expansion, `.zst` decompression, stream JSONL normalization (stringified `tool_calls` / `tools` for stable Arrow), and digest-keyed per-node caches; it cannot be combined with `subsets`/`splits`.
> 
> **Config and examples:** default `seq_len` 256 (fake data keeps 128), default `loss_impl` `liger_fused`, CI/example TOMLs pin `[renderer] name = "qwen3"`, and training docs describe renderer controls and `data_files`.
> 
> **Qwen3.5 dense** gets a custom `Qwen3_5ForCausalLM` (reusing MoE gated attention blocks), AutoModel registration, shared `normalize_qwen3_5_attn_implementation`, and ring/Ulysses patches for dense flash attention.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 29b63aadbb49e4e7ea5b8933dc9d120095c2e766. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/sft.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/sft.py
@@ -3,7 +3,7 @@ from pathlib import Path
 from typing import Annotated, Literal, TypeAlias
 
 from pydantic import Field, model_validator
-from renderers import RendererConfig
+from renderers import AutoRendererConfig, RendererConfig
 
 from prime_rl.configs.shared import (
     EnvVars,
@@ -30,7 +30,7 @@ class BaseDataConfig(BaseConfig):
     batch_size: int = Field(128, ge=1)
     """Global batch size."""
 
-    seq_len: int = Field(128, ge=1)
+    seq_len: int = Field(256, ge=1)
     """Sequence length."""
 
     pack_function: Literal["cat", "stack"] = "cat"
@@ -50,6 +50,12 @@ class BaseDataConfig(BaseConfig):
 
 class FakeDataConfig(BaseDataConfig):
     type: Literal["fake"] = "fake"
+
+    seq_len: int = Field(128, ge=1)
+    """Sequence length."""
+
+    pack_function: Literal["cat", "stack"] = "cat"
+    """Sample packing strategy."""
 
     length: Literal["fixed", "variable"] = "fixed"
     """Use fixed-length samples or variable-length samples."""
@@ -78,6 +84,9 @@ class SFTDataConfig(BaseDataConfig):
     name: str = "PrimeIntellect/Reverse-Text-SFT"
     """HF dataset name or path."""
 
+    data_files: list[str] | None = None
+    """Optional local dataset files passed to ``load_dataset``."""
+
     subsets: list[str] | None = None
     """Subsets to load from the HF dataset."""
 
@@ -102,6 +111,8 @@ class SFTDataConfig(BaseDataConfig):
 
     @model_validator(mode="after")
     def validate_subsets_and_splits(self):
+        if self.data_files is not None and (self.subsets is not None or self.splits is not None):
+            raise ValueError("data_files cannot be combined with subsets/splits")
         if self.subsets is not None or self.splits is not None:
             if self.subsets is not None and self.splits is not None:
                 if len(self.subsets) != len(self.splits):
@@ -175,13 +186,8 @@ class SFTConfig(BaseConfig):
 
     tokenizer: TokenizerConfig = TokenizerConfig()
 
-    renderer: RendererConfig | None = None
-    """Typed renderer config (``renderers.RendererConfig`` discriminated
-    union). When set, SFT tokenizes samples through the ``renderers``
-    library (single ``render()`` + ``message_indices`` mask) instead of
-    the default ``build_incremental_token_mask`` path. Required for chat
-    templates that render position-dependently (e.g. Qwen3, Qwen3.5).
-    ``None`` (default) uses the legacy tokenization path."""
+    renderer: RendererConfig = AutoRendererConfig()
+    """Renderer config. Defaults to auto-selecting from the tokenizer model name."""
 
     data: DataConfig = SFTDataConfig()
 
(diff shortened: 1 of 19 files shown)


packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -78,6 +84,9 @@ class SFTDataConfig(BaseDataConfig):
     name: str = "PrimeIntellect/Reverse-Text-SFT"
     """HF dataset name or path."""
 
+    data_files: list[str] | None = None
```
mikasenghaas, line 87: why do we need this? can you not use local path for `name`? i think this is natively supported in `load_dataset` no?

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -225,8 +231,8 @@ class SFTConfig(BaseConfig):
     dist_timeout_seconds: int = 3600
     """Timeout in seconds for torch distributed ops."""
 
-    loss_impl: Literal["liger", "torch", "liger_fused", "quack_fused"] = "torch"
-    """Cross-entropy loss implementation. ``liger_fused`` fuses the lm_head projection with the CE loss to avoid materializing full logits. ``quack_fused`` uses quack-kernels for chunked linear + CE with CuTe DSL CUDA kernels."""
+    loss_impl: Literal["liger", "torch", "liger_fused", "quack_fused"] = "liger_fused"
```
mikasenghaas, line 234: unrelated to this pr?

src/prime_rl/trainer/sft/data.py
```diff
@@ -114,21 +113,38 @@ def __iter__(self):
             yield fake_sample
 
 
+def _drop_null_fields(value: Any) -> Any:
```
mikasenghaas, line 116: argh i hate pyarrow

mikasenghaas: acc, @hallerite we should switch checks in renderers from `key in collection` to `key in collection and key is not None` so we dont have to do this ugliness here

src/prime_rl/trainer/sft/data.py
```diff
@@ -223,45 +234,36 @@ def should_mask(message: dict) -> bool:
...
-                    "be ignored. Further warnings suppressed for this dataset."
-                )
-                self._warned_chat_template_kwargs = True
-
-            input_ids, loss_mask = build_training_sample(
-                self.renderer,
-                messages,
-                role_to_mask=should_mask,
-                tools=tools,
+        if example.get("chat_template_kwargs") and not self._warned_chat_template_kwargs:
```
mikasenghaas, line 247: do we need this? would be inclined to remove

src/prime_rl/trainer/sft/data.py
```diff
@@ -537,15 +538,141 @@ def setup_and_interleave_datasets(
     return dataset
 
 
+def _normalize_oai_record(record: dict) -> dict:
```
mikasenghaas, line 541: dont like all of the utils down here, can we avoid them?

src/prime_rl/trainer/sft/train.py
```diff
@@ -58,6 +58,19 @@
 from torchtitan.distributed.utils import clip_grad_norm_
 
 
+def setup_renderer(tokenizer, config):
+    """Create the SFT renderer, rejecting the DefaultRenderer fallback."""
+    renderer = create_renderer(tokenizer, config)
+    if isinstance(renderer, DefaultRenderer):
```
mikasenghaas, line 64: can we do this check in pydantic?

PR #2889, opened 2026-06-26
feat(trainer): pack RL multimodal samples

hubert-marek:
## Summary
- Adds RL `pack_multimodal` support for eager multimodal samples while preserving run/LoRA and rank modality scheduling constraints.
- Makes `seq_lens` the first-class packed sample boundary contract and threads it through transport, RL data loading, trainer forward, and Qwen3.5 MRoPE.
- Gates packed multimodal training with an explicit model capability validator, varlen attention validation, and config-layer rejection for VLM context parallelism.

## Verification
- `uv run --no-project python -m py_compile src/prime_rl/trainer/batch.py src/prime_rl/trainer/rl/data.py src/prime_rl/transport/types.py tests/unit/orchestrator/test_batch.py`
- `uv run --no-project --with ruff==0.13.0 ruff check --fix --config=pyproject.toml <PR files>`
- `uv run --no-project --with ruff==0.13.0 ruff format --config=pyproject.toml <PR files>`
- `PYTHONPATH="src:packages/prime-rl-configs/src" uv run --no-project --with pytest --with numpy --with msgspec --with torch --with jaxtyping --with tomli-w --with pydantic --with psutil --with setproctitle --with pandas --with rich --with transformers --with wandb --with 'prime-pydantic-config[toml]' --with pyzmq --with loguru python -m pytest tests/unit/orchestrator/test_batch.py -q`
- `PYTHONPATH=src:packages/prime-rl-configs/src uv run --no-project --with ruff ruff check src/prime_rl/utils/vlm.py src/prime_rl/trainer/rl/train.py packages/prime-rl-configs/src/prime_rl/configs/trainer.py tests/unit/utils/test_vlm.py tests/unit/test_configs.py`
- `PYTHONPATH=src:packages/prime-rl-configs/src uv run --no-project --with pytest --with torch --with numpy --with transformers --with psutil --with setproctitle --with loguru pytest tests/unit/utils/test_vlm.py -q`
- Direct `TrainerConfig` validation check for VLM + `cp=2` + `pack_multimodal=true`

---

> [!NOTE]
> **Medium Risk**
> Changes batch packing, multimodal tensor layout, and MRoPE/varlen attention boundaries—incorrect packing could skew RL loss or attention masks, though CP/VLM and unsupported models are blocked at config/runtime.
> 
> **Overview**
[...]

PR #2944, opened 2026-07-03
feat(trainer): seq_lens as the packed-sample boundary contract (SFT + RL)

hubert-marek:
## Summary

Part 3/7 of the split of #2485 + #2889. **Based on #2943 (`feat/qwen35-dense-attention`)** — needs the dense Qwen3.5 model for its seq_lens consumption; the renderer-only SFT data path comes from #2942.

`seq_lens` becomes the packed-sample boundary contract for both trainers; models consume it for varlen attention.

**SFT + models (first commit):**
- CatDataset rework: `_finalize_pack` pads every pack to `seq_len` (pad tokens form their own loss-masked document, so packs are always full-length and CP-divisible), yields the tail pack instead of dropping it, and overflow samples start the next pack instead of being truncated mid-sample.
- `Sample`/`Batch` gain `seq_lens`; `cat_collate` forwards it for batch-size-1 packs; stack bucketing yields `seq_lens=None`.
- **`seq_lens` is a universal contract**: every custom `ForCausalLM` declares the typed `seq_lens` forward parameter and the trainer passes it unconditionally to any `PreTrainedModelPrimeRL` (generic HF models never receive it). Standard-attention families keep deriving `cu_seqlens` from `position_ids` on the flash path (boundary-equivalent — the packer resets positions per document) and now **reject packed rows loudly on non-flash paths** instead of silently attending across documents; gpt_oss/glm_moe_dsa honor boundaries via `position_ids` (HF packed-sequence masks / sparse MLA varlen indices) and accept the param; Qwen3.5 dense + MoE consume it for varlen attention and fla state resets.
- Dense + MoE Qwen3.5 build `cu_seqlens` from `seq_lens` on the **non-flash** path. Previously `cu_seqlens` stayed `None` there: fla's GatedDeltaNet carried conv/SSM state across packed samples and SDPA attended across document boundaries — silent cross-sample contamination for sdpa/eager + cat packing. Packed hybrid batches with `full_attention` layers require flash and raise otherwise.
- Config validator rejects `pack_function="stack"` for Qwen3.5 custom flash attention (its varlen kernel assumes batch 1; stack produces genuine batches).
- Under CP both trainers pass `seq_lens=None` for now — boundaries span the pre-shard pack; global semantics land in #2946.
- The HF-impl monkey-patch keeps deriving `cu_seqlens` from position_ids only; it never receives `seq_lens` (only custom `PreTrainedModelPrimeRL` models do), so no seq_lens branch is added there.

**RL (second commit, from #2889):** packed RL micro-batches carry the same boundaries end to end — `prepare_sample` stamps them, `_materialize_bin` accumulates and asserts them, `pad_micro_batch` records padding as its own segment, the RL dataloader tensorizes them into `forward()`.

## Data efficiency: old vs new packing (offline replay, 12k docs, real tokenizer path, seq 8192)

The dataset (long CoT, mean doc 6,783 tokens) makes the trade stark:

| | old packing (main) | new packing (this PR) |
|---|---|---|
| tokens trained per epoch | 41.4M | **52.9M (+27.7%)** |
| data discarded per epoch | **49.1%** (chopped remainders) | 0% (only the inherent overlong cap) |
| docs ending mid-sample | **42.1%** | 0% |
| pad overhead per row | 0% | 21.4% |

The old packer's "zero padding" chops nearly every pack's last document and discards its remainder — half the dataset's tokens are never trained, and 42% of docs train without their endings/EOS. The new packer pays 21.4% pad compute (hardware tok/s is unchanged: pads flow through as a masked document) to train 28% more tokens with correct boundaries. Pad fraction shrinks on shorter-doc datasets.

## E2E (W&B project [`pr2485`](https://wandb.ai/primeintellect/pr2485))

- [x] Text SFT dense + MoE with reworked cat packing: dense [9gjfrcx5](https://wandb.ai/primeintellect/pr2485/runs/9gjfrcx5), MoE [jyaqjzo1](https://wandb.ai/primeintellect/pr2485/runs/jyaqjzo1) — 10/10 steps each
- [x] RL text run at this tip (seq_lens end-to-end through transport → packer → dataloader → forward): 10/10 steps — [bec3d2415a](https://wandb.ai/primeintellect/pr2485/runs/bec3d2415afb47e6b7de0feb2648bba7)
- [x] Integration suite (SFT+RL subset) at this tip: 7/8 passed; the one failure is `test_loss_goes_down` — a marginal 5-step assertion (loss delta +0.0018) whose values match the #2485 reference bit-for-bit; see #2948 for the full-suite/main-control analysis

---

> [!NOTE]
> **Medium Risk**
> Changes packed-sequence attention boundaries and SFT cat packing across the core training forward path; mis-wiring could contaminate attention across documents, though multi-doc non-flash paths now fail loudly.
> 
> **Overview**
> Introduces **`seq_lens`** as the per-document length list for packed rows (sum equals row length), threaded from dataloaders through **`forward()`** into custom **`PreTrainedModelPrimeRL`** models only; generic HF models never receive it.
> 
> **SFT:** **`CatDataset`** no longer truncates mid-document on overflow—it yields full packs via **`_finalize_pack`** (pad tokens become their own masked segment), emits the tail pack, and **`cat_collate`** passes **`seq_lens`** for batch-size-1 cat packs; stack bucketing sets **`seq_lens=None`**.
> 
> **RL:** Packing stamps and merges **`seq_lens`** in **`prepare_sample`**, **`_materialize_bin`**, and **`pad_micro_batch`** (padding appended as its own segment); RL batches tensorize it into the model forward.
> 
> **Models:** Most custom attention stacks **error on multi-segment packs without flash** (avoids cross-document SDPA/eager attention). **Qwen3.5** dense/MoE build **`cu_seqlens`** from **`seq_lens`** on the non-flash path for hybrid layers (linear attention / fla state). **SFT and RL** drop **`seq_lens`** under context parallelism until global boundaries are wired.
> 
> CI reverse-text SFT integration configs bump **`max_steps`** / **`resume_step`** only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b35e828e5fbbacc67c8e3f90cfbd0a7554e6b571. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/sft.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/sft.py
@@ -287,20 +287,6 @@ class SFTConfig(BaseConfig):
         return self
 
     @model_validator(mode="after")
-    def validate_qwen3_5_stack_flash(self):
-        name = self.model.name.lower()
-        is_qwen3_5 = "qwen3.5" in name or "qwen3_5" in name
-        is_flash = self.model.attn in ("flash_attention_2", "flash_attention_3", "fa4")
-        if is_qwen3_5 and self.model.impl != "hf" and is_flash:
-            if self.data.pack_function == "stack":
-                raise ValueError("Qwen3.5 custom flash attention does not support pack_function='stack'; use 'cat'")
-            if self.val is not None and self.val.data.pack_function == "stack":
-                raise ValueError(
-                    "Qwen3.5 custom flash attention does not support pack_function='stack' for validation data; use 'cat'"
-                )
-        return self
-
-    @model_validator(mode="after")
     def validate_cp_seq_len(self):
         if self.model.cp > 1:
             if self.data.seq_len % self.model.cp != 0:
--- a/packages/prime-rl-configs/src/prime_rl/configs/sft.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/sft.py
@@ -287,6 +287,20 @@ class SFTConfig(BaseConfig):
         return self
 
     @model_validator(mode="after")
+    def validate_qwen3_5_stack_flash(self):
+        name = self.model.name.lower()
+        is_qwen3_5 = "qwen3.5" in name or "qwen3_5" in name
+        is_flash = self.model.attn in ("flash_attention_2", "flash_attention_3", "fa4")
+        if is_qwen3_5 and self.model.impl != "hf" and is_flash:
+            if self.data.pack_function == "stack":
+                raise ValueError("Qwen3.5 custom flash attention does not support pack_function='stack'; use 'cat'")
+            if self.val is not None and self.val.data.pack_function == "stack":
+                raise ValueError(
+                    "Qwen3.5 custom flash attention does not support pack_function='stack' for validation data; use 'cat'"
+                )
+        return self
+
+    @model_validator(mode="after")
     def validate_cp_seq_len(self):
         if self.model.cp > 1:
             if self.data.seq_len % self.model.cp != 0:
--- a/src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
+++ b/src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
@@ -26,7 +26,7 @@ from prime_rl.trainer.models.qwen3_5_moe.modeling_qwen3_5_moe import (
     Qwen3_5MoeRotaryEmbedding,
     normalize_qwen3_5_attn_implementation,
 )
-from prime_rl.utils.sequence import get_cu_seqlens_from_position_ids
+from prime_rl.utils.sequence import get_cu_seqlens_from_position_ids, get_cu_seqlens_from_seq_lens
 
 
 class Qwen3_5GatedSDPAAttention(Qwen3_5MoeGatedSDPAAttention):
@@ -196,6 +196,7 @@ class Qwen3_5Model(Qwen3_5PreTrainedModel):
         input_ids: Optional[torch.LongTensor] = None,
         position_ids: Optional[torch.LongTensor] = None,
         inputs_embeds: Optional[torch.FloatTensor] = None,
+        seq_lens: Optional[torch.LongTensor] = None,
     ) -> BaseModelOutputWithPast:
         if (input_ids is None) ^ (inputs_embeds is not None):
             raise ValueError("You must specify exactly one of input_ids or inputs_embeds")
@@ -210,6 +211,13 @@ class Qwen3_5Model(Qwen3_5PreTrainedModel):
         if flash_attn_enabled:
             cu_seqlens, max_seqlen = get_cu_seqlens_from_position_ids(position_ids)
             torch._dynamo.mark_dynamic(cu_seqlens, 0)
+        elif seq_lens is not None:
+            # seq_lens is only populated for batch size 1 (see cat_collate), so a
+            # single boundary list covers the whole packed row.
+            seq_lens = seq_lens.to(device=inputs_embeds.device)
+            if seq_lens.numel() > 1 and "full_attention" in self.config.layer_types:
+                raise ValueError("Packed Qwen3.5 batches with full_attention layers require flash attention")
+            cu_seqlens, max_seqlen = get_cu_seqlens_from_seq_lens(seq_lens, total_tokens=inputs_embeds.shape[1])
         else:
             cu_seqlens = None
             max_seqlen = None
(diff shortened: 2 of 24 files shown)


src/prime_rl/trainer/sft/data.py
```diff
@@ -347,29 +371,57 @@ def __init__(self, dataset: StatefulIterableDataset, seq_len: int):
...
+        seq_len = 0
         for sample in self.dataset:
-            # Add sample to packed samples
-            for key, value in sample.items():
-                assert isinstance(value, list), f"Value for key {key} must be a list"
-                packed_samples[key].extend(value)
+            sample_len = len(sample["input_ids"])
+            would_overflow = seq_len + sample_len > self.seq_len
+            if seq_len > 0 and would_overflow:
+                yield self._finalize_pack(packed_samples, self.seq_len)
```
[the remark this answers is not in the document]
hubert-marek: Doesn't matter

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -286,6 +286,20 @@ def validate_pack_function(self):
                 raise ValueError("Validation packing function must be 'cat' when CP is enabled")
         return self
 
+    @model_validator(mode="after")
+    def validate_qwen3_5_stack_flash(self):
+        name = self.model.name.lower()
+        is_qwen3_5 = "qwen3.5" in name or "qwen3_5" in name
+        is_flash = self.model.attn in ("flash_attention_2", "flash_attention_3", "fa4")
+        if is_qwen3_5 and self.model.impl != "hf" and is_flash:
```
mikasenghaas: we dont want model specific validators here. its likely a feature abt the model which makes this path not work, then we should also validate that feature so we dont have to repeat validators for each model

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -254,6 +262,15 @@ def set_decoder(self, decoder):
     def get_decoder(self):
         return self.model
 
+    def prime_forward_kwargs(
```
mikasenghaas: getting like code smell

src/prime_rl/trainer/sft/data.py
```diff
@@ -349,29 +373,57 @@ def __init__(self, dataset: StatefulIterableDataset, seq_len: int):
...
+        # not persisted, so a checkpoint resume drops at most one sample per
+        # dataloader worker.
         return {"dataset": self.dataset.state_dict()}
 
     def load_state_dict(self, state_dict: dict):
         self.dataset.load_state_dict(state_dict["dataset"])
 
     def __iter__(self):
-        packed_samples, seq_len = defaultdict(list), 0
+        packed_samples = _new_pack()
```
mikasenghaas: can we inline those functions again, old code was significantly easier to read

src/prime_rl/trainer/sft/data.py
```diff
@@ -411,9 +463,10 @@ def __iter__(self):
             # Truncate sample if it's longer than max area
             len_sample = len(sample["input_ids"])
             if len_sample > self.max_area:
-                for key, value in sample.items():
+                for key in ("input_ids", "position_ids", "loss_mask", "target_ids"):
```
mikasenghaas, line 450: why this change?

src/prime_rl/trainer/sft/data.py
```diff
@@ -484,22 +540,26 @@ def find_bucket_idx(len_sample: int) -> int:
 
 
 def stack_collate(samples: list[Sample]) -> Batch:
+    sample = samples[0]
```
mikasenghaas: lets remove style diffs from prs like this. makes it unnecessarily hard to review

packages/prime-rl-configs/src/prime_rl/configs/sft.py
eligotts: @hubert-marek in some of my prs i introduce the notion of a multimodal adapter per model, might make sense to get that in first, might allow us to keep core shared code cleaner and shift model specific stuff into these per model multimodal adapters

src/prime_rl/trainer/sft/data.py
hubert-marek: We don't want to truncate mm_kwargs, and mm_token_type_ids, and seq_lens - only the ("input_ids", "position_ids", "loss_mask", "target_ids")

PR #2947, opened 2026-07-03
feat(rl): pack multimodal samples with context parallelism

hubert-marek:
## Summary

Parts 6–7/7: RL multimodal packing plus context parallelism, combining the final two PRs in the original stack. **Based on #2946 (`feat/vlm-cp`)** — needs the seq_lens contract from #3032 (boundaries for packed MM rows) and the VLM models from #2945. Rebased onto the current #2946 at `76fff07a0`, including its explicit pre-shard boundary handling.

Multimodal RL samples no longer force one-sample micro-batches: eager image samples pack with text spans and compatible `mm_kwargs` samples from the same run/LoRA. Packing is **always on** — VLM training is custom-implementation-only as of #2945, so there is no hf path to protect and no config knob:

- `_MicroBatchBin.can_add` gains modality-compat/run/LoRA constraints; materialized bins concatenate `EncodedTensor` `mm_kwargs` along dim 0 and keep per-sample boundaries in `seq_lens`.
- `image_grid_thw` samples must carry `mm_token_type_ids` (MRoPE position computation needs them).
- The fail-loud contract for future custom VLMs: `supports_packed_multimodal_training` set on the custom Qwen3.5 dense + MoE VLM configs, checked at startup by `validate_multi_modal_pack` for any `[model.vlm]` run; a config validator requires varlen flash attention for VLM training. VLM context parallelism is supported with `cp_style='ulysses'`; the shared model loader requires CP to resolve to a custom implementation.
- Under Ulysses CP, MRoPE batches defer sharding until after the vision merge; `seq_lens_are_pre_shard` preserves packed boundaries, routed experts shard in the model, and LoRA token counts use the sharded label length.
- Required RL `seq_lens` + `padding_len` metadata stays intact through transport. Padding is folded into the final sequence length during batch construction, so custom CP forwards and Ulysses/full-attention setup consume the lengths directly; generic HF models receive no PrimeRL-only kwargs.
- The MM nightly (`multimodal_color_codeword`) now exercises the packed path directly.

Deviations from #2889: its `pack_multimodal` knob, `resolve_pack_multimodal`, and the param threading through packer/dataloader are dropped (custom-only VLMs make the gate pointless — the knob existed to preserve hf-VLM behavior); its `tokenizer` packer parameter was removed by main's dead-code cleanup; its `model.py` hunk passing `seq_lens` directly to generic MRoPE models is superseded by the `prime_forward_kwargs` hook from #2944; its seq_lens contract landed in #2944.

[...]

PR #2946, opened 2026-07-03
feat(cp): context parallelism for hybrid + VLM SFT

hubert-marek:
## Summary

Part 5/7 of the split of #2485. **Based on #3075 (`feat/sft-vlm`)**. Rebased onto its current head at `c7ad29e72`, preserving #3075's strict custom-VLM input contract, causal-shift-aligned multimodal truncation, VLM SFT path, and global `seq_lens` contract.

Context parallelism for hybrid (linear-attention) and VLM models in SFT:

- **`setup_model_cp`** replaces `setup_hybrid_cp`/`setup_nemotron_h_cp`: models owning linear-attention/Mamba layers expose `set_context_parallel_attributes` (dense + MoE Qwen3.5, NemotronH) and wire their own layers through activation-checkpoint wrappers; hybrid models without the hook are rejected instead of silently misconfigured.
- **`seq_lens` stays global under CP**: documents can straddle the shard cut, so `input_ids`/`position_ids` shard per rank while `seq_lens` keeps the full pre-shard boundaries, flagged via the `seq_lens_are_pre_shard` typed `forward()` parameter, which this PR adds to every custom model (extending #2944's universal `seq_lens` contract). A `cu_seqlens_are_pre_shard` provenance flag rides to fla's GatedDeltaNet instead of reading `cu_seqlens` back — the `.item()` read cost one CPU-GPU pipeline sync per linear-attention layer per microbatch (~288/step on 35B MoE cp2). The CP conv path switches to fla's `causal_conv1d` with a CP context so conv state passes between ranks with document-boundary resets. Qwen3.5 CP now fails immediately when full pre-shard boundary provenance is missing instead of fabricating a single-sequence boundary.
- **Explicit boundaries drive every CP consumer**: trailing padding is already folded into the final sequence length when batches are built; Ulysses/full attention, Qwen3.5 DeltaNet, and Nemotron-H Mamba receive the full pre-shard boundaries without forward-time normalization. Generic HF models still receive no PrimeRL-only kwargs.
[...]

PR #2948, opened 2026-07-03
feat(rl): multimodal context parallelism

hubert-marek:
## Summary

Part 7/7 — the only new code in the stack (in neither #2485 nor #2889; both deferred RL MM CP). **Based on #2947 (`feat/rl-multimodal-packing`)** — needs MM packing from #2947, the VLM CP machinery from #2946 (which is trainer-agnostic, in the modeling files), and the seq_lens contract from #2944. Rebased onto the current #2947 after its packing review fixes and the finalized #2946 CP cleanup.

Mirrors the SFT VLM CP path in the RL trainer, replacing the blanket `NotImplementedError`:

- MRoPE multimodal batches under ulysses defer sharding to the model: `input_ids`/`position_ids` stay global so the vision encoder and image-embed merge see the full sequence; the model root forward shards embeds/positions after merge (and `routed_experts` with them). Text batches keep the existing pre-shard path.
- `seq_lens` passes with `seq_lens_are_pre_shard=True` under CP — micro-batch boundaries span the pre-shard sequence, and `pad_to_multiple_of=cp` already guarantees shard divisibility (padding is its own `seq_lens` segment since #2944).
- The LoRA token-count CP adjustment derives `chunk_size` from the sharded length even when `input_ids` stays global.
- Gates relaxed: VLM + CP now requires `cp_style='ulysses'` and a custom model implementation instead of blanket rejection (`impl="auto"` passes the validator; `get_model` rejects it at load time if it resolves to hf).
- Docs: guide on adding a custom VLM implementation (VLM training is custom-only as of #2945; this documents the full contract — registration, seq_lens consumption, MRoPE, dummy-vision FSDP symmetry, packing capability flag, CP hooks, KL validation).

**Proposed follow-up** (not in this stack): the deferral condition and sharding choreography here deliberately mirror `sft/train.py`; extracting the shared logic into one helper would keep the two trainer loops from drifting.

## E2E

- [ ] RL MM cp=2 ulysses run (`multimodal_color_codeword` shape) — W&B link TBD

- [x] Integration suite on 8×H100 at this tip: **27 passed, 6 failed, 1 error — every failure also fails (or is family-flaky) on unmodified `main` on the same node**: 4× `test_benchmark_regression` (tolerances calibrated for the CI A6000 runners), `test_reverse_text_moe` (FileNotFoundError, env), `test_reverse_text_multi_run` (flaky family — main fails a sibling test). The one stack-correlated flip, `test_reverse_text_sft::test_loss_goes_down`, is a marginal assertion: 5 steps at batch 4, last-vs-first loss delta +0.0018, and this branch's loss/grad values match the #2485-head reference **bit-for-bit** (5.5987/73.2068). Suggest lengthening that CI run or comparing first-k/last-k means — flagged for review.

---

> [!NOTE]
> **Medium Risk**
> Touches distributed RL forward/CP sharding for multimodal batches; incorrect boundary or shard handling could skew logprobs/gradients, but scope is aligned with existing SFT behavior and gated to ulysses + custom models.
> 
> **Overview**
> **RL multimodal training can use context parallelism** when `cp_style='ulysses'`, instead of being rejected outright.
> 
> The RL trainer **mirrors the SFT VLM CP path**: for MRoPE batches (multimodal kwargs with `image_grid_thw`), it **skips pre-forward sharding** of `input_ids`/`position_ids` so vision merge and image embeds see the full sequence; the custom model shards after merge. Text-only CP batches still use the existing `setup_cp_params` path. **`seq_lens` stays on the wire** under CP with `seq_lens_are_pre_shard=True` so packed document boundaries refer to the pre-shard sequence. LoRA CP token counting uses **sharded `labels` length** for `chunk_size` when inputs stay global.
> 
> **Config validation** replaces blanket “VLM + CP forbidden” with **VLM + CP requires `cp_style='ulysses'`** (ring CP with VLM still errors).
> 
> **Docs** add a development guide for **custom VLM implementations** (registration, packed `seq_lens`, dummy vision for FSDP, optional CP hooks).
> 
> **Tests** swap the old rejection test for ring-vs-ulysses coverage and assert ulysses + VLM + `cp=2` validates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd09c630eb2d0f2f62760b60331d00837aeb0d01. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hubert-marek:
Folded into #2947 now that the RL multimodal CP delta is small and directly coupled to multimodal packing.

PR #2953, opened 2026-07-03
feat(trainer): MoE loss-free load balancing

faresobeid:
## Summary

Adds training-time **MoE load balancing** to prime-rl's custom model implementation. Opt-in and off by default.

### MoE load balancing

New `[model.moe_load_balance]` config (`mode` = `off` | `loss_free`, `coeff`), wired into both SFT and RL. Before this, `load_balance_coeff` only allocated the router `expert_bias` buffer (loaded from checkpoints and frozen) — there was no active balancing mechanism during training. (torchtitan implements this same path as an optimizer step-pre-hook; prime-rl never wired it up.)

**`loss_free`** — DeepSeek auxiliary-loss-free `expert_bias` update ([arXiv:2408.15664](https://arxiv.org/abs/2408.15664)), applied once per optimizer step: `expert_bias += coeff · sign(load_error)`, from a dedicated per-step usage accumulator. Details:
- The accumulator only grows on **training-mode** forwards, so SFT validation traffic isn't folded into the update.
- The bias delta is **recentered to zero mean** (matching torchtitan) so the bias vector doesn't drift a global offset over long runs. This also makes the update invariant to the uniform per-layer count doubling caused by full-block activation checkpointing (the router reruns in backward).
- Expert counts are summed across the `dp_cp` group so balancing targets the whole global batch. Default `coeff = 1e-3`.
- `get_model` forces the `expert_bias` buffer to exist when `loss_free` is requested (several models default `load_balance_coeff=None`), and raises a clear error if a model can't provide one (e.g. MiniMax with `use_routing_bias=False`) rather than silently no-op'ing.

SFT validation switches the model to `eval()` when loss-free balancing is active, so validation forwards don't skew the load-balance counts. With balancing off (the default), the model stays in train mode (no `torch.compile` recompilation).

## Files

- `configs/trainer.py` — `MoELoadBalanceConfig`
- `trainer/model.py` — `set_moe_load_balance_active`, `update_expert_bias`, loss-free `expert_bias` enable + validation
- `trainer/models/layers/moe.py` — per-step load accumulator (train-only)
- `trainer/{sft,rl}/train.py` — bias-update wiring; SFT eval-mode validation
- `docs/advanced.md` — MoE Load Balancing section

## Testing

Unit-tested on CPU: `update_expert_bias` lowers over-loaded / raises under-loaded experts, its delta sums to zero, and the update is bit-identical for 1× vs 2× (AC-doubled) counts; config surface validates (`mode` ∈ {off, loss_free}, `coeff` default `1e-3`). `ruff check` + `ruff format --check` clean. Not yet exercised in a live distributed run.

---

> [!NOTE]
> **Medium Risk**
> Changes MoE routing during training via mutable router biases and distributed all-reduces; misconfiguration or incompatible models fail at load time, but routing shifts can affect training dynamics and EP load.
> 
> **Overview**
> Adds opt-in **training-time MoE load balancing** for the custom MoE stack via `[model.moe_load_balance]` (`mode`: `off` | `loss_free`, `coeff`).
> 
> When **`loss_free`** is enabled, each optimizer step updates router **`expert_bias`** using DeepSeek’s auxiliary-loss-free rule (sign of load error, zero-mean delta, no extra loss). Usage is tracked in a dedicated **`expert_load_acc`** buffer (separate from logging’s `tokens_per_expert`), summed over the **`dp_cp`** group, and only accumulated on **training** forwards when balancing is active.
> 
> **SFT** toggles **`eval()`** during validation when balancing is on so val traffic does not skew counts; **RL** and **SFT** call **`update_expert_bias`** after **`optimizer.step()`**. **`get_model`** forces creation of **`expert_bias`** when needed and errors if the architecture has no routing bias buffer.
> 
> Docs add a **MoE Load Balancing** section under advanced custom modeling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3661b6cac197dcc02d7539b432c886e4a271e33a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/trainer/sft/train.py
```diff
@@ -448,6 +469,9 @@ def run_validation(step: int) -> None:
 
         logger.debug("Optimizer step")
         optimizer.step()
+        if config.model.moe_load_balance.mode == "loss_free":
+            lb = config.model.moe_load_balance
+            update_expert_bias(model, lb.coeff, dp_cp_group)
```
samsja: where do we use `lb` after ?

PR #2968, opened 2026-07-07
feat(sft): renderer-only SFT tokenization

samsja:
## Summary

Split from #2942 (part 1/7 of #2485) — renderer-only SFT tokenization, without the local `data_files` ingestion that ships separately.

Makes SFT tokenization renderer-only and deletes the broken incremental masking path:

- **Renderer-only SFT**: `build_training_sample` with a hand-coded renderer replaces the incremental `apply_chat_template` masking path, which corrupted loss masks under position-dependent chat templates (e.g. Qwen3 stripping past thinking blocks across user turns). Non-assistant roles opt into the loss via the renderer's body-only path (`content_sft_roles`).
- **Typed renderers are required for real SFT data**: the config validator rejects both `renderer.name='default'` and auto-resolution misses, since `DefaultRenderer` cannot provide the sampled-token/content masks the loss depends on. Fake-data runs without validation skip renderer construction and stay exempt.
- **`prime-qwen3` typed renderer** (PrimeIntellect-ai/renderers#100, on renderers `main`): `deps/renderers` is bumped to the merge commit for the custom `PrimeIntellect/Qwen3-0.6B` / `1.7B` chat template (byte-parity with `apply_chat_template`, full mask metadata, XML tool calls, bridging). Base checkpoints are in `MODEL_RENDERER_MAP`, so the CI + example reverse-text/wordle SFT configs drop their explicit `[renderer]` blocks and auto-resolve; only the renamed Wordle fine-tune keeps an explicit `name = "prime-qwen3"` (exact-match map by design), replacing its previous `qwen3` + `thinking_retention` selection.
- **Null-safe message resolution**: `messages` takes precedence over `prompt`/`completion` only when non-null — Arrow schema union adds `messages: null` to prompt/completion rows whenever other rows have a `messages` column, and the previous key-presence check silently resolved those rows to empty message lists.
- `_drop_null_fields` strips Arrow phantom nulls before tool-call deserialization, and preserves genuine nulls inside `tool_calls[].function.arguments` whether arguments arrive as JSON strings or already-decoded dicts.
- Defaults: `seq_len` 256 (FakeDataConfig keeps 128), `loss_impl` `liger_fused`.

## Tests

- `tests/unit/train/sft/test_sft_dataset.py` — 20 passed
- SFT renderer config validators + all migrated CI/example TOMLs — passed
- renderers submodule (upstream PR): dedicated `prime-qwen3` parity (18) + shared protocol/round-trip/bridge/tool-arg selections (103 passed, 2 skipped)

### DIFF
--- a/configs/ci/integration/reverse_text_sft/resume.toml
+++ b/configs/ci/integration/reverse_text_sft/resume.toml
@@ -6,8 +6,9 @@ resume_step = 5
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+# PrimeIntellect/Qwen3-0.6B ships its own chat template (distinct from Qwen3's)
 [renderer]
-name = "qwen3"
+name = "default"
 
 [data]
 name = "PrimeIntellect/Reverse-Text-SFT"
--- a/configs/ci/integration/reverse_text_sft/resume.toml
+++ b/configs/ci/integration/reverse_text_sft/resume.toml
@@ -6,6 +6,9 @@ resume_step = 5
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+[renderer]
+name = "qwen3"
+
 [data]
 name = "PrimeIntellect/Reverse-Text-SFT"
 batch_size = 4
--- a/configs/ci/integration/reverse_text_sft/start.toml
+++ b/configs/ci/integration/reverse_text_sft/start.toml
@@ -5,8 +5,9 @@ max_steps = 5
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+# PrimeIntellect/Qwen3-0.6B ships its own chat template (distinct from Qwen3's)
 [renderer]
-name = "qwen3"
+name = "default"
 
 [data]
 name = "PrimeIntellect/Reverse-Text-SFT"
--- a/configs/ci/integration/reverse_text_sft/start.toml
+++ b/configs/ci/integration/reverse_text_sft/start.toml
@@ -5,6 +5,9 @@ max_steps = 5
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+[renderer]
+name = "qwen3"
+
 [data]
 name = "PrimeIntellect/Reverse-Text-SFT"
 batch_size = 4
--- a/configs/ci/integration/reverse_text_sft_lora/resume.toml
+++ b/configs/ci/integration/reverse_text_sft_lora/resume.toml
@@ -9,8 +9,9 @@ save_adapter_separately = true
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+# PrimeIntellect/Qwen3-0.6B ships its own chat template (distinct from Qwen3's)
 [renderer]
-name = "qwen3"
+name = "default"
 
 [model.lora]
 rank = 8
--- a/configs/ci/integration/reverse_text_sft_lora/resume.toml
+++ b/configs/ci/integration/reverse_text_sft_lora/resume.toml
@@ -9,6 +9,9 @@ save_adapter_separately = true
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+[renderer]
+name = "qwen3"
+
 [model.lora]
 rank = 8
 target_modules = [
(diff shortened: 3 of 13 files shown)


configs/ci/integration/reverse_text_sft/resume.toml
```diff
@@ -6,6 +6,9 @@ resume_step = 5
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+[renderer]
+name = "qwen3"
```
hallerite: `PrimeIntellect/Qwen3-0.6B` should use the default renderer, as it has a different chat template than Qwen3

configs/ci/integration/reverse_text_sft/start.toml
```diff
@@ -5,6 +5,9 @@ max_steps = 5
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+[renderer]
+name = "qwen3"
```
hallerite: `PrimeIntellect/Qwen3-0.6B` should use the default renderer, as it has a different chat template than Qwen3

configs/ci/integration/reverse_text_sft_lora/resume.toml
```diff
@@ -9,6 +9,9 @@ save_adapter_separately = true
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+[renderer]
+name = "qwen3"
```
hallerite: `PrimeIntellect/Qwen3-0.6B` should use the default renderer, as it has a different chat template than Qwen3

configs/ci/integration/reverse_text_sft_lora/start.toml
```diff
@@ -8,6 +8,9 @@ save_adapter_separately = true
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+[renderer]
+name = "qwen3"
```
hallerite: `PrimeIntellect/Qwen3-0.6B` should use the default renderer, as it has a different chat template than Qwen3

examples/reverse_text/sft.toml
```diff
@@ -5,6 +5,9 @@ max_steps = 100
 [model]
 name = "PrimeIntellect/Qwen3-0.6B"
 
+[renderer]
+name = "qwen3"
```
hallerite: `PrimeIntellect/Qwen3-0.6B` should use the default parser, as it has a different chat template than Qwen3

examples/wordle/sft.toml
```diff
@@ -5,6 +5,9 @@ max_steps = 20
 [model]
 name = "PrimeIntellect/Qwen3-1.7B"
 
+[renderer]
+name = "qwen3"
```
hallerite: `PrimeIntellect/Qwen3-0.6B` should use the default renderer, as it has a different chat template than Qwen3

src/prime_rl/trainer/sft/data.py
```diff
@@ -163,16 +171,19 @@ def __init__(
         self.data_world_size = get_world().world_size // non_dp_size * num_workers
 
     def _process(self, example: dict) -> dict | None:
-        # Skip processing if no tokenizer was provided
         if self.tokenizer is None:
             return example
+        if self.renderer is None:
```
hallerite: if one is required, why not just make renderer be type `Renderer` instead of `Renderer | None`?

src/prime_rl/trainer/sft/data.py
```diff
@@ -223,45 +229,36 @@ def should_mask(message: dict) -> bool:
...
-                    "be ignored. Further warnings suppressed for this dataset."
-                )
-                self._warned_chat_template_kwargs = True
-
-            input_ids, loss_mask = build_training_sample(
-                self.renderer,
-                messages,
-                role_to_mask=should_mask,
-                tools=tools,
+        if example.get("chat_template_kwargs") and not self._warned_chat_template_kwargs:
```
hallerite: remove this

src/prime_rl/trainer/sft/data.py
hubert-marek: Laud fail but I can remove

src/prime_rl/trainer/sft/data.py
```diff
@@ -114,23 +108,44 @@ def __iter__(self):
...
+    return value
+
+
 class SFTDataset(StatefulIterableDataset):
     """A dataset wrapping a HF SFT dataset with prompt/completion or raw messages format."""
 
     def __init__(
         self,
         dataset: Dataset,
         tokenizer: PreTrainedTokenizer | None,
```
hallerite: I wonder if we shouldn't unify tokenizer and renderer here and make renderer mandatory, as we want to force users of using a renderer. that should make the file cleaner overall

PR #2978, opened 2026-07-09
feat: make zero_advantage pre-batch filter enforcing by default

samsja:
## Summary

The `zero_advantage` pre-batch filter now defaults to `enforce = true`, so rollouts whose advantage stream is all zero (e.g. a GRPO group where every rollout earned the same reward) are dropped before they enter the training batch — instead of merely being tracked as a detection metric.

This matches the behavior previously recommended in the CHANGELOG for configs migrating from the removed `online_difficulty_filtering = true`:

> To preserve `online_difficulty_filtering = true`, enforce the zero-advantage pre-batch filter: `[[orchestrator.pre_batch_filters]]` with `type = "zero_advantage"` and `enforce = true`

Making it the default means users get sensible behavior out of the box without needing to add the filter block manually.

## Changes

- `packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py`: Changed the `pre_batch_filters` default for `ZeroAdvantageFilterConfig` from `enforce=False` to `enforce=True`, and updated the docstring.
- `CHANGELOG.md`: Added a breaking-change entry.

## Behavior

| Filter | Pre-batch default | Post-batch default |
|---|---|---|
| gibberish | monitor (`enforce=false`) | enforce (class default `true`) |
| repetition | monitor (`enforce=false`) | enforce (class default `true`) |
| **zero_advantage** | **enforce (`enforce=true`)** ← changed | enforce (class default `true`) |

## Opting out

To restore the old monitor-only behavior:

```toml
[[orchestrator.pre_batch_filters]]
type = "zero_advantage"
enforce = false
```

## Testing

- All 23 filter unit tests pass (`tests/unit/orchestrator/test_filters.py`)
- All 37 non-env-dependent config tests pass (`tests/unit/test_configs.py`)
- Verified the default config: `gibberish: enforce=False`, `repetition: enforce=False`, `zero_advantage: enforce=True`

---

> [!NOTE]
> **Medium Risk**
> Changes default training batch composition for configs that relied on zero-advantage groups still entering batches; behavior is reversible via explicit config but existing runs may see different effective batch sizes and learning signal without edits.
> 
> **Overview**
> **Breaking default:** `orchestrator.pre_batch_filters` now registers `zero_advantage` with **`enforce = true`**, so GRPO-style groups whose advantage stream is all zero are **dropped before they occupy a training batch slot** instead of only contributing detection metrics.
> 
> Gibberish and repetition pre-batch filters stay in monitor mode (`enforce = false`). Post-batch filter defaults are unchanged.
> 
> **Migration:** To restore monitor-only behavior, add `[[orchestrator.pre_batch_filters]]` with `type = "zero_advantage"` and `enforce = false`. `CHANGELOG.md` documents the breaking change and ties it to the former `online_difficulty_filtering` migration guidance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7945fca85bd687359bb2ad36e903982747a4224. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas, line 1: should we need to turn it off as post batch filter now?

PR #2979, opened 2026-07-09
feat: top-p/top-k train sampling with sampling replay

faresobeid:
Adds **top-p and top-k sampling support for train rollouts** (both were hardcoded off). Truncated sampling renormalizes the rollout distribution over the surviving "kept set" of tokens; our rollout logprobs already reflect that (`logprobs_mode = "processed_logprobs"`), but the trainer normalizes over the full vocabulary — so every importance ratio is biased, and runs with truncated sampling collapse. This PR makes truncation safe by recording the kept set at sampling time and renormalizing trainer logprobs over the same set: DeepSeek V3.2's "Keep Sampling Mask" ([arXiv:2512.02556](https://arxiv.org/abs/2512.02556) §3.1), also described in Cognition's [SWE-1.7 post](https://cognition.com/blog/swe-1-7) as "sampling distribution replay".

## Usage

```toml
[orchestrator.train.sampling]
top_p = 0.95
top_k = 512   # optional — defaulted to 512 when truncation is on, since it bounds the kept sets
```

That's the whole config — there are no replay flags. Truncated train sampling (`top_p < 1` and/or `top_k`) implies **sampling replay** end to end:

- Every truncating policy-sourced sampling config gets a top-k bound (`top_k` respected if set, else defaulted to 512, which rarely binds at typical top_p), so kept sets are never larger than k. Truncation knobs must be the typed fields — smuggling them via `extra_body` is rejected. Frozen-source envs are exempt (external endpoints, no importance ratios).
- `inference.kept_tokens` (`None` = capture off, int = capture width) is derived as the largest configured top_k, so no kept set ever overflows — replay is exact at every position. The field is auto-set (`use_pd_kv_transfer`-style) and persisted into per-node configs; hand-setting is only for standalone-launched servers.
- The trainer is data-driven: it replays masks whenever a batch carries them, like any other per-token stream. The orchestrator enforces that truncating envs actually produce masks (fails fast if the server isn't capturing).
- Consumers of rollout logprobs that break under renormalization are rejected at config time: `opd`/`opsd` (reference logprobs are full-vocab prefill scores), and the gibberish/repetition filters (removed from the default lists, rejected if explicitly configured — their full-softmax thresholds misfire when singleton kept sets read as probability 1.0).

## How it works

**Inference** (`src/prime_rl/inference/vllm/kept_tokens.py`, monkey patches over the stock vLLM 0.24 wheel — no upstream support exists yet, unlike routed experts, which 0.24 returns natively via `enable_return_routed_experts`; upstreaming would shrink this to API-layer glue the same way):

- The kept set is read off the sampler's processed logprobs — the exact tensor the token was sampled from — so membership holds by construction. (slime and SGLang instead recompute the nucleus and must force-keep the sampled token against kernel boundary disagreements.)
- vLLM's inter-process output structs are fixed msgspec schemas, so the kept ids ride the existing logprobs channel as a `-1`-separated row extension, at a fixed device-side width (no host syncs). An API-process patch splits the extension back off before vLLM builds logprob dicts — chat/eval consumers see byte-identical logprobs — and `/inference/v1/generate` returns base64 `{ids, counts}` per choice, like `routed_experts`. Kept sets are decode-only, so PD-disaggregated serving needs no router changes.
- Incompatible setups fail at startup instead of running silently biased: speculative decoding, `logprobs_mode` overrides, `VLLM_USE_V2_MODEL_RUNNER=1`.

**Trainer** (data-driven — replays masks whenever the batch carries them):

- Masked positions compute `logprob = logits[label]/T - logsumexp(logits[kept]/T)` in both the chunked fused LM head (backward restricted to kept ids) and the vanilla path. Positions without a mask (context tokens, non-policy samples) use full-vocab logprobs.
- Singleton kept sets (top token above the top-p threshold) give logprob 0 and exactly zero gradient — the entropy-preserving property from the SWE-1.7 post.
- Entropy stays full-vocab (it's a collapse diagnostic; matches slime).
- Gemma-family softcapped lm_heads don't implement kept-set renormalization and fail loudly on their head assert.

**Transport**: `KeptTokens {ids, counts}` (int32 bytes, CSR-style) on `TrainingSample`/`MicroBatch`, appended last to keep the positional wire layout stable; packed/truncated/padded alongside the other per-token streams; tensorized as `[1, seq, max_kept]` with `-1` padding.

## Paired dep PRs (merged)

Both dep PRs are merged; the submodules pin their upstream `main` merge commits (branch is also synced with prime-rl `main`):
- PrimeIntellect-ai/renderers#102 — parse `kept_tokens` from `/generate` responses (base64 splice fast path).
- PrimeIntellect-ai/verifiers#1978 — `KeptTokens` dataclass: `TurnTokens` carrier → graph attribution → `Branch.kept_tokens`.

## Verification

All CPU-level, against the real installed vLLM 0.24:

- Fused and vanilla logprob paths match a dense masked-renormalization reference, forward and backward — including singleton zero-grad, misaligned-mask fallback, temperature ≠ 1, and an exp-overflow regression case.
- The patched `Sampler.forward` produces kept sets matching an independent implementation of vLLM's top-p semantics; the returned sampled-token logprob equals the renormalized reference; greedy/untruncated steps stay position-aligned.
- Serialize → parse → branch spreading → transport → pack/truncate/pad → tensorize round-trips, with mixed masked/maskless samples.
- Config policy: top-k defaulting (explicit, per-env, inherited), derived `inference.kept_tokens` (including explicit-value override), `extra_body` truncation rejection, opd/filter guards, frozen-source exemption, standalone-server warning.

End-to-end GPU validation (H200, `top_p 0.97 / top_k 512`):

- **reverse-text** (Qwen3-0.6B, 20 steps): reward 0.10 → 0.78, `mismatch_kl` bounded 0.0005–0.0094, entropy rising, 0% errors.
- **hendrycks sanity** (R1-Distill-Qwen-1.5B, 200 steps, batch 512 × 8k ctx, W&B `hendrycks-sanity/topp0.97-topk512-replay`): **`mismatch_kl` 0.0003–0.0004 at every step, flat from step 1 to 200** — *below* an untruncated control run's noise floor (0.00042–0.00046), since singleton kept sets contribute exactly zero mismatch. Train reward 0.49 → 0.68; AIME2024 eval 0.1875 → 0.2458; entropy flat (no collapse); DPPO trust-region masking 0.0000 in both runs; replay costs ≈5% MFU vs control.
- Measured mask load (2.5M-token steps): kept-set sizes mean ~8 / median 2 / p99 ~82, **45–48% singletons** (matching Cognition's observation), `top_k=512` never binds, 100.00% mask coverage on sampled tokens, ~32 B/token on the wire (~40% of the trainer payload, trivial in absolute bandwidth).

Known gap from the ablation: **legacy v0 envs are not wired for kept tokens** — truncated sampling on a v0 env fails loudly at the first group (the orchestrator's missing-mask guard). Needs the v0 `ResponseTokens` path wired for parity (routed_experts supports v0) or documented v1-only support.

## Merge update (2026-08-03)

Merged current `origin/main` into the branch (post config-shim removal / TOML section expansion). Conflict resolutions worth noting:

- `TrainSamplingConfig.max_completion_tokens` keeps main's shim-free form (no `max_tokens` alias); `top_p`/`top_k` unchanged.
- `setup_truncated_sampling` / `auto_setup_kept_tokens_capture` ported from `train.env` to main's `train.source`, alongside main's new `validate_env_algorithms`.
- `kept_tokens` CP sharding follows main's deferred-VLM guard (`not defer_vlm_cp_to_model`), mirroring `routed_experts`.

Both dep PRs (PrimeIntellect-ai/renderers#102, PrimeIntellect-ai/verifiers#1978) are ancestors of main's submodule pins, which the merge adopts.

Local checks: `uv lock --check`, ruff on changed files, `tests/unit/test_configs.py` (108 passed), `tests/unit/inference/` (18 passed), `tests/unit/train/rl/` + `tests/unit/orchestrator/` (117 passed, 1 pre-existing main failure: qwen3-vl fixture, #3161).

Post-merge GPU validation — reverse-text, 20 steps, `top_p 0.97` (auto `top_k 512`, derived `inference.kept_tokens = 512`), 1 trainer + 1 inference GPU: reward 0.13 → 0.74, mismatch_kl bounded 0.0006–0.0146, entropy ~1.0 (no collapse), 0% rollout errors, 100% trainable; traces carry `top_p 0.97` / `top_k 512`.

---

> [!NOTE]
> **High Risk**
> Touches core RL correctness (importance ratios), vLLM monkey-patches, and training loss math; misconfiguration previously collapsed training and is only partly blocked at config/startup time.
> 
> **Overview**
> Adds **train rollout `top_p` / `top_k`** and end-to-end **sampling replay** so rollout logprobs (renormalized under truncation) match trainer logprobs over the same per-token kept sets.
> 
> **Config & policy:** `TrainSamplingConfig` exposes truncation fields (no more hardcoded `top_p = 1`), rejects truncation via `extra_body`, and `OrchestratorConfig` defaults missing `top_k` to 512, blocks `opd`/`opsd`, and strips or rejects gibberish/repetition filters. `RLConfig` sets `inference.kept_tokens` from the largest train `top_k`.
> 
> **Inference:** New vLLM patches capture kept sets from processed logprobs, ship them on `/inference/v1/generate`, with startup guards for speculative decoding and wrong `logprobs_mode`.
> 
> **Pipeline:** `KeptTokens` flows orchestrator → transport → packer → trainer; truncated policy envs fail fast if masks are missing.
> 
> **Trainer:** Fused and vanilla LM heads renormalize logprobs (and fused backward) over kept sets; Gemma softcap heads explicitly reject replay.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 06147dd28f97e89d59317167fbb8860cc2db7da8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/inference.md
```diff
@@ -282,3 +282,17 @@ enable_return_routed_experts = true
 This however is not free, it adds a significant overhead to the HTTP requests as this payload can grow quite large. We reccomend increasing `orchestrator.*.env.num_workers` to allow for more parallelization on the verifiers side.
 
 Currently this feature is also not supported with CPU KV cache offload, which can have negative impact on the inference throughput.
+
+### Sampling Mask Replay
```
mikasenghaas: i kinda like "Sampling Replay", so you ahve router + sampling replay as 2 stability flags. wdyt?

docs/inference.md
```diff
@@ -282,3 +282,17 @@ enable_return_routed_experts = true
...
+
+Truncated sampling (`top_p < 1`, `top_k`, `min_p > 0`) renormalizes the sampling distribution over the surviving "kept set" of tokens. The rollout logprobs reflect that (`logprobs_mode = "processed_logprobs"`), so the trainer must renormalize over the same set — otherwise every importance ratio is biased and training collapses (DeepSeek V3.2's "Keep Sampling Mask", [arXiv:2512.02556](https://arxiv.org/abs/2512.02556) §3.1; Cognition's [SWE-1.7 post](https://cognition.com/blog/swe-1-7)). prime-rl handles this automatically: the kept-set token ids are recorded at sampling time and the trainer renormalizes its logprobs over them.
```
mikasenghaas: do we support min p? thinking we shouldnt support it

packages/prime-rl-configs/src/prime_rl/configs/inference.py
```diff
@@ -430,6 +430,9 @@ class InferenceConfig(BaseConfig):
     enable_return_routed_experts: bool = False
     """Return routed experts in responses. Forwarded as ``--enable-return-routed-experts``."""
 
+    kept_tokens: int | None = Field(None, ge=1)
```
mikasenghaas, line 404: if its auto-derived lets not put it on the config?

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -53,6 +53,20 @@ class TrainSamplingConfig(BaseConfig):
...
+    """Nucleus (top-p) sampling for train rollouts. Values below 1.0 truncate the sampling
+    distribution; the ``rl`` entrypoint auto-enables sampling-mask replay so trainer and
+    rollout distributions stay consistent — see docs/inference.md (Sampling Mask Replay)."""
+
+    top_k: int | None = Field(None, ge=1)
+    """Top-k sampling for train rollouts. Truncation triggers sampling-mask replay, and
+    a default top-k is injected when only top-p/min-p truncate so kept sets stay
+    bounded — see docs/inference.md (Sampling Mask Replay)."""
+
+    min_p: float = Field(0.0, ge=0, lt=1.0)
```
mikasenghaas: do we care abt this ?

mikasenghaas: would personally be fine with not exposing this

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -53,6 +53,20 @@ class TrainSamplingConfig(BaseConfig):
     temperature: float = Field(1.0, ge=0, le=2.0)
     """Sampling temperature."""
 
+    top_p: float = Field(1.0, gt=0, le=1.0)
```
mikasenghaas, line 57: should we default to < 1.0 or nah?

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -63,18 +77,48 @@ class TrainSamplingConfig(BaseConfig):
...
 
+    def truncates_distribution(self) -> bool:
+        return self.top_p < 1.0 or self.top_k is not None or self.min_p > 0.0
+
+    @model_validator(mode="after")
+    def validate_no_extra_body_truncation(self):
+        """Truncating values must come from the typed fields — the replay policy reads
+        them. Disabled values pass so resolved configs (where ``resolve_env_config``
+        stamped the ``top_k = -1`` / ``min_p = 0.0`` sentinels) re-validate cleanly."""
+        smuggled = [
```
mikasenghaas, line 83: lol funny name fable

src/prime_rl/inference/vllm/kept_tokens.py
mikasenghaas, line 1: argh in the great year of 2026, we still have to monkeypatch vllm for stuff like this?! do we do the same r3?

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -53,6 +53,16 @@ class TrainSamplingConfig(BaseConfig):
     temperature: float = Field(1.0, ge=0, le=2.0)
     """Sampling temperature."""
 
+    top_p: float = Field(1.0, gt=0, le=1.0)
+    """Nucleus (top-p) sampling for train rollouts. Values below 1.0 truncate the sampling
+    distribution; the ``rl`` entrypoint auto-enables sampling replay so trainer and
+    rollout distributions stay consistent — see docs/inference.md (Sampling Replay)."""
+
+    top_k: int | None = Field(None, ge=1)
```
mikasenghaas, line 62: should we have a `le`?

mikasenghaas (dismissed):
lgtm!

mikasenghaas:
Superseded by #3235 — same feature re-cut as a single commit on current main (vLLM 0.26 / pass-through inference config / post-v0), with an ids-only engine extension and upstream alignment notes (vllm-project/vllm#49577).

PR #2984, opened 2026-07-09
feat(models): dense Qwen3.5 custom attention

hubert-marek:
## Summary
- add a dense custom `Qwen3_5ForCausalLM` implementation using PrimeRL attention modules
- register dense Qwen3.5 model/config dispatch and context-parallel attention substitutions
- add unit coverage for dispatch, state-dict conversion, and attention selection

## Validation
- `uv run --no-sync ruff check ...` — passed
- `uv run --no-sync pytest tests/unit/train/models/test_qwen3_5.py -q --tb=short` — blocked during collection because `flash_attn` is not installed in this environment

The custom implementation previously matched the HF path to 1e-4 loss at 0.8B and 9B in the linked stacked-PR runs, with mismatch KL 0.0003 across 20 RL steps.

Made with [Cursor](https://cursor.com)


hubert-marek:
Superseded by #2943; keeping the existing stacked PR and rebasing it onto main after its prerequisite merges.

samsja (approved):

PR #2988, opened 2026-07-10
feat(sft): require renderer-based tokenization

hubert-marek:
## Summary

Split from #2942 (part 1/7 of #2485) — renderer-only SFT tokenization, without the local `data_files` ingestion that ships separately.

Makes SFT tokenization renderer-only and deletes the broken incremental masking path:

- **Renderer-only SFT**: `build_training_sample` with a hand-coded renderer replaces the incremental `apply_chat_template` masking path, which corrupted loss masks under position-dependent chat templates (e.g. Qwen3 stripping past thinking blocks across user turns). `DefaultRenderer` is rejected in `setup_renderer` (it falls back to incremental tokenization and doesn't fix the problem); fake data skips renderer construction so models without a hand-coded renderer can still benchmark step time. Non-assistant roles opt into the loss via the renderer's body-only path (`content_sft_roles`).
- **Null-safe message resolution**: `messages` takes precedence over `prompt`/`completion` only when non-null — Arrow schema union adds `messages: null` to prompt/completion rows whenever other rows have a `messages` column, and the previous key-presence check silently resolved those rows to empty message lists.
- `_drop_null_fields` strips Arrow phantom nulls before tool-call deserialization so genuine nulls inside argument strings survive.
- Defaults: `seq_len` 256 (FakeDataConfig keeps 128), `loss_impl` `liger_fused`; CI + example reverse-text/wordle configs pin `[renderer] name = "qwen3"` (PrimeIntellect/* fine-tunes miss the auto-renderer's exact-match lookup).

## Tests

All 18 tests in `test_sft_dataset.py` pass:
```
18 passed, 3 warnings in 18.17s
```

---

> [!NOTE]
> **Medium Risk**
> Changes core SFT tokenization and loss masking for all non-fake runs; misconfigured or unmapped renderers now fail at validation, but behavior shifts away from incremental masking for every real dataset.
> 
> **Overview**
> **SFT is renderer-only for real data.** `SFTDataset` always takes a `Renderer` and builds samples via `build_training_sample` (sampled-mask loss, optional `content_sft_roles`, `ensure_final_stop`). The legacy incremental `apply_chat_template` / `build_incremental_token_mask` path and related helpers are removed from `chat_template.py`.
> 
> **Config defaults and guards.** `[renderer]` defaults to `auto` (was optional `None`). `validate_typed_renderer` rejects `default` or unmapped auto resolution when training or validation uses real samples; fake-only runs can still use `default`. VLM SFT is blocked whenever `model.vlm` is set.
> 
> **Trainer wiring.** A renderer is created when `data.type != "fake"` or `[val]` is set; fake-only training skips it for benchmarking. Runtime `DefaultRenderer` rejection in `train.py` moves to config validation.
> 
> **Dataset robustness.** Rows use non-null checks for `messages` vs `prompt`/`completion` (Arrow union null `messages`). `_drop_null_fields` strips phantom nulls before rendering while preserving nulls inside tool-call argument payloads.
> 
> **Examples/docs.** Training docs describe renderer-only SFT and `[renderer]` TOML controls; Wordle RL switches orchestrator renderer to `prime-qwen3`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b2b39f806e03ee2acf2dae90e2de851d0d7cdaea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


samsja (approved):

samsja (dismissed):

src/prime_rl/trainer/sft/train.py
```diff
@@ -162,17 +161,12 @@ def train(config: SFTConfig):
     logger.info(f"Initializing tokenizer ({config.tokenizer})")
     tokenizer = setup_tokenizer(config.tokenizer)
 
+    # Fake data never renders messages, so a model without a hand-coded renderer
+    # can still be used to benchmark step time / memory. Validation data is
+    # always real, so it needs the renderer even when training data is fake.
     renderer = None
-    if config.renderer is not None:
+    if config.data.type != "fake" or config.val is not None:
         renderer = create_renderer(tokenizer, config.renderer)
```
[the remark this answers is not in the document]
hubert-marek: @hallerite should we add a custom renderer for `PrimeIntellect/Qwen3-0.6B and 1.7B` or can I change those tests to normal qwen or we don't care?

samsja (dismissed):

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -51,6 +52,12 @@ def validate_batch_size(self):
 class FakeDataConfig(BaseDataConfig):
     type: Literal["fake"] = "fake"
 
+    seq_len: int = Field(128, ge=1)
+    """Sequence length."""
+
+    pack_function: Literal["cat", "stack"] = "cat"
+    """Sample packing strategy."""
```
samsja: hmmm why change here

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -225,8 +227,8 @@ class SFTConfig(BaseConfig):
     dist_timeout_seconds: int = 3600
     """Timeout in seconds for torch distributed ops."""
 
-    loss_impl: Literal["liger", "torch", "liger_fused", "quack_fused"] = "torch"
-    """Cross-entropy loss implementation. ``liger_fused`` fuses the lm_head projection with the CE loss to avoid materializing full logits. ``quack_fused`` uses quack-kernels for chunked linear + CE with CuTe DSL CUDA kernels."""
+    loss_impl: Literal["liger", "torch", "liger_fused", "quack_fused"] = "liger_fused"
+    """Cross-entropy loss implementation. Defaults to fused Liger loss to avoid materializing full logits."""
```
samsja: why change here, not about this pr

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -271,6 +273,28 @@ def validate_deployment(self):
...
         return self
 
+    @model_validator(mode="after")
+    def validate_typed_renderer(self):
+        """Require a typed renderer whenever SFT renders real samples."""
+        if self.data.type == "fake" and self.val is None:
+            return self
+
+        model_id = self.tokenizer.name or self.model.name
+        if isinstance(self.renderer, AutoRendererConfig):
```
samsja, line 277: in which case is it not instance of AutoRenderConfig ? shouldn;t it be just if self.renderer is not None ?

hubert-marek: It needs to resolve later by acutally checking if there is an renderer to be selected by auto

src/prime_rl/trainer/sft/data.py
```diff
@@ -114,38 +108,52 @@ def __iter__(self):
             yield fake_sample
 
 
+def _drop_null_fields(value: Any, path: tuple[str, ...] = ()) -> Any:
```
hallerite, line 111: hmm

hallerite (approved):

src/prime_rl/trainer/sft/data.py
hubert-marek: https://arrow.apache.org/docs/python/generated/pyarrow.unify_schemas.html

PR #3004, opened 2026-07-11
fix(trainer): broadcast resumed policy weights before training loop

samsja:
On resume, the orchestrator advances to v{checkpoint_step} and pauses inference to receive those weights over NCCL. Since #2896 moved the in-loop weight broadcast to the END of a step, the trainer would start step {checkpoint_step+1} first — which blocks on rollouts the paused inference engines cannot produce until they get v{checkpoint_step}. Both sides then wait forever (confirmed on GLM-5.2 resume: orchestrator "Created NCCL_READY marker" then silent, trainer stuck "Looking for batches" for the next step).

Restore the pre-#2896 top-of-loop sync with a one-shot broadcast right after checkpoint load. Runs are only registered during batch collection (packer.discover_runs), so discover run 0 first (mirroring optim.py's wait-for-run-0 pattern) to populate used_idxs before broadcasting.

---

> [!NOTE]
> **Medium Risk**
> Touches critical trainer/orchestrator startup and weight-sync ordering; mis-timed broadcasts could still deadlock or serve wrong policy versions, but the change aligns both sides symmetrically.
> 
> **Overview**
> Fixes a **resume (and fresh-start) deadlock** where the trainer waited for rollouts before inference had the policy version those rollouts require, after in-loop weight sync moved to the **end** of each step (#2896).
> 
> The **trainer** now does a **one-shot broadcast** of `v{progress.step - 1}` at the **top of the first loop iteration** (after checkpoint load), including `wait_for_run(0)` and marking runs ready—**before** `dataloader.wait_for_batch()`.
> 
> The **orchestrator** always **syncs inference at startup** (v0 from scratch or v{resume_step} on resume), with a **1200s default** wait for the trainer’s startup broadcast when `ckpt.wait_for_weights_timeout` is unset.
> 
> **`MultiRunManager.wait_for_run`** centralizes the “discover run 0 from disk” poll (used by optimizer setup and startup broadcast). **RL config** now **rejects LoRA + explicit NCCL** weight broadcast with a clear error (filesystem required for adapters).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4dfdbaaad2a54ea62602778028098eda72f3b918. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/trainer/rl/train.py
```diff
@@ -226,6 +226,24 @@ def load_run_checkpoint(_optimizer, idx: int) -> None:
         progress.step += 1
         logger.info(f"Resuming training from checkpoint step {checkpoint_step}")
 
+        # Sync inference to the resumed policy (v{checkpoint_step}) BEFORE the loop. On resume the
+        # orchestrator advances to v{checkpoint_step} and pauses inference to receive those weights
+        # over NCCL. Since #2896 moved the in-loop broadcast to the END of a step, the trainer would
+        # otherwise wait for step {checkpoint_step+1} rollouts the paused engines cannot produce, so
+        # both sides deadlock. This one-shot broadcast restores the pre-#2896 top-of-loop sync.
+        if weight_broadcast is not None:
```
mikasenghaas: lol when is this false?

src/prime_rl/trainer/rl/train.py
```diff
@@ -226,6 +226,24 @@ def load_run_checkpoint(_optimizer, idx: int) -> None:
...
+            logger.info(f"Broadcasting resumed policy weights (v{checkpoint_step}) to inference engines")
+            # Runs are only registered during batch collection (packer.discover_runs), which hasn't
+            # happened yet before the loop. Discover run 0 now so used_idxs is populated (mirrors optim.py).
```
mikasenghaas: oufff

mikasenghaas (commented):
can not solve this by *always* broadcasting weights on the first step (whether resume or not)

samsja:
Good call — reworked to always broadcast on the first step instead of the resume-only special case.

- **trainer**: broadcasts `v{step-1}` at the top of the loop's first iteration — fresh → `v0`, resume → `v{checkpoint_step}`. No more pre-loop block or `if resume` branch.
- **orchestrator**: now *always* pauses + syncs inference at startup (was resume-only). This is the piece that makes "whether resume or not" actually work: on a fresh start inference isn't paused, so an unconditional trainer broadcast would hang with no `NCCL_READY` rendezvous. Syncing `v0` at startup keeps both sides symmetric.
- the `while 0 not in idx_2_id` loop is gone from the diff — factored into `MultiRunManager.wait_for_run()` and reused (also dedupes the copy that was already in `optim.py`).

Resume path validated end-to-end on GLM-5.2 (broadcast fires → inference resumes → step 11 trains). Fresh-start path is covered by the symmetric orchestrator sync but I haven't run a from-scratch job with it yet — worth a look in review/CI.

PR #3014, opened 2026-07-13
fix(trainer): complete Nemotron-H adapter-to-HF key mapping

snimu:
## Problem

`NemotronHPreTrainedModel.convert_adapter_to_hf` renames adapter keys from the custom trainer namespace to Hugging Face Nemotron-H names, but only had two rules: `.mlp. → .mixer.` and `.self_attn. → .mixer.`. Missing (compare the full-weights converter in `converting_nemotron_h.py`, which the adapter path does not use):

- `.mamba. → .mixer.` — adapter keys on Mamba projections export as `...mamba.in_proj.lora_A...`, matching no HF module;
- `mlp.shared_expert → mixer.shared_experts` — the generic mlp rule produces the singular `mixer.shared_expert`, also matching nothing;
- `model. → backbone.` — HF Nemotron-H roots at `backbone`, so every exported key kept the wrong prefix.

This converter is not export cosmetics: `setup_lora` registers it for every `PreTrainedModelPrimeRL`, and it runs inside `MultiRunManager.get_state_dict_for_run`, which feeds the ordinary LoRA weight broadcast (`trainer/rl/broadcast/filesystem.py`) and LoRA checkpoint export (`ckpt.py::get_run_adapter_state_dict`). Keys that don't match HF module names are silently dropped by vLLM/PEFT — a LoRA RL run on Nemotron-H would sample from a policy whose adapter weights never (or only partially) update. Latent today (no repo config combines Nemotron-H with LoRA), but Nemotron-H is a supported model family.

## Fix

Extend the rename table to five rules, mirroring `converting_nemotron_h.py`'s key grammar:

1. `(.layers.N).mlp.shared_expert. → .mixer.shared_experts.` — ordered before the generic mlp rule so the pluralization happens exactly once;
2. `(.layers.N).mlp. → .mixer.` (existing);
3. `(.layers.N).self_attn. → .mixer.` (existing);
4. `(.layers.N).mamba. → .mixer.`;
5. `^model. → backbone.` — anchored so only the root prefix is rewritten and `lora_A`/`lora_B` suffixes can't be touched.

The PEFT `base_model.model.` prefix is added *after* this converter runs (`ckpt.py`), yielding `base_model.model.backbone.layers...`, which matches HF Nemotron-H's top-level `backbone` attribute.

## Test

Added `test_nemotron_h_adapter_key_conversion` to the existing Nemotron-H suite: pure dict-in/dict-out over five representative adapter keys (Mamba projection, shared expert, latent projection, per-expert MoE key, attention projection) asserting the exact converted key set. Note the module carries `pytest.mark.gpu`, so it runs with the GPU suite.

## Caveat

Statically mirrors the in-repo full-weights converter (the best available ground truth); an end-to-end validation with a real Nemotron-H checkpoint + PEFT load is still worthwhile before relying on Nemotron-H LoRA serving.

---

> [!NOTE]
> **Medium Risk**
> Changes affect trainer–orchestrator startup coordination and LoRA weight paths for Nemotron-H; wrong gating could stall NCCL runs or leave filesystem runs unsynced on resume if misconfigured.
> 
> **Overview**
> **Nemotron-H LoRA export** extends `convert_adapter_to_hf` so trainer-side adapter keys match the live HF module tree: `.mamba.` and `.self_attn.` → `.mixer.`, `mlp.shared_expert` → `mixer.shared_experts` (before the generic `.mlp.` rule), while keeping the `model.` root prefix. A unit test locks the expected key set for Mamba, MoE, and attention LoRA tensors.
> 
> **Startup policy sync** is no longer unconditional. The orchestrator loads/syncs weights before step 1 only on **checkpoint resume** or **NCCL** weight broadcast; fresh **filesystem** runs skip waiting for `broadcasts/step_0` because v0 is already the base model and late-registered multi-run jobs would never see a one-shot startup broadcast. The trainer mirrors this by gating the first-step **NCCL** startup broadcast on `weight_broadcast.type == "nccl"`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 04d88c5c153026a0c16fb5a405cf080268b43325. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -323,25 +323,23 @@ class Orchestrator:
         else:
             get_logger().info("Training from scratch")
 
-        # Sync inference to the incoming policy before the first step, whether fresh (v0 = base) or
-        # resumed (v{resume_step}). This pauses inference and rendezvouses with the trainer's
-        # first-step broadcast (train.py), so inference serves the correct policy from step 1. Keeping
-        # it unconditional makes trainer/orchestrator startup symmetric and avoids the resume deadlock
-        # where inference waited for weights the trainer only sent at the end of a step.
-        sync_version = self.resume_step if self.resume_step is not None else 0
-        check_exists = config.weight_broadcast.type != "nccl"
-        # Wait for the trainer's startup broadcast to land. On a fresh start there is no ckpt block,
-        # so fall back to a default timeout instead of not waiting at all (which would raise before
-        # broadcasts/step_0/STABLE is written when using filesystem weight broadcast).
-        wait_timeout = config.ckpt.wait_for_weights_timeout if config.ckpt else STARTUP_WEIGHT_WAIT_TIMEOUT_S
-        weights_path = get_weight_dir(
-            config.output_dir, sync_version, check_exists=check_exists, wait_timeout=wait_timeout
-        )
-        await self.policy_inference.update_weights(weights_path, lora_name=self.lora_name, step=sync_version)
-        if self.lora_name is not None:
-            self.policy_inference.update_model_name(self.lora_name)
-            self.policy.model_name = self.lora_name
-        self.policy.version = sync_version
+        # Sync inference to the incoming policy before the first step when resuming or when using
+        # NCCL, which rendezvouses with the trainer's first-step broadcast (train.py). A fresh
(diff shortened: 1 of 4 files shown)


samsja:
closing as only for lora

PR #3026, opened 2026-07-14
fix(utils): revalidate NFS directory caches in wait_for_path polls

faresobeid:
## Problem

On multi-node deployments with NFS-backed shared storage, a file created on one node stays invisible to another node's `path.exists()` for up to `acdirmax` (commonly **60 seconds**), because the NFS client caches directory lookups (negative dentries + attributes). Every `wait_for_path` poll site pays this as a fixed, invisible latency:

- **`FileSystemMicroBatchReceiver.wait`** — remote trainer ranks poll for their per-rank micro-batch files written by the master on another node, and discover them up to 60s late. Result: a constant **~60s added to every training step**. We measured `time/wait_for_batch` at 60.013s ± 1ms across consecutive steps — a timer signature, not workload — while the batch files' mtimes predated the waits by minutes. On a ~4-minute step that's ~25% of step time.
- **NCCL_READY marker waits** during weight broadcast (trainer ranks polling marker files written by inference nodes) — inflates `time/broadcast_weights`.
- **Orchestrator watcher** awaiting broadcast `STABLE` markers — delays policy adoption by inference when the orchestrator isn't colocated with the trainer master.

## Fix

READDIR the target's parent directory before each existence check. A directory read forces the NFS client to revalidate and drop stale negative entries — the standard "`ls` makes the file appear" behavior. Implemented as a small `_revalidate_dir` helper used by both `sync_wait_for_path` and the async `wait_for_path`.

## Safety / regression review

- Read-only syscall; loop exit condition unchanged; `OSError` (e.g. parent not created yet) is ignored, degrading to exactly the previous behavior — verified with a missing-parent functional test.
- Parents at all call sites are small step/marker directories (tens of entries; a READDIR is ~ms). The tightest poll site (0.1s cadence during broadcast handshake) adds ~10 tiny RPCs/s for a few seconds.
- No-op cost on local filesystems; portable (plain `os.listdir`).
- The async variant already performs blocking `path.exists()` inline; the added call matches that existing pattern and cost profile.

## Validation

A/B/A in production GLM-4.5-Air multi-node runs (2 trainer nodes, NFS home): with this fix deployed, the instrumented batch-wait sync phase measured **2–7s**; after removing it, the constant **60.0s** plateau returned; re-deploying restored 2–7s. Net ~16% wall-clock on our runs.

**Alternatives considered** (why this layer): `lookupcache=positive` mount option is the correct ops-level fix but the codebase can't assume users' mounts; TCPStore-signaling doesn't help because `open()` hits the same cached negative dentry after the signal; the ZMQ transport avoids NFS entirely but is a much larger change covering only one of the three affected sites. The poll sites check-first and revalidate only on a miss, so the fix costs nothing when files are already visible.

---

> [!NOTE]
> **Low Risk**
> Read-only, best-effort cache refresh with unchanged exit conditions; degrades to previous behavior if the parent is missing.
> 
> **Overview**
> **`sync_wait_for_path` and `wait_for_path`** now call **`os.listdir` on the target’s parent** before each existence check so NFS clients drop stale negative directory cache entries (files written on another node can otherwise stay invisible for ~60s).
> 
> **`OSError` from a missing parent is suppressed**, so behavior matches the prior loop when the directory is not created yet. Call sites (micro-batch files, NCCL ready markers, broadcast **`STABLE`** markers) are unchanged; only the poll path is updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e851f549d38b2d407596c029e25c79d15eb3707. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/utils/pathing.py
+++ b/src/prime_rl/utils/pathing.py
@@ -203,6 +203,11 @@ def sync_wait_for_path(path: Path, interval: int = 1, log_interval: int = 10) ->
     wait_time = 0
     logger.debug(f"Waiting for path `{path}`")
     while True:
+        # Check first (free when the file is already visible), revalidate only on a miss —
+        # the miss may be a stale NFS negative-dentry rather than true absence.
+        if path.exists():
+            logger.debug(f"Found path `{path}`")
+            break
         _revalidate_dir(path)
         if path.exists():
             logger.debug(f"Found path `{path}`")
@@ -218,6 +223,11 @@ async def wait_for_path(path: Path, interval: int = 1, log_interval: int = 10) -
     wait_time = 0
     logger.debug(f"Waiting for path `{path}`")
     while True:
+        # Check first (free when the file is already visible), revalidate only on a miss —
+        # the miss may be a stale NFS negative-dentry rather than true absence.
+        if path.exists():
+            logger.debug(f"Found path `{path}`")
+            break
         _revalidate_dir(path)
         if path.exists():
             logger.debug(f"Found path `{path}`")
--- a/src/prime_rl/utils/pathing.py
+++ b/src/prime_rl/utils/pathing.py
@@ -1,4 +1,5 @@
 import asyncio
+import os
 import shutil
 import time
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


src/prime_rl/utils/pathing.py
```diff
@@ -179,11 +180,35 @@ def clean_future_steps(output_dir: Path, resume_step: int) -> None:
...
     logger = get_logger()
     wait_time = 0
     logger.debug(f"Waiting for path `{path}`")
     while True:
+        # Check first (free when the file is already visible), revalidate only on a miss —
+        # the miss may be a stale NFS negative-dentry rather than true absence.
+        if path.exists():
+            logger.debug(f"Found path `{path}`")
+            break
+        _revalidate_dir(path)
```
mikasenghaas: can we just `os.listdir(path.parent)` here and below? should be +2loc

mikasenghaas: can put a small comment like # refresh nfs cache

faresobeid: Done — inlined at both sites with a one-line comment. Kept a `contextlib.suppress(OSError)` around it: the parent dir often doesn't exist yet when the wait starts (e.g. remote ranks start polling for rank files before the master creates the step dir), and a bare `listdir` would raise there.

PR #3042, opened 2026-07-15
feat(trainer): ulysses CP beyond num_kv_heads via GQA KV-head replication

mikasenghaas:
## Problem

Ulysses CP head-shards Q/K/V, capping the CP degree at `num_key_value_heads`. Hybrid-Mamba GQA models — NemotronH Super/Ultra/Nano all have **2 KV heads** — are therefore stuck at `cp <= 2` (ring CP is softmax-only and rejects Mamba). At cp=2, 131k context puts 65k tokens/GPU through Super's top-22/512-expert MoE and is memory-infeasible on H200 regardless of node count. NVIDIA post-trained Super at 256k–512k with CP=64 (Megatron's sequence-sharded, Mamba-aware CP).

## Approach

**GQA KV-head replication** (the DeepSpeed-Ulysses GQA scheme), ~20 lines in `ulysses_attn.py`:

- When `num_kv_heads < cp`, replicate each KV head `cp / num_kv_heads` times (`repeat_interleave`) before the *existing* seq→head all-to-all.
- GQA groups query heads contiguously: after the Q all-to-all, rank `r` holds query heads `[r·Hq/cp, (r+1)·Hq/cp)`, which all attend KV head `⌊r·Hkv/cp⌋` — exactly the replica the all-to-all lands on rank `r` (index algebra in the docstring).
- Backward of `repeat_interleave` sums gradients over replicas — the exact GQA KV gradient.
- Q path and the `num_kv_heads >= cp` case are byte-identical to before. Ring path untouched. Mamba layers already head-shard independently via `cp_mamba` (NemotronH: 128 mamba heads, `n_groups=8` → mamba caps cp at 8).
- Constraints (asserted): `cp % num_kv_heads == 0` (replication path), `num_attention_heads % cp == 0` (existing).
- The RL trainer uses the same `substitute_ulysses_attn` and inherits the fix.

Separate commit: **fuse `relu2` via `torch.compile`** — the eager `F.relu(x).square()` materialized an extra full-size copy of the MoE intermediate (the empirical OOM site for Super; the expert path graph-breaks out of the layer-level compile at `@expert_parallel` / `_grouped_mm`). Measured: Super @16k tokens/GPU on 2 nodes goes from OOM to 126.5/139.8 GiB peak.

## Verification

[...]

PR #3045, opened 2026-07-15
fix(trainer): stage CPU-offloaded weights through CUDA in NCCL broadcast

mikasenghaas:
Follow-up to #3042 (found during bring-up of the first run to combine the two settings; the commit just missed the merge).

With `fsdp_cpu_offload` the gathered state-dict tensors live on CPU, and `broadcast_state_dict` passed them to the NCCL communicator directly:

```
AssertionError: this nccl communicator is created to work on cuda:0, but the input tensor is on cpu
```

Pre-existing: RL never combined `fsdp_cpu_offload` (params on CPU) with NCCL weight broadcast before — muon (the usual RL optimizer) rejects FSDP CPU offload, and it's the memory profile that makes 196k-context NemotronH runs fit on 2 nodes (adamw). Stage the per-layer, per-dtype concatenated buffer through CUDA before broadcasting (+4 lines; a transient per-layer GPU buffer, ~3 GB max for a 120B/88-layer model).

Verified: the 196k cp=8 Nemotron Super RL run that hit this assert proceeds past weight broadcast with the fix (run in flight — will confirm rollouts/step 1 here).

### DIFF
--- a/src/prime_rl/trainer/rl/broadcast/nccl.py
+++ b/src/prime_rl/trainer/rl/broadcast/nccl.py
@@ -61,6 +61,10 @@ def broadcast_state_dict(state_dict: dict[str, Tensor], communicator: PyNcclComm
         # Flatten all tensors and concatenate
         flat_tensors = [value.flatten() for _, value in items]
         concatenated = torch.cat(flat_tensors)
+        if not concatenated.is_cuda:
+            # fsdp_cpu_offload keeps parameters on CPU; NCCL requires device tensors,
+            # so stage the (per-layer, per-dtype) buffer through CUDA transiently.
+            concatenated = concatenated.cuda()
         communicator.broadcast(concatenated, src=0)
         del concatenated
         # Clean up individual tensors
--- a/src/prime_rl/trainer/rl/broadcast/nccl.py
+++ b/src/prime_rl/trainer/rl/broadcast/nccl.py
@@ -168,7 +168,13 @@ class NCCLWeightBroadcastSender:
     def _resolve_dtensors(self, state_dict: dict[str, Tensor]) -> dict[str, Tensor]:
         for key, value in list(state_dict.items()):
             if isinstance(value, DTensor):
-                state_dict[key] = cast(DTensor, value.to(self.dtype)).full_tensor()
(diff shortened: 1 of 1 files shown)


mikasenghaas:
Superseded by the removal PR — team decided to remove fsdp_cpu_offload entirely rather than fix it.

mikasenghaas:
Reopening: #3049 (fsdp_cpu_offload removal) is being reverted in #3052 after 2-node long-context measurements, so this staging fix is needed again — it's the only fast weight-broadcast path with FSDP CPU offload (filesystem broadcast full-gathers CPU-resident fp32 weights over Gloo and effectively hangs; measured on a Nemotron Super 131k run).

mikasenghaas:
Second fix pushed (a6bdfe8): the original staging happened *after* `_resolve_dtensors`, but `full_tensor()` on CPU DTensors already gathers over **Gloo** — measured ~1h+ stall at the v0 broadcast for Super's 480 GB fp32 until the NCCL watchdog killed the job. Now each shard moves to CUDA before `full_tensor()` so the gather runs over NCCL per tensor. Validating e2e on the Super 131k run.

mikasenghaas:
Status: the per-tensor CUDA staging (a6bdfe8) is insufficient — `DTensor.to("cuda")` moves the local shard but the DeviceMesh stays CPU/Gloo, so `full_tensor()` still gathers over Gloo (observed: ranks stuck 4.5h in the gather, no exception). A correct fix needs the gather itself to run on a CUDA mesh (or explicit NCCL collectives on staged shards) — leaving this PR as WIP; the immediate runs proceed without fsdp_cpu_offload instead.

mikasenghaas:
Deprioritized on our side: even with async pinned staging (bb8a777, unvalidated — test run cancelled), FSDP CPU offload implies re-staging the full CPU-resident weights every step's broadcast, a recurring per-step tax the team judged not worth it for RL-with-inference. The branch is rebased onto main with the mesh-correct NCCL gather + async staging if anyone wants to pick it up; trainer-only/SFT users of fsdp_cpu_offload are unaffected (no broadcast).

PR #3068, opened 2026-07-17
fix(trainer): skip no-op weight conversion for dense models

samsja:
## Problem

Training a dense (non-MoE) model like `Qwen3-0.6B-Reverse-Text-SFT` from a read-only HF model cache crashes at startup:

```
OSError: [Errno 30] Read-only file system: '/model-cache/models--PrimeIntellect--Qwen3-0.6B-Reverse-Text-SFT/snapshots/c97a910849ec6aa962add3dc253a0817d61c0210/prime'
```

## Root cause

The auto-conversion logic in `load_dcp_from_hf` is designed for MoE models, where HF and PrimeRL store expert weights differently (per-expert tensors `expert.0.w1` vs stacked `experts.w1`). It detects a format mismatch, converts, and writes the result to `<snapshot>/prime`.

For **dense models** (non-MoE Qwen3, Llama, etc.), `is_hf_state_dict` and `is_prime_state_dict` both unconditionally return `True` — the key names are identical in both formats, so the same state dict is valid in both. This causes the conversion branch to fire anyway, performing a no-op `convert_to_prime` and trying to `save_state_dict` the unchanged weights into the snapshot directory. When the HF cache is mounted read-only (as in Docker/cluster environments), this crashes with `OSError: [Errno 30] Read-only file system`.

## Fix

Add `not` guards on each conversion branch so conversion only triggers when the snapshot format **genuinely differs** from the model format:

- **HF → Prime branch**: only if `is_hf_state_dict(snapshot)` AND `NOT is_prime_state_dict(snapshot)` — i.e. the snapshot is HF-only and needs conversion.
- **Prime → HF branch**: only if `is_prime_state_dict(snapshot)` AND `NOT is_hf_state_dict(snapshot)` — i.e. the snapshot is Prime-only and needs conversion.

When both formats match (dense models), both branches are skipped and the code loads directly from `source_path` — no conversion, no write.

## Impact on MoE models

MoE models are **unaffected**. Their `is_hf_state_dict` / `is_prime_state_dict` checks are mutually exclusive (they inspect for `expert.{e}.` vs `experts.` key patterns), so the `not` guards are always satisfied when a genuine format difference exists.

| Model type | `is_hf_state_dict(sd)` | `is_prime_state_dict(sd)` | Both can be True? | Behavior with fix |
|---|---|---|---|---|
| MoE | pattern-specific check | pattern-specific check | No | Converts when mismatch ✅ |
| Dense | always `True` | always `True` | Yes | Skips conversion ✅ |

## Validation

- `uv run ruff check src/prime_rl/trainer/model.py` ✅
- Python syntax check ✅
- Existing MoE conversion tests (`test_nemotron_h_weight_conversion_roundtrip`, etc.) test the conversion methods themselves, which are unchanged — only the branching conditions were modified.

## Workaround note

PR #3023 added a `conversion_dir` config to redirect converted weights to a writable path. This fix addresses the root cause: the conversion should not happen at all when there is no format difference to convert.

---

> [!NOTE]
> **High Risk**
> Removes SDPA/eager fallbacks across all custom architectures and changes default attention and weight-load behavior; misconfigured environments without flash-attn will fail at startup instead of degrading gracefully.
> 
> **Overview**
> Trainer **`model.attn`** now only accepts **`flash_attention_2`**, **`flash_attention_3`**, and **`flash_attention_4`** (default **FA3**). **`eager`**, **`sdpa`**, and the internal **`fa4`** alias are removed, along with the transformers **`fa4`** dummy registration.
> 
> Custom model stacks route attention exclusively through **Flash varlen** paths: shared **`SDPAAttention`** and model-specific SDPA classes (AFMoE, Laguna, Qwen3.5 gated) are deleted, **`_supports_sdpa`** is set false, and CP/ring/ulysses wiring uses **`flash_attention_4`** naming consistently. **`PreTrainedModelPrimeRL`** overrides **`_check_and_adjust_attn_implementation`** so init does not require installed flash kernels on CPU-only hosts.
> 
> **`load_dcp_from_hf`** only runs HF↔Prime conversion when the snapshot is genuinely one format or the other (guards with **`not snapshot_is_prime`** / **`not snapshot_is_hf`**), avoiding no-op writes into read-only dense-model caches. **GPT-OSS** no longer suggests **`eager`** on non-Hopper GPUs—it errors if flash attention cannot run on the device.
> 
> Unit tests and debug configs switch from **`sdpa`** to **`flash_attention_2`** where flash is required.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0bff16dc68235a2bac75f7b165657711017f8e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/configs/debug/moe/sft/train.toml
+++ b/configs/debug/moe/sft/train.toml
@@ -4,7 +4,7 @@ max_steps = 5
 name = "PrimeIntellect/GLM-0.5B" 
 moe_use_grouped_mm = false
 impl = "custom"
-attn = "sdpa"
+attn = "flash_attention_2"
 
 [model.compile]
 
--- a/packages/prime-rl-configs/src/prime_rl/configs/trainer.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/trainer.py
@@ -18,14 +18,8 @@ from prime_rl.utils.config import BaseConfig
 
 # -- Shared trainer configs (used by both SFT and RL trainers) --
 
-AttnImplementation: TypeAlias = Literal["eager", "sdpa", "flash_attention_2", "flash_attention_3", "fa4"]
+AttnImplementation: TypeAlias = Literal["sdpa", "flash_attention_2", "flash_attention_3", "flash_attention_4"]
 EPCommBackend: TypeAlias = Literal["torch", "deepep"]
-
-# User-facing name -> internal name. Users set `flash_attention_4` in configs,
-# which gets rewritten to `fa4` before pydantic validation.
-# We use `fa4` internally because `flash_attention_*` triggers transformers
-# to attempt installing a kernel from hub.
-_ATTN_ALIASES = {"flash_attention_4": "fa4"}
 
 
 class GCConfig(BaseConfig):
@@ -123,7 +117,7 @@ class ModelConfig(BaseModelConfig):
     seq_len: int = 2048
     """Sequence length the model is trained on."""
 
-    attn: AttnImplementation = "flash_attention_2"
+    attn: AttnImplementation = "flash_attention_3"
     """Attention implementation. With CP enabled, ring attention uses the matching kernel family (FA2/FA3/FA4)."""
 
     compile: CompileConfig | None = CompileConfig()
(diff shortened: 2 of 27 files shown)


samsja:
Closing in favor of a rebased version on top of main.

PR #3032, opened 2026-07-14
feat(trainer): use seq_lens as the packed-sample boundary contract

hubert-marek:
## Summary

Rebased replacement for #2944, now targeting `main` directly.

- Remove the unused `stack` SFT path and `pack_function`; SFT now has one cat-packed data path with an explicit single-row contract.
- Carry required `seq_lens` through internal SFT/RL batches, transport, dataloaders, and trainer forward. Generic Hugging Face models never receive PrimeRL-only kwargs.
- Derive custom-model FlashAttention `cu_seqlens` only from explicit lengths; public custom-model forwards require keyword-only `seq_lens` and never infer boundaries from `position_ids`.
- Rework SFT cat packing to preserve overflow samples, emit the tail pack, and checkpoint the read-ahead sample so resume cannot skip data.
- Merge trailing padding directly into the final `seq_lens` segment when batches are built, so models receive valid attention lengths without separate padding metadata or forward-time normalization.
- Dense and MoE Qwen3.5 consume boundaries for non-flash linear-attention state resets. Under existing CP, trainers use global lengths for attention setup and derive explicit shard-local lengths for custom model forwards.

## Tests

- GitHub CI: CPU/GPU unit jobs, Ruff, CodeQL, and all integration jobs passed
- Final-stack Slurm H100: reverse-text SFT 4/4, CP-focused 10/10, and RL boundary/packing 35/35 passed; broader model coverage passed 66 tests, with two Qwen3.5 MoE backward tests blocked by the node image missing `cuda/atomic` for TileLang compilation
- Cleanup validation: 30 batch/sequence tests and 22 SFT dataset tests passed; Ruff passed on all changed files
- Direct checkpoint/resume simulation confirms resumed packed output exactly matches uninterrupted execution
- H100 control: resumed steps 11–20 matched uninterrupted training exactly (max loss difference `0.0`); PR and `main` had nearly identical convergence slopes
- Resume integration runs through step 20 so the assertion covers ten post-checkpoint updates
- H100 boundary microbenchmark: explicit `seq_lens` took ~82–83 µs versus ~138–143 µs for position-reset inference across 1K–16K and packed rows

Historical end-to-end validation from #2944:
- Dense and MoE text SFT: 10/10 steps
- RL text run with `seq_lens` end-to-end: 10/10 steps

---

> [!NOTE]
> **High Risk**
[...]

PR #3075, opened 2026-07-17
feat(sft): multimodal (VLM) SFT

hubert-marek:
## Summary

Rebased replacement for #2945, now targeting `main` after #3032 merged.

- Carry renderer-produced multimodal samples end to end through SFT, including image-safe truncation, `mm_token_type_ids`, multimodal processor outputs, and cat packing.
- Load the model processor, attach it to the renderer, and save it with weight checkpoints.
- Add the dense Qwen3.5 composite VLM body (HF vision encoder + custom PrimeRL text model) with MRoPE positions derived from renderer metadata.
- Require custom PrimeRL implementations for VLM training and keep vision collectives symmetric across text-only and multimodal microbatches.
- Freeze vision encoders for text-only VLM training and fail invalid LoRA configurations with no matching targets.
- Enforce the VLM dtype, micro-batch-size, LoRA, and no-CP constraints in SFT config validation.
- Update the multimodal nightly config and add dense and MoE VLM SFT examples.

The branch preserves #3032's explicit `seq_lens` contract: trailing padding is merged into the final sequence segment and no separate padding metadata is transported.

## Validation

- `23 passed` — existing SFT dataset suite after the image-safe truncation fix
- `32 passed, 5 deselected` — focused non-GPU VLM model/forward/SFT tests using an import-only `ring_flash_attn` shim; no attention kernels were exercised
- `41 passed, 80 deselected` — config unit tests excluding external-plugin config loading
- Ruff format/check and diff checks passed
- Historical multimodal KL run: [custom Qwen3.5 VLM](https://wandb.ai/primeintellect/pr2485/runs/11cf49aa353d456fb5892a1944b30260), mean mismatch KL `0.0138`, step-0 KL `0.0025`
- Historical HF control: [Qwen3-VL](https://wandb.ai/primeintellect/pr2485/runs/4902c84180644e8ba8d5416bbab5a3c3), mean mismatch KL `0.1145`

Post-rebase CI fixes: the exact `examples/vlm_sft_moe/sft.toml` config test and an `sft --dry-run` validation both pass. The nightly RL VLM config now sets `impl = "custom"` in `[trainer.model]`, and obsolete GPU tests for removed modeling-layer validation were dropped.

Latest main merge: merged `main` at `445392294`; the README overlap retains the current multimodal docs anchor and keeps VLM CP marked unsupported, while main's obsolete debug-config deletion is preserved.

Cleanup: multimodal CUDA transfer is inlined in the singleton SFT collate path; the single-use helper was removed.

API cleanup: `[model.vlm]` now requires `model.impl = "custom"`. The dense and MoE custom VLM bodies accept token IDs only, share embedding accessors with their text-only bodies, and no longer carry unsupported `inputs_embeds` fallbacks or redundant modeling-layer validation. Multimodal samples without `[model.vlm]` still fail at the lazy dataset boundary.

Truncation follow-up: multimodal samples are now causal-shifted before image-safe truncation, keeping placeholder ranges, modality IDs, targets, and processor sidecars on the same model-input boundary.

Review follow-up: the multimodal processor is now loaded from `model.name`, so overriding `tokenizer.name` does not discard model-specific image preprocessing. `configure_trainable_parameters` is limited to LoRA application and frozen-vision discovery; router transforms and EP remain explicit in `setup_model`. LoRA keeps its existing initial freeze and logging, while EP performs the same conditional re-freeze as `main` after replacing parameters with DTensors. The VLM-specific vision freeze runs afterward.

Weight-tying follow-up: Qwen3.5 honors the checkpoint's standard tied-weight mapping. Checkpoints that omit `lm_head.weight` load the shared language embedding through the existing tied-model path without custom loader handling.

LoRA documentation follow-up: `freeze_vision_encoder` now promises parameter freezing only. The docs explicitly note that the default Qwen3.5 LoRA targets do not match vision modules, without claiming that arbitrary user-supplied target patterns are filtered.

---

> [!NOTE]
> **Medium Risk**
> Touches core training paths (model setup, forward, checkpoints, packing) and distributed VLM collectives; misconfiguration is mostly caught by validators, but multimodal truncation and vision/LoRA interactions need careful review.
> 
> **Overview**
> Adds **end-to-end multimodal SFT** for Qwen3.5 dense and MoE when `[model.vlm]` is set with `model.impl = "custom"`.
> 
> The SFT pipeline now carries renderer-produced `mm_kwargs`, `mm_token_type_ids`, and image-safe truncation through packing and the training forward; it loads an `AutoProcessor` from `model.name`, wires it into the renderer, and saves it with HF weight checkpoints. Config validation enforces VLM bfloat16 dtypes, `micro_batch_size = 1`, no CP, and rejects `freeze_vision_encoder=false` with LoRA; the old blocker that forbade VLM in renderer-only SFT is removed.
> 
> **Modeling:** Registers dense `qwen3_5` in the custom VLM map and extends `Qwen3_5ForCausalLM` / MoE VLM bodies with HF vision + custom text, MRoPE from renderer metadata, and dummy vision passes on text-only steps for distributed collective symmetry. Trainability is centralized in `configure_trainable_parameters` (LoRA + optional vision freeze after EP).
> 
> Docs and CI shift the multimodal story to Qwen3.5 (CP marked unsupported for VLMs); new example TOMLs cover dense and MoE VLM SFT.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c882df962ec2cfb5e683ee5d1f42fc04233e6eb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -221,24 +224,166 @@ def forward(
...
+        inputs_embeds: torch.FloatTensor | None = None,
+        pixel_values: torch.Tensor | None = None,
+        image_grid_thw: torch.LongTensor | None = None,
+        mm_token_type_ids: torch.LongTensor | None = None,
+        *,
+        seq_lens: torch.LongTensor,
+    ) -> tuple[torch.FloatTensor, torch.LongTensor]:
+        if inputs_embeds is None:
+            if input_ids is None:
+                raise ValueError("input_ids are required when inputs_embeds are not provided")
```
S1ro1: can we avoid checks like this in the modelling code? would either remove them fully as I assume they can never happen or centralize them?

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -221,24 +224,166 @@ def forward(
...
+
+        vision_output = self.visual(pixel_values, grid_thw=vision_grid_thw, return_dict=True)
+        image_embeds = vision_output.pooler_output.to(inputs_embeds.device, inputs_embeds.dtype)
+
+        if has_images:
+            image_mask = input_ids == self.config.image_token_id
+            image_token_count = int(image_mask.sum().item())
+            image_feature_count = int(image_embeds.shape[0])
+            if image_token_count != image_feature_count:
+                raise ValueError(
```
S1ro1: likewise here, and many other places

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -221,24 +224,166 @@ def forward(
...
         super().__init__(config, **kwargs)
-        self.model = Qwen3_5Model(config)
-        self.vocab_size = config.vocab_size
-        self.lm_head = nn.Linear(config.hidden_size, config.vocab_size, bias=False)
+        self._is_vlm = hasattr(config, "vision_config")
+
+        if self._is_vlm:
+            self.model = Qwen3_5VLMModel(config)
+            text_config = config.text_config
+            self._tied_weights_keys = {"lm_head.weight": "model.language_model.embed_tokens.weight"}
```
S1ro1, line 341: I think tying weights breaks things in a lot of cases, do we need it?

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -221,24 +224,166 @@ def forward(
...
 
     def get_input_embeddings(self):
+        if self._is_vlm:
+            return self.model.get_input_embeddings()
         return self.model.embed_tokens
 
     def set_input_embeddings(self, value):
-        self.model.embed_tokens = value
+        if self._is_vlm:
+            self.model.set_input_embeddings(value)
```
S1ro1: lets add this method to the text only model too, so all these if self._is_vlm get collapsed

src/prime_rl/trainer/models/qwen3_5_moe/modeling_qwen3_5_moe.py
```diff
@@ -799,9 +794,17 @@ def forward(
                 )
             image_mask = image_mask.unsqueeze(-1).expand_as(inputs_embeds).to(inputs_embeds.device)
             inputs_embeds = inputs_embeds.masked_scatter(image_mask, image_embeds)
+        else:
+            inputs_embeds = inputs_embeds + image_embeds.sum() * 0.0
 
         if position_ids is None:
             if image_grid_thw is not None:
+                if input_ids is None:
```
S1ro1: again here

S1ro1 (commented):
would be a bit more skeptical about the llm generated tests also, they dont seem to test a lot and just pollute code
also data part needs checking

examples/vlm_sft_moe/sft.toml
```diff
@@ -0,0 +1,55 @@
...
+[model.vlm]
+vision_encoder_attr = "model.visual"
+language_model_attr = "model.language_model"
+freeze_vision_encoder = true
+
+[model.lora]
+rank = 32
+alpha = 32
+target_modules = ["q_proj", "k_proj", "v_proj", "o_proj", "gate_proj", "up_proj", "down_proj", "experts"]
+
```
samsja, line 25: prob don't need example with lora here

samsja (commented):
wondering if it wouldn;'t be easier to just not support lora for vlm, kinda trying to keep the lora path as is for now

src/prime_rl/trainer/model.py
```diff
@@ -1099,6 +1123,43 @@ def apply_ep(model: nn.Module, config: ModelConfig, parallel_dims: ParallelDims)
...
+
+    if config.lora is not None:
+        excluded_modules = (frozen_vision_encoder,) if frozen_vision_encoder is not None else ()
+        apply_lora_to_model(model, config.lora, excluded_modules=excluded_modules)
+
+    if config.moe_router_dtype == "float32":
+        apply_fp32_moe_router(model)
+
+    if parallel_dims.ep_enabled:
+        apply_ep(model, config, parallel_dims)
```
samsja: hmm seems off to me that we have parraelism apply code in a `configure_trainable_parameters` function

src/prime_rl/trainer/model.py
```diff
@@ -1099,6 +1123,43 @@ def apply_ep(model: nn.Module, config: ModelConfig, parallel_dims: ParallelDims)
...
+    if config.vlm is not None and config.vlm.freeze_vision_encoder:
+        frozen_vision_encoder = get_vision_encoder(model, override=config.vlm.vision_encoder_attr)
+    elif config.vlm is None:
+        frozen_vision_encoder = get_vision_encoder(model)
+        if frozen_vision_encoder is not None:
+            get_logger().info("Training a VLM checkpoint on text-only data; freezing the vision encoder")
+
+    if config.lora is not None:
+        excluded_modules = (frozen_vision_encoder,) if frozen_vision_encoder is not None else ()
+        apply_lora_to_model(model, config.lora, excluded_modules=excluded_modules)
```
samsja: this function should only do this and keep the rest as is

samsja (commented):
``` ### 3. model.py — freeze_moe_router reordered relative to apply_fp32_moe_router
 and apply_ep

 Before: apply_lora → freeze_moe_router → apply_fp32_moe_router →
 freeze_sparse_indexer → apply_ep

 After: apply_lora → apply_fp32_moe_router → apply_ep → freeze_all_except_lora →
 freeze_moe_router → freeze_sparse_indexer
```

lets keep the otherd the same as before

src/prime_rl/trainer/sft/data.py
```diff
@@ -254,11 +304,40 @@ def should_mask(message: dict) -> bool:
...
+
+        if was_mm_truncated and not set(self.renderer.get_stop_token_ids()) & set(input_ids):
+            return None
 
         # Causal shift: model predicts next token from current.
         target_ids = input_ids.copy()[1:]
         loss_mask = loss_mask[1:]
         input_ids = input_ids[:-1]
+        if mm_token_type_ids is not None:
+            mm_token_type_ids = mm_token_type_ids[:-1]
```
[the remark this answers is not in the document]
eligotts: i think this is legit that should be fixed

src/prime_rl/trainer/lora.py
```diff
@@ -129,16 +136,43 @@ def freeze_all_except_lora_and_specified(model: nn.Module, config: LoRAConfig) -
...
+    logger.info(f"LoRA enabled: {lora_adapter_params:,} adapter params adapting {lora_adapted_params:,} base params")
+    logger.info(f"LoRA: {fully_trainable:,} fully trainable parameters")
+    logger.info(f"LoRA: {adapted_or_trainable:,} adapted or fully trainable out of {total_params:,} parameters")
+
+
+def apply_lora_to_model(
+    model: nn.Module,
+    config: LoRAConfig,
+    *,
+    excluded_modules: tuple[nn.Module, ...] = (),
```
eligotts: i think this excluded modules stuff might be overkill, its overly defensive on vision tower having naming collision

src/prime_rl/trainer/model.py
hubert-marek: It was because ep wswitches required grad to true and we need to re turn it off but I can refractor it into separate functions

src/prime_rl/trainer/model.py
hubert-marek: ok

src/prime_rl/trainer/model.py
```diff
@@ -1099,6 +1123,43 @@ def apply_ep(model: nn.Module, config: ModelConfig, parallel_dims: ParallelDims)
...
+
+    if parallel_dims.ep_enabled:
+        apply_ep(model, config, parallel_dims)
+
+    if config.lora is not None:
+        # Enforce the final policy after transforms that may replace parameters.
+        freeze_all_except_lora_and_specified(model, config.lora)
+
+    # Explicit freezes take precedence over LoRA modules_to_save and EP replacements.
+    if config.freeze_moe_router:
```
eligotts: Agree with keeping the main setup order. EP remaking expert params as trainable DTensors is already handled on main by re-freezing after apply_ep. For VLM+LoRA we only need to append freeze_vision_encoder() (and maybe vision exclude) onto that path — not move freeze_moe_router / sparse indexer after EP or fold EP into configure_trainable_parameters.
Suggested shape: leave main order as-is; after the existing post-EP LoRA re-freeze, call freeze_vision_encoder when freezing vision. Drop this reorder.

S1ro1 (approved):
like it!

PR #3083, opened 2026-07-18
fix: count length penalty per trace (last branch) instead of per branch (summed)

samsja:
## Summary

The GRPO length penalty was calculated using trace-level token/turn counts, which **sum across all branches** of the message graph. For multi-branch rollouts (tree search, forking), this inflates the penalty proportionally to the number of branches, incentivizing the model to compact to avoid the disproportionate penalty.

## Fix

Use the last branch (`branches[-1]` — the main root-to-leaf path) instead of the trace-level sums for `num_output_tokens`, `num_total_tokens`, and `num_turns` in the length penalty calculation. This follows the existing convention where `branches[-1]` is the main path (used for `transcript`, `tool_messages`, etc.).

For single-branch rollouts (the common linear case), behavior is unchanged — the fix only affects multi-branch traces.

### DIFF
--- a/src/prime_rl/orchestrator/algo/grpo.py
+++ b/src/prime_rl/orchestrator/algo/grpo.py
@@ -27,9 +27,10 @@ class GRPOAlgorithm(Algorithm):
         if length_penalty is None:
             advantages = rewards - rewards.mean()
         else:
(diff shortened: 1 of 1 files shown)


samsja:
Closing — the original code already sums across all branches (all turns including pre-compaction). No change needed.

PR #3103, opened 2026-07-21
feat(rl): pack multimodal samples with context parallelism

hubert-marek:
## Summary

Rebased successor to #2947, now targeting `main` after #3075 and #3099 merged. The five RL feature commits are patch-identical to the reviewed stack; follow-up cleanup commits align validation and tests with current `main`.

Multimodal RL samples no longer force one-sample micro-batches. Image samples pack with text spans and compatible `mm_kwargs` samples from the same run/LoRA:

- `_MicroBatchBin` enforces length, routed-expert, run/LoRA, and multimodal-sidecar compatibility. Materialized bins concatenate eager `EncodedTensor` sidecars along dimension 0 and preserve per-sample boundaries in `seq_lens`.
- Worker balancing does not split multimodal bins, and modality groups remain aligned across ranks. One shared multimodal-sidecar validator serves both `prepare_sample()` and per-run `MultiPacker` rejection, so the contract is not duplicated.
- Packing is always enabled for VLM RL; there is no packing config knob. Current VLM training already requires `model.impl = "custom"`.
- Qwen3.5 dense and MoE VLMs advertise packed-multimodal support. RL checks that capability once at startup, while transported multimodal batches still fail loudly if `[model.vlm]` is absent.
- The shared model config requires Ulysses for VLM context parallelism. MRoPE batches defer sequence sharding until after the full-sequence vision merge; `seq_lens_are_pre_shard` preserves packed boundaries, routed experts shard inside the model, and LoRA token counts use the sharded label length.
- Validation relies on the canonical attention choices already defined by `ModelConfig`; this PR does not add a second attention whitelist or restore removed attention fallbacks.

## Impact

Packed multimodal RL substantially reduces trainer step time by combining compatible samples while preserving document boundaries and distributed collective ordering.

Existing packed-vs-unpacked evidence with identical debug config/data shape (one train GPU, batch 256, custom Qwen3.5-0.8B):

| | unpacked | packed |
|---|---:|---:|
| Trainer step time, step 3 | 2m44s | **29.6s (5.5× faster)** |
| 15-step wall clock | timed out at step 14 (40 min) | **15/15 in 12 min** |

W&B: packed [be835a1999](https://wandb.ai/primeintellect/pr2485/runs/be835a19995749a49ad8184a2b77e6e8), unpacked [526133c55a](https://wandb.ai/primeintellect/pr2485/runs/526133c55ad44fc293b83dba7ea5ded8).

## Validation

- Original five commits verified unchanged with `git range-diff`.
- `uv run pytest tests/unit/test_configs.py tests/unit/orchestrator/test_batch.py tests/unit/train/rl/test_packer.py -q -k 'not test_load_configs'`: 68 passed, 57 external-plugin config cases deselected.
- Ruff check and format check passed across all changed Python files.
- `git diff --check` passed.
- Existing final-stack H100 RL boundary, multimodal packing, and forward suite: 35 passed.

---

> [!NOTE]
> **Medium Risk**
> Changes core RL batching, distributed CP/MRoPE forwarding, and multimodal tensor layout; mistakes could cause silent wrong training or collective hangs, but behavior is guarded by validation and unit tests.
> 
> **Overview**
> **Multimodal RL packing** no longer forces one sample per micro-batch. Compatible image and text spans from the same run/LoRA can share a packed row: `mm_kwargs` sidecars concatenate on dim 0, `seq_lens` keep document boundaries, and multimodal bins are not split for worker balancing.
> 
> **Trainer and validation** reject `image_grid_thw` without `mm_token_type_ids`, require `[model.vlm]` when multimodal batches arrive, and fail at startup if a VLM lacks `supports_packed_multimodal_training` (set on Qwen3.5 dense/MoE VLMs). VLM + CP must use **Ulysses** (`ModelConfig.vlm_cp_requires_ulysses`; removed from SFT-only validator). For MRoPE batches with CP, sequence sharding is deferred until after the vision merge; `seq_lens_are_pre_shard` and LoRA token counts follow the sharded label length.
> 
> **Docs** add a guide for custom VLM implementations (packed boundaries, registration, optional CP).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64361c3a4382e06cbaeb484a1ffaf91c4880f4c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/trainer/rl/train.py
```diff
@@ -371,30 +379,30 @@ def load_run_checkpoint(_optimizer, idx: int) -> None:
...
-                    input_ids,
-                    position_ids,
-                    cp_rank,
-                    cp_size,
-                    cp_group,
-                    seq_lens=seq_lens,
-                    cp_style=config.model.cp_style,
-                )
+                # MRoPE batches must merge image embeddings before sharding.
+                defer_vlm_cp_to_model = mm_kwargs is not None and "image_grid_thw" in mm_kwargs
```
eligotts, line 386: there are a few qwen specific artifacts in here like "image_grid_thw" but i think its fine to get this in now, ill have prs coming after which will generalize this a bit

eligotts (approved):

src/prime_rl/trainer/batch.py
```diff
@@ -230,17 +240,45 @@ def from_sample(cls, lora_idx: int, sample: MicroBatch) -> "_MicroBatchBin":
...
+            return False
+        if (first_sample.routed_experts is None) != (sample.routed_experts is None):
+            return False
+
+        sample_is_mm = _is_multimodal_sample(sample)
+        existing_mm_sample = self.first_multimodal_sample
+        if existing_mm_sample is None and not sample_is_mm:
+            return True
+        if self.first_lora_idx != lora_idx:
+            return False
```
samsja, line 269: hmm what is this case here and why do we return false ? shouldn't we just failed ?

src/prime_rl/trainer/batch.py
```diff
@@ -230,17 +240,45 @@ def from_sample(cls, lora_idx: int, sample: MicroBatch) -> "_MicroBatchBin":
...
+        if existing_mm_sample is not None and sample_is_mm:
+            dst = existing_mm_sample.mm_kwargs
+            src = sample.mm_kwargs
+            assert dst is not None and src is not None
```
samsja, line 273: can we add a comment in the assert assert x!=y, "hello"

src/prime_rl/trainer/rl/train.py
samsja, line 386: this seems a bit code smell tbh, but fine for now

eligotts: this will be fixed in a coming pr

src/prime_rl/trainer/batch.py
eligotts: this is for multitenant lora, ensuring we dont pack samples that belong to different runs

PR #3108, opened 2026-07-22
fix(orchestrator): gate group-relative algorithms to a single trainable agent

hallerite:
## What

`GRPOAlgorithm.score_group` / `MaxRLAlgorithm.score_group` now refuse a group whose trainable traces span more than one `agent_name` (new `assert_single_trainable_agent` helper in `algo/base.py`). Echo inherits the gate through GRPO's `score_group`; per-rollout algorithms (opd/opsd/sft) are untouched.

## Why

Group-relative advantages assume the group is exchangeable attempts by **one** agent. After the multi-agent episode wire (#3104), a group's trainable survivors can span several agents, and mean-centering them together mixes different agents' reward scales into one baseline: silently wrong credit, no error.

The gate is deliberately the condition that matters, not "single-agent envs only": multi-agent envs with exactly one trainable agent keep working unchanged — e.g. a trainable solver rewarded by a frozen agentic judge. A group mixing trainable agent names is refused with an actionable message: mark the other agents untrainable in the env's `setup()` (`agents.<name>.trainable = False`), or implement your own algorithm that assigns credit across agents.

Cross-agent credit assignment is an open question prime-rl shouldn't answer silently in a base class — this refuses instead of guessing. If a well-founded scheme lands later, it arrives as a named `Algorithm`, and this gate is exactly the code it replaces.

Companion verifiers PR: PrimeIntellect-ai/verifiers#2101 (enforces standing at the source: an agent off the trainable model context refuses on training runs and is demoted on eval runs).

## Test plan

- `uv run pytest tests/unit/orchestrator/test_advantage.py tests/unit/orchestrator/test_algorithms.py` — green, including the new `test_group_algos_refuse_multiple_trainable_agents` (mixed-agent group raises for both algos; same-agent multi-trace group still centers to zero).
- `uv run ruff check` / `ruff format --check` clean on the touched files.

### DIFF
--- a/docs/algorithms.md
+++ b/docs/algorithms.md
@@ -285,6 +285,8 @@ The per-token training signal is set by `algo.type` and the [algorithm](#the-alg
 
 The default advantage is per-group reward minus per-group baseline (DR-GRPO without std normalization). For each prompt's group of `group_size` rollouts, every token in rollout $i$ receives advantage $s_i - \bar{s}$ where $\bar{s}$ is the group mean.
 
+The group-relative algorithms (`grpo`, `max_rl`, `echo`) require the group's trainable traces to come from **one** agent — the baseline assumes exchangeable attempts by a single agent. Multi-agent envs work as long as exactly one agent is trainable (e.g. a trainable solver rewarded by a frozen agentic judge); a group spanning several trainable agent names is refused at scoring time. Credit assignment across multiple trainable agents is not something prime-rl prescribes — implement your own algorithm for it (see [Authoring an Algorithm](#authoring-an-algorithm)).
+
 This is intentionally simple — it does the right thing for most envs. Write a named algorithm class when you need group-aware shaping that depends on trajectory metadata (sub-agent rollouts, relative-rank shaping, …) — see [Authoring an Algorithm](#authoring-an-algorithm).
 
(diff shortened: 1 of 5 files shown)


tests/unit/orchestrator/test_advantage.py
```diff
@@ -183,6 +183,22 @@ def test_max_rl_mean_normalized():
     assert _max_rl(_make_group(rewards=[1.0, 1.0])) == pytest.approx([0.0, 0.0])
 
 
+def test_group_algos_refuse_multiple_trainable_agents():
```
mikasenghaas, line 186: remove

docs/algorithms.md
```diff
@@ -285,6 +285,8 @@ The per-token training signal is set by `algo.type` and the [algorithm](#the-alg
 
 The default advantage is per-group reward minus per-group baseline (DR-GRPO without std normalization). For each prompt's group of `group_size` rollouts, every token in rollout $i$ receives advantage $s_i - \bar{s}$ where $\bar{s}$ is the group mean.
 
+The group-relative algorithms (`grpo`, `max_rl`, `echo`) require the group's trainable traces to come from **one** agent — the baseline assumes exchangeable attempts by a single agent. Multi-agent envs work as long as exactly one agent is trainable (e.g. a trainable solver rewarded by a frozen agentic judge); a group spanning several trainable agent names is refused at scoring time. Credit assignment across multiple trainable agents is not something prime-rl prescribes — implement your own algorithm for it (see [Authoring an Algorithm](#authoring-an-algorithm)).
```
mikasenghaas, line 288: do we have no such constraint for say opd?

hallerite: no, because for OPD, we just do a teacher prefill for each trainable trace

PR #3109, opened 2026-07-22
feat: chunked per-layer optimizer step with stream-overlapped CPU offload

samsja:
## Summary

The `CPUOffloadOptimizer` can optionally perform `optimizer.step()` **per-transformer-layer** with **stream-overlapped** H2D/D2H transfers instead of all-at-once. This is opt-in via `optim_cpu_offload_chunked` — the default remains the simple all-at-once path.

### Problem

With all-at-once offload, the **entire** model's optimizer states land on GPU simultaneously during `step()`. When `weight + grad + all_opt_states > VRAM` — even without activations — this causes OOM on large models.

### Config

```toml
[trainer.model]
optim_cpu_offload = true             # already the default
optim_cpu_offload_chunked = true     # opt-in per-layer chunked stepping
```

`optim_cpu_offload_chunked` defaults to `false`. When enabled, the optimizer step is performed per-transformer-layer with stream-overlapped H2D/D2H transfers: while layer *i* computes, layer *i+1* is prefetched and layer *i-1* is evicted. The prefetch waits for the eviction to complete, so **at most two layers' optimizer states are on GPU at any time**.

### Implementation

- **Per-layer chunking**: Parameters grouped by transformer layer (regex `layers.{N}.`). Non-layer params (embeddings, lm_head, norm) go into a trailing "misc" chunk.
- **Stream-overlapped**: H2D/D2H on dedicated CUDA streams (created once in `__init__`, reused across steps); next layer prefetched while current computes; previous layer evicted while next computes. Ends with `torch.cuda.synchronize()`.
- **Non-chunked path** (default): Simple all-at-once — move all states to GPU, step, move back to CPU. No CUDA streams.
- **First-step**: The chunked init step also processes per-layer, avoiding materializing all optimizer states on GPU at once.
- **Pinned memory**: D2H pre-allocates pinned CPU destination and async-copies into it. `state_dict()`, `load_state_dict()`, and ckpt paths sync after D2H to prevent reading incomplete copies.
- **Muon step counter sync**: Per-group `step` integer synced after all chunks since `param_groups` are temporarily swapped.

### Performance (measured on RTX PRO 6000, float32, AdamW, 10 steps)

| Model (layers, hidden) | All-at-once peak | All-at-once step | Chunked peak | Chunked step | Mem savings | Speedup |
|---|---|---|---|---|---|---|
| Small (14, 768) | 1,308 MiB | 26.5 ms | 882 MiB | 23.9 ms | 426 MiB | 1.1× |
| Medium (28, 1024) | 4,798 MiB | 87.4 ms | 2,891 MiB | 63.3 ms | 1,907 MiB | 1.4× |
| Large (40, 2560) | 41,031 MiB | 861.7 ms | 25,083 MiB | 597.3 ms | 15,948 MiB | 1.4× |

Chunked+stream is **faster** than all-at-once because stream overlap hides H2D/D2H transfer latency behind compute. Memory savings are 33–39%.

### Validation

- Full reverse-text RL training (20 steps, Qwen3-0.6B, 2 GPUs) completed successfully
- Parity tests: chunked vs non-chunked AdamW — params, optimizer states, and state_dict round-trip all match (`tests/unit/train/test_optim_offload.py`, 5 tests)
- All integration tests pass (reverse_text, sft, lora, moe, multi_run, benchmark_regression)
- `base_optimizer`, `state`, `param_groups`, `zero_grad`, `state_dict`, `load_state_dict` APIs unchanged

### Bugbot issues addressed

1. **Async D2H checkpoint corruption** (HIGH): Added `torch.cuda.synchronize()` after `_move_states("cpu")` in `state_dict()`, `load_state_dict()`, and `ckpt.py`
2. **Incomplete stream dependency** (MEDIUM): Added `torch.cuda.synchronize()` at end of streamed step; `h2d_stream.wait_stream(d2h_stream)` before prefetch to prevent pin buffer races
3. **First step OOM** (HIGH): Chunked init step processes per-layer instead of full `optimizer.step()`
4. **Closure run per chunk** (LOW): Closure passed only to first chunk
5. **Stream memory leak** (MEDIUM): CUDA streams created once in `__init__` and reused across steps instead of allocating new streams per `step()`

### DIFF
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -146,7 +146,15 @@ optim_cpu_offload = true   # already the default
 
 Mutually exclusive with `fsdp_cpu_offload`. Also incompatible with `trainer.max_concurrent_runs > 1` (multi-tenant training) — set `optim_cpu_offload = false` for multi-run. Muon doesn't support `fsdp_cpu_offload` but does support `optim_cpu_offload`.
 
-The optimizer step is performed per-transformer-layer ("chunked"): each layer's optimizer states are moved to GPU, the step runs for that layer only, and the states move back to CPU before the next layer. This bounds peak GPU optimizer-state memory to a single layer's worth, preventing OOM when `weight + grad + all_opt_states` exceeds available VRAM even without activations. H2D and D2H transfers are pipelined on dedicated CUDA streams so the next layer's states are prefetched while the current layer computes.
+The optimizer step is performed per-transformer-layer ("chunked"): each layer's optimizer states are moved to GPU, the step runs for that layer only, and the states move back to CPU before the next layer. This bounds peak GPU optimizer-state memory to a single layer's worth, preventing OOM when `weight + grad + all_opt_states` exceeds available VRAM even without activations. H2D and D2H transfers are pipelined on dedicated CUDA streams so the next layer's states are prefetched while the current layer computes:
+
+```toml
+[trainer.model]
+optim_cpu_offload_chunked = true   # already the default
+optim_cpu_offload_stream = true    # already the default
+```
+
+Set `optim_cpu_offload_chunked = false` to fall back to all-at-once offloading (loads the full model's optimizer states to GPU in one shot — simpler but requires enough VRAM to hold them all simultaneously). Set `optim_cpu_offload_stream = false` to disable stream overlap within the chunked step, falling back to a simple sequential move→step→move loop.
 
 ### LM Head Chunking
 
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -146,6 +146,8 @@ optim_cpu_offload = true   # already the default
 
 Mutually exclusive with `fsdp_cpu_offload`. Also incompatible with `trainer.max_concurrent_runs > 1` (multi-tenant training) — set `optim_cpu_offload = false` for multi-run. Muon doesn't support `fsdp_cpu_offload` but does support `optim_cpu_offload`.
 
+The optimizer step is performed per-transformer-layer ("chunked"): each layer's optimizer states are moved to GPU, the step runs for that layer only, and the states move back to CPU before the next layer. This bounds peak GPU optimizer-state memory to a single layer's worth, preventing OOM when `weight + grad + all_opt_states` exceeds available VRAM even without activations. H2D and D2H transfers are pipelined on dedicated CUDA streams so the next layer's states are prefetched while the current layer computes.
+
 ### LM Head Chunking
 
 The vanilla LM head materializes a `[batch * seq, vocab]` logits tensor on every step — a major memory tax when the vocabulary is large (often >100K). `fused_lm_head_token_chunk_size` swaps in a custom fused linear + logprob/entropy kernel that streams through `chunk_size` tokens at a time, avoiding the materialization. It defaults to `1024` for RL training:
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -154,7 +154,7 @@ optim_cpu_offload_chunked = true
 optim_cpu_offload_stream = true
 ```
 
-Set `optim_cpu_offload_chunked = false` to fall back to all-at-once offloading (loads the full model's optimizer states to GPU in one shot — simpler but requires enough VRAM to hold them all simultaneously). Set `optim_cpu_offload_stream = false` to disable stream overlap within the chunked step, falling back to a simple sequential move→step→move loop.
+Both default to `false`. Set `optim_cpu_offload_chunked = true` to enable per-layer chunking. Set `optim_cpu_offload_stream = true` to additionally overlap H2D/D2H transfers with optimizer compute on dedicated CUDA streams. When chunked is enabled but stream is not, the step uses a simple sequential move→step→move loop.
 
 ### LM Head Chunking
 
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -150,8 +150,8 @@ The optimizer step is performed per-transformer-layer ("chunked"): each layer's 
 
 ```toml
 [trainer.model]
-optim_cpu_offload_chunked = true   # already the default
-optim_cpu_offload_stream = true    # already the default
+optim_cpu_offload_chunked = true
+optim_cpu_offload_stream = true
 ```
 
 Set `optim_cpu_offload_chunked = false` to fall back to all-at-once offloading (loads the full model's optimizer states to GPU in one shot — simpler but requires enough VRAM to hold them all simultaneously). Set `optim_cpu_offload_stream = false` to disable stream overlap within the chunked step, falling back to a simple sequential move→step→move loop.
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -146,7 +146,7 @@ optim_cpu_offload = true   # already the default
 
 Mutually exclusive with `fsdp_cpu_offload`. Also incompatible with `trainer.max_concurrent_runs > 1` (multi-tenant training) — set `optim_cpu_offload = false` for multi-run. Muon doesn't support `fsdp_cpu_offload` but does support `optim_cpu_offload`.
 
-The optimizer step is performed per-transformer-layer ("chunked"): each layer's optimizer states are moved to GPU, the step runs for that layer only, and the states move back to CPU before the next layer. This bounds peak GPU optimizer-state memory to a single layer's worth, preventing OOM when `weight + grad + all_opt_states` exceeds available VRAM even without activations. H2D and D2H transfers are pipelined on dedicated CUDA streams so the next layer's states are prefetched while the current layer computes:
+The optimizer step is performed per-transformer-layer ("chunked"): each layer's optimizer states are moved to GPU, the step runs for that layer only, and the states move back to CPU before the next layer. This reduces peak GPU optimizer-state memory from the full model's states to about one layer's worth (two layers with stream overlap: the current layer being computed plus the next layer being prefetched). Set `optim_cpu_offload_chunked = true` to enable:
 
 ```toml
 [trainer.model]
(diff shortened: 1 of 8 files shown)


src/prime_rl/trainer/optim.py
```diff
@@ -20,58 +21,270 @@ class CPUOffloadOptimizer:
...
+    negligible compared to the savings from not loading the full model's states.
     """
 
-    def __init__(self, optimizer: Optimizer, pin_memory: bool = True):
+    def __init__(
+        self,
+        optimizer: Optimizer,
+        named_params: list[tuple[str, nn.Parameter]] | None = None,
+        pin_memory: bool = True,
+        stream: bool = True,
```
S1ro1: why do we default this to true when the global is false, seems weird

src/prime_rl/trainer/optim.py
```diff
@@ -20,58 +21,270 @@ class CPUOffloadOptimizer:
...
         self._initialized = False
 
+        # Build per-layer chunks from the optimizer's actual param_groups so
+        # we only include params the optimizer tracks (excludes frozen params
+        # that Muon / _create_optimizer filtered out).
+        self._chunks: list[list[nn.Parameter]] | None = None
+        if named_params is not None:
+            self._chunks = self._build_chunks(named_params)
+
+    # ------------------------------------------------------------------ #
```
S1ro1: can we remove this slop?

src/prime_rl/trainer/optim.py
```diff
@@ -20,58 +21,270 @@ class CPUOffloadOptimizer:
...
+                    by_layer.setdefault(layer_idx, []).append(param)
+                else:
+                    misc.append(param)
+
+        chunks = [by_layer[k] for k in sorted(by_layer)]
+        if misc:
+            chunks.append(misc)
+        return chunks or [misc]  # edge-case: no layer params at all
+
+    def _chunk_param_ids(self, chunk_idx: int) -> set[int]:
```
S1ro1: likewise comment about what this does, seems to make no sense really

S1ro1 (changes requested):
can we in general unslopify this? it's very hard to read and follow, logic seems sane but code eh

docs/scaling.md
```diff
@@ -146,6 +146,8 @@ optim_cpu_offload = true   # already the default
 
 Mutually exclusive with `fsdp_cpu_offload`. Also incompatible with `trainer.max_concurrent_runs > 1` (multi-tenant training) — set `optim_cpu_offload = false` for multi-run. Muon doesn't support `fsdp_cpu_offload` but does support `optim_cpu_offload`.
 
+The optimizer step is performed per-transformer-layer with stream-overlapped H2D/D2H transfers: while layer *i* computes, layer *i+1* is prefetched and layer *i-1* is evicted. The prefetch waits for the eviction to complete, so at most two layers' optimizer states are on GPU at any time. This reduces peak GPU optimizer-state memory from the full model's states to about two layers' worth, preventing OOM when `weight + grad + all_opt_states > VRAM`.
```
samsja: can remove this

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -170,7 +170,7 @@ class ModelConfig(BaseModelConfig):
     """Enable FSDP CPU offloading for parameters, gradients, and optimizer states. Uses pinned memory for efficient CPU↔GPU transfers."""
 
     optim_cpu_offload: bool = True
-    """Offload only optimizer states (momentum, variance) to CPU, keeping weights on GPU. Avoids the H2D all-gather overhead of FSDP CPU offload while still saving GPU memory."""
+    """Offload only optimizer states (momentum, variance) to CPU, keeping weights on GPU. Avoids the H2D all-gather overhead of FSDP CPU offload while still saving GPU memory. The optimizer step is performed per-transformer-layer with stream-overlapped H2D/D2H transfers, keeping at most two layers' optimizer states on GPU at a time."""
```
S1ro1: I'm really not a fan of making it even slower when it's default, like I think it should be a separate flag

src/prime_rl/trainer/optim.py
```diff
@@ -15,80 +16,163 @@
...
 
-        # Run optimizer step
+        For D2H, the old GPU tensor is recorded on the stream so the caching
+        allocator won't reuse its storage until the async copy finishes.
+        """
+        chunk_ids = {id(p) for p in self._chunks[chunk_idx]}
+        with torch.cuda.stream(stream):
+            for p, state in self.optimizer.state.items():
+                if id(p) not in chunk_ids:
+                    continue
```
S1ro1, line 96: this branch looks suspicious to say the least

S1ro1 (changes requested):
Streams in torch are known to be buggy and cause silent correctness issues, on top of this it's a default that makes training slower without opt-out.

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -172,6 +172,9 @@ class ModelConfig(BaseModelConfig):
     optim_cpu_offload: bool = True
     """Offload only optimizer states (momentum, variance) to CPU, keeping weights on GPU. Avoids the H2D all-gather overhead of FSDP CPU offload while still saving GPU memory."""
 
+    optim_cpu_offload_chunked: bool = False
+    """When optimizer CPU offload is enabled, perform the optimizer step per-transformer-layer with stream-overlapped H2D/D2H transfers instead of all-at-once. At most two layers' optimizer states are on GPU at any time. Prevents OOM when weight + grad + all_opt_states exceeds VRAM at the cost of slower stepping."""
```
mikasenghaas: how do those related? im assuming you can only do the chunking if cpu is enabled. should we make this explicit with `OptimCPUOffloadConfig` or so which has a `chunk: bool = False` field?

src/prime_rl/trainer/sft/train.py
```diff
@@ -180,7 +180,11 @@ def train(config: SFTConfig):
     # Set up the optimizer
     logger.info(f"Initializing optimizer ({config.optim})")
     optimizer = setup_optimizer(
-        config.optim, list(model.named_parameters()), parallel_dims, cpu_offload=config.model.optim_cpu_offload
+        config.optim,
+        list(model.named_parameters()),
+        parallel_dims,
+        cpu_offload=config.model.optim_cpu_offload,
+        cpu_offload_chunked=config.model.optim_cpu_offload_chunked,
```
mikasenghaas: acc feels like those should be on `OptimizerConfig`

src/prime_rl/trainer/optim.py
```diff
@@ -15,80 +16,173 @@
...
+        self.optimizer.param_groups = self._original_param_groups
+        return result
+
+    def _sync_step_counters(self, original_steps: list):
+        """Increment the original param_groups' step counters by one.
+
+        Muon stores a per-group ``step`` that ``step()`` increments. Because we swap
+        in per-chunk copies, the original groups never see the increment. Standard
+        optimizers (AdamW, SGD, SignSGD) keep step in ``state[p]['step']`` which is
+        per-parameter and already correct.
```
mikasenghaas, line 131: bit annoying to have this branchign factor. feel like this is a repeating pattern with muon? anythign we can do abt it?

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
samsja: its faster since its done by chunk

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -274,8 +271,8 @@ def selective_ac_only_with_custom_impl(self):
 
     @model_validator(mode="after")
     def cpu_offload_mutual_exclusion(self):
-        if self.fsdp_cpu_offload and self.optim_cpu_offload:
-            raise ValueError("Cannot enable both fsdp_cpu_offload and optim_cpu_offload. Use one or the other.")
+        if self.fsdp_cpu_offload and self.optim.cpu_offload.enabled:
+            raise ValueError("Cannot enable both fsdp_cpu_offload and optim.cpu_offload. Use one or the other.")
```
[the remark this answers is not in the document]
S1ro1: seems legitimate?

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -351,6 +348,14 @@ class CosineSchedulerConfig(BaseConfig):
 ]
 
 
+class CPUOffloadConfig(BaseConfig):
+    enabled: bool = True
+    """Offload optimizer states (momentum, variance) to CPU, keeping weights on GPU. Avoids the H2D all-gather overhead of FSDP CPU offload while still saving GPU memory."""
```
S1ro1, line 353: Do we need the enabled? Shouldn't it again be None | ...?

src/prime_rl/trainer/rl/train.py
```diff
@@ -170,7 +170,8 @@ def train(config: TrainerConfig):
             list(model.named_parameters()),
             parallel_dims,
             lora=config.model.lora is not None,
-            cpu_offload=config.model.optim_cpu_offload,
+            cpu_offload=config.optim.cpu_offload.enabled,
```
S1ro1, line 173: can pass the full config object here no?

src/prime_rl/trainer/optim.py
```diff
@@ -114,6 +208,7 @@ def setup_optimizer(
     parallel_dims: ParallelDims,
     lora: bool = False,
     cpu_offload: bool = False,
+    cpu_offload_chunked: bool = False,
```
S1ro1, line 211: ditto here, we can just pass the whole config object and enable if not None

src/prime_rl/trainer/sft/train.py
```diff
@@ -180,7 +180,11 @@ def train(config: SFTConfig):
     # Set up the optimizer
     logger.info(f"Initializing optimizer ({config.optim})")
     optimizer = setup_optimizer(
-        config.optim, list(model.named_parameters()), parallel_dims, cpu_offload=config.model.optim_cpu_offload
+        config.optim,
+        list(model.named_parameters()),
+        parallel_dims,
+        cpu_offload=config.optim.cpu_offload.enabled,
+        cpu_offload_chunked=config.optim.cpu_offload.chunked,
```
S1ro1, line 187: ditto here

PR #3110, opened 2026-07-22
feat(wandb): log full RLConfig to W&B instead of per-component subconfig

samsja:
## Summary

The `rl` entrypoint splits the full `RLConfig` into per-component subconfigs (`trainer.toml`, `orchestrator.toml`, `inference.toml`) before launching subprocesses. Each process only loads its subconfig, so `wandb.init(config=...)` only received the orchestrator config (the primary in shared W&B mode). The trainer, inference, slurm, deployment, and shared sections were missing entirely from the W&B run config.

## Changes

- **`rl.py`**: The `rl()` entrypoint writes the full resolved `RLConfig` as JSON (`output_dir/configs/full_config.json`) and sets `WANDB_FULL_CONFIG_PATH` in `os.environ` so all locally-launched subprocesses inherit it.
- **`multi_node_rl.sbatch.j2`**: Exports `WANDB_FULL_CONFIG_PATH` directly (srun tasks don't inherit the launcher's `os.environ`).
- **`wandb.py`**: `WandbMonitor` reads the full config JSON when `WANDB_FULL_CONFIG_PATH` is set, falling back to the per-component subconfig otherwise. After `wandb.init`, also uploads all config files (`full_config.json` + subconfig TOMLs) to the W&B run so they're downloadable from the UI Files tab.

All three launch modes are covered:
- **Local**: `os.environ` propagates to subprocesses via `**os.environ` in `Popen` env dicts
- **Single-node SLURM**: sbatch template re-runs `uv run rl` -> `rl()` -> sets `os.environ` -> `rl_local()`
- **Multi-node SLURM**: sbatch template exports `WANDB_FULL_CONFIG_PATH` alongside existing `WANDB_SHARED_*` vars

## Verification

Tested end-to-end with `WANDB_MODE=offline uv run rl @ examples/basic/reverse-text/rl.toml --max-steps 3 --output-dir /tmp/rl-test --wandb`:
- `full_config.json` written with all 17 top-level keys (trainer, orchestrator, inference, slurm, deployment, wandb, ckpt, model, etc.)
- Orchestrator log confirms: `Using full config from /tmp/rl-test/configs/full_config.json for W&B`
- All config files (`full_config.json`, `trainer.toml`, `orchestrator.toml`, `inference.toml`) uploaded to the W&B run's Files tab

### DIFF
--- a/src/prime_rl/entrypoints/rl.py
+++ b/src/prime_rl/entrypoints/rl.py
@@ -38,6 +38,7 @@ from prime_rl.utils.process import (
 
 RL_TOML = "rl.toml"
 RL_SBATCH = "rl.sbatch"
+FULL_CONFIG_JSON = "full_config.json"
 
 TRAINER_TOML = "trainer.toml"
 ORCHESTRATOR_TOML = "orchestrator.toml"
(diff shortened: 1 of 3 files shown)


src/prime_rl/entrypoints/rl.py
```diff
@@ -38,6 +38,7 @@
 
 RL_TOML = "rl.toml"
 RL_SBATCH = "rl.sbatch"
+FULL_CONFIG_JSON = "full_config.json"
```
mikasenghaas, line 41: call this `rl.json`, will land as `<output>/configs/rl.json`

mikasenghaas: and is consistent with other artifacts

PR #3139, opened 2026-07-24
fix(orchestrator): fail fast when an env server dies

eligotts:
## Problem

If a spawned environment server died during startup, the orchestrator kept waiting on its address queue for up to 600 seconds. The real exception remained in the env log. A server that died later could also leave untimed rollout requests pending indefinitely.

## Changes

- Read spawned-server addresses through `verifiers.v1.serve.wait_for_address`, which polls the child exit code and includes the env log tail in startup failures.
- Check owned env-server processes on every orchestrator loop pass so mid-run deaths fail the run promptly.
- Tie each spawned server to the orchestrator with a death pipe so abrupt orchestrator exits do not orphan brokers, workers, sandboxes, or tunnels.
- Register shutdown before spawning the first environment, ensuring partial startup is cleaned up.
- Keep the monitoring skill aligned with the new failure messages.

The `deps/verifiers` pin points to the conflict-resolved head of [verifiers#2131](https://github.com/PrimeIntellect-ai/verifiers/pull/2131), which provides the readiness handshake, worker reaping, and shared address-wait helper. Repoint it to `main` after that PR merges.

## Verification

- `uv sync --all-packages --locked`
- `uv run ruff check` and `uv run ruff format --check` on the touched Python files
- `uv run pre-commit run --all-files`
- 263 orchestrator, config, and utility tests passed
- Spawned-process smoke checks covered a startup import failure, a healthy pooled server, and mid-run server death

The remaining trainer-model tests require the CI-only `flash_attn` extra and are left to the PR checks.

---

> [!NOTE]
> **Medium Risk**
> Changes orchestrator env process supervision and main-loop behavior; mis-detection could abort healthy runs, but scope is limited to spawned env servers and improves failure visibility over silent stalls.
> 
> **Overview**
> **Env server lifecycle** is tightened so dead servers are detected quickly instead of hanging startup or stalling the pipeline mid-run.
> 
> Spawned env servers now report their address through `wait_for_address` (verifiers), which polls the child exit code and surfaces startup failures with a tail of the env log. Each spawn gets a **death pipe** so the server exits when the orchestrator dies abruptly, and shutdown registers **before** the first spawn so partial startup still cleans up non-daemon children.
> 
> The orchestrator main loop calls `check_alive()` on train/eval env collections each iteration and raises if an owned server has exited. The monitor-run skill documents the new startup vs mid-run vs worker-pool failure messages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46ad01a062867a98e47e10de2c560727e5d0d8c8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skills/training/monitor-run/SKILL.md
+++ b/skills/training/monitor-run/SKILL.md
@@ -161,6 +161,7 @@ The binary batches consumed by the trainer still live at `{output_dir}/rollouts/
 A few warnings are normal. Escalate when errors are persistent, growing, or hit a large fraction of rollouts.
 
 - **Env workers**: exceptions in env code, timeouts, sandbox errors, OOM kills (most common source — runs user code).
+- **Env server death**: an env server that dies is reported, not waited on. At startup the orchestrator raises `Env server <name> died on startup with exit code N` with the tail of `logs/envs/{train,eval}/<name>.log` quoted inline; mid-run it raises `Env server <name> exited with code N`. A worker inside a live server dying instead shows up as errored rollouts (`env server worker died (exit code N)`) plus `EnvServerPool worker N died` in the env log — the pool replaces it up to 3 times, then every rollout for that env fails with `no live workers`.
 - **Orchestrator**: empty/errored rollout spikes, weight-broadcast failures, checkpoint errors.
 - **Trainer**: NCCL/CUDA errors, OOM, NaN loss or gradients.
 - **Inference**: NCCL/CUDA errors, OOM, request timeouts.
--- a/src/prime_rl/orchestrator/envs.py
+++ b/src/prime_rl/orchestrator/envs.py
@@ -23,16 +23,16 @@ import asyncio
 import atexit
 import multiprocessing as mp
 import os
-import queue
 import sys
 from collections.abc import Iterator, Sequence
 from itertools import islice
+from multiprocessing.connection import Connection
 from multiprocessing.process import BaseProcess
 from pathlib import Path
 from typing import Generic, TypeVar
 
 import verifiers.v1 as vf
-from verifiers.v1.serve import EnvClient, env_config_data
+from verifiers.v1.serve import EnvClient, env_config_data, wait_for_address
 
 from prime_rl.configs.orchestrator import EnvConfig, EvalEnvConfig, TrainEnvConfig
 from prime_rl.orchestrator.algo import Algorithm, build_algorithm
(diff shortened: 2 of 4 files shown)


eligotts:
Heads up on CI: the failures here are expected and are not this change. Leaving them red rather than working around them — the fix imports `wait_for_address` from verifiers instead of keeping a private copy, and that means it needs a real verifiers upgrade in this repo.

Blocked on three things, in order:

**1. PrimeIntellect-ai/verifiers#2131 must merge.** Not just for the helper: with the previously pinned verifiers, a *pooled* env server (the default, since `pool` defaults to `elastic`) reports its address as soon as its socket is bound, before any worker has loaded the env. So in the common case — the process that dies is a pool *worker* — an exit-code poll here sees a healthy broker and cannot help. #2131 makes readiness a worker handshake and reaps dead workers.

**2. Configs need migrating for verifiers#2106** ("move the runtime onto the agent"), which is in the 12-commit delta between the old pin and current verifiers main. `env.agent.harness.runtime` becomes `env.agent.runtime` (likewise `env.judge.harness.runtime`). This repo still nests it in 48 config/doc/test files, so RL integration tests fail config validation:

```
1 validation error for RLConfig
--orchestrator.train.env[0].env.agent.harness.runtime
Extra inputs are not permitted
```

**3. Test helpers need migrating for verifiers#2119** ("store raw score and weight on trace rewards"), also in that delta. `Trace.rewards` is now `dict[str, Reward]` rather than `dict[str, float]`, which fails 78 unit tests:

```
ValidationError: 1 validation error for Rollout[TaskData]
rewards.reward
  Input should be a valid dictionary or instance of Reward [input_value=1.0, input_type=float]
```

Worth noting this one is confined to tests — prime-rl source never reads `Trace.rewards`, so it is roughly ten lines across six test files wrapping floats in `Reward(score=..., weight=...)`.

`uv.lock` **is** updated here (17/4 lines: the new `kuhn-poker-v1` env package and `prime-sandboxes>=0.2.33`), since `uv sync --locked` aborts before any test runs otherwise and that part is purely mechanical. Both migrations above are mechanical too, but they are a verifiers *upgrade*, not this bugfix, and they belong in their own review — happy to take that as a separate PR if useful.

Order: merge verifiers#2131 → bump + migrate (2 and 3) → this goes green.

PR #3149, opened 2026-07-28
feat: default rollout transport to zmq and inject host for multi-node

mikasenghaas:
## Summary

Makes ZMQ the default rollout transport and fixes the multi-node hang that blocked using it.

- **ZMQ is the default** (`trainer.rollout_transport` / `orchestrator.rollout_transport` → `ZMQTransportConfig()`): rollout batches stream orchestrator → trainer over sockets instead of being written to disk as `.bin` files. Filesystem stays available via `rollout_transport.type = "filesystem"`.
- **Shared config** (`configs/rl.py` + `utils/validation.py`): `rollout_transport` is a shared `RLConfig` field (like `weight_broadcast`), propagated to `trainer`/`orchestrator`. `auto_setup_rollout_transport` resolves the shared field from the sub-configs so the launcher gate is correct whether zmq comes from the default, the shared block, or a sub-config, and rejects mismatched sub-configs.
- **Launcher** (`entrypoints/rl.py` + `multi_node_rl.sbatch.j2`): when the resolved transport is zmq, inject `--rollout_transport.host $MASTER_ADDR` into the trainer torchrun and the orchestrator (mirrors `weight_broadcast.host`).
- Docs + monitor-run skill updated for the new default.

## Problem

ZMQ rollout transport binds/connects a single `rollout_transport.host` (default `localhost`) for both the training-batch PUSH/PULL (orchestrator → trainer) and the micro-batch PUB/SUB (data-world coordinator → each data rank).

On a **multi-node trainer** deployment the data ranks live on different nodes, so a receiver on a non-head node connects to *its own* `localhost` and never reaches the coordinator on the head node → the micro-batch sender blocks forever in `_wait_for_ready()` and training hangs at step 0. The launcher injects `--weight_broadcast.host $MASTER_ADDR` for nccl but had **no equivalent for `rollout_transport`**, and the host can't be set in TOML (`$MASTER_ADDR` is only known at launch).

## Breaking

- **Rollout `.bin` files are no longer written by default.** Anything reading `{output_dir}/rollouts/step_N/train_rollouts.bin` must set `rollout_transport.type = "filesystem"` (shared block, or identically on both `trainer` and `orchestrator`). Traces under `rollouts/step_N/{train,eval}/…/traces.jsonl` are unaffected.
- **The trainer now binds TCP ports by default** — `rollout_transport.port` (5555) plus `+1` / `+2` for the micro-batch PUB and READY sockets. Two RL runs on the same host now collide on bind; give one of them a different `rollout_transport.port`.
- Setting `rollout_transport` on only one of `trainer` / `orchestrator` is now a config error instead of a silent mismatch (previously both defaulted to filesystem, so the mismatch was reachable the same way).

## Verification

**reverse-text, 2 train + 1 infer:**
| | endpoint | steps | rollout `.bin` |
|---|---|---|---|
| before (main), zmq | `tcp://localhost:5556` | 0 in 11 min (hang) | — |
| after (PR), zmq | `tcp://…:5555` | 8 steps, ~2s each | **0** |
| filesystem | n/a | trains | 18 `.bin`/step |

**At scale — GLM-4.5-Air (100B glm4_moe), SWE (scaleswe), 2 train + 1 infer, batch 256×16, cp=4, debug logs.** Full production (R3) batches ship orchestrator→trainer over zmq and **complete real trainer optimizer steps** end-to-end:
- the **off-head** trainer node's micro-batch receiver connects to the **trainer-head hostname** (not `localhost`); the READY handshake completes (`4 ranks READY, starting PUB`) and the off-head node receives micro-batches **every step** — the exact link that hung pre-fix;
- **5/5 trainer steps `SUCCESS`** — Loss ≈ 0, Grad Norm 0.05–0.08, MFU 21.4%, ~9.3k tok/s; final checkpoint written;
- **0 rollout `.bin`** at production payload. Filesystem wrote ~17 GB/rank/step at this scale → zmq saves **multiple TB per run**.

Also unit-tested standalone: the exact transport transforms (msgspec encode → ZMQ PUSH/PULL `send_multipart(copy=False)`, HWM=10 → decode) round-trip reconstructed R3 batches at 2.9 / 5.8 / 11.6 GB with routed-experts payloads, no hang.

Gating verified in the generated sbatch (2 host injections trainer+orch; `bash -n` clean); filesystem / per-subconfig / shared-block all resolve correctly. The at-scale runs above predate the default flip and set zmq explicitly; the flip itself was verified through config resolution (`RLConfig()` → zmq on shared + both sub-configs, shared `filesystem` still overrides) and `tests/unit/test_configs.py`.

---

> [!NOTE]
> **Medium Risk**
> Changes the default rollout transport from filesystem to ZMQ (behavior and disk usage), and touches multi-node training startup wiring; incorrect host injection would still cause hangs, but the change aligns with existing NCCL host injection patterns.
> 
> **Overview**
> **Makes ZMQ the default orchestrator→trainer rollout link** and fixes multi-node SLURM jobs that hung at step 0 because every rank used `localhost` for ZMQ instead of the trainer head.
> 
> Trainer and orchestrator defaults switch from filesystem to `ZMQTransportConfig`. `RLConfig` gains a shared `[rollout_transport]` field (propagated like weight broadcast), plus `auto_setup_rollout_transport` to resolve the shared value from sub-configs and reject trainer/orchestrator type mismatches so the launcher can gate host injection correctly.
> 
> On multi-node SLURM, when rollout transport is ZMQ, the sbatch template now passes `--rollout_transport.host $MASTER_ADDR` to **torchrun** and the **orchestrator**, mirroring NCCL `weight_broadcast.host`. Docs and the monitor-run skill note ZMQ as default and that `train_rollouts.bin` only appears with `rollout_transport.type = "filesystem"`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17e731a0f6fc70d53cc8ee8a0760332326bed203. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


S1ro1 (approved):
bullish

PR #3152, opened 2026-07-29
fix: compute trainable rate over effective rollouts

mikasenghaas:
## Summary

Observed in a live run:

```
Only 260/296944 generated rollouts are trainable (0.1%) — consider reviewing task difficulty / filter config
```

The denominator was the full arrival window (`batch.rollouts` — errored, filtered and payload-less rollouts included), so the rate tracked error/filter volume rather than task difficulty, and the warning fired on healthy runs.

- Compute `Trainable` (the `Step …` line) and the low-trainable warning over `batch.rollouts.effective`, the clean trained-on subset — the only pool where "carries a gradient" is meaningful.
- Hoist `effective = batch.rollouts.effective` in `finalize_train_batch` so the warning and the downstream record/metrics paths share one view.
- Guard the warning against an empty effective window (a batch shipping only overflow from a previous window divides by zero today).

The per-env `Ratio` in the multi-env lines still uses the full window, unchanged.

---

> [!NOTE]
> **Low Risk**
> Logging and warning logic only; no changes to training payloads, filters, or shipping behavior.
> 
> **Overview**
> **Trainable** metrics and the low-trainable warning now use `batch.rollouts.effective` (the clean, trained-on cohort) instead of the full arrival window, so rates reflect gradient-bearing rollouts rather than error/filter volume.
> 
> In `finalize_train_batch`, `effective` is computed once and reused for the warning, trace save, and related paths. The warning is skipped when `effective` is empty to avoid divide-by-zero. The per-step **Step …** log’s **Trainable** fraction matches this definition; multi-env **Ratio** lines still use the full window. `TrainBatch` docstring notes trainable counts come from `rollouts.effective`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32f1579eb8fd91aa251ca41e3d2c5d2adf025d2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/orchestrator.py
+++ b/src/prime_rl/orchestrator/orchestrator.py
@@ -597,11 +597,15 @@ class Orchestrator:
                 )
             return
         self.consecutive_empty_batches = 0
-        n_trainable = sum(1 for r in batch.rollouts if r.is_trainable)
-        if n_trainable / len(batch.rollouts) <= 0.1:
+        # The clean, trained-on subset — the only pool where "carries a gradient" is meaningful.
+        # Errored, filtered and payload-less rollouts never reach the trainer, so counting them
+        # would make the rate track error/filter volume instead of task difficulty.
+        effective = batch.rollouts.effective
+        n_trainable = sum(1 for r in effective if r.is_trainable)
+        if effective and n_trainable / len(effective) <= 0.1:
             get_logger().warning(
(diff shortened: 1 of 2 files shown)


S1ro1 (approved):
'hy,

PR #3154, opened 2026-07-29
Add custom vendored kernels

MarioSieg:
Adds infrastructure for locally vendored CUDA kernels and ships the first one: a fused MoE forward kernel for Blackwell.

## Kernel infrastructure (the `kernels/` submodule)

- Kernels live in their own monorepo, [PrimeIntellect-ai/prime-kernels](https://github.com/PrimeIntellect-ai/prime-kernels), checked out here as the `kernels/` submodule. That repo is the wheel root: `kernels.toml` declares each kernel (sources, architectures), `setup.py` builds them, and the `prime_kernels` package exposes `is_available()` / `load()` / `status()` so callers never import CUDA extensions blindly. flash_moe's C++/CUDA sources are committed there too, so one non-recursive submodule init brings in everything a source build needs. Nothing about a kernel lives in prime-rl, which stays a pure-Python wheel.
- `.github/workflows/build_kernels.yaml` cross-builds wheels for x86_64 and aarch64 (CUDA 12.8, torch pinned from `uv.lock`, ABI stamped into the wheel version). It triggers on the `kernels` gitlink, so every submodule bump is build-verified. `tag-and-release.yaml` now builds and attaches the wheels to the draft release *before* publishing, so `[tool.uv.sources]` pins never point at a release without assets.
- prime-rl consumes the wheels via a new `kernels` extra, pinned to the v0.8.0 release assets. The package installs fine on non-Blackwell GPUs and simply reports its kernels unavailable.
- New `kernels` skill (plus updates to the `install`/`release` skills) documenting how to add, build, and publish kernels.

## First kernel: `flash_moe` (SM 100a / Blackwell tcgen05)

- `moe_align`, `fused_moe_bf16`, and `fused_moe_mxfp8` registered under `torch.ops.prime_moe`, with Python wrappers and `register_fake` for torch.compile.
- New trainer config field `model.moe_fused_kernel` (default **false**): routed experts run through the fused kernel in forward; backward recomputes the reference grouped-mm path. The mxfp8 variant is picked when `quantization` is MXFP8 with `enable_grouped_gemm`, otherwise bf16. Constraints (Blackwell-only, `ep=1`, `impl='custom'`, `score_before_experts=False`, divisibility requirements) are documented on the config field.

## MXFP8 correctness fixes surfaced along the way

- Token groups are now aligned to 32 for the mxfp8 fused path, and empty expert groups are clamped — torchao's wgrad leaves empty groups uninitialized, which produced NaNs.
- torchao's dim0-quantize Triton kernel indexes with int32 and corrupts memory when `numel >= 2^31`; large inputs are now chunked before quantization (`mxfp8_grouped_gemm.py`).

---

> [!NOTE]
> **High Risk**
> Touches MoE training numerics (forward kernel + hand-written/autograd backward) and release CI that uploads wheels with contents:write. Default is off, but enabling it changes expert compute and is Blackwell/EP-constrained.
> 
> **Overview**
> Adds infrastructure to vendor CUDA kernels from the `prime-kernels` submodule and ships the first one: a fused MoE forward (`flash_moe`) for Blackwell.
> 
> **Kernel packaging.** New `kernels` extra installs prebuilt wheels from release assets (x86_64/aarch64, CUDA 12.8, torch ABI stamped). `build_kernels.yaml` cross-compiles on submodule bumps; `tag-and-release.yaml` attaches wheels to the draft **before** publishing so installs never fall back to compiling CUDA. Source builds stay a manual `uv pip install -e deps/prime-kernels`.
> 
> **Trainer path.** `model.moe_fused_kernel` (default **false**) routes custom-impl MoE experts through the kernel in forward; backward is a closed-form bf16 grouped-mm or autograd through the mxfp8 reference. MXFP8 is chosen when grouped-GEMM quantization is on. Setup fails fast on EP, `score_before_experts`, missing MoE layers, or unsupported shapes.
> 
> Also chunks torchao dim0 MXFP8 quantize for tensors with `numel >= 2^31` to avoid int32 index corruption. GPU tests and a bench script compare fused vs grouped-mm.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 810558bdf8d597444e9eee0745da49eb68fec0fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kernels/prime_kernels/flash_moe/mxfp8.py
```diff
@@ -0,0 +1,19 @@
+import torch
+
+BLOCK = 32
+
+
+def pack_scales_blocked(scales: torch.Tensor) -> torch.Tensor:
```
S1ro1, line 6: any reason this is in a separate file? feels not necessary

kernels/prime_kernels/kernels.toml
```diff
@@ -0,0 +1,28 @@
...
+#   upstream     where the sources were originally developed, if anywhere; provenance only
+#   sources      C++/CUDA files to compile into prime_kernels.<name>._C
+#   arch         compute capabilities to compile for; matched exactly against the device
+#   min-cuda     oldest CUDA toolkit that can build it
+#   cxx-std      C++ standard
+
+[flash_moe]
+description = "Fused MoE forward (bf16 + mxfp8) on Blackwell tcgen05"
+ops = "prime_moe"
+upstream = "https://github.com/PrimeIntellect-ai/prime-flash-moe"
```
S1ro1: this feels useless no?

kernels/prime_kernels/kernels.toml
```diff
@@ -0,0 +1,28 @@
...
+# csrc/ copied from upstream main @ fe06a93 (2026-07-30); update when you copy again.
+sources = [
+    "csrc/torch_interface.cpp",
+    "csrc/kernels/fused_moe_bf16_kernel.cu",
+    "csrc/kernels/fused_moe_mxfp8_kernel.cu",
+]
+include-dirs = ["csrc/kernels"]
+# tcgen05 + block scaled MMA are architecture specific: sm_100a only, CUDA >= 12.8.
+arch = ["10.0a"]
+min-cuda = "12.8"
```
S1ro1: we will always support just 1 cuda version, which is increment only, would remove

kernels/pyproject.toml
```diff
@@ -0,0 +1,13 @@
...
+# The release build appends a `+cu<toolkit>torch<version>` local version, because torch C++
+# extensions are bound to the torch and CUDA they were compiled against.
+version = "0.1.0"
+description = "CUDA kernels vendored into prime-rl"
+readme = "README.md"
+requires-python = "~=3.12.0"
+dependencies = ["torch>=2.9.0"]
+
+[build-system]
+requires = ["setuptools>=77", "wheel", "ninja", "torch>=2.9.0"]
```
S1ro1, line 12: can we link the same deps as prime-rl has?

pyproject.toml
```diff
@@ -138,6 +143,13 @@ exclude = [
     "deps/research-environments/environments/tool_use/tau3_bench_v1",
 ]
 
+# Declared statically so `uv lock` never has to build the kernels (which needs nvcc) just to
+# read their metadata. Keep in sync with kernels/pyproject.toml.
+[[tool.uv.dependency-metadata]]
+name = "prime-kernels"
+version = "0.1.0"
+requires-dist = ["torch>=2.9.0"]
```
S1ro1: likewise, pin the same as prime has

pyproject.toml
```diff
@@ -216,6 +233,9 @@ flash-linear-attention = { git = "https://github.com/fla-org/flash-linear-attent
 flash_attn_3 = { index = "pytorch-cu128-test" }
 
 [tool.uv.extra-build-dependencies]
+# prime-kernels needs no entry here: `no-build-isolation-package` builds it against the
```
S1ro1: the comment seems to be a ai artifact, can remove

pyproject.toml
```diff
@@ -77,6 +77,11 @@ disagg = [
 gpt-oss = [
     "kernels",
 ]
+# CUDA kernels under `kernels/` (see kernels/README.md). Built from source, so it needs
```
S1ro1: likewise, ai generated comment, doesn't really say much

skills/install/SKILL.md
```diff
@@ -35,6 +35,16 @@ When bumping a package past the workspace-wide `exclude-newer = "7 days"` window
 
 ## Optional extras
 
+### CUDA kernels
```
S1ro1, line 38: the install in --all-extras should install pre-released wheels, no sync should build really, build should be manual

skills/kernels/SKILL.md
```diff
@@ -0,0 +1,161 @@
...
+kernel's Python surface *and* its C++/CUDA sources under `csrc/`, all declared in the single
+manifest `prime_kernels/kernels.toml`. See `kernels/README.md` once the submodule is
+initialized.
+
+Nothing about a kernel lives in prime-rl. prime-rl pins a prime-kernels commit, builds the
+wheel from it, and installs the result — and stays a pure-Python wheel itself; never add
+compiled extensions to it.
+
+Everything under `kernels/` is excluded from ruff (`tool.ruff.extend-exclude` in
+`pyproject.toml`), like `deps/`.
```
garrett361: Small organizational question: why not put it under `deps/kernels`? alongside other prime submodules

skills/kernels/SKILL.md
```diff
@@ -0,0 +1,161 @@
...
+
+```python
+import prime_kernels
+
+if prime_kernels.is_available("flash_moe"):
+    flash_moe = prime_kernels.load("flash_moe")
+```
+
+`prime_kernels.status()` maps every kernel to `"available"` or the reason it is not — log
+it once at startup rather than failing a run halfway through.
```
garrett361: very nice

skills/kernels/SKILL.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Changing or adding a kernel
+
+The work happens in the prime-kernels repo, not here. Inside `kernels/`:
+
+1. Commit the sources under `prime_kernels/<name>/csrc/`.
+2. Add a `[<name>]` table to `prime_kernels/kernels.toml` — `sources`, `include-dirs`,
+   `arch`, `cxx-std`; paths are relative to the kernel folder.
+3. Write `prime_kernels/<name>/__init__.py` — `from . import _C`, one wrapper and one
+   `torch.library.register_fake` per op.
```
garrett361: Checking, as I'm more familiar with python-only side custom op registration: we only need the `torch.library.register_fake` call, and not any `torch.library.custom_op` decorator [like here](https://docs.pytorch.org/tutorials/advanced/python_custom_ops_registrations.html#adding-training-and-other-registrations-to-python-custom-operators), because the latter is already handled cpp side?

garrett361: Maybe also a small comment that a `<op>.register_autograd` is also needed for any training kernels

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -181,6 +184,263 @@ def _run_experts_grouped_mm_impl(
...
+    w2: torch.Tensor,
+    w3: torch.Tensor,
+    selected_experts_indices: torch.Tensor,
+    top_scores: torch.Tensor,
+    num_experts: int,
+) -> torch.Tensor:
+    hidden_dim = w1.shape[1]
+    assert hidden_dim % _FUSED_MOE_BLOCK_M == 0, (
+        f"moe_intermediate_size {hidden_dim} must be a multiple of {_FUSED_MOE_BLOCK_M} for the fused MoE kernel"
+    )
```
garrett361: Feels to me like it might be `prime_kernels`'s responsibility to do these kinds of checks and raises, and have all of the `_FUSED_MOE*` env vars in that lib instead of `prime_rl`.  

With the current pattern, any changes in kernel requirement shapes would lead to changes in both libs, but could we isolate to `prime_kernels` only?

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -181,6 +184,263 @@ def _run_experts_grouped_mm_impl(
...
+    grad_w3 = torch._grouped_mm(grad_b.t(), x_sorted, offs=offs).type_as(w3) if needs_w3 else None
+    grad_w2 = None
+    if needs_w2:
+        grad_y = (g_sorted.float() * scores_sorted).to(torch.bfloat16)
+        grad_w2 = torch._grouped_mm(grad_y.t(), h, offs=offs).type_as(w2)
+    return grad_x, grad_w1, grad_w2, grad_w3, grad_scores
+
+
+class _FusedMoE(torch.autograd.Function):
+    """Forward runs the fused MoE kernel and as the fused kernel is forward only, backward is done by hand."""
```
garrett361, line 390: Doesn't look like this is tested? Would like to see some tests that outputs and grad numerics are in reasonable alignment w/ the non-fused MoE outputs/grads. 

Note sure what our usual testing stringency is @samsja @S1ro1

tests/unit/train/models/test_fused_moe.py
```diff
@@ -0,0 +1,105 @@
...
+    grad_out = torch.randn_like(out)
+    out.backward(grad_out)
+    expected.backward(grad_out)
+
+    for name, actual, reference in zip(("x", "w1", "w2", "w3", "top_scores"), fused, grouped_mm):
+        assert _rel_err(actual.grad, reference.grad) < BF16_TOL, f"grad_{name} disagrees with the grouped-mm path"
+
+
+def test_fused_mxfp8_backward_populates_every_grad(moe_inputs):
+    """The mxfp8 backward differentiates the reference, so only its wiring is worth asserting."""
```
garrett361: I get that the backwards is running through a vanilla reference, but might still be worth verifying the grad values are as expected to verify all this glue code

https://github.com/PrimeIntellect-ai/prime-rl/blob/9281fbd8f2f512ad4532caa25f509e5bc4b5372f/src/prime_rl/trainer/models/layers/moe.py?plain=1#L410-L428

up to you

tests/unit/train/models/test_fused_moe.py
```diff
@@ -0,0 +1,105 @@
...
+    pytest.mark.skipif(_UNAVAILABLE is not None, reason=f"fused MoE kernel unavailable: {_UNAVAILABLE}"),
+]
+
+NUM_EXPERTS, TOP_K, DIM, HIDDEN_DIM, NUM_TOKENS = 4, 2, 256, 128, 512
+
+# Both paths are bf16 GEMMs with fp32 accumulation but different reduction orders, so they
+# agree to bf16 precision, not bitwise. mxfp8 additionally quantizes both operands to e4m3
+# with one e8m0 scale per 32 elements, which costs roughly another order of magnitude.
+BF16_TOL = 3e-2
+MXFP8_TOL = 1.5e-1
```
garrett361: Do you know how tight these are? The MXFP8_TOL seems especially lenient, but I don't know what's typical for mxfp8

garrett361 (dismissed):
Looks good, thanks for the changes! Left minor comments, up to you whether to address. Cursor has one question, too.

Did you happen to benchmark the speedup for the full fwd + backward through the custom autograd vs the default, btw? Curious about the actual speedups.

samsja (dismissed):

MarioSieg:
> Looks good, thanks for the changes! Left minor comments, up to you whether to address. Cursor has one question, too.
> 
> Did you happen to benchmark the speedup for the full fwd + backward through the custom autograd vs the default, btw? Curious about the actual speedups.
Good point I will measure - we are clearly faster on the forward but fwd+backward we might actually be a bit slower - this is a case for a fused backward kernel in the future probably @samsja?

Doing a benchmark right now.

samsja (dismissed):

samsja (approved):

PR #3155, opened 2026-07-29
fix: flush stale eval epochs when a newer one finishes first

mikasenghaas:
## Summary

When eval epochs of the same env overlap (eval faster than `interval` allows, or a straggling rollout), an epoch triggered at step `y` can finish before the epoch at step `x < y`. Today the older epoch just keeps waiting: its cohort sits in the sink, a single stuck rollout holds the final drain hostage, and if the run is interrupted meanwhile the step-`x` scores are lost without a trace. The out-of-order completion also logs eval metrics with a regressing `step`, which leaves the W&B run summary (what run tables and reports display) stuck on the older epoch's scores.

This PR makes a newer epoch's completion a completion barrier for older epochs of the same env:

- `EvalSink.flush_before(env, step)` force-finalizes every pending epoch of `env` older than `step` (partial groups included) and marks it closed; straggler episodes arriving for a closed epoch are dropped.
- `RolloutDispatcher.cancel_eval_step(env, step)` cancels the stale epoch's in-flight rollouts and drops its group state without emitting markers (the sink already closed the bucket).
- The orchestrator logs a warning (`Eval {env} @ step=y finished before step=x - cancelled its N remaining rollouts and logging step x as-is`) and finalizes the stale epochs **before** the newer one, so eval metrics are logged monotonically in `step` and the W&B summary stays fresh.

Epochs of *different* envs are left alone — independent per-env cadences finishing at different times is normal and W&B keys them separately.

## Verification

Toy repro on reverse-text (2 GPUs, `Qwen3-0.6B-Reverse-Text-SFT`, eval `interval=1`, 16 examples × 2 rollouts), with a temporary injection delaying eval step 2's task-0 rollouts by 90s:

**Before** ([run](https://wandb.ai/primeintellect/oor-step-repro/runs/bd594a19070241f0b9a49b4013022dc9)): evals finalize in order 3, 4, 5, 6, then 2 (91s late). All points do reach W&B history correctly positioned (the `define_metric("*", step_metric="step")` custom x-axis handles out-of-order data), but the run summary ends on step 2's scores (`eval/reverse-text/all/avg@2 = 0.099` instead of step 6's `0.317`, `summary["step"] = 2`), and the final drain blocks on the straggler for the full 90s.

**After** ([run](https://wandb.ai/primeintellect/oor-step-repro/runs/23e8e94a55e04120bb95835f709df20b)): when step 3's epoch completes, the orchestrator warns (`Eval reverse-text @ step=3 finished before step=2 - cancelled its 2 remaining rollouts and logging step 2 as-is (30 rollouts)`), and logs step 2 before step 3. All epochs land in ascending step order, the summary reflects the newest epoch (`avg@2 = 0.258`, `step = 6`), and the step loop finishes in 31.5s instead of blocking 90s on the straggler.

### DIFF
--- a/src/prime_rl/orchestrator/dispatcher.py
+++ b/src/prime_rl/orchestrator/dispatcher.py
@@ -657,6 +657,27 @@ class RolloutDispatcher:
             await safe_cancel_all(tasks_to_cancel)
         return cancelled
 
+    async def cancel_eval_step(self, env_name: str, eval_step: int) -> int:
+        """Cancel all in-flight rollouts of one eval epoch and drop its group
+        state. No markers are emitted — the sink has already closed the
+        epoch's bucket, so late arrivals are dropped rather than counted."""
+        for group_id, group in list(self.groups.items()):
+            if group.kind == "eval" and group.env_name == env_name and group.eval_step == eval_step:
+                self.groups.pop(group_id, None)
+        tasks: list[asyncio.Task] = []
+        cancelled = 0
+        for task, meta in list(self.inflight.items()):
+            if meta.kind == "eval" and meta.env_name == env_name and meta.eval_step == eval_step:
+                del self.inflight[task]
+                self.release(meta.rollout_count)
+                cancelled += meta.rollout_count
+                tasks.append(task)
+        if cancelled:
+            self.metrics.record_cancellation(kind="eval", env_name=env_name, n=cancelled)
+        if tasks:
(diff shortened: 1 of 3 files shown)


mikasenghaas:
Closing — out-of-order eval epoch completion turned out to be a non-issue (wandb's custom step metric already handles it; see PR description for the investigation).

PR #3164, opened 2026-07-30
fix: checkpoint and resume the orchestrator train data position

mikasenghaas:
## Summary

- Data loading on the orchestrator was stateless across restarts: `TrainSource` shuffled once with a fixed seed and its cursors lived only in memory, so a resumed run restarted every finite taskset from the top of the same permutation, re-sampling tasks the previous run already saw.
- `TrainSource` now tracks a per-env data position `{epoch, cursor}`, exposed via `state_dict()` / `load_state_dict()` — mirroring how the SFT dataloader checkpoints its position. For finite tasksets, epochs are 1-indexed and each epoch's shuffle is seeded with the epoch number from the canonical table order, so a permutation is a pure function of the epoch and can be replayed on resume. Legacy server-side datasets get the same treatment over their index table.
- For infinite (generator-backed) tasksets the epoch stays 1 and the cursor counts generator pulls; restore fast-forwards the generator by that many pulls — an exact replay when the generator is deterministic, a harmless discard when it isn't.
- The data position is a first-class part of the orchestrator checkpoint: `CheckpointManager.save/load` take the `TrainSource` alongside `Progress` and store its state under its own `train_source` key in `progress.pt` (`Progress` itself is unchanged). On resume, `load` restores it and logs one line per env: `Resumed data position for env <name> - epoch=E, cursor=C/N` (`N` is `infinite` for generator-backed tasksets). `--ckpt.skip-progress` skips both the counters and the data position.
- The interval save runs synchronously on the event loop (no `asyncio.to_thread`): the payload is tiny, and snapshotting inline keeps the dispatcher from mutating `TrainSource` mid-save, which would tear the snapshot.
- The cross-env dispatch sequence is reproducible too: the env-choice RNG advances exactly once per dispatched example — a draw whose permit cost doesn't fit is held as `pending_env` (head-of-line) instead of being redrawn — and the RNG state + pending draw are part of the `TrainSource` state. Without the hold, permit-rejected attempts would consume draws at timing-dependent points, making the interleave irreproducible even within a single run. The trade-off is head-of-line blocking: a pending group draw waits for `group_size` permits rather than letting a cheaper env slip through. (`TrainSource` also no longer takes a `seed` — the RNG only drives env choice now that shuffles are epoch-seeded, so the previously hardcoded call-site value moved inside.)

Known approximation: cursors advance when the dispatcher pulls an example, ahead of batch shipping, so tasks in flight at checkpoint time are counted as consumed — a resume skips them for that epoch (up to ~`max_inflight_episodes`). Trading that for duplicates would need shipped-task tracking; not worth the complexity. (Completion order and batch composition also stay timing-dependent — what's reproducible is the dispatch sequence, not the training batches.)

## Breaking

- Orchestrator checkpoints written before this PR can't be resumed: their `progress.pt` has no `train_source` key, so `CheckpointManager.load` fails with `KeyError: 'train_source'`. Migration: finish (or restart) in-flight runs rather than resuming them across this change, or resume with `--ckpt.skip-progress` (which skips loading the counters and the data position).

## Verification

reverse-text RL (`examples/basic/reverse-text/rl.toml`, 1000 tasks, batch 128 / group 16 → 8 tasks per step), 10 steps + resume for 10 more:

```
# run 1: uv run rl @ examples/basic/reverse-text/rl.toml --max-steps 10 --no-wandb --ckpt
step-10 checkpoint: progress=Progress(step=10, ...)
                    train_source={'rng': <625-word MT state>, 'pending_env': None,
                                  'envs': {'reverse-text': {'epoch': 1, 'cursor': 100}}}

# run 2: … --max-steps 20 --ckpt.resume-step 10
INFO  Resumed data position for env reverse-text - epoch=1, cursor=100/1000
step-20 checkpoint: envs={'reverse-text': {'epoch': 1, 'cursor': 199}}
```

Cross-checking `task.data.idx` in the saved traces: steps 1–10 dispatched 85 distinct tasks, steps 11–20 dispatched 84, with **zero overlap** — the resumed run continued the epoch-1 permutation instead of restarting it (before this change, run 2 restarted from cursor 0 and re-sampled run 1's tasks). Unit round-trips additionally cover the multi-env interleave (an interrupted+resumed source replays the uninterrupted dispatch sequence exactly, including a pending draw held across a checkpoint), infinite-generator fast-forward, and stale env names in old state.

### DIFF
--- a/src/prime_rl/orchestrator/train_source.py
+++ b/src/prime_rl/orchestrator/train_source.py
@@ -14,11 +14,12 @@ import random
 from collections.abc import Iterator
 
 import verifiers.v1 as vf
+from torch.distributed.checkpoint.stateful import Stateful
 
 from prime_rl.orchestrator.envs import TrainEnvs
 
 
-class TrainSource:
+class TrainSource(Stateful):
     """``next_example(available_permits)`` picks a weighted-RR env and
     returns its next example (or ``None`` when the env's per-call permit
     cost doesn't fit — the dispatch loop retries when permits free up).
@@ -26,7 +27,7 @@ class TrainSource:
     whose data is shipped to the env server at dispatch).
 
     An env's data position is ``{epoch, cursor}``, round-tripped through a
-    checkpoint via ``get_state()`` / ``load_state()``. For a finite env,
+    checkpoint via ``state_dict()`` / ``load_state_dict()``. For a finite env,
     epochs are 1-indexed and seed that epoch's shuffle; for an infinite env
     the epoch stays 1 and the cursor counts generator pulls, replayed on
     restore by fast-forwarding the generator (exact iff it's deterministic).
@@ -78,12 +79,12 @@ class TrainSource:
         random.Random(self.epochs[env_name]).shuffle(rows)
         return rows
 
-    def get_state(self) -> dict[str, dict[str, int]]:
+    def state_dict(self) -> dict[str, dict[str, int]]:
         """Per-env data position ``{epoch, cursor}``."""
         return {name: {"epoch": self.epochs[name], "cursor": self.cursors[name]} for name in self.epochs}
 
-    def load_state(self, state: dict[str, dict[str, int]]) -> None:
-        for name, position in state.items():
+    def load_state_dict(self, state_dict: dict[str, dict[str, int]]) -> None:
+        for name, position in state_dict.items():
             if name not in self.base_rows:
                 continue
             self.epochs[name] = position["epoch"]
--- a/src/prime_rl/orchestrator/train_source.py
+++ b/src/prime_rl/orchestrator/train_source.py
@@ -25,18 +25,22 @@ class TrainSource:
     Returned dicts carry ``env_name`` + ``task_idx`` (+ ``task`` for v1 envs,
     whose data is shipped to the env server at dispatch).
 
-    An env's data position is ``{epoch, cursor}``, round-tripped through a
-    checkpoint via ``state_dict()`` / ``load_state_dict()``. For a finite env,
-    epochs are 1-indexed and seed that epoch's shuffle; for an infinite env
-    the epoch stays 1 and the cursor counts generator pulls, replayed on
-    restore by fast-forwarding the generator (exact iff it's deterministic).
-    The cursor advances at dispatch time (ahead of shipped batches), so a
-    resume skips the tasks that were in flight at checkpoint time."""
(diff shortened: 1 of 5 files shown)


src/prime_rl/orchestrator/train_source.py
```diff
@@ -22,40 +23,79 @@ class TrainSource:
...
+    An env's data position is ``{epoch, cursor}``, round-tripped through a
+    checkpoint via ``state_dict()`` / ``load_state_dict()``. For a finite env,
+    epochs are 1-indexed and seed that epoch's shuffle; for an infinite env
+    the epoch stays 1 and the cursor counts generator pulls, replayed on
+    restore by fast-forwarding the generator (exact iff it's deterministic).
+    The cursor advances at dispatch time (ahead of shipped batches), so a
+    resume skips the tasks that were in flight at checkpoint time."""
+
+    def __init__(self, train_envs: TrainEnvs) -> None:
+        # Env-choice RNG; not checkpointed (i.i.d. sampling, restarting the
```
hallerite: why not also keep that for proper reproducibility?

hallerite (commented):
almost lgtm

src/prime_rl/orchestrator/train_source.py
mikasenghaas: this is just the seed to sample from envs. i think its fine to not replay that tbh

rawwerks:
Feel free to ignore this comment if it's already accounted for, or if I'm completely mis-understanding the plan...

I'm experimenting with custom training-example selection and was reading this alongside #2722. This PR moves the per-env cursor and env-choice RNG **into** `TrainSource.state_dict()`, while #2722 moves that same state **out** into `ShuffledCursorSampler` / `WeightedRoundRobin`. #2722's strategy ABCs don't have a state contract, so if it ever gets rebased on top of this, custom strategies would quietly lose checkpoint/resume.

Not suggesting anything should change here — just wondering whether `state_dict()` is meant to stay `TrainSource`'s, or eventually become something selection strategies contribute to.

(Related to: https://github.com/PrimeIntellect-ai/prime-rl/issues/3173)

hallerite (approved):

mikasenghaas:
hi @rawwerks! i do agree that the responsibility for data state (and thus, checkpointing) will prob move out of the train source once we have the env sampler. but since this is a feature we need to get in and right now we dont have the abstraction, it will live where its most at home today, which is the train source. but ya, expecting this to move over most likely

PR #3175, opened 2026-08-03
fix(nccl): preserve checkpoint layer boundaries

biswapanda:
## Summary

**Why:** Prime-RL broadcasts checkpoint weights in layer/state-dict groups. Flattening those groups before vLLM reloads them loses the checkpoint boundaries needed for correct reconstruction and can apply weights incorrectly.

**Changes:**

- preserve each received checkpoint state dict as a separate stream
- initialize and finalize vLLM's layerwise reload once while loading each group independently
- retain the existing flattened stream for kernel-format updates
- add regression coverage for grouped and ungrouped weight loading

## Scope

This is an independent correctness fix shared by native Prime-RL and external inference backends.

## Test plan

- `ruff check` on the changed NCCL worker and test
- `pytest -q tests/unit/inference/test_nccl_weight_update.py`

---

> [!NOTE]
> **Medium Risk**
> Touches live inference weight-update paths over NCCL; the change is narrow and regression-tested but incorrect grouping would break model weights.
> 
> **Overview**
> Fixes checkpoint-format NCCL weight reloads by **not flattening** multiple trainer broadcasts into one tensor stream.
> 
> `NCCLWeightBroadcastReceiver` now exposes **`receive_state_dicts()`**, yielding one generator per layer broadcast, while **`receive_state_dict()`** still flattens streams for kernel-format loading. Non-quantized updates call **`load_weight_groups_checkpoint_layerwise`**, which runs vLLM’s **`initialize_layerwise_reload` / `finalize_layerwise_reload`** once and invokes **`model.load_weights`** separately for each group.
> 
> Adds unit tests that assert per-layer streams stay separate and that grouped reload runs init → multiple loads → finalize in order.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6cca92c5724bf8d29e4940ff93698e6f254a4bbc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/inference/vllm/worker/nccl.py
+++ b/src/prime_rl/inference/vllm/worker/nccl.py
@@ -78,15 +78,22 @@ class NCCLWeightBroadcastReceiver:
         self.communicator = PyNcclCommunicator(pg, device=device)
 
     @torch.no_grad()
-    def receive_state_dict(self):
-        """Receives the state dict of a model from the trainer master rank using NCCL communicator."""
+    def receive_state_dicts(
+        self,
+    ) -> Generator[Generator[tuple[str, torch.Tensor], None, None], None, None]:
+        """Receive each trainer-broadcast state dict as a separate stream."""
         logger.info("Receiving weights from trainer")
         num_state_dict_to_receive = receive_integer(self.communicator)
         logger.info(f"Receiving {num_state_dict_to_receive} layer state dicts")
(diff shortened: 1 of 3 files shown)


S1ro1:
I'm quite suspicious of this, we use this path in every run and never had any issues. Do you have some examples to prove this?

PR #3177, opened 2026-08-03
fix(inference): map external engine ranks

biswapanda:
## Summary

**Why:** An external inference endpoint can represent several TP/PP/DP ranks, so assigning one collective rank per admin URL is incorrect. Multi-engine weight transfer needs deterministic, non-overlapping rank spans or collectives can deadlock or update the wrong workers.

**Changes:**

- calculate a global rank for each vLLM worker from its TP, PP, and DP coordinates
- assign each external engine a rank offset and world-size span
- pass those spans into NCCL and NIXL collective initialization
- handle dense and expert-parallel vLLM worker layouts
- preserve the existing locally managed inference topology

## Scope

This is backend-neutral topology support. It does not introduce Dynamo configuration or discovery.

## Test plan

- `ruff check` on rank, client, worker, and test files
- `pytest -q tests/unit/inference/test_nccl_rank.py tests/unit/utils/test_external_engine_topology.py`

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 6fe4249ed816917086ae35a09f6663ef8ad5dfda. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/inference/vllm/ranks.py
+++ b/src/prime_rl/inference/vllm/ranks.py
@@ -1,0 +1,39 @@
+def global_inference_rank(
+    *,
+    rank_offset: int,
+    data_parallel_index: int,
+    data_parallel_size: int,
+    worker_rank: int,
+    tensor_parallel_size: int,
+    pipeline_parallel_size: int,
+    inference_world_size: int,
+    prefill_context_parallel_size: int = 1,
+    engine_world_size: int | None = None,
+) -> int:
+    """Map one vLLM worker to its rank in Prime's inference NCCL group."""
+    model_parallel_size = tensor_parallel_size * pipeline_parallel_size * prefill_context_parallel_size
+    logical_data_parallel_size = data_parallel_size
+    if engine_world_size is not None:
+        if engine_world_size <= 0 or engine_world_size % model_parallel_size:
+            raise ValueError(
+                f"engine world size {engine_world_size} is not divisible by model parallel size {model_parallel_size}"
+            )
+        logical_data_parallel_size = engine_world_size // model_parallel_size
+        # Dense vLLM EngineCore processes retain their global DP index but rewrite
+        # data_parallel_size to one. MoE EngineCore processes preserve the logical
+        # size, so keep validating that value rather than masking bad discovery.
+        if data_parallel_size != 1 and data_parallel_size != logical_data_parallel_size:
+            raise ValueError(
+                f"data parallel size {data_parallel_size} does not match engine-derived size "
+                f"{logical_data_parallel_size}"
+            )
+    if not 0 <= data_parallel_index < logical_data_parallel_size:
+        raise ValueError(
+            f"data parallel index {data_parallel_index} is outside logical data parallel size "
+            f"{logical_data_parallel_size}"
+        )
+
+    rank = rank_offset + data_parallel_index * model_parallel_size + worker_rank % model_parallel_size
+    if not 0 <= rank < inference_world_size:
+        raise ValueError(f"calculated inference rank {rank} is outside inference world size {inference_world_size}")
+    return rank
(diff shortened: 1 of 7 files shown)


src/prime_rl/inference/vllm/worker/nccl.py
```diff
@@ -101,27 +102,37 @@ def init_broadcaster(
...
-        # The previous dp_group-based computation broke in vLLM v1 multiprocess
-        # DP mode where each worker is a separate process with a singleton
-        # DP group (rank_in_group is always 0).
-        local_rank = self.device.index
-        global_rank_inference = rank_offset + local_rank
+        if engine_world_size is None:
+            global_rank_inference = rank_offset + self.device.index
+        else:
+            parallel_config = self.parallel_config
+            global_rank_inference = global_inference_rank(
```
S1ro1, line 120: Would be for having a single path be used here, we can maybe consolidate this?

src/prime_rl/inference/vllm/worker/nixl.py
```diff
@@ -96,9 +97,24 @@ def init_broadcaster(
...
+        engine_world_size: int | None = None,
     ) -> None:
-        del inference_world_size, quantize_in_weight_transfer
-        global_rank = rank_offset + self.device.index
+        del quantize_in_weight_transfer
+        if engine_world_size is None:
+            global_rank = rank_offset + self.device.index
+        else:
+            parallel_config = self.parallel_config
+            global_rank = global_inference_rank(
```
S1ro1, line 107: Likewise

src/prime_rl/inference/vllm/ranks.py
```diff
@@ -0,0 +1,39 @@
...
+    pipeline_parallel_size: int,
+    inference_world_size: int,
+    prefill_context_parallel_size: int = 1,
+    engine_world_size: int | None = None,
+) -> int:
+    """Map one vLLM worker to its rank in Prime's inference NCCL group."""
+    model_parallel_size = tensor_parallel_size * pipeline_parallel_size * prefill_context_parallel_size
+    logical_data_parallel_size = data_parallel_size
+    if engine_world_size is not None:
+        if engine_world_size <= 0 or engine_world_size % model_parallel_size:
```
S1ro1, line 17: These 3 ValueErrors seem to never be able to happen

src/prime_rl/utils/client.py
```diff
@@ -524,32 +526,46 @@ async def init_nccl_broadcast(
     """
     logger = get_logger()
 
+    has_explicit_engine_world_sizes = engine_world_sizes is not None
```
S1ro1, line 529: again feels quite clunky by a way we have 2 separate paths, can we consolidate to 1 that is agnostic?

src/prime_rl/utils/client.py
```diff
@@ -524,32 +526,46 @@ async def init_nccl_broadcast(
...
+            inference_world_size = sum(engine_world_sizes)
+        else:
+            inference_world_size = len(admin_clients)
         logger.warning(
             f"inference_world_size not provided, defaulting to {inference_world_size} (one GPU per admin client)"
         )
 
-    gpus_per_server = inference_world_size // len(admin_clients)
+    if engine_world_sizes is None:
+        if inference_world_size % len(admin_clients) != 0:
```
S1ro1, line 540: These runtime checks seem quite eh, would again make it "always work" and be a single path for both raw vllm and dynamo

src/prime_rl/utils/client.py
```diff
@@ -571,31 +589,61 @@ async def init_nixl_broadcast(
...
     session_id: str,
+    *,
+    engine_world_sizes: list[int] | None = None,
 ) -> None:
     """Configure every vLLM worker for NIXL + ModelExpress pulls."""
-    workers_per_server = inference_world_size // len(admin_clients)
-
-    async def initialize(admin_client: AsyncClient, rank_offset: int) -> None:
+    has_explicit_engine_world_sizes = engine_world_sizes is not None
+    if engine_world_sizes is None:
```
S1ro1, line 597: Same concern as above, is very weird to do these checks at runtime and also not a fan of the split path

S1ro1 (changes requested):

PR #3094, opened 2026-07-20
fix: dp=1 EP replicas launch dense-style (vLLM rejects external-LB args at dp=1)

mikasenghaas:
## Problem
Multi-node inference with `enable_expert_parallel` and per-node data-parallel size 1 (e.g. `tp=8` on 8-GPU nodes → `dp_local=1`; the "etp8" shape: experts EP-sharded across the node's 8 GPUs, attention at full tp=8) fails to launch. The EP branch in `multi_node_rl.sbatch.j2` unconditionally passes vLLM's external-LB DP args, which vLLM rejects when `data_parallel_size == 1`:

```
pydantic_core.ValidationError: data_parallel_external_lb can only be set when data_parallel_size > 1
```

## Fix
Guard the external-LB EP launch on `INFER_GROUP_DP > 1`; for dp=1 EP replicas, launch dense-style (independent single-engine server). `enable_expert_parallel` in the engine config alone still shards experts across the TP group — no external-LB coordination needed within a single DP rank.

Split out of #3067 as an independent bug fix (the global-router change there is separate and still under evaluation). Validated: etp8 launches and serves cleanly on 8 nodes with this fix; without it, launch crashes at engine init.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Narrow SLURM launcher branch change for dp=1 EP; multi-DP EP behavior unchanged when `INFER_GROUP_DP > 1`.
> 
> **Overview**
> Fixes **multi-node RL inference launch** when `enable_expert_parallel` is on but the replica’s effective data-parallel size is **1** (e.g. one node with `tp=8` → `INFER_GROUP_DP=1`).
> 
> The EP branch in `multi_node_rl.sbatch.j2` now requires **`INFERENCE_ENABLE_EXPERT_PARALLEL` and `INFER_GROUP_DP > 1`** before passing external load-balancer DP fields (`data_parallel_rank`, `data_parallel_address`, RPC port, `INFER_GROUP_DP`). Otherwise it uses the **dense-style** `launch_inference_rank` path (`dp=1`, no DP coordinator args), because vLLM rejects `data_parallel_external_lb` when `data_parallel_size == 1`.
> 
> Expert sharding within a single TP group still comes from **`enable_expert_parallel` in the inference config**; only cross-rank external-LB coordination is skipped for dp=1 EP.
> 
[...]

PR #3184, opened 2026-08-03
fix: standalone inference at dp=1 EP (template) + poolside_v1 tool-call regex

mikasenghaas:
Same bug as #3094, same fix, in the sibling template: the multi-node branch of `inference.sbatch.j2` still passes the external-LB DP args (`data_parallel_rank`/`data_parallel_address`) when `dp_per_node == 1` (e.g. `tp = 8` on 8-GPU nodes), which vLLM ≥0.24 rejects at startup with `data_parallel_external_lb can only be set when data_parallel_size > 1`. #3094 guarded this in `multi_node_rl.sbatch.j2` only.

Hit in practice by the research-prod `int5/inference/{super,ultra}.toml` deployments (PrimeIntellect-ai/research-prod#158) at `num_nodes > 1`: every replica crash-looped on the ValidationError. With this guard, dp=1 EP replicas launch dense-style (EP shards experts across the TP group; no DP linkage needed), matching the RL template's behavior. Verified by relaunching both deployments at 2 and 4 replicas.

---

> [!NOTE]
> **Low Risk**
> Template-only launch-path change; aligns inference with the RL template and fixes crash-loops without altering training or auth paths.
> 
> **Overview**
> **Fixes multi-node standalone inference crash** when expert parallel is on but each node runs a single DP rank (`dp_per_node == 1`, e.g. `tp=8` on 8 GPUs).
> 
> The multi-node branch in `inference.sbatch.j2` now only passes `data_parallel_rank` / `data_parallel_address` when **both** `EP=1` and `DP_PER_NODE > 1`. Otherwise it launches a single-engine server with `data_parallel_size=1` and no external-LB JSON—matching the guard already added in `multi_node_rl.sbatch.j2` (#3094). vLLM ≥0.24 rejects external-LB when `data_parallel_size == 1`; EP within one TP group still comes from `enable_expert_parallel` in config.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0620f22d571681ef318ee8e2389c0121fcd15fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/inference/patches.py
+++ b/src/prime_rl/inference/patches.py
@@ -15,7 +15,6 @@ def apply_shared_vllm_patches():
     _patch_qwen35_moe_lora_format()
     monkey_patch_nano_v3_reasoning_parser()
     monkey_patch_qwen3_coder_param_newline_trim()
-    monkey_patch_poolside_v1_tool_call_name_boundary()
     monkey_patch_minimax_m2_think_end_passthrough()
     monkey_patch_vllm_padded_input_scrub()
     monkey_patch_return_routed_experts_with_nixl_connector()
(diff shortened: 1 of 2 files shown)


mikasenghaas:
Added a second, related fix on this branch: `poolside_v1`'s non-streaming `func_detail_regex` requires a newline after the tool name, but Laguna S 2.1 emits `<tool_call>name<arg_key>...` without one (upstream vLLM ≥0.25 accepts both) — every tool call leaked verbatim into `content`. Patched via `apply_shared_vllm_patches` following the existing parser-patch pattern; verified against live Laguna-S-2.1-FP8 output. Both fixes are exercised by the research-prod int5 inference deployments.

samsja (dismissed):

PR #3189, opened 2026-08-04
feat: remove multi-tenant

mikasenghaas:
## Summary

Removes the multi-tenant infrastructure end to end and simplifies every codepath that existed for it:

- **No more packer / double fs hop**: bin-packing moves into the orchestrator (`orchestrator/packing.py::BatchPacker`, wrapping the now torch-free `trainer/batch.py::prepare_batch`), which ships per-DP-rank micro batches directly to the trainer — one transport hop instead of two. The trainer-side packer (`trainer/rl/packer.py`), the hop-1 `TrainingBatch` sender/receiver pairs, and the packer watchdog are deleted. ZMQ data now flows on `port`, the READY barrier on `port + 1` (`port + 2` freed), and the transport binds in the orchestrator process (the multi-node sbatch template addresses the orchestrator's node via `ORCH_ADDR`).
- **No more run discovery / run manager**: `MultiRunManager` (run discovery, eviction, the `control/orch.toml` handshake, `synchronize_state`), `MultiCheckpointManager`, `MultiLoRAOptimizer`, `MultiLoRAScheduler`, `MultiPacker`, and the per-run Prometheus metrics are deleted. A slim `LoRAState` in `trainer/lora.py` keeps single-tenant LoRA (RL + SFT) working: adapter module registry, adapter state dicts, resets, and the shared `lora_num_tokens`/scaling tensors with `n_adapters = 1`.
- **No more run dir**: the `run_default` nesting is gone — `orchestrator.output_dir == trainer.output_dir`, one flat experiment dir for rollouts, broadcasts, checkpoints, traces, and token exports.
- **No more elastic inference pool**: `utils/elastic.py` (DNS discovery + per-tenant adapter sync) is deleted along with the `InferencePool` protocol/factory and client-side load balancing. `client.base_url` is a single string; `admin_base_url` stays a list for per-server admin operations (weight updates, health checks, NCCL/NIXL rank offsets).
- **No more branching on checkpointing**: the trainer has exactly one checkpoint path (interval saves + final save + weight checkpoints); the resume-step `STABLE`-vs-all branch and `get_stable_ckpt_steps` are gone.
- **Always broadcast step 0**: the trainer's startup broadcast of the incoming policy (v0 from scratch, v{resume_step} on resume) runs for every transport, and the orchestrator's startup weight sync is unconditional to match — a broken broadcast path now fails at startup, and fresh starts align with the restart flow.
- Configs, docs, tests, skills, CI matrix, and the k8s chart are cleaned up accordingly (`reverse-text-multi-run` CI scenario deleted; dead `HEADLESS_SERVICE` env vars removed).

## Breaking

- **`client.base_url` is a single string** (was `list[str]`). Multi-replica data planes need a router in front (the multi-node template already routes through `vllm-router`); per-server admin traffic goes through `client.admin_base_url` (still a list). `client.elastic` and `client.router_url` are removed.
- **`orchestrator.output_dir` must equal `trainer.output_dir`** — the `run_default` subdirectory is gone and the on-disk layout flattens (`broadcasts/`, orchestrator `checkpoints/`, `rollouts/`, `token_exports/` move up one level). Resuming a pre-existing experiment dir across this boundary is unsupported.
- **Multi-tenant training is removed entirely**: `trainer.max_concurrent_runs`, run discovery/eviction, per-run checkpoints and token-export dirs, the orchestrator `[optim]` block, and the `trainer_run_*` / `trainer_runs_*` Prometheus gauges.
- **Transport wire/ports changed**: the orchestrator ships packed per-rank micro batches (no `TrainingBatch`); ZMQ data moved from `port + 1` to `port`, READY from `port + 2` to `port + 1`, and the sockets bind in the orchestrator process. Mixed-version orchestrator/trainer pairs cannot talk.
- **`orchestrator.num_train_workers` and the new `orchestrator.pad_to_multiple_of` are load-bearing**: they define the packed grid (trainer DP world size, cp padding). The `rl` entrypoint auto-fills both (the multi-node fill now correctly divides by `trainer.model.cp`); standalone orchestrator runs must set them.
- **Filesystem-transport runs block at startup** until the trainer broadcasts step 0 (default wait 1200s; `ckpt.wait_for_weights_timeout` overrides). Large models on slow shared filesystems may need the timeout raised.
- **RL LoRA optimizers now include `modules_to_save` params** (previously silently excluded in RL, included in SFT) — intentional alignment with the SFT behavior.

## Verification

- `pytest tests/unit -m "not gpu"`: 442 passed (one pre-existing `test_qwen3_vl_e2e` failure reproduces on unmodified `main` on the same box — local dependency drift, unrelated).
- End-to-end smoke: `uv run rl @ configs/ci/integration/reverse-text/start.toml` (ZMQ rollout transport + filesystem weight broadcast, 15 steps) — exercises orchestrator-side packing, the single-hop transport, the flat output dir, and the new unconditional step-0 broadcast/startup sync.

---

> [!NOTE]
> **High Risk**
> Large architectural change to rollout transport, ZMQ ports/bind host, output-dir layout, and weight-sync startup—mixed-version or resumed old run dirs will break; training correctness depends on orchestrator/trainer topology fields matching.
> 
> **Overview**
> **Removes multi-tenant training** end to end: `MultiRunManager`, trainer-side packer, per-run checkpoints/optimizers, elastic inference discovery, and the `reverse-text-multi-run` CI scenario. LoRA is **single-adapter** via `LoRAState` (`n_adapters=1`); packing no longer tracks per-run `lora_num_tokens` in micro-batches.
> 
> **Rollout path is one hop:** the orchestrator **`BatchPacker`** calls `prepare_batch` and ships a **per-DP-rank micro-batch grid** (`MicroBatchSender`); the trainer only receives. Weight broadcast and checkpoints assume **one experiment** under a shared flat `output_dir` (no `run_default` nesting).
> 
> **Config / client breaking changes:** `client.base_url` and frozen-model `base_url` are a **single string** (not a list); `client.elastic` is removed. `trainer.max_concurrent_runs` and orchestrator `[optim]` are gone. The `rl` entrypoint sets `orchestrator.pad_to_multiple_of` from `trainer.model.cp` and fixes multi-node `num_train_workers` to account for CP.
> 
> **Startup alignment:** the trainer **always** broadcasts the incoming policy at step start (all transports); the orchestrator **always** waits for that weight dir before the first step (configurable timeout). ZMQ rollout binds on the orchestrator node (`ORCH_ADDR` in SLURM); k8s splits `INFERENCE_URL` (first replica) vs `ADMIN_INFERENCE_URLS` (all replicas).
> 
> Docs/skills updated for flat outputs, filesystem rollouts as `rank_<rank>.bin`, and dropped multi-tenant sections.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 005a9fcbc8975894c25ee45c38e7c7cff03989af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


samsja (approved):
now its all green

PR #3192, opened 2026-08-04
perf(nemotron-h): use fla causal_conv1d for packed Mamba conv

hubert-marek:
## Problem

Both NemotronH Mamba paths derive their packed-document metadata inside the layer:

- `models/nemotron_h/modeling_nemotron_h.py:93` (`_varlen_mamba_forward`, non-CP) — `cu_seqlens.tolist()`, a per-document `nn.Conv1d` loop, a `torch.cat`, and a second Python loop to build `seq_idx`.
- `models/layers/cp_mamba.py:183` (`mamba_cp_forward`, Ulysses CP) — the same `.tolist()`, `_causal_conv1d_varlen`, and `_build_seq_idx`.

`.tolist()` on a CUDA tensor is a host-device sync. It runs once per Mamba layer, so on Nemotron-3-Super (40 `M` layers in `hybrid_override_pattern`) that is 40 stream drains per forward, 80 per micro-step with activation recompute. Each one blocks the Python thread until the stream drains, so the CPU cannot run ahead and enqueue the next layer's work, and per-step jitter on any rank propagates to all of them instead of being absorbed in queue depth.

The per-document conv loop compounds it: kernel launches scale with the document count (4–58 per rank on a 131k pack), which is also the most imbalanced work in the layer.

## Change

Use fla's Triton `causal_conv1d` and `prepare_sequence_ids` in both paths, matching the call shape already used for Qwen3.5 GDN in `trainer/model.py:203` and `qwen3_5_moe/modeling_qwen3_5_moe.py:168`:

```python
hidden_states_B_C, _ = fla_causal_conv1d(
    x=hidden_states_B_C,
    weight=self.conv1d.weight.squeeze(1),
    bias=self.conv1d.bias,
    activation=self.activation,
    cu_seqlens=cu_seqlens,
)
...
seq_idx = prepare_sequence_ids(cu_seqlens).to(torch.int32).unsqueeze(0)
```

Both fla helpers take `cu_seqlens` directly and are `@tensor_cache`d by argument identity, so the 40 layers of a forward share one build. `_build_seq_idx` and `_causal_conv1d_varlen` are deleted; the `[B, T, D]` layout fla expects is what the code already holds, so the transposes and the `torch.cat` go too, and `activation=` folds in the separate `silu`.

`flash-linear-attention` is already a required dependency (`pyproject.toml:40`) and its Triton backend needs no arch-specific CUDA compilation — that was the stated reason in `_patch_mamba2_use_triton_ssd`'s docstring for avoiding a fused conv here, so the docstring is updated too.

`cp_context` is deliberately left unset in `mamba_cp_forward`: after the seq-to-head all-to-all the conv is head-parallel and local over the full sequence, so fla's sequence-parallel CP conv is not the right primitive there.

## Measured effect

Nemotron-3-Super-120B-A12B, 8×H200 nodes (64 GPUs), seq_len 131072, cp=2 ulysses, full activation checkpointing. Four runs resumed from the same checkpoint over the **same 106 training steps with the same data order**, so the only variable is this patch:

| run | conv | mean step | MFU | slowest step |
|---|---|---|---|---|
| **this PR** | **fla** | **133.8 s** | **48.0%** | **150 s** |
| baseline a | stock | 143.7 s | 44.2% | 199 s |
| baseline b | stock | 145.7 s | 44.1% | 231 s |
| baseline c | stock | 144.9 s | 44.3% | 211 s |

**7.6% faster** (144.8 s → 133.8 s), **+3.8 MFU points**. The tail matters more than the mean: worst-case step drops from 199–231 s to 150 s, which is the run-ahead effect — without a per-layer sync, one slow rank no longer re-synchronizes the whole cluster at every layer.

## Correctness

Numerical equivalence against the per-document `nn.Conv1d` loop, forward and backward, relative L2:

| dtype | C | docs | T | fwd | dx | dw |
|---|---|---|---|---|---|---|
| fp32 | 128 | 4 | 2096 | 8.2e-08 | 7.8e-08 | 1.6e-07 |
| fp32 | 5120 | 2 | 4096 | 8.1e-08 | 7.7e-08 | 2.4e-07 |
| bf16 | 5120 | 4 | 4096 | 2.9e-03 | 2.7e-03 | 3.9e-03 |
| bf16 | 5120 | 1 | 4096 | 2.9e-03 | 2.7e-03 | 2.5e-03 |
| bf16 | 2048 | 3 | 4096 | 2.9e-03 | 2.7e-03 | 3.7e-03 |

fp32 agrees to fp32 rounding; bf16 to bf16 rounding (eps ≈ 7.8e-03), as expected from a different accumulation order. On the full 120B model this shows up as **1e-4**: resuming the same checkpoint into the same data, the first step logs loss 0.5301 / grad-norm 0.1257 with this patch versus 0.5300 / 0.1253–0.1255 across the three stock baselines. Not bit-identical, so it is best adopted at a phase boundary rather than mid-run.

## Tests

`tests/unit/train/models/test_cp_mamba.py` now needs a GPU (Triton), so it takes `pytestmark = [pytest.mark.gpu]` like the other model tests. `test_mamba_cp_forward_rejects_local_boundaries` covered a validation branch inside the deleted `_build_seq_idx`; it is replaced by `test_mamba_cp_forward_builds_seq_idx_for_uneven_packs`, which asserts the document ids fla produces for an uneven 3-document pack.

On one H200: `test_cp_mamba.py` 2 passed. `test_nemotron_h.py` + `test_nemotron_h_kl.py` 9 passed, 5 failed, 1 xfailed — the 5 failures (`AssertionError: varlen mamba expects batch_size=1 (packed)`) reproduce identically on unmodified `main` and are unrelated.

## What this does not do

This was written while chasing a deterministic multi-node hang in a Mamba CP all-to-all, on the theory that the per-layer sync was the mechanism. **It is not.** With this patch applied and the data order held fixed, the hang reproduces at exactly the same training step. The patch does change where ranks block — inside `all_to_all_single` rather than behind `cu_seqlens.tolist()` in a `cudaMemcpy` — which makes the stranded ranks visible to the NCCL flight recorder instead of hiding them. Useful, but the root cause is elsewhere and is not addressed here.

### DIFF
--- a/src/prime_rl/trainer/models/layers/cp_mamba.py
+++ b/src/prime_rl/trainer/models/layers/cp_mamba.py
@@ -16,45 +16,8 @@ from __future__ import annotations
 
 import torch
 import torch.distributed as dist
-
-
-def _build_seq_idx(boundaries: list[int], seq_len: int, device: torch.device) -> torch.Tensor:
-    """Return the packed-document id for each token in a full CP sequence."""
-    if boundaries[0] != 0 or boundaries[-1] != seq_len:
-        raise ValueError(f"cu_seqlens must span the full sequence length {seq_len}, got {boundaries}")
-
-    seq_idx = torch.empty(seq_len, dtype=torch.int32, device=device)
-    for sequence_id, (start, end) in enumerate(zip(boundaries[:-1], boundaries[1:])):
-        if end <= start:
-            raise ValueError(f"cu_seqlens must be strictly increasing, got {boundaries}")
-        seq_idx[start:end] = sequence_id
-    return seq_idx.unsqueeze(0)
-
-
-def _causal_conv1d_varlen(
-    hidden_states: torch.Tensor,
-    weight: torch.Tensor,
-    bias: torch.Tensor | None,
-    boundaries: list[int],
-) -> torch.Tensor:
-    """Apply a depthwise causal convolution independently to packed documents."""
-    batch_size, _, channels = hidden_states.shape
-    if batch_size != 1:
-        raise ValueError(f"packed CP Mamba expects batch_size=1, got {batch_size}")
-
-    inputs = hidden_states.transpose(1, 2)
-    outputs = []
-    for start, end in zip(boundaries[:-1], boundaries[1:]):
-        convolved = torch.nn.functional.conv1d(
-            inputs[:, :, start:end],
-            weight,
-            bias,
-            groups=channels,
-            padding=weight.shape[2] - 1,
-        )
-        outputs.append(convolved[:, :, : end - start])
-
-    return torch.cat(outputs, dim=-1).transpose(1, 2)
+from fla.modules.conv import causal_conv1d as fla_causal_conv1d
+from fla.ops.utils import prepare_sequence_ids
 
 
 class _SeqToHeadParallel(torch.autograd.Function):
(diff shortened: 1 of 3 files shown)


S1ro1 (approved):
goat lgtm

PR #3196, opened 2026-08-04
feat: stream FSDP gradients through CPU offload

samsja:
## Summary

Adds gradient CPU offload on top of optimizer CPU offload and includes the transfer-stream correctness fix previously split into #3200.

- evicts finalized FSDP2 sharded gradients to pinned CPU accumulators during backward
- accumulates microbatches on an event-gated CPU worker
- computes clipping directly from completed CPU accumulators
- applies deferred scale and clipping factors while gradients use the optimizer state's H2D prefetch stream
- reuses optimizer H2D/D2H streams across steps so allocator pools do not grow per iteration
- skips the optional zero-gradient-ratio scan for offloaded gradients
- integrates the path into SFT and RL training
- includes the GLM-5.2 four-node gradient-offload debug configurations

## Motivation

Optimizer-state offload alone still leaves complete gradient shards on GPU. At large model sizes those shards exceed device capacity, so they must be evicted as FSDP units finish backward rather than after the entire backward pass.

The transfer streams must also be persistent. Creating fresh streams each step creates new CUDA caching-allocator pools, causing reserved memory to grow and eventually starving the default stream.

## Impact

Qwen3-4B SFT with random weights and fake data on two RTX PRO 6000 GPUs, sequence length 1024, FP32 reduction, compile, full activation checkpointing, optimizer CPU offload, and gradient accumulation 2:

| Variant | Step time | Throughput |
| --- | ---: | ---: |
| Optimizer CPU offload | 3.585 s | 1101.93 tok/s |
| Optimizer + gradient CPU offload | 3.909 s (+9.1%) | 1018.61 tok/s (-7.6%) |

With FP32 gradients and two-way sharding, Qwen3-4B allocates 7.49 GiB pinned accumulators plus 7.49 GiB staging per rank when accumulation is greater than one.

## Validation

- fake-model update and gradient-norm parity with accumulation 3
- BF16 and FP32 reduction paths
- FP32 router unit
- 2x1 HSDP mesh
- compiled layers without graph breaks
- Qwen3-4B SFT comparison
- GLM-5.2 four-node trainer smoke test: 10/10 steps
- GLM-5.2 full RL smoke test: 20/20 steps and flat peak memory
- Ruff, formatting, compileall, and config validation


src/prime_rl/trainer/rl/train.py
```diff
@@ -563,20 +567,30 @@ def load_run_checkpoint(_optimizer, idx: int) -> None:
...
         # across dp_cp so the final gradient is the true per-token mean over the global batch.
-        for param in model.parameters():
-            if param.grad is not None:
-                param.grad.mul_(parallel_dims.fsdp_gradient_divide_factor)
+        if config.model.grad_cpu_offload:
+            optimizer.scale_gradients_(parallel_dims.fsdp_gradient_divide_factor)
+        else:
+            for param in model.parameters():
+                if param.grad is not None:
+                    param.grad.mul_(parallel_dims.fsdp_gradient_divide_factor)
```
samsja: can we create a scale gradient function that take care of this for both case ?

src/prime_rl/trainer/sft/train.py
```diff
@@ -473,12 +480,19 @@ def run_validation(step: int) -> None:
...
             logger.debug(f"Clipping gradients with max norm {config.optim.max_norm}")
-            grad_norm = clip_grad_norm_(
-                model.parameters(), max_norm=config.optim.max_norm, ep_enabled=parallel_dims.ep_enabled
-            )
+            if config.model.grad_cpu_offload:
+                grad_norm = optimizer.clip_grad_norm_(config.optim.max_norm)
+            else:
+                grad_norm = clip_grad_norm_(
+                    model.parameters(), max_norm=config.optim.max_norm, ep_enabled=parallel_dims.ep_enabled
+                )
```
samsja: same can we have one function doing it for both and doing the if statment under ghhood ?

skills/training/start-run/SKILL.md
```diff
@@ -20,6 +20,9 @@ All entrypoints run via `uv run <command>` and accept TOML configs via `@ path/t
 - Validation aliases let renamed fields keep working; legacy keys can be remapped in a `model_validator(mode="before")`.
 - Auto-generated `--help` panels from `Field(description=...)` or PEP 224 docstrings.
 - Friendly errors: required-field boxes, validator errors point at the offending flag, unknown flags get a "did you mean" hint.
+- Experimental gradient offload requires both `model.optim_cpu_offload = true` and
+  `model.grad_cpu_offload = true`. Budget pinned CPU RAM for an accumulator plus
+  a same-sized staging set when gradient accumulation is greater than one.
```
samsja, line 25: remove this

PR #3197, opened 2026-08-04
fix: accept seq_lens contract kwargs in GlmMoeDsaModel.forward

samsja:
## Summary

GLM-5.2 (`glm_moe_dsa`) crashes on any run with `fused_lm_head_token_chunk_size` set:

```
TypeError: GlmMoeDsaModel.forward() got an unexpected keyword argument 'seq_lens'
```

The fused LM head wrapper (`_patch_model_forward` in `models/layers/lm_head.py`) replaces the causal-LM forward and passes all remaining kwargs — including the trainer's universal `seq_lens` / `seq_lens_are_pre_shard` contract kwargs — straight to the inner `self.model`. Every other custom model accepts these on its inner `Model.forward` (e.g. `Glm4MoeModel`, `Qwen3Model`), but `glm_moe_dsa` only accepted them on `GlmMoeDsaForCausalLM`.

This adds the same accept-and-ignore keyword-only args to `GlmMoeDsaModel.forward` — document boundaries derive from `position_ids` (sparse MLA builds its varlen indices from them), matching the existing rationale on the CausalLM wrapper.

Hit on a 4-node GLM-5.2 fake-data trainer run (65k seq len, cp8, fused LM head) while testing #3196; the failure is independent of that PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Signature and documentation only; behavior unchanged because the new kwargs are ignored.
> 
> **Overview**
> Fixes a **`TypeError`** when training GLM-5.2 with **`fused_lm_head_token_chunk_size`**: the fused LM head forwards the trainer’s packed-batch kwargs (`seq_lens`, `seq_lens_are_pre_shard`) into the inner backbone, but **`GlmMoeDsaModel.forward`** did not accept them.
> 
> This PR adds the same **keyword-only, accept-and-ignore** parameters to **`GlmMoeDsaModel.forward`** (with docstrings noting they are unused), aligning with other PrimeRL models. Document boundaries for this architecture still come from **`position_ids`** for sparse MLA varlen indexing, as on the CausalLM wrapper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97aff825c901f7cfa465e969675db452186f070d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3200, opened 2026-08-05
fix: reuse offload transfer streams + GLM-5.2 4-node debug configs

samsja:
## Summary

Stacked on #3196 (base: `fix/grad-cpu-offload-poc`). Two things from validating gradient CPU offload with GLM-5.2 on 4 H200 nodes:

### Fix: reuse optimizer offload transfer streams across steps

`CPUOffloadOptimizer.step()` created two fresh CUDA streams per call. Each new stream lands its H2D/D2H staging in a new per-stream caching-allocator pool, so reserved memory grew ~2.3 GiB per step (observed 118.9 → 134.3 GiB over 8 steps before the allocator flushed under pressure) and those cached blocks starve the default stream (post-step weight broadcast OOMed until #3199's `empty_cache()`). Streams are now created once in `__init__`.

### Debug configs (`examples/advanced/glm-5.2/debug/`)

- `trainer-4n-fake.toml` — trainer-only smoke test: fake data, random init, 65k seq, cp8/ep8, fp8, sign_sgd, optimizer+gradient CPU offload. Verified: 10/10 steps, peak 134.3 GiB GPU, ~1.44 TiB host RAM.
- `wordle-pd-4n.toml` — full RL smoke test: wordle-v1 through 1 disaggregated replica (2 prefill + 2 decode, NIXL), 4 trainer nodes, per-step NCCL fp8 weight broadcast, router replay. Verified: 20/20 steps, 0% rollout errors, peak flat at 129.9 GiB.

Both document the hard-won constraint inline: gradient accumulation must stay at 1 at this scale — accumulation > 1 lazily pins a second ~700 GiB/node staging set, exhausting 2 TiB hosts and stalling kernel launches until the NCCL watchdog kills the job (masquerading as a collective deadlock).

Full-loop validation also depends on #3197 (GLM-5.2 fused-LM-head seq_lens fix) and #3199 (broadcast memory fixes).


samsja:
Consolidated into #3196 so optimizer + gradient CPU offload, persistent transfer streams, and the GLM-5.2 validation configs live in one PR. The branch remains available for history.

PR #3202, opened 2026-08-05
feat!: vllm pass-through inference config

mikasenghaas:
## Summary

- Consolidates all vLLM pass-through fields into a single flat `VllmConfig` under `[vllm]` (standalone inference) / `[inference.vllm]` (RL), using vLLM's own argument names (`model`, `tensor_parallel_size`, `data_parallel_size`, `max_model_len`, ...).
- `VllmConfig` uses `extra="allow"`: **any** vLLM argument passes through to the server verbatim, typed by prime-rl or not — `[inference.vllm] max_num_seqs = 256` or `--vllm.max-num-seqs 256` just work. This replaces the `vllm_extra` escape hatch. CLI values for untyped keys are JSON-coerced, so dict-valued vLLM args work too: `--vllm.compilation-config '{"cudagraph_mode": "NONE"}'`.
- Removes `ModelConfig`, `ParallelConfig`, `vllm_extra`, and the `to_vllm()` rename table (`rgetattr`/`rsetattr` translation). `InferenceConfig.to_namespace()` now dumps the `server` + `vllm` sections straight into a vLLM `Namespace`.
- Parser auto-detection (`tool_call_parser`/`reasoning_parser` `"auto"` defaults), LoRA rank rounding, and the api-server-count auto-setup move onto `VllmConfig`; disaggregated/kv-offload auto-setup on `InferenceConfig` now mutates `self.vllm.*`.
- Non-vLLM orchestration knobs stay on `[inference]`: `server`, `router`, `backend_port`, `deployment`, `slurm`, `weight_broadcast`, `kv_cache_offload`, `env_vars`, `use_deep_gemm`, `enable_fp32_lm_head`, `enable_fp32_router_logits`.
- The SLURM per-rank launch helper passes structural args as `--vllm.*` flags and delivers per-rank overrides (`data_parallel_rank`, role extras) as a JSON file via `--vllm @ overrides.json` instead of `--vllm-extra`.
- Migrates all checked-in TOMLs, docs, skills, and README/example launch commands.

## Examples

Single node — `vllm serve` and the prime-rl entrypoint now read the same:

```bash
# native vLLM
vllm serve Qwen/Qwen3-0.6B \
  --data-parallel-size 2 --max-model-len 2048 --max-num-seqs 8

# prime-rl (same names under --vllm.*, plus the router + RL admin endpoints)
uv run inference --vllm.model Qwen/Qwen3-0.6B \
  --vllm.data-parallel-size 2 --vllm.max-model-len 2048 --vllm.max-num-seqs 8
```

or as a config file (`uv run inference @ infer.toml`):

```toml
[vllm]
model = "Qwen/Qwen3-0.6B"
data_parallel_size = 2
max_model_len = 2048
max_num_seqs = 8          # untyped — any vLLM arg passes through
```

Multi-node (wide-EP over 2 nodes, SLURM) — the per-node engine args vLLM would take via `vllm serve --tensor-parallel-size 2 --data-parallel-size 8 --enable-expert-parallel ...` live in the same flat section; deployment topology stays prime-rl-side:

```toml
[vllm]
model = "PrimeIntellect/INTELLECT-3"
tensor_parallel_size = 2
data_parallel_size = 8
enable_expert_parallel = true

[deployment]
type = "multi_node"
num_nodes = 2

[slurm]
partition = "cluster"
```

In an RL config the same section nests under the component: `[inference.vllm]`, CLI `--inference.vllm.tensor-parallel-size 8`.

## Breaking

Every inference-related config key moves; there are no aliases for old spellings (unknown keys fail validation with a did-you-mean hint).

| Before | After |
|---|---|
| `[inference.model] name = "..."` | `[inference.vllm] model = "..."` |
| `[inference.model]` other keys (`max_model_len`, `dtype`, `enforce_eager`, `trust_remote_code`, `chat_template`, `tool_call_parser`, `reasoning_parser`, `rope_scaling`) | same keys under `[inference.vllm]` |
| `[inference.parallel] tp` / `dp` | `[inference.vllm] tensor_parallel_size` / `data_parallel_size` |
| flat `[inference]` vLLM fields (`enable_lora`, `max_loras`, `max_cpu_loras`, `max_lora_rank`, `lora_target_modules`, `enable_prefix_caching`, `gpu_memory_utilization`, `quantization`, `api_server_count`, `data_parallel_size_local`, `data_parallel_rpc_port`, `seed`, `enable_expert_parallel`, `all2all_backend`, `enable_eplb`, `enable_dbo`, `enable_return_routed_experts`) | same keys under `[inference.vllm]` |
| `[inference.vllm_extra] key = ...` | key inline in `[inference.vllm]` |
| standalone inference: `[model]`, `[parallel]`, flat vLLM keys | `[vllm]` |
| CLI `--model.name`, `--parallel.tp`, `--gpu-memory-utilization`, `--vllm-extra '{...}'` (standalone) | `--vllm.model`, `--vllm.tensor-parallel-size`, `--vllm.gpu-memory-utilization`, `--vllm.<any-arg>` |
| shared `[model] name` propagation target `inference.model.name` | `inference.vllm.model` |
| `InferenceConfig.to_vllm()` | `InferenceConfig.to_namespace()` |

`[inference.model].vlm` is gone: nothing on the inference side consumed it (shared `[model] vlm` still propagates to trainer/orchestrator).

## Verification

On a 2-GPU box (RTX PRO 6000 Blackwell), via the router at `:8000` (engine at `:8100`):

- `uv run inference --vllm.model Qwen/Qwen3-0.6B --vllm.enforce-eager --vllm.max-model-len 2048 --vllm.gpu-memory-utilization 0.3 --vllm.max-num-seqs 8` — boots, `/v1/models` and `/v1/completions` respond; vLLM's `non-default args` dump shows the untyped `max_num_seqs: 8` (int-coerced from CLI), auto-resolved `tool_call_parser: hermes`, `logprobs_mode: processed_logprobs`, and the fp32 `additional_config` flags.
- Same via TOML with `data_parallel_size = 2` and untyped `max_num_seqs = 8` — both engines come up behind the router (`api_server_count` auto-raised to 2), completions route correctly. (The router's `/v1/models` proxy 500s at dp=2 — reproduced identically on `main` with the old config shape, so pre-existing vllm-router behavior, not from this change.)
- `uv run pytest tests/unit -m "not gpu"` — 502 passed on the current branch (config + parser suites updated to the new shape).


packages/prime-rl-configs/src/prime_rl/configs/inference.py
```diff
@@ -23,61 +24,192 @@ class ServerConfig(BaseConfig):
...
+            return data
+        for key in [k for k in data if k not in cls.model_fields]:
+            value = data.pop(key)
+            if isinstance(value, str):
+                try:
+                    value = json.loads(value)
+                except json.JSONDecodeError:
+                    pass
+            data[key.replace("-", "_")] = value
+        return data
```
samsja, line 161: vllm have as pydantic object similar to our config, wondering if we could consume it instead of replicating the key ?

mikasenghaas: they have pydantic?!?! wild

anyways, then we would need vllm as a dep in our configs package right?

samsja: hmm you made a good point, not ideal lets do like this

samsja (approved):

PR #3204, opened 2026-08-06
fix(slurm): pin UCX_NET_DEVICES only when the UCX build has IB support

samsja:
## Summary

The sbatch templates enumerate ACTIVE IB/RoCE ports and pin `UCX_NET_DEVICES` to them (plus a hard `mlx5_0:1` pin on decode ranks). When `third_party/ucx` is a TCP-only build — which is exactly what `scripts/install_nixl_from_source.sh` produces on hosts without the rdma-core dev headers — that pin leaves UCX with **no usable transport at all**: NIXL backend creation fails on every rank (`nixl_agent.cpp createBackend ... NIXL_ERR_BACKEND`, preceded by `UCX WARN network devices 'mlx5_0:1',... are not available`), engines never come up, and the trainer hangs at weight-broadcast init until `dist_timeout` kills the job.

Guard both pin sites on `third_party/ucx/lib/ucx/libuct_ib.so` existing, falling back to `UCX_NET_DEVICES=all` (TCP) otherwise. IB-capable builds keep the exact current behavior.

Hit while bringing up a GLM-5.2 disaggregated (NIXL) run on a cluster with ACTIVE IB ports but a TCP-only UCX build; with this guard the run proceeds over TCP, and with an IB-enabled build the pinning works as before.

### DIFF
--- a/src/prime_rl/templates/inference.sbatch.j2
+++ b/src/prime_rl/templates/inference.sbatch.j2
@@ -211,8 +211,15 @@ srun --kill-on-bad-exit=1 bash -c '
             echo "$(basename "$dev"):$(basename "$port")"
         done | paste -sd,
     }
-    UCX_NET_DEVICES=$(rdma_ports InfiniBand)
-    [ -n "$UCX_NET_DEVICES" ] || UCX_NET_DEVICES=$(rdma_ports Ethernet)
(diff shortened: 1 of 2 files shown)


samsja:
Not needed for our setup (root cause was pointing third_party/ucx at a TCP-only build instead of the IB-capable ucx-rdma build); closing.

PR #3210, opened 2026-08-06
feat: offload FP32 master weights behind persistent BF16 model

samsja:
## Summary

Adds an experimental middle ground between gradient offload and full FSDP CPU offload:

- keep persistent BF16 compute weights on GPU
- keep FP32 master weights and fused AdamW state on CPU
- offload BF16 model gradients directly into FP32 pinned buffers
- copy refreshed BF16 weights to GPU once per optimizer step

The mode requires optimizer-state and gradient CPU offload, supports AdamW only, and rejects resumable optimizer checkpoints. SFT weights-only checkpoints remain supported.

This PR is stacked on #3200, which is stacked on #3196.

## Qwen3-4B result

One RTX PRO 6000, `Qwen/Qwen3-4B-Instruct-2507`, I3-SFT-10K `math`, sequence length 2048, batch size 2, micro-batch size 1, three steps:

| Metric | Optimizer + gradient offload | + CPU master-weight offload | Difference |
| --- | ---: | ---: | ---: |
| Mean step time, steps 2–3 | 5.769 s | 5.629 s | -2.4% |
| Mean forward/backward | 1.708 s | 1.566 s | -8.3% |
| Mean throughput | 691.5 tok/s | 723.4 tok/s | +4.6% |
| Peak GPU reserved memory | 35.379 GiB | 20.088 GiB | -15.291 GiB (-43.2%) |

The first implementation used default CPU AdamW and converted accumulated BF16 gradients after backward, resulting in a 22.3s step. Two changes remove that bottleneck: gradients now land directly in FP32 pinned buffers, and CPU AdamW uses PyTorch's fused implementation.

## Transfer and phase measurements

Pinned-memory microbenchmarks on this host measured 50–53 GiB/s across PCIe:

- 14.98 GiB FP32 master → BF16 GPU refresh: approximately 0.30s
- two accumulated BF16 GPU → FP32 CPU gradient copies: approximately 0.60s total

The real steady optimizer phases measured approximately 0.44s for gradient preparation, 2.18–2.91s for fused CPU AdamW, and 0.30s for BF16 refresh. This tracks the observed 5.63s total step after adding 1.57s forward/backward and clipping/other overhead.

The one-time FP32-to-BF16 conversion explicitly releases the old CUDA allocator cache. Without that, `max_memory_reserved()` counts stale blocks from the original 15 GiB FP32 model and badly overstates the treatment peak.

## Numerical check

The saved checkpoints contain 398 BF16 tensors and 4,022,468,096 elements. All tensors match at `rtol=1e-5, atol=1e-5`; mean absolute difference is `4.57e-9` and maximum absolute difference is `7.63e-6`.

## Validation

- `ruff check` and `ruff format --check` on changed Python files
- two consecutive optimizer steps on a small FSDP2 model, checking that masters and every AdamW state tensor remain on CPU, gradient buffers are FP32, and the BF16 shadow is refreshed
- config validation for the required offload combination and weights-only checkpoint restriction
- pinned H2D/D2H and CPU AdamW microbenchmarks
- matched three-step Qwen3-4B SFT benchmark and full saved-checkpoint comparison

`tests/unit/test_configs.py` was also attempted in the isolated worktree; it stops on the existing missing `alphabet-sort-v1` taskset dependency while loading `configs/debug/algo/echo.toml`.

### DIFF
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -159,7 +159,7 @@ grad_cpu_offload = true
 master_weight_cpu_offload = true
 ```
 
-This stores FP32 master weights in pinned CPU memory and refreshes the persistent BF16 GPU weights after each optimizer step. It currently supports AdamW and weights-only checkpoints; resumable optimizer checkpoints are rejected.
+This stores FP32 master weights and gradients in pinned CPU memory, updates them with fused CPU AdamW, and refreshes the persistent BF16 GPU weights after each optimizer step. Gradient accumulation requires two FP32 gradient-sized pinned buffers. It currently supports AdamW and weights-only checkpoints; resumable optimizer checkpoints are rejected.
 
 ### LM Head Chunking
 
--- a/docs/scaling.md
+++ b/docs/scaling.md
@@ -87,6 +87,8 @@ FSDP2 is the default model sharding strategy. By default the trainer fully shard
 | `trainer.model.reshard_after_forward` | If `true` (default), parameters are resharded after the forward pass to free memory; the backward pass re-gathers. Set `false` to keep params resident — faster but more memory. |
 | `trainer.model.fsdp_cpu_offload` | Offload params + grads + optimizer state to CPU. Big memory win, large throughput hit. |
 | `trainer.model.optim_cpu_offload` | Offload only optimizer state. Mid-ground — small throughput cost, decent memory savings, especially at low GPU count. |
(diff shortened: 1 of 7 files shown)


samsja:
Consolidated into #3212, which now contains the FP32 master-weight/fused AdamW implementation plus the public FSDP post-accumulate-hook refactor. The branch remains available for history.

PR #3105, opened 2026-07-22
chore: single global router in front of engines for all deployments

mikasenghaas:
## Summary

Unifies inference deployment so that every shape — single-node, multi-node, and P/D-disaggregated — runs a **single global router** as the client-facing endpoint in front of the vLLM engines (first step towards a single-URL client, see [RES-1156](https://linear.app/primeintellect/issue/RES-1156/single-node-router-deployments), [RES-1028](https://linear.app/primeintellect/issue/RES-1028/revert-base-url-to-a-single-string-drop-multi-client)):

- **Single node (local + SLURM)**: `inference_local` now spawns a `vllm-router` on `server.port` (default `8000`) fronting the engine on `backend_port` (default `8100`). The router's lifecycle is tied to the inference process (torn down on exit; a router crash tears down the engine via a watchdog thread). The RL entrypoint auto-wires `orchestrator.model.client.admin_base_url` to the engine port so admin ops (pause/update_weights/resume) bypass the router, matching multi-node semantics.
- **Multi-node + P/D (SLURM)**: the per-replica routers are collapsed into one global router on inference node 0 fronting all per-rank endpoints across replicas (both `vllm-router` and `llm-d`). `INFER_URLS` is now a single URL; `ADMIN_URLS` still enumerate every rank directly. llm-d endpoint discovery now enumerates all replicas (prefill nodes across replicas first, then decode nodes).
- **Config**: `router` and `backend_port` move from the deployment configs to `InferenceConfig` top level — they describe the same client-facing layout in every deployment. The router listens on `server.port` (the per-router `port` field is gone). `router = "None"` runs a bare engine; the SLURM launchers write that into per-rank engine configs.
- **Readiness**: `StaticInferencePool.wait_for_ready` now health-checks the client-facing (router) endpoint in addition to the admin (engine) endpoints — the router only starts serving once its workers pass its own health check, so checking engines alone let rollouts race ahead of the router and fail with connection errors.

## Breaking

- `[inference.deployment.router]` → `[inference.router]` (standalone entrypoint: `[deployment.router]` → `[router]`).
- `inference.deployment.backend_port` → `inference.backend_port`.
- `inference.deployment.router.port` is removed — the router always binds `inference.server.port`.
- Multi-node / P/D runs now expose **one** router URL instead of one per inference replica; the orchestrator receives a single `--model.client.base-url`.

## Verification

On a 2-GPU node:

[...]

PR #3211, opened 2026-08-06
Revert "chore: single global router in front of engines for all deployments (#3105)"

mikasenghaas:
Reverts #3105 (single global router in front of engines for all deployments).

## Why

Under full RL SWE load the single global router stalls, and the run makes no progress. Observed twice, back-to-back, on `chore/bump-vf-envs` (which includes current main) with GLM-4.5-Air, 768 inflight rollouts + 500 concurrent eval rollouts on 4 replicas:

| run | solver `ProviderError: Request timed out.` | trainer steps | duration |
|---|---|---|---|
| 1st | 8,283 | 0 | 1h13m |
| 2nd (clean reproduction) | 1,573 and climbing at kill | 0 | ~45m |

Signature, from the logs:

- Timeouts are the client's **600s read timeout** on `/inference/v1/generate`. Traced one end-to-end: the router mapped the session to a worker at `05:31:20`; the client gave up at `05:41:40`. Never on a rollout's first call (median: 3rd call) — short early calls return, long later ones die.
- The **engines were healthy and unsaturated throughout**: max queue depth 9, KV cache ~1%, zero preemptions, throughput flowing. The stall is between client and engine.
- The router logged hard errors sending to its workers (`Failed to send typed request worker_url=http://...:8100 route=/v1/chat/completions`), and starts with `request queue with size: 100, timeout: 60s` — all streaming traffic for ~1,270 concurrent rollouts funnels through this one process.
- Same config, same load, same models ran cleanly for 400+ steps on the pre-#3105 topology (per-replica routers, client pool balancing across them).

## Conflicts resolved

- `docs/training.md`: the log-tree section drifted after #3105 (env-server log layout changed separately) — restored the per-replica `router_*.log` line, kept the newer env log line.
- `examples/advanced/glm-5.2/{swe-llmd,infer/pd-llmd}.toml`: these gained `[inference.router.*]` scorer subsections after #3105; moved to `[inference.deployment.router.*]` to match the restored schema.

## Checks

`tests/unit/test_configs.py`, `tests/unit/utils/test_client.py`, `tests/unit/utils/test_elastic.py`: 140 passed.

A fix-forward that keeps the single-router topology but survives this load is presumably possible (queue sizing, per-worker connection pools, streaming backpressure) — but until then every multi-replica RL run on main is broken at scale, so reverting first.

---

> [!NOTE]
> **High Risk**
> Changes the production inference routing path for multi-replica RL and disaggregated SLURM runs; misconfigured URLs or client DP headers could break rollouts, though this restores a topology that previously survived large-scale SWE loads.
> 
> **Overview**
> Reverts the **single global router** topology (#3105) that funneled all multi-replica RL traffic through one `vllm-router` process and stalled under high concurrent load.
> 
> **Routing topology:** SLURM multi-node and P/D jobs again start a **router per inference replica** (logs under `inference/router_*.log`), with `launch_router` scoped to each replica’s nodes for llm-d endpoint discovery. Client-facing URLs become **one router URL per replica** (orchestrator can balance across them) instead of a single global URL on inference node 0.
> 
> **Config schema:** `router` and `backend_port` move from top-level `[inference]` into **`[inference.deployment]`** (and `[deployment.router]` for standalone inference). Router backends gain an explicit **`port`** field for the client-facing listener.
> 
> **Local inference:** `uv run inference` no longer spawns `vllm-router` in-process; it binds vLLM directly on **`inference.server.port`**.
> 
> **Orchestrator client wiring:** Adds **`dp_rank_count`** on `ClientConfig`—single-node RL auto-expands logical clients with **`X-data-parallel-rank`** for KV stickiness; multi-node SLURM sets **`dp_rank_count = 1`** because the router handles DP balancing. Removes auto **`admin_base_url`** bypass and separate health checks on the router URL.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d2fa3a0467c4fe2b6131e52d086a7229469c8d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/inference.md
+++ b/docs/inference.md
@@ -30,9 +30,7 @@ We support 3 distinct deployment shapes:
 
 Most of the features are supported for all deployment shapes, with few exceptions. These exceptions are rejected on validation.
 
-Every deployment shape has the same client-facing layout: a single global router listens on `inference.server.port` and fronts all vLLM engines, which listen on `inference.backend_port` (+ rank offset). Clients always talk to one URL, regardless of how many engines run behind it.
-
-You can select the deployment shape with `InferenceDeploymentConfig` in your config file. This is a config-field that allows you to set the deployment shape and topology knobs such as `num_nodes` and `num_replicas`.
+You can select the deployment shape with `InferenceDeploymentConfig` in your config file. This is a config-field that allows you to set the deployment shape, deployment-specific knobs such as `num_nodes`, `num_replicas`, `backend_port`, and a `[...deployment.router]` block, etc.
 
 ```toml
 [inference.deployment]
@@ -57,8 +55,6 @@ The single-node deployment is the default deployment shape. It runs the inferenc
 [inference.deployment]
 type = "single_node"
 ```
-
-The launcher starts a `vllm-router` on `inference.server.port` (default `8000`) fronting the vLLM engine on `inference.backend_port` (default `8100`). Clients connect to the router URL; admin operations (weight updates, health checks) bypass the router and hit the engine port directly — the RL entrypoint wires `orchestrator.model.client.admin_base_url` accordingly.
 
 This deployment shape runs the inference server on a single node, if configured with NVLink enabled, it allows you more freedom in terms of parallelism configurations.
 
@@ -106,7 +102,7 @@ tp = 2
 dp = 4
 ```
 
-This configuration will run 2 independent vLLM replicas, each with `tp=2` and `dp=4`. Routing is handled by a single global router running on the first inference node, fronting the per-rank endpoints of all replicas — either `vllm-router` (default) or the upstream `llm-d` EPP+Envoy, selected via the `[inference.router]` block. You can read more about the supported routing options in the [router](#router) section.
+This configuration will run 2 independent vLLM replicas, each with `tp=2` and `dp=4`. Routing is handled by a router instance running on the same node as the 1st replica — either `vllm-router` (default) or the upstream `llm-d` EPP+Envoy, selected via the `[...deployment.router]` block. You can read more about the supported routing options in the [router](#router) section.
 
 ### Wide-EP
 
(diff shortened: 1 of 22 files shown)


mikasenghaas:
## Update after running the revert at full load

Ran the same 5-step judge config from this branch (job on 2 train + 4 infer replica nodes, 768 inflight + 500 eval rollouts). Result refines the diagnosis:

- **The revert works as a blast-radius fix**: 4 solver timeouts in the first 50 minutes, vs 1,573+ at the same age on the single-router topology. Engines busy, rollouts completing.
- **But at ~50 min, ONE of the four per-replica routers wedged** — `router_3`'s log output stops mid-stream at 19:50:09 (after ~70–98k lines/10min), its engine drains to 0 running and stays healthy/idle, and every session consistent-hashed to it starts dying at the 600s client timeout (~230 in the next ten minutes, ~¼ of traffic). No error is logged by the router; it just stops.

So the root cause is **the router process wedging under sustained load**, not the topology per se. #3105 made the wedge fatal (one router = 100% of traffic, collapse at ~25 min); per-replica routers contain it to ¼ (onset ~50 min).

**Why now, on the same `vllm-router` 0.1.26 that ran clean for 4 days:** the ancestry discriminates exactly one change. The 4-day clean run's verifiers pin (base Jul 29) predates verifiers#2218 ("build one client per rollout", Aug 3); every run that wedged a router carries it. #2218 makes each rollout build **and close** its own httpx client, so the router sees constant TCP connection churn from ~768 concurrent rollouts instead of a shared keepalive pool — a classic way to wedge a hyper/axum accept loop or connection pool.

Recommendation: land this revert for blast-radius (single wedge should not kill the whole run), and separately either give verifiers' per-rollout clients a shared transport/connection pool, or harden the router against connection churn. Can provide logs for both wedge instances.

PR #3212, opened 2026-08-06
feat: fuse CPU offload with public FSDP hooks

samsja:
## Summary

Adds the final fused CPU-offload design on top of #3196:

- expose one user-facing `OptimizerOffloadingConfig` and pass that same validated object to the optimizer implementation
- provide four modes: no offload, optimizer-state-only offload, gradient offload with a GPU step, and full optimizer offload
- make `{ gradients = true }` accumulate finalized gradients in pinned CPU buffers and prefetch them alongside optimizer-state chunks for the GPU step
- make `{ full = true }` offload finalized gradients into FP32 pinned buffers and run the optimizer on CPU-resident FP32 master weights
- keep persistent BF16 compute weights on GPU in full mode and refresh them once per optimizer step
- always use pinned CPU storage and persistent CUDA transfer streams
- replace lag-1 module hooks and private FSDP state/event access with public sharded-parameter post-accumulate hooks
- replace transformer-layer name parsing with model-independent, size-bounded optimizer chunks
- checkpoint CPU masters and optimizer state as CPU-backed DTensors under the original model FQNs, preserving FSDP sharding metadata and resumability
- explicitly remove hook handles and stop the gradient-copy worker after training

The user-facing modes are:

```toml
optim_cpu_offload = false           # no offload
optim_cpu_offload = true            # optimizer state only
optim_cpu_offload = { gradients = true } # gradients and optimizer state; GPU step
optim_cpu_offload = { full = true } # gradients, FP32 masters, state, and optimizer step
```

AdamW, SGD, and SignSGD support full offload. Muon supports optimizer-state and gradient offload but rejects full offload because its optimizer step would run on CPU.

## Why

The GPU-step optimizer-state path transfers Adam state to GPU and back for every update. That is roughly 16 bytes per parameter per step across PCIe: 8 bytes H2D for the two FP32 moments and 8 bytes D2H to evict the updated moments.

Gradient offload additionally transfers each finalized FP32 gradient shard to CPU during every backward and returns the final accumulated gradient for the GPU step. With Adam state offload, this is approximately `(4A + 20)` bytes per parameter per optimizer step at gradient accumulation `A`, although gradient and state transfers are chunked and overlapped.

Full offload avoids returning gradients and Adam state to GPU. It transfers the FP32 gradient shards during every backward and refreshes BF16 compute weights after the CPU optimizer step, approximately `(4A + 2)` bytes per parameter per step. It saves substantially more GPU memory, but the CPU AdamW step can dominate runtime.

FSDP invokes parameter post-accumulate hooks after assigning each finalized sharded gradient. This provides the correct tensor and stream boundary directly, removing the private `_get_fsdp_state()` / `_post_reduce_event` dependency and the PyTorch-version guard from the earlier implementation.

## Modes and measured impact

One RTX PRO 6000, Qwen3-4B, I3-math, sequence length 2048, micro batch size 1, current PR revision, mean of steady steps 3-6:

| Setup | Config | Grad. accum. | Step time | Forward/backward | Peak GPU |
| --- | --- | ---: | ---: | ---: | ---: |
| Normal | `optim_cpu_offload = false` | 2 | 1.18 s | 0.80 s | 65.07 GiB |
| Optimizer offload | `optim_cpu_offload = true` | 2 | 1.92 s | 0.80 s | 43.48 GiB |
| Gradient offload | `optim_cpu_offload = { gradients = true }` | 2 | 2.89 s | 1.49 s | 37.29 GiB |
| Full offload | `optim_cpu_offload = { full = true }` | 2 | 4.30 s | 1.50 s | 20.09 GiB |

### Gradient accumulation scaling

| Grad. accum. | Optimizer offload | Gradient offload | Full offload |
| ---: | ---: | ---: | ---: |
| 1 | 1.45 s | 1.90 s | 3.91 s |
| 2 | 1.92 s | 2.89 s | 4.30 s |
| 4 | 2.89 s | 4.49 s | 7.52 s |
| 8 | 4.73 s | 7.67 s | 10.08 s |

The optimizer-state-only path adds a nearly constant 1.12 s outside forward/backward because its Adam-state transfer happens once per optimizer step. Gradient offload has the same per-microbatch D2H scaling as full offload, but its chunked GPU step costs about 1.40 s outside forward/backward instead of the slower CPU update. Full offload remains the maximum-memory-saving mode.

## Validation

- `ruff check` and `ruff format --check` on changed Python files
- `pytest -q tests/unit/test_configs.py -k 'not test_load_configs'`: 42 passed
- six-step Qwen3-4B/I3-math SFT benchmark for optimizer-only and full offload at gradient accumulation 1, 2, 4, and 8
- six-step Qwen3-4B/I3-math SFT benchmark for gradient offload at gradient accumulation 1, 2, 4, and 8
- two-step Qwen3-4B/I3-math Muon smoke run with gradient offload
- gradient H2D allocations are recorded on the compute stream before release so the CUDA allocator cannot recycle them while GPU optimizer kernels are still reading them
- three-step Qwen3-4B/I3-math SFT run with `{ full = true }`: completed successfully at 20.1 GiB peak GPU memory
- two-step Qwen3-4B/I3-math SignSGD run with `{ full = true }`: completed successfully at 20.1 GiB peak GPU memory
- full Qwen3-4B DCP checkpoint after step 1, fresh-process load, resumed step 2, and a second full checkpoint all completed successfully
- checkpoint metadata contained FP32 masters and both Adam moments for all 398 trainable FSDP parameters under their original FQNs
- resumed step-2 loss/gradient norm was `0.6222` / `11.3333`; uninterrupted control was `0.6231` / `11.3486` under the small run-to-run nondeterminism of the compiled FlashAttention setup
- full step-3 checkpoint comparison across 4.02B parameters from the earlier offload parity run: maximum absolute difference `7.63e-6`, zero elements above `1e-5` versus no offload

The complete config-file parameterization was also run: 62 cases passed and 48 optional-environment cases could not import taskset packages such as `reverse_text_v1`, `alphabet_sort_v1`, and `math_env_v1` in this environment.

## Deadlock + robustness fixes (GLM-5.2 4-node validation, 2026-08-07)

Validating this PR on GLM-5.2 (78L, 65k-131k, cp8/ep8, 4 H200 nodes) surfaced a distributed deadlock with gradient accumulation ≥ 2: the post-accumulate offload hook runs inside FSDP `foreach_reduce` on the autograd thread, and its lazy `cudaHostAlloc` (staging buffer, first needed at micro-batch 1) wedges against the driver while peer ranks' NCCL kernels spin — all GPUs at 100% SM / 0% memory traffic, no watchdog on the spinning rank (faulthandler-confirmed stack). Deterministic at 78 layers; 30/60-layer models win the race.

- `fix: preallocate pinned gradient buffers before the first collective` — pinning is only safe before any NCCL kernel exists, so accumulator + staging are allocated at manager init (staging is now always resident; a config gate could refine that for accum-1 workloads)
- `perf: slab-allocate pinned gradient offload buffers` — one pinned slab per dtype instead of ~800 per-param allocations (measured 52 s for 85 GiB/rank on 1 node)
- `fix: propagate dist timeout to device-mesh sub process groups` — sub-PGs silently used torch's 10-min default instead of `dist_timeout_seconds`, so init-time pinning skew across nodes tripped peer watchdogs
- plus cherry-picks this branch needed to run GLM-5.2 at all: the `seq_lens` forward fix (already in main) and the 4-node weight-broadcast memory fix (#3199)

Validated: 1-node repro (12L, accum 2, offload) 6/6 steps; repro config for the deadlock lives at `research-prod/rlm5/repro-grad-offload-deadlock/`. Full-scale 78L/65k/accum-2 verdict run in progress (repeatedly interrupted by cluster node failures, not by the code).


packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -54,6 +54,25 @@ class ActivationOffloadingConfig(BaseConfig):
...
+    full: bool = False
+    """Offload gradients and run the optimizer step on CPU-resident FP32 master weights."""
+
+
+def _normalize_optimizer_offloading(value: Any) -> Any:
+    if value is True:
+        return {}
+    if value is False:
+        return None
+    return value
```
samsja, line 70: tbh not sure we want this why keeping ti ?

skills/training/start-run/SKILL.md
```diff
@@ -7,6 +7,13 @@ description: How to launch prime-rl training runs — the `rl`, `sft`, and `infe
 
 All entrypoints run via `uv run <command>` and accept TOML configs via `@ path/to.toml` plus CLI overrides.
 
+When running from a fresh git worktree, initialize its pinned dependencies first with
+`git submodule update --init --recursive`. If reusing a synced environment from another
+checkout, set `UV_PROJECT_ENVIRONMENT` to that environment and prepend the worktree's
+`src`, `packages/prime-rl-configs/src`, `deps/renderers`, `deps/verifiers`, and
+`deps/pydantic-config/src` directories to `PYTHONPATH`. Verify representative imports
+with `inspect.getfile` before launching so the run cannot silently use another checkout.
+
```
samsja, line 16: lest remove

skills/training/start-run/SKILL.md
```diff
@@ -20,9 +27,16 @@ All entrypoints run via `uv run <command>` and accept TOML configs via `@ path/t
...
+- Gradient offload with a GPU optimizer step uses
+  `model.optim_cpu_offload = { gradients = true }`. It accumulates gradients in
+  pinned CPU RAM and streams the final gradient chunks back alongside optimizer
+  state. With gradient accumulation, budget two FP32 gradient-sized buffers.
+- Full optimizer offload uses `model.optim_cpu_offload = { full = true }`.
+  It keeps a persistent BF16 compute model on GPU, stores FP32 master weights and
+  gradients in pinned CPU RAM, runs the optimizer on CPU, and refreshes the BF16
+  weights once per optimizer step. Muon is not supported. Resumable checkpoints
+  include the FP32 masters and optimizer state under the original FSDP parameter
+  names.
```
samsja, line 39: hmm lerts remoe tjhis

src/prime_rl/trainer/sft/train.py
```diff
@@ -423,6 +428,7 @@ def run_validation(step: int) -> None:
 
             with maybe_record_function("backward"):
                 scaled_loss.backward()
+                finish_backward(gradient_manager, wait_for_copies=micro_step == grad_accum_steps - 1)
```
samsja, line 431: remove finish abckwrtd as a finction and make the if stastenmt explicit please

samsja:
Folded into #3234: the optimizer-in-backward work built on this branch ended up reworking and partially removing what this PR introduced (state-only offload mode is gone entirely, offload is now full-offload-only and disabled by default). #3234 now targets main and carries the full combined diff — reviewing it alone gives the complete, current picture.

PR #3215, opened 2026-08-07
chore(nixl): use canonical VRAM mem_type

yafshar:
## Change

`prep_xfer_dlist(mem_type=...)` passed `"cuda"`, which nixl marks deprecated in
favor of `"VRAM"` (`src/api/python/_api.py`):

    "VRAM": nixlBind.VRAM_SEG,
    "cuda": nixlBind.VRAM_SEG,  # deprecated

Both resolve to `VRAM_SEG`, so this is a no-op at runtime. `"VRAM"` is also the
spelling nixl's own `_tensor_mem_type()` helper emits.

## Note

The accelerator-derived `UCX_TLS` defaults this PR originally carried are dropped
per review. Since `set_ucx_env_defaults()` uses `setdefault`, a non-CUDA
accelerator supplies its own copy transport from the launcher environment.


yafshar:
@S1ro1 @mikasenghaas — mind a look? Small one: `set_ucx_env_defaults()` hardcodes `cuda_copy`, so NIXL registration fails on XPU. Now derived from `torch.accelerator.current_accelerator()`; CUDA `UCX_TLS` is byte-identical to before. Rebased onto the `prime_rl.transports` move (#3308).

Two out-of-scope items flagged in the description I could use a call on: whether inference should use the derived string instead of `UCX_TLS=all`, and that `install_nixl_from_source.sh` builds UCX with `--with-ze=no`.

S1ro1 (changes requested):
we're trying to keep prime-rl very opinionated - defaults are for our research team to use and benefit from. As it's all `.set_default` you can override this yourself from the slurm launcher or via exported env vars. We're fine accepting "cuda" -> "vram" rename but not the added complexity of accelerator based defaults

yafshar:
> we're trying to keep prime-rl very opinionated - defaults are for our research team to use and benefit from. As it's all `.set_default` you can override this yourself from the slurm launcher or via exported env vars. We're fine accepting "cuda" -> "vram" rename but not the added complexity of accelerator based defaults

@S1ro1 Makes sense — trimmed to just the mem_type rename, accelerator map dropped.

Worth recording for anyone who does need that override: it works on the trainer path but not inference. `entrypoints/inference.py` calls `os.environ.update()` with `DEFAULT_INFERENCE_ENV_VARS` merged over the ambient
env ("existing os.environ loses", per the comment there), so an exported `UCX_TLS` gets replaced by `UCX_TLS=all`. [`inference.env_vars`] is the knob that wins. 

Not a problem in practice, since `all` already covers every copy transport — just means the export advice is trainer-only.

yafshar:
@S1ro1 can we merge this fix?

S1ro1 (approved):

PR #3232, opened 2026-08-10
feat(monitor): upload native v1 episodes without orchestration refactor

mrmoxon:
## Goal

Preserve complete native v1 Episodes in Prime-RL sample uploads without making Episode the orchestration unit.

## What changes

- retain the original v1 Episode envelope as excluded, in-memory metadata when `Env.run` converts its traces to existing `Rollout` objects
- deduplicate that shared envelope in `PrimeMonitor` and serialize it through verifiers' existing `build_samples()` compatibility path
- preserve fixed-agent, failed, and other sibling traces in `info.native_wrapper`, while the existing trainable trace remains the flat summary for older Platform consumers
- keep legacy/group rollout paths working through a one-trace Episode fallback
- treat serialization failures as best-effort monitor failures rather than crashing training

## What does not change

- Rollout remains the unit used by dispatch, scoring, training, metrics, and storage
- no Episode-first orchestration foundation is included
- no dependency on the earlier stacked Episode PRs
- the existing Platform upload endpoint and Parquet schema remain unchanged

## Validation

- focused PrimeMonitor tests: 6 passed
- Ruff check and format check passed
- `git diff --check` passed

---

> [!NOTE]
> **Medium Risk**
> Changes telemetry serialization and Platform payload shape (`info`, episode grouping) but keeps training on rollouts; failures are best-effort and should not block training.
> 
> **Overview**
> **Prime Monitor** now uploads **native v1 Episodes** to Platform instead of flattening each rollout with `trace_to_sample`, so multi-trace episodes (judge/solver/critic, failed siblings, etc.) stay intact on the wire while training still runs on rollouts.
> 
> Each v1 `Env.run` attaches an in-memory **`native_episode`** on every `Rollout` (full `WireEpisode` with orchestrator `Rollout` traces). `Rollout` gains an excluded **`native_episode`** field for monitors only.
> 
> `log_samples` rebuilds episodes via **`_rollouts_to_episodes`** (dedupe shared envelopes; legacy paths synthesize one-trace episodes), then **`build_samples`** projects rows and layers run/step/advantage/`env_name` as before. **`info`** comes from verifiers (e.g. `native_wrapper`, `native_trace_index`). Parquet build failures **warn and skip upload** instead of crashing training.
> 
> Unit tests cover multi-trace retention, legacy fallback, and serialization failure handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 208ba210fe58db6e8a6c087ada938741519cd813. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/types.py
+++ b/src/prime_rl/orchestrator/types.py
@@ -94,6 +94,9 @@ class Rollout(vf.Trace[DataT], Generic[DataT]):
     # Links the traces of one episode; stamped into ``info`` on arrival so
     # saved records keep their grouping.
     episode_id: str = Field(default="", exclude=True)
+    # The original v1 envelope, retained only in memory so monitors can upload the
+    # complete multi-trace Episode without making Episode the orchestrator's unit.
+    native_episode: vf.WireEpisode | None = Field(default=None, exclude=True, repr=False)
     policy_version: int = Field(default=0, exclude=True)
     off_policy_steps: int = Field(default=0, exclude=True)
     samples: list[TrainingSample] = Field(default_factory=list, exclude=True)
--- a/src/prime_rl/orchestrator/envs.py
+++ b/src/prime_rl/orchestrator/envs.py
@@ -130,8 +130,10 @@ class Env:
             detail = f"{error.type}: {error.message}" if error is not None else "no traces and no error recorded"
             raise RuntimeError(f"episode failed before any trace was produced — {detail}")
         rollouts = [ROLLOUT_TYPE.model_construct(**dict(wire)) for wire in episode.traces]
+        native_episode = episode.model_copy(update={"traces": rollouts})
         for rollout in rollouts:
             rollout.episode_id = episode.id
+            rollout.native_episode = native_episode
             if not episode.ok and rollout.ok:
                 error = episode.last_error or vf.Error(
                     type="EpisodeFailed", message="A sibling trace in this episode failed"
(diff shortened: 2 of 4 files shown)


src/prime_rl/orchestrator/types.py
```diff
@@ -94,6 +94,9 @@ class Rollout(vf.Trace[DataT], Generic[DataT]):
     # Links the traces of one episode; stamped into ``info`` on arrival so
     # saved records keep their grouping.
     episode_id: str = Field(default="", exclude=True)
+    # The original v1 envelope, retained only in memory so monitors can upload the
+    # complete multi-trace Episode without making Episode the orchestrator's unit.
+    native_episode: vf.WireEpisode | None = Field(default=None, exclude=True, repr=False)
```
mikasenghaas, line 99: wait each rollout stores its back-stores its episode? this seems like a lot of redundancy and could lead to perf issues

imo this is a braoder smell of the episode-type not being fully native on prl orch yet. this is a known limitation @hallerite and me want to tackle this week. wondering if we should delay this pr until then

mikasenghaas: e..g do we even dedup the episodes if one episode makes multiple rollouts?

mikasenghaas: if we want to get in today, then we should try to pass a flat list of episodes directly to the prime monitor

hallerite: I think we should. the PR will be much cleaner after the refactor

mrmoxon: will leave it for now then

PR #3255, opened 2026-08-12
feat: default vllm-router balancing to cache_aware

mikasenghaas:
## Summary

- Default the vllm-router balancing policy to `cache_aware` (was `consistent_hash`), for both single-node (local `start_router`) and multi-node/disaggregated (SLURM `launch_router`) deployments.
- Expose the `cache_aware` tuning knobs on `VllmRouterConfig` — `cache_threshold`, `balance_abs_threshold`, `balance_rel_threshold` — and plumb them through both launch paths. Unset fields use the router binary's defaults.
- Type `policy` as a `Literal` of the policies the router binary supports today (`cache_aware`, `consistent_hash`, `round_robin`, `power_of_two`, `random`), so typos fail at config validation.
- Update `docs/inference.md` routing-policies section accordingly.

Motivation: post-mortem of two 4-node GLM Air runs showed `consistent_hash` locking a saturated engine into a KV thrash spiral — preemption → re-prefill churn → prefix-cache eviction (98% → 5% hit rate) → ~4× slower generation — while its hash-pinned sessions kept it saturated and the other three engines drained idle. `consistent_hash` never looks at load, so no work migrates.

`cache_aware` keeps the property we want from stickiness (a rollout's turns follow their own growing token prefix, so KV re-use survives) but monitors per-worker in-flight counts and switches to shortest-queue routing while load is imbalanced (`max − min > balance_abs_threshold` and `max > balance_rel_threshold × min`), so a saturated engine sheds new work instead of hoarding it. `consistent_hash` remains available via `inference.router.policy`.

Stacked on #3240 to allow side-by-side comparison of per-engine metrics.

## Breaking

- The default routing policy changes from `consistent_hash` to `cache_aware`. Affinity is now derived from request token prefixes instead of the `X-Session-ID` header (the header is ignored by `cache_aware`; the orchestrator still sends it). To keep the old behavior, set `inference.router.policy = "consistent_hash"`.

## Verification

All on this change (verified atop main before restacking onto #3240 — the stack adds only the metrics passthrough), fork router v0.1.26 binary:

- **Multi-node (render)**: `uv run inference --slurm --deployment.type multi_node --dry-run` renders `launch_router regular ... "cache_aware"`; with `--router.cache-threshold 0.7 --router.balance-abs-threshold 16` the sbatch gains `cmd+=(--cache-threshold 0.7)` / `cmd+=(--balance-abs-threshold 16)`; without knobs, no extra flags.
- **Single-node (live)**: standalone `uv run inference --vllm.model Qwen/Qwen3-0.6B --router.type vllm-router --router.balance-abs-threshold 16` starts the router with `policy: CacheAware { balance_abs_threshold: 16, ... }`; chat completions through it succeed.
- **RL e2e**: reverse-text example (`uv run rl @ examples/basic/reverse-text/rl.toml`), 20/20 steps through the `cache_aware` router, 0% errors, reward 0.14 → 0.73, clean shutdown.

### DIFF
--- a/docs/inference.md
+++ b/docs/inference.md
@@ -181,7 +181,7 @@ non_cached_tokens = 16          # llm-d only: below this many non-cached prompt 
 "kv-cache-utilization-scorer" = 2.0
 ```
 
-- **`vllm-router`** (default) — our fork of [vllm-router](https://github.com/PrimeIntellect-ai/router). Knob: `policy`. The only backend supported for single-node (local) deployments.
+- **`vllm-router`** (default) — our fork of [vllm-router](https://github.com/PrimeIntellect-ai/router). Knobs: `policy` plus the `cache_aware` thresholds (`cache_threshold`, `balance_abs_threshold`, `balance_rel_threshold`). The only backend supported for single-node (local) deployments.
 - **`llm-d`** — the upstream [llm-d](https://llm-d.ai) Endpoint Picker (EPP) + Envoy proxy (multi-node / disaggregated SLURM deployments only). Routing combines **prefix-cache affinity** (grouped rollouts reuse a cached prefix and skip prefill) with the **`active-request-scorer`** — an in-flight load balancer that spreads requests across ranks immediately, unlike the metrics-scraped `queue-scorer` / `kv-cache-utilization-scorer` / `load-aware-scorer` (which lag and concentrate bursts of same-prefix requests). The scorer weights follow the upstream llm-d P/D guide; tune via `scorers` (base) + `prefill_scorer_overrides` / `decode_scorer_overrides` (per-profile, P/D). Does not support `enable_return_routed_experts` (router replay).
 
 Both backends support the 2 most important things:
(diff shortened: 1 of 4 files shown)


mikasenghaas:
Closing after e2e A/B testing: `cache_aware` is clearly worse than `consistent_hash` for RL workloads, so the default stays `consistent_hash`.

**Setup**: GLM-4.5-Air, 2×etp=8 inference deployment on scaleswe at 512 concurrency (deliberately pushing the deployment), 4 runs: {fork v0.1.26, upstream 0.1.15} × {cache_aware, consistent_hash}.

**Results**: https://wandb.ai/primeintellect/router-balance/workspace?nw=45payatkk26 (see the *inference* section)

- `cache_aware`: heavy cross-engine imbalance on both router builds — engines saturate quickly, and the prefix-cache hit rate is consistently lower and much more volatile.
- `consistent_hash`: much better balance and cache behavior; no meaningful difference between fork and upstream.

Takeaway: RL traffic is mostly unique prompts per group/step, so the radix-tree affinity `cache_aware` relies on buys little, while per-rollout session stickiness is exactly what the KV cache wants. Notably, neither llm-d nor Dynamo offers a client-keyed consistent-hash mode, so `consistent_hash` on vllm-router is currently the strongest routing option for RL.

PR #3256, opened 2026-08-12
feat: online evals during sft

mikasenghaas:
## Summary

Online evals for the `sft` entrypoint, reusing the orchestrator's eval machinery. The handoff is the filesystem, not NCCL: the trainer writes HF weight checkpoints, the inference server reloads them from disk, and a new evaluator process schedules the evals.

- **Config**: `SFTConfig` gains `[eval]` (`OnlineEvalConfig` — the orchestrator's `EvalConfig` shape: multiple `[[eval.source]]` envs with per-source `interval` / `num_examples` / `group_size` / sampling overrides, plus the inference `client`) and `[inference]` (the existing `InferenceConfig`). Single-node: `deployment.num_train_gpus` + `deployment.num_infer_gpus` (alias: `num_eval_gpus`) split GPUs between the trainer and inference (same API as `rl`), with `inference.vllm` DP auto-filled from the GPU count (mirroring `RLConfig`). Multi-node: `deployment.num_infer_nodes` sizes a dedicated eval job. The eval client is auto-wired from `[inference]` (router bypass for admin ops included); `[ckpt]` is auto-enabled since weight checkpoints are the handoff.
- **Evaluator**: new `evaluator` entrypoint (`EvaluatorConfig`, `src/prime_rl/orchestrator/evaluator.py`). It watches `weights/step_{n}` for stable HF checkpoints, points the inference server at each eligible one (`/update_weights` from disk, `weight_broadcast.type = "filesystem"`), and runs the due evals sequentially per checkpoint so every epoch measures exactly one policy version. Reuses `EvalEnvs` / `EvalSource` / `EvalSink` / `EvalRollouts` and logs metrics and traces exactly like the RL orchestrator: `eval/{env}/{all,effective}/...` + `eval/{env}/policy_version` (= checkpoint step), episode traces (via the file monitor) under `rollouts/step_{n}/eval/{all,effective}/`. The base model is evaluated at step 0 before training (`eval.skip_first_step` disables), and the final checkpoint always fires every env. It also runs standalone against any weights dir + OpenAI-compatible server.
- **Trainer**: the SFT loop saves an HF weight checkpoint at every step an eval env is due (in addition to `ckpt.interval`) — that write is how the SFT loop tells inference a new policy is ready. Weight-checkpoint cleaning always keeps the newest checkpoint so the evaluator's final eval can't be deleted out from under it.
- **Single-node launcher**: with `[eval]` set, `uv run sft` composes the full deployment — inference server, one env server per eval source (same deterministic-address contract as `rl`), evaluator, and the torchrun trainer — with process monitoring, a SIGTERM handler, and shared-W&B wiring.
- **Multi-node launcher**: with `[eval]` on a `multi_node` deployment, the launcher submits two decoupled SLURM jobs — the trainer (`num_train_nodes`) and an eval job (`num_infer_nodes`, alias `num_eval_nodes`: one vLLM engine per DP rank behind a router, env servers, evaluator). The only coupling is weight checkpoints on the shared filesystem, so the trainer job releases its allocation when training finishes while the eval job drains remaining checkpoints and exits after the final one. The eval job is held until the trainer job starts (`--dependency=after`, `sbatch --parsable`); a failed second submission cancels the first.
- **Shared W&B run**: trainer and evaluator log to one run keyed by `$PRL_RUN_ID` (`WANDB_RUN_ID`), split into creator and finisher: the trainer (`WANDB_SHARED_PRIMARY`) creates the run so it exists even if the eval side never starts; the evaluator (`WANDB_SHARED_FINISHER`), which outlives the trainer, writes the final state so the run isn't marked finished while final evals still log. A new SFT-flavored W&B overview view groups loss/eval/data/perf sections.
- **Example**: `examples/basic/reverse-text/sft.toml` (and the `configs/basic` dev mirror) now train with online reverse-text evals on 2 GPUs (1 train + 1 infer).

Robustness (review findings addressed in-branch):
- Evals are auxiliary: a failed weight reload skips that step's evals with an error instead of killing the run (queued examples drained so they can't leak into a later epoch), and eval-due checkpoints deleted by `ckpt.keep_last`/`keep_interval` cleaning before evaluation are reported with a warning instead of silently gapping the eval curve.
- Stale eval artifacts from a previous run — `weights/step_{n}` checkpoints and `rollouts/step_{n}` trace dirs — are cleaned at launch (all on fresh start, past the resume step on resume), so the evaluator never replays another run's checkpoints and append-only trace files never mix two policies. Abandoned STABLE-less checkpoints (crash mid-save) are skipped instead of wedging the watcher.
- Plain SFT (no `[eval]`) keeps its old launcher behavior: no GPU probing, no `CUDA_VISIBLE_DEVICES` pinning.
- Pre-existing `is_last_step` off-by-one fixed: resuming from a checkpoint at `step >= max_steps` no longer trains forever.

Known follow-ups: consumer-aware weight-checkpoint retention (delete-after-eval); trainer-failure propagation to the decoupled eval job (it currently drains until walltime if the trainer dies mid-run); `wait -n -p` in the eval sbatch requires bash >= 5.1.

## Breaking

- SFT `deployment.num_gpus` is renamed to `deployment.num_train_gpus` and multi-node `deployment.num_nodes` to `deployment.num_train_nodes` (matching the `rl` launcher's API). Update SFT TOMLs and CLI overrides; the old keys now fail as unknown fields.

## Verification

- Single-node e2e on 2×RTX PRO 6000 (1 train + 1 infer GPU), pre- and post-merge with `main` (run-dir layout, JSON configs, `$PRL_RUN_ID` identity): base model evaluated at step 0, weights reloaded from disk at each eval step and the forced final checkpoint, reward climbing across steps (0.038 → 0.055 → 0.093 on a 6-step smoke), one shared W&B run containing both trainer and eval metrics (trainer `primary=True/finisher=False`, evaluator `primary=False/finisher=True`, run ends `finished`), traces under `rollouts/step_{n}/eval/`.
- Multi-node decoupled e2e on 2 nodes (Qwen3-30B-A3B-Thinking-2507 SFT + AIME25 evals): both SLURM jobs exit 0, evals drain after the trainer job releases its allocation.
- Trainer failure mid-run tears down inference, env servers, and evaluator with a non-zero launcher exit; plain SFT (fake data, no `[eval]`) unchanged; `uv run rl --dry-run` still resolves.

---

> [!NOTE]
> **High Risk**
> Large cross-cutting change (launcher, SLURM, trainer checkpoint cadence, inference weight reload, shared W&B across jobs) with breaking SFT deployment field renames; misconfiguration or checkpoint races can hang evaluators or gap eval curves.
> 
> **Overview**
> **Online evals during SFT** reuse the RL orchestrator’s eval config shape (`[eval]` + `[[eval.source]]`, optional `[inference]`). The trainer saves HF **weight checkpoints** at eval-due steps (not only `ckpt.interval`); a new **`evaluator`** entrypoint watches `weights/step_{n}` with `STABLE` markers, calls inference **`/update_weights`** from disk (`filesystem` weight broadcast), and runs rollouts through the same eval stack (`EvalSource` / `EvalSink`, `eval/{env}/...` metrics and rollout traces). Base model is evaluated at step 0; the final step force-runs every env.
> 
> **Launching:** Single-node `uv run sft` can partition GPUs via `deployment.num_train_gpus` and `deployment.num_infer_gpus` (alias `num_eval_gpus`) and start inference, per-source env servers, evaluator, and trainer with shared W&B (`WANDB_SHARED_PRIMARY` / `FINISHER`). Multi-node submits a **decoupled** eval SLURM job (`num_infer_nodes`, new `multi_node_sft_eval.sbatch.j2`) that outlives the trainer; jobs couple only through shared checkpoint paths.
> 
> **Config / API:** SFT deployment renames **`num_gpus` → `num_train_gpus`** and **`num_nodes` → `num_train_nodes`**. `SFTConfig` validates online-eval wiring (ckpt weights required, inference model must match `model.name`, multi-node needs router). Stale weight checkpoints and rollout dirs are cleaned on launch/resume; checkpoint retention keeps the newest weights so final evals are not deleted early.
> 
> **Monitoring / training metrics:** W&B overview supports an **SFT** flavor (loss/perplexity sections); trainer logs `loss/perplexity` and `val/perplexity`. `EvalSource.trigger` gains **`force`** for final-checkpoint evals.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e8b5e188653441ebd06ef9c260f3c7c7cbde8646. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


configs/basic/reverse-text/sft.toml
```diff
@@ -1,9 +1,11 @@
-# SFT warmup for reverse-text on 2 GPUs. Dev-sized mirror of examples/basic/reverse-text/sft.toml.
+# SFT warmup for reverse-text on 2 GPUs (1 train + 1 infer for online evals).
+# Dev-sized mirror of examples/basic/reverse-text/sft.toml.
 
 max_steps = 100
 
 [deployment]
-num_gpus = 2
+num_gpus = 1
```
mikasenghaas: rename to `num_train_gpus` (same api as rl training afaik)

docs/training.md
```diff
@@ -176,6 +176,57 @@ uv run sft @ examples/basic/reverse-text/sft.toml --wandb
...
+
+Eval metrics log exactly like RL evals — `eval/{env}/{all,effective}/...` plus `eval/{env}/policy_version` (the checkpoint step) — into the same W&B run as the trainer, and traces land under `rollouts/step_{n}/eval/`. Online evals require HF weight checkpoints (`[ckpt]` is auto-enabled). Eval-step weight checkpoints follow the usual `ckpt.keep_last` / `ckpt.keep_interval` retention; if cleaning outpaces the evaluator, the deleted steps are skipped with a warning instead of evaluated. Without an `[inference]` block the launcher starts no server and the evaluator connects to `eval.client.base_url`; the server must run with `weight_broadcast.type = "filesystem"`, and if a router fronts it, set `eval.client.admin_base_url` to the engine URLs (admin ops bypass routers). The evaluator also runs standalone against any weights directory: `uv run evaluator @ <config>` (`EvaluatorConfig`).
+
+#### Multi-Node (Decoupled Trainer and Inference Pool)
```
mikasenghaas, line 209: too much detail. remove from docs

docs/training.md
```diff
@@ -325,7 +378,7 @@ uv run rl @ rl.toml --wandb \
...
+prime-rl deliberately logs a **large number of metrics** for maximum observability: every rollout metric is emitted per subset (`all`/`effective`), per statistic (`mean`/`max`/`min`/`p10`/`p90`), and per environment alongside a cross-env aggregate, so a multi-env run can emit thousands of series. To keep that navigable, W&B mode **auto-creates an `overview` saved view** on the first run into a project — curating the handful of metrics that matter into `train`, `eval`, `stability`, and `performance` sections (with per-env breakdowns). The view is created once per project and adapts to the run's environments; if a later run uses a different set of environments, a new versioned view (`overview-v2`, …) is created instead of overwriting the first. SFT runs get the same treatment with an SFT-shaped `train` section — loss and perplexity (train + validation) instead of rollout metrics — alongside the same per-env `eval` sections and SFT-specific `stability` / `performance` panels.
```
mikasenghaas: remove the added sentence, just weave in that each training run gets an overview view with selected metrics

examples/basic/reverse-text/sft.toml
```diff
@@ -8,7 +8,35 @@ name = "PrimeIntellect/Qwen3-0.6B"
...
 seq_len = 4096
-batch_size = 32 
+batch_size = 32
 
 [optim]
-lr = 2e-5
\ No newline at end of file
+lr = 2e-5
+
+# Online evals: the trainer writes an HF weight checkpoint every `interval` steps,
```
mikasenghaas: remove comment

packages/prime-rl-configs/src/prime_rl/configs/evaluator.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class EvaluatorConfig(BaseConfig):
+    """``uv run evaluator``: watch a weights directory for new HF checkpoints, point
+    the inference server at each one (``/update_weights`` from disk), and run the
+    configured evals against the updated weights. The ``sft`` launcher writes this
+    config; it can also be run standalone against any trainer that writes
+    ``weights/step_{n}`` HF checkpoints with ``STABLE`` markers."""
+
+    model: BaseModelConfig = BaseModelConfig()
```
mikasenghaas: how is this used? this should not be configurable right since its just the ckpts of the run. can we not auto derive this to smth like {model_name}-{step} or smth

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -145,7 +148,10 @@ class SingleNodeDeploymentConfig(BaseDeploymentConfig):
     type: Literal["single_node"] = "single_node"
 
     num_gpus: int = 1
```
mikasenghaas: should be `num_train_gpus` i think

configs/basic/reverse-text/sft.toml
mikasenghaas: Renamed `deployment.num_gpus` → `deployment.num_train_gpus` across the SFT surface (config, launcher, examples, docs, integration tests) in a9e98ebba; listed under `## Breaking` in the description.

packages/prime-rl-configs/src/prime_rl/configs/sft.py
mikasenghaas: Renamed to `num_train_gpus` in a9e98ebba.

docs/training.md
mikasenghaas: Removed the paragraph in a9e98ebba.

docs/training.md
mikasenghaas: Removed the SFT sentence and folded it into the existing paragraph — it now opens with "every training run (RL and SFT) gets an auto-created overview saved view" (a9e98ebba).

examples/basic/reverse-text/sft.toml
mikasenghaas: Removed in a9e98ebba.

packages/prime-rl-configs/src/prime_rl/configs/evaluator.py
mikasenghaas: It's the served-model name: the `model` field on every eval request plus the startup model check. The sft launcher always auto-fills it from `model.name` — it's only a config field because `EvaluatorConfig` doubles as the standalone `uv run evaluator` entrypoint. A per-step name like `{model_name}-{step}` isn't possible with this reload mechanism: `/update_weights` swaps weights in place under one served name (registering a new vLLM model alias per checkpoint would defeat that), so per-step results are told apart by `eval/{env}/policy_version` instead. Flattened `model: BaseModelConfig` → `model_name: str` and documented this on the field in a9e98ebba.

packages/prime-rl-configs/src/prime_rl/configs/evaluator.py
```diff
@@ -0,0 +1,83 @@
...
+    client: ClientConfig = ClientConfig()
+    """Client of the inference server evals run against. Auto-wired from the
+    ``[inference]`` block when the launcher manages the server."""
+
+    env_server_base_port: int = Field(5000, ge=1, le=65535)
+    """First port of the env-server port range: the eval source at position ``i`` is
+    served at ``tcp://127.0.0.1:<base + i>``. Sources with an explicit ``serve.address``
+    keep it instead, without shifting the other sources' ports."""
+
+    max_inflight_episodes: int = Field(64, ge=1)
```
mikasenghaas: put to 128

packages/prime-rl-configs/src/prime_rl/configs/evaluator.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class EvaluatorConfig(BaseConfig):
+    """``uv run evaluator``: watch a weights directory for new HF checkpoints, point
+    the inference server at each one (``/update_weights`` from disk), and run the
+    configured evals against the updated weights. The ``sft`` launcher writes this
+    config; it can also be run standalone against any trainer that writes
+    ``weights/step_{n}`` HF checkpoints with ``STABLE`` markers."""
+
+    model_name: str = "Qwen/Qwen3-0.6B"
```
mikasenghaas: just `model`

packages/prime-rl-configs/src/prime_rl/configs/evaluator.py
```diff
@@ -0,0 +1,83 @@
...
+    """Name the inference server serves the model under — the ``model`` field of every
+    eval request and the startup model check. Auto-filled from ``model.name`` by the
+    ``sft`` launcher; the name stays fixed across checkpoint reloads (weights are
+    swapped in place), so per-step results are told apart by ``eval/{env}/policy_version``."""
+
+    eval: OnlineEvalConfig
+    """Eval sources, sampling, intervals, and the inference client."""
+
+    weights_dir: Path | None = None
```
mikasenghaas, line 56: this is hardcoded on sft entrypoint no? if so, let's hardcode the output_dir/weights assumption here too

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -142,22 +145,31 @@ class BaseDeploymentConfig(BaseConfig):
...
+        if self.num_train_gpus > self.gpus_per_node:
+            raise ValueError(f"num_train_gpus ({self.num_train_gpus}) exceeds gpus_per_node ({self.gpus_per_node}).")
         return self
 
 
 class MultiNodeDeploymentConfig(BaseDeploymentConfig):
     type: Literal["multi_node"] = "multi_node"
 
-    num_nodes: int = 2
+    num_nodes: int = Field(2, ge=1)
```
mikasenghaas: rename to `num_train_nodes`

skills/training/start-run/SKILL.md
```diff
@@ -58,6 +58,15 @@ uv run sft @ examples/basic/reverse-text/sft.toml --dry-run
 - Config: `SFTConfig` (`packages/prime-rl-configs/src/prime_rl/configs/sft.py`)
 - Entrypoint: `src/prime_rl/entrypoints/sft.py`
 - SLURM: single- and multi-node
+- Online evals: an `[eval]` block (same shape as `[orchestrator.eval]`) plus `[inference]`
```
mikasenghaas: remove

src/prime_rl/templates/multi_node_sft_eval.sbatch.j2
mikasenghaas, line 1: should we not have a single multi node sft template, and weave in the inference + evaluator deployment conditionally into the main one?

packages/prime-rl-configs/src/prime_rl/configs/evaluator.py
mikasenghaas: Done in 55c9b8cc1.

packages/prime-rl-configs/src/prime_rl/configs/evaluator.py
mikasenghaas: Renamed to `model` in 55c9b8cc1.

packages/prime-rl-configs/src/prime_rl/configs/evaluator.py
mikasenghaas: Yes, always launcher-derived. Dropped the field — `weights_dir` is now a property hardcoded to `<output_dir>/weights` (55c9b8cc1). Since that removes the only way to point the evaluator at a redirected checkpoint volume, `ckpt.output_dir` is now rejected when `[eval]` is set.

packages/prime-rl-configs/src/prime_rl/configs/sft.py
mikasenghaas: Renamed in 55c9b8cc1; added to `## Breaking`.

skills/training/start-run/SKILL.md
mikasenghaas: Removed in 55c9b8cc1.

src/prime_rl/templates/multi_node_sft_eval.sbatch.j2
mikasenghaas: I'd keep two templates. The RL template weaves inference + trainer into one script because they share a single SLURM allocation; here the decoupling is the feature — two independent jobs with different `#SBATCH` headers (node counts, job names, log files) and fully disjoint bodies (torchrun vs. engines/router/env-servers/evaluator + the completion sentinel), submitted separately so the trainer job releases its allocation while the eval job drains. One file would have to be rendered twice with a top-level trainer/eval switch, which braids two scripts together without sharing anything real. Happy to merge them if you'd rather have the single file, though.

packages/prime-rl-configs/src/prime_rl/configs/evaluator.py
mikasenghaas: Follow-up: weight checkpoints follow `ckpt.output_dir` when it redirects checkpoints to another volume, so the hardcoded `<output_dir>/weights` couldn't cover that deployment — brought the launcher-filled `weights_dir` field back and dropped the `ckpt.output_dir` rejection (latest commit).

docs/training.md
```diff
@@ -176,6 +176,55 @@ uv run sft @ examples/basic/reverse-text/sft.toml --monitors.wandb
 
 Multi-GPU and multi-node use torchrun under the hood (the `sft` entrypoint manages this for you — see [Scaling § SFT and Torchrun](scaling.md#sft-and-torchrun) for non-default layouts; multi-node SFT goes through [SLURM](scaling.md#slurm)).
 
+### Online Evals
```
samsja, line 179:
```suggestion
### Online Evals for SFT 
```
?

docs/training.md
```diff
@@ -176,6 +176,55 @@ uv run sft @ examples/basic/reverse-text/sft.toml --monitors.wandb
...
+num_examples = 32
+
+[[eval.source]]
+name = "reverse-text"
+
+[eval.source.env.taskset]
+id = "reverse-text"
+
+[eval.source.env.agent.harness]
+id = "null"
```
samsja, line 195: is this the new recommanded way ? no more taskets = {id= "null"} I kinda like it for being minimal

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -142,24 +145,33 @@ class BaseDeploymentConfig(BaseConfig):
 class SingleNodeDeploymentConfig(BaseDeploymentConfig):
     type: Literal["single_node"] = "single_node"
 
-    num_gpus: int = 1
-    """GPUs to use."""
+    num_train_gpus: int = 1
+    """GPUs allocated to the trainer."""
+
+    num_infer_gpus: int = 1
```
samsja: infer or eval ?

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -142,24 +145,33 @@ class BaseDeploymentConfig(BaseConfig):
...
 
 
 class MultiNodeDeploymentConfig(BaseDeploymentConfig):
     type: Literal["multi_node"] = "multi_node"
 
-    num_nodes: int = 2
+    num_train_nodes: int = Field(2, ge=1)
     """Training nodes."""
 
+    num_infer_nodes: int = Field(0, ge=0)
```
samsja: infer or eval nodes

src/prime_rl/orchestrator/evaluator.py
```diff
@@ -0,0 +1,320 @@
...
+
+        if reload_weights:
+            get_logger().info(f"Updating inference weights to checkpoint step {step} ({weight_dir})")
+            try:
+                await self.pool.update_weights(weight_dir, step=step)
+            except Exception as exc:
+                # Evals are auxiliary to training: a failed reload (checkpoint cleaned
+                # under us, misconfigured admin endpoint) skips this step's evals
+                # instead of killing the run. Drain the queued examples so they don't
+                # leak into a later step's epoch with the wrong eval_step.
```
samsja: a  bit verbose ai comment ?

src/prime_rl/orchestrator/evaluator.py
```diff
@@ -0,0 +1,320 @@
...
+        error-marker rollout so the sink's count-based finalization still fires."""
+        async with semaphore:
+            try:
+                rollouts = await env.run(
+                    client=self.pool.eval_client,
+                    model_name=self.pool.model_name,
+                    cache_salt=str(step),
+                    task_data=example["task"].data.model_dump(mode="json"),
+                )
+            except Exception as exc:
```
samsja, line 241: wondering it there are more finagrained expection here but otherwise fine by me

packages/prime-rl-configs/src/prime_rl/configs/sft.py
hallerite: eval better here imo

packages/prime-rl-configs/src/prime_rl/configs/sft.py
hallerite: eval

docs/training.md
mikasenghaas: its what we currently do, can revert back to more "dict-like syntax" if you want. agree its a bit verbose rn

packages/prime-rl-configs/src/prime_rl/configs/sft.py
mikasenghaas: hmm rly?

src/prime_rl/orchestrator/evaluator.py
mikasenghaas: think its fine tbh, we dont want to err on this

docs/training.md
mikasenghaas: its grouped under ## SFT so this is implied

docs/training.md
mikasenghaas: will do in follow up

packages/prime-rl-configs/src/prime_rl/configs/sft.py
```diff
@@ -287,6 +324,131 @@ def validate_deployment(self):
...
+        if self.eval is None:
+            if self.inference is not None:
+                raise ValueError("[inference] is only used for online evals — add an [eval] block or remove it.")
+            return self
+
+        # The trainer's HF weight checkpoints are how inference picks up new policies.
+        if self.ckpt is None:
+            self.ckpt = CheckpointConfig()
+        if self.ckpt.output_dir is not None:
+            raise ValueError("Online evals watch <run_dir>/weights; ckpt.output_dir is not supported with [eval].")
```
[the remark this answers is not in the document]
mikasenghaas: Valid — the removal was intended in 711f97057 but the scripted edit silently missed (the format hook had reshaped the raise). Actually removed in 45d8c3aac; verified a dry-run with `--ckpt.output-dir` now resolves and the evaluator watches `<ckpt.output_dir>/weights`.

src/prime_rl/orchestrator/evaluator.py
```diff
@@ -0,0 +1,320 @@
...
+
+        if config.resume_step is None:
+            # Base-model eval: the inference server starts with the untrained weights,
+            # so no reload is needed. The first trigger fires every env (policy v0)
+            # unless ``skip_first_step``.
+            await self.maybe_run_evals(step=0)
+        elif config.eval.retrigger_on_resume:
+            # Re-fire interval-aligned evals at the resume step (e.g. after a crash
+            # that lost in-flight evals). Requires the resume step's weights on disk.
+            await self.maybe_run_evals(step=config.resume_step, reload_weights=True)
```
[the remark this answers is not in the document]
mikasenghaas: Fixed in 45d8c3aac — `retrigger_on_resume` now force-fires every env when the resume step is at/past `max_steps`, same as the watch loop's final-checkpoint handling.

src/prime_rl/orchestrator/evaluator.py
mikasenghaas: Trimmed in 45d8c3aac.

packages/prime-rl-configs/src/prime_rl/configs/sft.py
mikasenghaas: will alias it

packages/prime-rl-configs/src/prime_rl/configs/sft.py
mikasenghaas: will alias

samsja: idk just like eval since its doing eval

mikasenghaas: Aliased in cd8379d55: `num_infer_gpus` stays canonical (rl-launcher parity), `num_eval_gpus` is accepted everywhere (TOML and `--deployment.num-eval-gpus`), including as a CLI override on top of a TOML that uses the canonical spelling.

packages/prime-rl-configs/src/prime_rl/configs/sft.py
mikasenghaas: Same in cd8379d55: `num_infer_nodes` canonical, `num_eval_nodes` accepted as an alias.

mikasenghaas:
Alias precedence fix landed at the right layer: pydantic-config#23 teaches `_normalize_alias_keys` to recurse into discriminated-union variants (and `Optional` models), so `num_eval_*`/`num_infer_*` merge onto one key with normal CLI-over-file precedence in every layering — the deployment pre-validator collapse is gone (435bd378c). Verified: TOML-alias + CLI-canonical now resolves to the CLI value (previously silently lost), plus the reverse direction, alias-only layers, and multi-node.

Merge order: PrimeIntellect-ai/pydantic-config#23 first, then this PR (the submodule pointer references that branch's commit).

mikasenghaas:
pydantic-config#23 is merged and released as prime-pydantic-config **0.4.3**; the submodule now points at the squash-merged main commit (c7299df21), so the merge-order constraint above is gone — this PR is self-contained again.

samsja (approved):

PR #3234, opened 2026-08-11
feat: overlap CPU optimizer with backward

samsja:
## What it does

- Preserves `optim_cpu_offload` as the existing state-only mode, enabled by default.
- Adds opt-in `full_offload` for gradients, FP32 masters, Adam state, and CPU optimizer execution.
- Keeps BF16 compute weights on GPU while preserving configured FP32 parameters such as MoE routers; FP32 masters, Adam moments, gradients, and optimizer arithmetic stay on CPU.
- Runs native multi-tensor CPU AdamW chunks as gradients become ready during backward; offload code and the CPU kernel live under `trainer/optim/`.
- Uses bounded pinned mixed-dtype transfer rings for gradient offload and weight refresh.
- Full offload automatically budgets CPU threads and applies NUMA affinity per rank.
- Supports gradient accumulation, validation, and resumable DCP checkpoints, including model-only resume.
- Full offload only supports AdamW and disables gradient clipping.

## Performance

One 8×H200 node, fake data, random initialization:

| Model | Sequence | No offload | Full offload | TPS retained | Peak HBM |
| --- | ---: | ---: | ---: | ---: | ---: |
| Qwen3-30B-A3B | 8K | 1.75 s | 3.96 s | 44% | 63.1 → 17.5 GiB |
| Qwen3-30B-A3B | 16K | 5.17 s | 5.73 s | 90% | 67.0 → 21.5 GiB |
| Qwen3-30B-A3B | 32K | 16–21 s | 15.4–16.7 s | ~100% | 73.7 → 29.0 GiB |
| Qwen3-30B-A3B | 64K | 38.7 s | 34.1 s | ~100% | 90.3 → 44.9 GiB |
| GLM-5, 8 layers | 16K | 2.1 s | 6.0 s | 38% | 122.9 → 53.9 GiB |
| GLM-5, 8 layers | 32K | 7.7 s | 7.1 s | ~100% | 134.1 → 67.0 GiB |
| GLM-5, 8 layers | 64K | OOM | 10.7 s | — | OOM → 91.6 GiB |
| GLM-5, 20 layers | 16K | OOM | 19.4 s / 3.4k TPS | — | OOM → 102 GiB |
| GLM-5, 20 layers | 32K | OOM | 23.1 s / 5.8k TPS | — | OOM → 115 GiB |
| GLM-5, 20 layers | 64K | OOM | 37.2 s / 7.3k TPS | — | OOM → 134 GiB |

[...]

PR #3258, opened 2026-08-12
perf: keep scalar optimizer step counters on CPU in state offload

samsja:
## Summary

`optim_cpu_offload` (on by default) moves optimizer state between CPU and GPU around each step. `_move_states` also shipped the **0-dim `step` counters** to CUDA — and torch's AdamW reads each one back with `.item()` per parameter per step, turning every read into a device sync (587 syncs/step on Qwen3-30B-A3B, visible as `aten::item` ≈ 1.6 s in profiler traces at seq 8K on 8×H200).

This keeps scalar step counters on CPU — their canonical `torch.optim` location — so the optimizer reads them without touching the device.

Measured effect is modest (8K step median 3.18 → 3.13 s) because the syncs were mostly *waiting on* the state transfers rather than adding to them — the dominant cost of state offload remains the ~2×30 GiB/rank PCIe streaming itself. The change removes the per-parameter serialization points and keeps step counters in the standard representation for checkpoints.

Losses unchanged (verified on paired 5-step runs).

### DIFF
--- a/src/prime_rl/trainer/optim.py
+++ b/src/prime_rl/trainer/optim.py
@@ -43,6 +43,12 @@ class CPUOffloadOptimizer:
                     new_dtensor._local_tensor = new_local
                     state[k] = new_dtensor
                 elif isinstance(v, torch.Tensor):
+                    if v.dim() == 0:
+                        # Scalar step counters stay on CPU: on CUDA the optimizer's
+                        # per-parameter `.item()` read becomes a device sync.
+                        if v.is_cuda:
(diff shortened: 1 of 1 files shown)


samsja:
Heads-up: #3234 has since removed the state-offload mode entirely (offload is now full-offload-only and disabled by default), which deletes the `_move_states` code this PR patches. This fix is still correct for `main` as long as the current default state-offload path exists there — merge order decides whether it lands or becomes moot.

PR #3157, opened 2026-07-29
chore: adapt to the verifiers trace record cleanups

mikasenghaas:
## Summary

Companion to PrimeIntellect-ai/verifiers#2172 (trace record cleanups). `deps/verifiers` points at vf `main` (`f646beb37`, the vf#2172 merge commit).

- `Trace.stamp` → `record_run`, `capture_error` → `record_error` (orchestrator, dispatcher).
- `Trace.error` → `last_error`, and the removed `trainable` / `agent_name` passthroughs read through the now-required `agent` info: `rollout.agent.trainable` / `rollout.agent.name` (metrics, sinks, RAE and hierarchical-GRPO baselines).
- `RunInfo.id` is required: the orchestrator stamps `monitor.run_id`, falling back to a run-local uuid when no W&B identity exists.
- `Trace.agent` is required: the dispatcher's synthesized error/cancel marker rollouts (and the unit-test rollout factories) record the default agent.
- `uv.lock`: the pinned verifiers now floors `renderers>=0.1.9.dev9` (vf#2175), matching the `deps/renderers` pin already on `main`.

## Verification

- `tests/unit`: 483 passed, 5 skipped. Two local-only exclusions, both broken on plain `main` too: `tests/unit/train/models` (`test_nemotron_h_backward` hard-crashes on this box) and the qwen3-vl fake-engine test (stale fixture after the renderers 0.1.9.dev9 bump — fixed separately in #3161; CI skips it without the cached HF snapshot).
- `ruff check` / `ruff format --check` / `uv lock --check` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Mechanical API alignment with verifiers trace schema; behavior should be unchanged aside from stricter required fields on synthetic rollouts and persisted traces.
> 
> **Overview**
> Updates prime-rl for **verifiers trace record cleanups** (companion to verifiers#2172): renamed trace helpers and required `agent` / run metadata on every rollout.
> 
> **Trace API renames** — `Trace.stamp` → `record_run`, `capture_error` → `record_error`, and `Trace.error` → `last_error` in the dispatcher, orchestrator trace persistence, and metrics error breakdowns.
> 
> **Agent-centric fields** — Removed top-level `trainable` / `agent_name` passthroughs; hierarchical GRPO peer keys and RAE baselines use `rollout.agent.name`, and train/eval sinks plus metrics “effective” subsets gate on `rollout.agent.trainable`.
> 
[...]

PR #3277, opened 2026-08-15
feat(orchestrator): record the scalar advantage on trace records

### LINKED ISSUE #3170: [Feature Request] Include advantage values in rollout trace records (`traces.jsonl`)
### Summary

Currently, rollout trace files written by the orchestrator (via `save_rollouts` + `Rollout.to_record()`) do not include the computed advantage values. The `advantages` field on `Rollout` is declared with `Field(..., exclude=True)`, so it is dropped during serialization, and unlike other orchestration metadata (e.g. `env_name`, `policy_version`), it is never mirrored into the record.

This makes offline analysis of training dynamics harder than it needs to be — advantages are already computed and used for W&B distribution logging (`log_distributions` uses `r.scalar_advantage()`), but they never reach the persisted traces.

### Motivation

Having the advantage in the trace records enables:

- **Offline debugging of credit assignment**: quickly spot degenerate groups (all-zero advantage, i.e. no training signal) directly from the JSONL files.
- **Post-hoc filtering / dataset curation**: select high-advantage rollouts for SFT distillation or analysis without re-running the pipeline.
- **Cross-checking**: the persisted values would match the W&B advantage distributions exactly, since both would use the same `scalar_advantage()` scalarization.

### Proposed change

Override `to_record()` on `Rollout` in `src/prime_rl/orchestrator/types.py`:

```python
def to_record(self) -> dict:
    record = super().to_record()
    record["advantage"] = self.scalar_advantage()
    return record
```

This is a minimal change that covers all three call sites in `orchestrator.py` at once (the per-step `all` file in `main_loop`, and the `effective` files in `finalize_train_batch` / `finalize_eval_batch`), without touching `save_rollouts` — `orjson` serializes `float | None` natively.

Optionally, a config flag could also persist the full per-token stream (`record["advantages"] = self.advantages`), though this significantly increases file size and probably shouldn't be the default.

### Expected behavior / semantics

| File | `advantage` value |
|---|---|
| `train/.../effective` | Real scalar value ✅ (written at ship time, after `assign_advantages()` has run) |
| `train/.../all` | `null` — expected, since these are written on arrival in `main_loop`, *before* the TrainSink pipeline computes advantages |
| `eval/...` | `null` — eval rollouts don't go through credit assignment |

Note on semantics: `scalar_advantage()` returns the mean over non-zero positions, so `0.0` means "group had identical rewards / signal cancelled out" while `null` means "never assigned". Downstream consumers should distinguish the two.

### Alternatives considered

- Patching only the `finalize_train_batch` call site to enrich records there — works, but leaves eval/all records inconsistent and duplicates logic if more call sites are added later.
- Removing `exclude=True` on `advantages` — would dump the full per-token list everywhere by default, bloating trace files.

Happy to submit a PR if this direction sounds good.

Mantissagithub:
## Problem

Persisted rollout artifacts need the scalar advantage used for training so offline readers can inspect credit assignment without reconstructing the pipeline.

Since this PR opened, #3297 replaced the `Rollout` wrapper with native `vf.Episode` records. Current `main` now derives the scalar from node-level advantages and stamps it into each effective trace as `info.advantage` through `TrainEpisodes.effective.vf_episodes`.

## Change

This branch is updated onto current `main` and preserves the remaining coverage around that behavior:

- test that an assigned scalar is written to `traces[].info.advantage`
- test that unassigned credit remains absent
- test that assigned `0.0` remains distinguishable from unassigned credit
- document the episode-shaped field and an updated `jq` query in the monitoring skill

The per-token advantage stream stays excluded from persisted records.

## Semantics

| file | `traces[].info.advantage` |
|---|---|
| `train/.../effective` | assigned scalar, including `0.0` |
| `train/.../all` | absent because the episode is written before scoring; `jq` resolves it as `null` |
| `eval/...` and algorithms without RL credit | absent; `jq` resolves it as `null` |

`0.0` means credit was assigned and the signal cancelled out. A missing value means credit was never assigned.

## Validation

- three focused record-shape tests: passed
- `uvx ruff@0.15.14 check --config pyproject.toml tests/unit/orchestrator/test_advantage.py`
- `uvx ruff@0.15.14 format --check --config pyproject.toml tests/unit/orchestrator/test_advantage.py`
- `git diff --check origin/main...HEAD`

Closes #3170


Mantissagithub:
whenever you have a moment @mikasenghaas, would appreciate your eyes on this one, since you wrote the all/effective split in #2983 and last touched the record shape in #3157. the part i'm least sure about is the `null` vs `0.0` distinction: i kept the scalar unset on `all` records rather than defaulting it to `0.0`, so readers can tell "not scored yet" apart from "scored, group cancelled out". happy to change that if you'd rather the key just be absent on those files instead of present-and-null.

@Alex-Shan does this match what you were after in #3170? it mirrors the scalar only, the per-token stream stays excluded.

Alex-Shan:
This looks good and matches what I expected. Adding `advantage` to `to_record()` cleanly covers both the `all` and `effective` write paths without touching `save_rollouts`, and the semantics around `null` vs `0.0` are clearly documented.

PR #3285, opened 2026-08-17
feat: adaptive concurrency

### LINKED ISSUE #2902: feat: adaptive concurrency control for rollout dispatch

mikasenghaas:
## Summary

Adds `ConcurrencyController` to the orchestrator which continuously sizes a cap on in-flight episodes. Replaces a static `max_inflight_episodes`. Motivated by two things:

- Previously `max_inflight_episodes` had to be manually tuned to the workload and inference deployment, wasting some experiment cycles on finding a performant setting. Under-provisioning doesn't maximize inference TPS, over-provisioning risks KV thrash which death-spirals on current main
- Further, there exists no single optimal `max_inflight_episodes` in multi-env training / evals since the optimal inflight depends on the expected tokens/ episode which may vary between envs. Hence, previously, this value had to be set pessimistically to the "most costly" env.

### Config surface

```toml
[orchestrator.concurrency]
initial_inflight = 512
min_inflight = 1
max_inflight = 1024
```

Defaults (`initial_inflight = None`, `min_inflight = 1`, `max_inflight = 1024`) are the right choice for 90%+ of workflows. Set

- **`initial_inflight`** to skip the initial ramp when a good value is known (unset → derived at startup as `C / max_model_len`, pessimistic but safe),
- **`max_inflight`** if you have hard external resource constraints (e.g. sandboxes); `None` removes the ceiling,
- **`min_inflight`** only as `min_inflight = max_inflight` to pin the cap and recover fixed-concurrency behavior (escape hatch; not recommended in practice).

The cap is clamped to `[min_inflight, max_inflight]` throughout. This deprecates `orchestrator.max_inflight_episodes` and `orchestrator.oversampling_factor`.

### Algorithm

The controller finds the maximum number of in-flight episodes the engines can hold **without KV thrash** (past the thrash cliff, more concurrency *reduces* throughput: prefix-cache eviction → re-prefill churn → pile-up).

- **C** (KV capacity in tokens, summed over decode engines) is scraped from vLLM but used **once**: to derive the starting cap when `initial_inflight` is unset. Everything after is feedback.
- **Grow**: ×1.25 per **turnover** while the engines are clear, KV usage < 0.6, and the cap binds admission (`inflight ≥ 0.9 × cap` — no growing past demand). Turnover approximates pipeline cycles: each completion advances it by `1/inflight`, so 100 finished episodes at concurrency 10 are 10 turnovers but 0.1 at concurrency 1000. Ramp speed thereby scales with the pipeline's demonstrated stability — batch-size, episode-length, and step-cadence independent.
- **Trim** (steady-state regulator): KV usage > 0.8 → resize cap *and pool* to `inflight × 0.7 / usage` and **cancel the excess episodes** (youngest first, least sunk cost), on a 30 s cooldown. Cap moves alone cannot relieve pressure from episodes growing in place — they only block admissions — and thrash onset is a cliff (prefix hit 0.97 → 0.06 within ~2 min of the working set crossing capacity). The 0.8-trigger / 0.7-target gap is hysteresis: each trim buys growth headroom so it doesn't re-fire every cooldown.
- **Cut** (emergency, immediate on the poll): preemptions (single-turn workloads) → 0.8× inflight; capacity-queued requests > 0.5× running sustained 30 s → 0.9× inflight (pool units — agentic episodes idle between turns, so engine `running` undercounts the pool). Agentic workloads **never preempt** — vLLM's admission control parks the excess in the waiting queue (measured: 0 lifetime preemptions during full thrash), hence the queue signal. Cuts cancel the excess, then a drain latch blocks further cuts until inflight settles under the new cap; a repeat overload within a short grace window after the drain cuts at 0.5×.

What makes cancellation work: **rollout abort propagates end-to-end** (verifiers `feat/rollout-abort`) — cancelling the orchestrator-side task sends a `cancel` to the env server, which kills the episode and its engine requests, so downward resizes take effect in seconds instead of draining for tens of minutes at thrashed throughput. Refills are burst-capped on **net pool growth** (completions refund the window) so recovery never lands a cap's worth of prefills at once, while fast-turning workloads are never throttled at steady state. Admission stays FIFO and env-unbiased — the controller only sizes the shared permit pool.

## E2E verification

A 2×3 matrix (W&B project [`adaptive-concurrency`](https://wandb.ai/primeintellect/adaptive-concurrency)) covering both workload extremes × the three cap-lifecycle scenarios. Rationale: single-turn and agentic workloads stress opposite signal paths (preemptions vs capacity-queueing), and the three starts exercise every controller phase — turnover ramp from the derived bound, overload cut + cancel from a deliberately over-provisioned start, and steady-state absorption of eval-wave mix shifts.

| scenario | hendrycks-math (single node, 4 train + 4 infer GPUs, R1-Distill-Qwen-1.5B) | GLM-4.5-Air scaleswe (2 nodes, bash harness, prime sandboxes) |
|---|---|---|
| `initial-none` — derived start, turnover ramp into the usage band | [run](https://wandb.ai/primeintellect/adaptive-concurrency/runs/941612aa865c464ab90d53547f006f97) | [run](https://wandb.ai/primeintellect/adaptive-concurrency/runs/a37cebe48ff44735b51c02a9e359cbec) |
| `initial-high` — over-capacity start, overload cut + cancel, settle | [run](https://wandb.ai/primeintellect/adaptive-concurrency/runs/326606d74e8846578035d77e5db47654) | [run](https://wandb.ai/primeintellect/adaptive-concurrency/runs/f1a93f8e277e4b4399a8efee87e18836) |
| `evals` — fixed start + periodic evals (AIME avg@4 / 10-turn SWE-Bench Verified) | [run](https://wandb.ai/primeintellect/adaptive-concurrency/runs/536814d17b9c4d879cbeb20d3d6d0521) | [run](https://wandb.ai/primeintellect/adaptive-concurrency/runs/3cd448fd48004cbe99fa574904da491f) |

Screenshot 2026-08-18 at 8 14 40 PM

Screenshot 2026-08-18 at 8 04 53 PM

Screenshot 2026-08-18 at 8 05 35 PM

## Follow ups

- Continuous capacity tracking for elastic deployments (proportional cap rescale on engine join/leave)
- Shipping the controller into a multi-env evals entrypoint (standalone eval runs + the SFT online-eval loop)

## Breaking

- `orchestrator.max_inflight_episodes` → `orchestrator.concurrency.max_inflight` (now a hard ceiling above the adaptive cap; `None` allowed).
- `orchestrator.oversampling_factor` removed — no derivation from `batch_size`. Repo configs are migrated to `concurrency.initial_inflight`.
- `InferenceMetricsCollector` polls `/metrics` unconditionally now; `collect_inference_metrics` only gates the W&B mirroring. Its waiting-queue overload warning is replaced by the controller's signals.
- `RolloutDispatcher` renamed to `Dispatcher`; `inflight_permits` to `current_inflight`.

Supersedes #2908. Closes #2902.

---

> [!NOTE]
> **High Risk**
> Changes core orchestration throughput and overload behavior—including mass episode cancellation—and replaces a widely used config knob; mis-tuning or controller bugs could stall training or thrash inference KV cache.
> 
> **Overview**
> Introduces **`ConcurrencyController`** and **`[orchestrator.concurrency]`** (`initial_inflight`, `min_inflight`, `max_inflight`) as the way to bound how many episodes run against inference at once. **`max_inflight_episodes`** and **`oversampling_factor`** are removed from orchestrator config; example TOMLs are migrated to explicit concurrency blocks (some runs pin the cap at a fixed value).
> 
> The controller sizes the cap from vLLM load: **grow** on pipeline turnover while KV usage is low and the cap binds admissions; **trim** when usage crosses a threshold (resize cap and cancel youngest in-flight train episodes); **cut** on preemptions or sustained capacity-queue overload, with drain/escalation logic to avoid cut cascades. **`Dispatcher`** (renamed from `RolloutDispatcher`) exposes **`set_limit`**, **`cancel_inflight`**, burst-smoothed admissions, and reports completions for growth pacing.
> 
> **`InferenceMetricsCollector`** always polls **`/metrics`** to feed the controller ( **`collect_inference_metrics`** now only gates W&B mirroring); parsing adds KV capacity, max model len, and capacity-specific waiting counts. Docs add an **Adaptive Concurrency** section; monitor metric paths use slash-separated dispatcher keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7b2436dfbcbfc087bbcb5a1ee8978a311301fec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


configs/ci/nightly-fft/wiki-search.toml
```diff
@@ -17,7 +17,9 @@ name = "Qwen/Qwen3-4B-Instruct-2507"
 [orchestrator]
 batch_size = 512
 group_size = 16
-oversampling_factor = 2.0
+
+[orchestrator.concurrency]
+initial_inflight = 1024
```
mikasenghaas: lets make 1024 max, and 512 initial

docs/inference.md
```diff
@@ -72,7 +73,7 @@ data_parallel_size = 4
 type = "single_node"
 ```
 
-We reccomend choosing your parallelism based on the expected throughput and latency requirements. High `dp` might create high latency, however it will also give you the highest throughput. This is a tradeoff you need to make based on your use case and required `orchestrator.max_inflight_requests`. Setting `tp` to a higher value will usually give you lower latency, but the inference server also will become saturated faster with lower number of requests.
+We reccomend choosing your parallelism based on the expected throughput and latency requirements. High `dp` might create high latency, however it will also give you the highest throughput. This is a tradeoff you need to make based on your use case and required rollout concurrency (see [Adaptive Concurrency](#adaptive-concurrency)). Setting `tp` to a higher value will usually give you lower latency, but the inference server also will become saturated faster with lower number of requests.
```
mikasenghaas: just "concurrency" not rollout concurrency

docs/inference.md
```diff
@@ -203,6 +204,37 @@ X-Session-ID = "trajectory_id" # this is the default - each rollout has a unique
 - `round_robin` - this policy will round-robin the requests between the available replicas. This is useful if you want to balance the load between the replicas. This might give you better results if you don't have enough rollouts to make `consistent_hash` hashing saturated.
 
 
+## Adaptive Concurrency
+
+The orchestrator sizes its in-flight episode cap adaptively. A static cap is workload-dependent: too low starves the engines; too high triggers KV thrash (preemption → re-prefill churn → prefix-cache eviction → pile-up), and recovery is slow. Reactive throttling alone cannot prevent this: episodes run for minutes to hours and build KV the whole time, so once overload is visible, the pressure already in flight is unsheddable. Safety must come from admission-time budgeting; feedback only trims the budget's error.
```
mikasenghaas: remove 

Reactive throttling alone cannot prevent this: episodes run for minutes to hours and build KV the whole time, so once overload is visible, the pressure already in flight is unsheddable. Safety must come from admission-time budgeting; feedback only trims the budget's error.

examples/advanced/glm-4.5-air/search.toml
```diff
@@ -71,7 +71,9 @@ type = "muon"
 batch_size = 256
 group_size = 16
 max_off_policy_steps = 32
-max_inflight_episodes = 512
+
+[orchestrator.concurrency]
+initial_inflight = 512
```
mikasenghaas: make this max 512 and initial = batch size = 256 as well

examples/advanced/glm-4.5-air/swe.toml
```diff
@@ -82,7 +82,9 @@ type = "muon"
 batch_size = 256
 group_size = 16
 max_off_policy_steps = 32
-max_inflight_episodes = 512
+
+[orchestrator.concurrency]
+initial_inflight = 512
```
mikasenghaas: same here

examples/advanced/glm-4.5-air/terminal.toml
```diff
@@ -72,7 +72,9 @@ type = "muon"
 batch_size = 256
 group_size = 16
 max_off_policy_steps = 32
-max_inflight_episodes = 512
+
+[orchestrator.concurrency]
+initial_inflight = 512
```
mikasenghaas: and here

examples/advanced/glm-5.2/swe-llmd.toml
```diff
@@ -113,9 +113,11 @@ weight_decay = 0.0
 [orchestrator]
 batch_size = 256
 group_size = 16
-max_inflight_episodes = 2048
 max_off_policy_steps = 8 # 16/32 can be better, 8 is more CPU stable
 
+[orchestrator.concurrency]
+initial_inflight = 2048
```
mikasenghaas: here

examples/advanced/glm-5.2/swe.toml
```diff
@@ -65,9 +65,11 @@ weight_decay = 0.1
 [orchestrator]
 batch_size = 4096
 group_size = 16
-oversampling_factor = 3
 max_off_policy_steps = 16
 
+[orchestrator.concurrency]
+initial_inflight = 12288
```
mikasenghaas: here

src/prime_rl/inference/vllm/server.py
```diff
@@ -22,6 +22,7 @@
     monkey_patch_dp_coordinator_startup_timeout,
     monkey_patch_harmony_stop_token_propagation,
     monkey_patch_nano_v3_reasoning_parser,
+    monkey_patch_out_of_vocab_detokenization,
```
mikasenghaas: remove this patch entirely

src/prime_rl/monitors/wandb/overview.py
```diff
@@ -70,6 +70,9 @@
     ["inference/agg/prefix_cache_hit_rate/pooled", "inference/agg/prefix_cache_hit_rate/min"],
     ["inference/agg/generation_tokens_total:rate/sum", "inference/agg/generation_tokens_total:rate/min"],
     ["inference/agg/prompt_tokens_total:rate/sum", "inference/agg/prompt_tokens_total:rate/max"],
+    # Orchestrator-side load on the engines: in-flight episodes by kind.
```
mikasenghaas: remove comment

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,499 @@
+"""ConcurrencyController: adaptive in-flight episode cap for the dispatcher.
```
mikasenghaas: lets not name rollout / episode specifically here. let's treat this more as a black-box "unit" since one unit may mean very diff inference workload

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,499 @@
...
+from enum import Enum, auto
+
+from prime_rl.configs.orchestrator import ConcurrencyConfig
+from prime_rl.utils.logger import format_time, get_logger
+from prime_rl.utils.utils import format_num
+
+# Per-episode EWMA smoothing for the cost/duration estimates: an effective
+# window of ~1/alpha episodes per env, independent of batch size and step
+# cadence, wide enough to average out heavy-tailed episode lengths.
+ESTIMATE_ALPHA = 1 / 512
```
mikasenghaas: lets do 1/ 1024

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,499 @@
...
+# Multiplicative backoff: the first cut, and the harsher follow-up when
+# overload survives a full drain
+BACKOFF_FACTOR = 0.75
```
mikasenghaas: lets do 0.8

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,499 @@
...
+ESCALATED_BACKOFF_FACTOR = 0.5
+# Per-poll growth while the engines are clear and the cap binds. Clocked on
```
mikasenghaas: make comment only describe the meaning of the var. also use docstrings. never explain what the code *does not* do

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,499 @@
...
+# polls, not steps: step duration varies from seconds to tens of minutes
+# across workloads, so step-clocked growth adapts at wildly different speeds
+# — and a binding check sampled only at the boundary instant misses caps
+# that bind all step long (episodes complete in bursts at boundaries).
+GROWTH_FACTOR = 1.003
+# Grow only when the cap actually constrains admission
+BINDING_FRACTION = 0.9
+KAPPA_MIN = 0.25
+KAPPA_MAX = 16.0
```
mikasenghaas: max to 8

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,499 @@
...
+# cooldown lets each trim propagate before the next is sized.
+KV_USAGE_TARGET = 0.85
+KV_TRIM_COOLDOWN_POLLS = 6
+# Queue-overload HARD: capacity-queued requests exceed this fraction of
+# running requests for QUEUE_PERSISTENCE_POLLS consecutive polls. Agentic
+# rollouts overload by queueing, not preempting — admission control parks
+# excess load in the waiting queue, so preemptions alone miss it. The
+# persistence window filters natural turn-completion bursts.
+QUEUE_RATIO = 0.5
+QUEUE_PERSISTENCE_POLLS = 12
```
mikasenghaas: lets do 6 so ew can react faster (within 1min)

src/prime_rl/orchestrator/dispatcher.py
```diff
@@ -262,7 +319,7 @@ async def start(self) -> None:
 
     async def stop(self) -> None:
         self.stopped.set()
-        await self.cancel_inflight_rollouts()
+        await self.cancel_inflight_episodes()
```
mikasenghaas, line 346: lets just call this cancel_inflight. no need to name episodes here

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -386,20 +378,46 @@ async def setup(self) -> None:
...
+        )
         self.dispatcher = RolloutDispatcher(
             train_envs=self.train_envs,
             eval_envs=self.eval_envs,
             train_source=self.train_source,
             eval_source=self.eval_source,
             policy_pool=self.policy_inference,
             policy=self.policy,
-            max_inflight_episodes=config.max_inflight_episodes,
+            max_inflight_episodes=self.concurrency.max_inflight,
```
mikasenghaas: rename to current_inflight and max_inflight on dispatcher. also rename `RolloutDispatcher` -> `Dispatcher`

configs/ci/nightly-fft/wiki-search.toml
mikasenghaas: Done in d1cc8e968.

docs/inference.md
mikasenghaas: Done in d1cc8e968.

docs/inference.md
mikasenghaas: Done in d1cc8e968.

examples/advanced/glm-4.5-air/search.toml
mikasenghaas: Done in d1cc8e968.

examples/advanced/glm-4.5-air/swe.toml
mikasenghaas: Done in d1cc8e968.

examples/advanced/glm-4.5-air/terminal.toml
mikasenghaas: Done in d1cc8e968.

examples/advanced/glm-5.2/swe-llmd.toml
mikasenghaas: Done in d1cc8e968.

examples/advanced/glm-5.2/swe.toml
mikasenghaas: Done in d1cc8e968.

src/prime_rl/inference/vllm/server.py
mikasenghaas: Done in d1cc8e968.

src/prime_rl/monitors/wandb/overview.py
mikasenghaas: Done in d1cc8e968.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in d1cc8e968.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in d1cc8e968.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in d1cc8e968.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in d1cc8e968.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in d1cc8e968.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in d1cc8e968.

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: Done in d1cc8e968.

src/prime_rl/orchestrator/orchestrator.py
mikasenghaas: Done in d1cc8e968.

src/prime_rl/orchestrator/dispatcher.py
```diff
@@ -130,8 +135,10 @@ def __init__(
         policy_pool: InferencePool,
         policy: Policy,
         max_inflight_episodes: int,
```
mikasenghaas: whats the diff between? 
```
max_inflight_episodes: int,
max_inflight_ceiling: int | None
```

src/prime_rl/orchestrator/dispatcher.py
```diff
@@ -189,7 +208,45 @@ def inflight_eval_count(self) -> int:
...
+        the current in-flight count sheds nothing — admissions just stay
+        blocked until enough episodes finish."""
+        self.max_inflight = max_inflight
+
+    def cancel_inflight(self, n: int) -> None:
+        """Cancel roughly ``n`` in-flight train episodes, youngest groups
+        first (least inference spend so far). Called on an overload cut so
+        the working set shrinks immediately instead of waiting for the
+        over-admitted episodes to finish at thrashed throughput."""
+        if n <= 0:
```
mikasenghaas: does this ever trigger? should we move _cancel_inflight inline?

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: dont see it renmaed

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,281 @@
...
+        self.can_grow = False
+        self.prev_waiting: dict[str, int] = {}
+        self.queue_overload_polls = 0
+        self.trim_cooldown = 0
+        # After a cut, ignore further cuts until inflight has drained below
+        # the new cap — the overload during drain is stale
+        self.draining = False
+        self.escalated = False
+
+        self._set_limit: Callable[[int], None] | None = None
```
mikasenghaas: dont use _ for this

mikasenghaas: use a small docstring for each hook, what it does and where it comes from

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,281 @@
...
+    def record_episode(self, env_name: str, kind: str, tokens: int, duration: float) -> None:
+        """One completed unit (from the dispatcher). Advances the turnover
+        clock and, while the last poll was green, grows the cap — so growth
+        is paced by the pipeline itself at any concurrency scale."""
+        inflight = self._get_inflight() if self._get_inflight is not None else 0
+        fraction = 1 / max(inflight, self.floor, 1)
+        self.turnover += fraction
+        if self.can_grow and inflight >= BINDING_FRACTION * self.max_inflight:
+            self.cap = self.clamp_f(self.cap * TURNOVER_GROWTH**fraction)
+            if int(self.cap) > self.max_inflight:
```
mikasenghaas: why do we do this in a branch, can we not just clamp(val, min, max)?

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,281 @@
...
+        self.max_inflight = n_max
+        if self._set_limit is not None:
+            self._set_limit(n_max)
+
+    def clamp(self, n_max: float) -> int:
+        return int(self.clamp_f(n_max))
+
+    def clamp_f(self, n_max: float) -> float:
+        ceiling = self.config.max_inflight or math.inf
+        return min(max(n_max, float(self.floor), float(self.config.min_inflight or 1)), float(ceiling))
```
mikasenghaas: whats the diff between those 2, can we unify?

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,281 @@
...
+        """Total KV tokens across decode engines; None until the first poll."""
+        return sum(self.capacity_by_engine.values()) or None
+
+    # ── observability ────────────────────────────────────────────────────────
+
+    def gauges(self) -> dict[str, float]:
+        return {
+            "concurrency/max_inflight": float(self.max_inflight),
+            "concurrency/turnover": self.turnover,
+            "concurrency/capacity_tokens": float(self.capacity or 0),
```
mikasenghaas: just call this capacity

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,281 @@
...
+            "concurrency/signal": float(
+                {Signal.CLEAR: 0, Signal.SOFT: 1, Signal.HARD: 2}[self.signal],
+            ),
+            "concurrency/queue_overload_polls": float(self.queue_overload_polls),
```
mikasenghaas: remove this metric

src/prime_rl/inference/vllm/server.py
```diff
@@ -35,6 +35,8 @@
 # NOTE: Register Nano V3 reasoning parser so configs can use
 # `reasoning_parser = "nano_v3"` without a vLLM plugin file.
 monkey_patch_nano_v3_reasoning_parser()
+# NOTE: Padded-vocab models (Qwen) can sample out-of-vocab ids under RL drift;
```
mikasenghaas: remove this note

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: Done in a5bbc2207.

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: Done in a5bbc2207.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in a5bbc2207.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in a5bbc2207.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in a5bbc2207.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in a5bbc2207.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in a5bbc2207.

src/prime_rl/inference/vllm/server.py
mikasenghaas: Done in a5bbc2207.

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: They're different things: the first is the starting value of the dynamic cap (the controller moves it afterwards), the second is the configured hard maximum, used only to bound out_q. Renamed to initial_max_inflight + documented the distinction in a5bbc2207.

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: The n <= 0 guard was dead (resize_down only calls with a positive excess) — removed. The sync wrapper itself has to stay: observe() is sync so it can't await; create_task is the sync-to-async bridge for the actual cancellation work.

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,317 @@
...
+            total_queued += sample.waiting_capacity if sample.waiting_capacity is not None else sample.waiting
+        self.prev_waiting = {sample.engine_id: sample.waiting for sample in samples}
+
+        if total_running > 0 and total_queued > QUEUE_RATIO * total_running:
+            self.queue_overload_polls += 1
+        else:
+            self.queue_overload_polls = 0
+        queue_overload = self.queue_overload_polls >= QUEUE_PERSISTENCE_POLLS
+        if queue_overload or max_usage > KV_USAGE_GROW:
+            worst = max(worst, Signal.HARD if queue_overload else Signal.SOFT, key=lambda s: s.value)
```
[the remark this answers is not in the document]
mikasenghaas: Fixed in 1678186dc — overload signals now read decode engines only (prefill samples are filtered before the signal loop; capacity/max_model_len collection unchanged).

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -605,29 +627,15 @@ def resolve_batching(self):
...
-            if self.max_inflight_episodes is None:
-                self.max_inflight_episodes = resolved_max_inflight_episodes
-
-        if self.max_inflight_episodes is not None and self.max_inflight_episodes < self.group_size:
-            raise ValueError("max_inflight_episodes must be at least the number of rollouts per example")
+
+        for field in ("max_inflight", "initial_inflight"):
+            value = getattr(self.concurrency, field)
+            if value is not None and value < self.group_size:
+                raise ValueError(f"concurrency.{field} must be at least the number of rollouts per example")
```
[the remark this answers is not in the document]
mikasenghaas: Fixed in 1678186dc — divisibility is checked after the default lands (`if self.batch_size is not None`), so group_size-only configs are validated against the 128 default.

src/prime_rl/orchestrator/dispatcher.py
```diff
@@ -201,7 +227,47 @@ def inflight_eval_count(self) -> int:
...
+            group_age[meta.group_id] = max(group_age.get(meta.group_id, 0.0), meta.started_at)
+        shed = 0
+        for group_id in sorted(group_age, key=lambda gid: group_age[gid], reverse=True):
+            if shed >= n:
+                break
+            # Count only live cancellations toward the excess: drop_group's
+            # return includes never-dispatched markers, which free no permits
+            live = sum(1 for meta in self.inflight.values() if meta.group_id == group_id)
+            await self.drop_group(group_id)
+            shed += live
```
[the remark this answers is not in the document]
mikasenghaas: Fixed in 1678186dc — group age is now the oldest member's start (min), so youngest-first genuinely sheds the least sunk cost.

src/prime_rl/orchestrator/dispatcher.py
```diff
@@ -201,7 +227,47 @@ def inflight_eval_count(self) -> int:
...
+            get_logger().warning(f"Cancelled {shed} youngest in-flight episodes after overload cut")
+
+    def admission_budget(self) -> int:
+        """Admissions still allowed in the current burst window."""
+        now = time.monotonic()
+        if now - self.admission_window_start >= self.admission_window_s:
+            self.admission_window_start = now
+            self.admissions_in_window = 0
+        burst_cap = max(self.min_burst, self.max_inflight // 10)
+        return burst_cap - self.admissions_in_window
```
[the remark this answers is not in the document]
mikasenghaas: Partially real; fixed the harmful half in 1678186dc. The admission-window refund no longer applies to cancelled episodes (natural completions only), so a mass shed cannot hand its burst budget to the refill. The residual interleaving (refill while later groups drop) is bounded by the already-lowered cap, the per-window burst cap, and the drain latch that blocks further cuts until the pool settles — admissions under the new cap during a shed are legitimate replacements, not a violation of the cut.

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -420,6 +420,31 @@ class NIXLWeightBroadcastConfig(InMemoryWeightBroadcastConfig):
 ]
 
 
+class ConcurrencyConfig(BaseConfig):
+    """Adaptive in-flight concurrency control. The orchestrator sizes the
+    in-flight episode cap from engine KV capacity and learned per-env episode
+    costs; these fields only bound and seed it."""
+
+    initial_inflight: int | None = Field(None, ge=1)
```
samsja, line 428: should we make this scale with number of nodes initialy ? or not worth it because anyway its adaptative

samsja:
>What makes cancellation work: rollout abort propagates end-to-end

does rollout cancelation has a mechanism to trigger early group formation ? thinking if the downscaling here can lead to off policiy issue ?

That being said the fact that we cancel the newly created first should help avoid this issue

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -420,6 +420,31 @@ class NIXLWeightBroadcastConfig(InMemoryWeightBroadcastConfig):
...
+    """In-flight cap to start from. Set it when a good value is known to skip the initial ramp; None derives a pessimistic bound at runtime — engine KV capacity divided by the engine's max context length (``max_model_len``, whether set explicitly or taken from the model config)."""
```
samsja: hmm we don't have a default value here ? seems off

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -420,6 +420,31 @@ class NIXLWeightBroadcastConfig(InMemoryWeightBroadcastConfig):
...
+
+    min_inflight: int | None = Field(None, ge=1)
```
samsja: why not 0 as min defefault here ?

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -420,6 +420,31 @@ class NIXLWeightBroadcastConfig(InMemoryWeightBroadcastConfig):
...
+    """Hard floor on the cap (None auto-sets to 1). Escape hatch: ``min_inflight = max_inflight`` pins the cap for a fixed concurrency; not recommended in practice."""
+
+    max_inflight: int | None = Field(None, ge=1)
```
samsja: wondering if we shouldn't still bound this to some value

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -420,6 +420,31 @@ class NIXLWeightBroadcastConfig(InMemoryWeightBroadcastConfig):
...
+    """Hard ceiling on in-flight episodes — one episode is one agent run at a time, whatever the env's agents are. None leaves the cap purely capacity-driven."""
+
+    @model_validator(mode="after")
+    def validate_bounds(self):
+        if self.max_inflight is not None and self.initial_inflight is not None:
```
samsja: yeah lets remoe this None stuff would make this function trival to nbe written tbh

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,321 @@
...
+runaway growth creates), growth freezes instead of compounding blind."""
+
+
+class Signal(Enum):
+    CLEAR = auto()
+    SOFT = auto()
+    HARD = auto()
+
+
+@dataclass(frozen=True)
```
samsja, line 92: why frozen ?

src/prime_rl/orchestrator/concurrency.py
```diff
@@ -0,0 +1,321 @@
...
+escalated fraction; past the grace window the system is considered recovered
+and escalation resets."""
+
+GROWTH_GATE_TTL_S = 15.0
+"""The growth gate expires this long after the poll that set it — if the
+metrics path stalls (an API server too loaded to answer is exactly the state
+runaway growth creates), growth freezes instead of compounding blind."""
+
+
+class Signal(Enum):
```
samsja: why not literal here instead of Enum ?

src/prime_rl/orchestrator/dispatcher.py
```diff
@@ -201,7 +227,51 @@ def inflight_eval_count(self) -> int:
...
+    def cancel_inflight(self, n: int) -> None:
+        """Cancel roughly ``n`` in-flight train episodes, youngest groups
+        first (least inference spend so far). Called on an overload cut so
+        the working set shrinks immediately instead of waiting for the
+        over-admitted episodes to finish at thrashed throughput."""
+        asyncio.create_task(self._cancel_inflight(n))
+
+    async def _cancel_inflight(self, n: int) -> None:
+        # A group's age is its OLDEST member's start: max() would make a
+        # long-running group look young the moment it schedules another
```
samsja, line 246: ehhh can we rewrite this comment, I think its qite usefull, but its a bit confusing

samsja:
still need to wrap my head around the turnover logic, maybe a bit of docs on it could help and on the algorithm to upscale downscale concurrency. 

Code looks good tho, surprisingly small tbh

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: i would like for most ppl to never set this tbh. i think i might agree that it would make sense to make this scale with inference so that config doesn't have to chnage if the deployment size changes

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: its auto-derived as `C / max_ctx` which can only be determined at runtime

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: we could do 1, yes

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: ya, possibly wdyt is a good value? 1k? 2k?

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: its just data packet, should never need to set any values in place

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: can do

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: fair

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: its a little complicated i think, would keep it global for now

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: Done in 84d4e6acd — defaults to 1 (a plain int now, no None-meaning-1).

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: Done in 84d4e6acd — defaults to 1024; None still opts out of the ceiling. Examples that start above 1024 set their own ceiling explicitly (a7b2436df).

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: Simplified in 84d4e6acd: min_inflight is a plain int so its None-guard is gone. initial_inflight and max_inflight keep None by design — initial derives at runtime from measured KV capacity, and max_inflight = None is the explicit opt-out of the ceiling.

src/prime_rl/orchestrator/concurrency.py
mikasenghaas: Done in 84d4e6acd — Signal is a Literal with a severity map.

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: Rewritten in 84d4e6acd: the pool may only grow by burst_cap per window; replacing a completed episode is always free, only net expansion is metered.

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: Data point for keeping it global: the unset default already scales with deployment size in the natural unit — it derives from summed decode-engine KV capacity, so 3 FP8 replicas derived 113 and 1 replica derived 37 with zero config, and it stays correct across TP/DP shapes, quantization, and P/D (prefill engines excluded). Node count would be the wrong ruler for the same reason a GLM-Air node and a 1.5B node differ ~20x in what they hold.

samsja (dismissed):

PR #3294, opened 2026-08-18
Remove pipeline sync in inner loss function loop by replacing the bra…

MarioSieg:
This removes a often call CPU<->GPU sync by replacing a branch if(..) with branchless code

---

> [!NOTE]
> **Medium Risk**
> Changes numerics-equivalent math and autograd behavior on the training loss path; logic is localized but mistakes would affect logprob/entropy gradients.
> 
> **Overview**
> **`_SequenceChunkedLogProbEntropyFn`** in `lm_head.py` no longer branches on `torch.any(mask)` inside the per-vocab-chunk loop. Target logit reads and gradient scatter use **`gather` / `scatter_add_` with `torch.where` and masks** so the hot path stays on GPU without CPU sync from `bool(Tensor)`.
> 
> Forward now calls **`ctx.set_materialize_grads(False)`** so unused entropy gradients arrive as **`None`** instead of zero tensors. Backward only checks **`grad_entropy is None`** (no `torch.all(grad_entropy == 0.0)`), avoiding another sync when entropy is not differentiated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d205f0809540c49464f526fc6bdfd3fa7122bf5e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/models/layers/lm_head.py
+++ b/src/prime_rl/trainer/models/layers/lm_head.py
@@ -136,16 +136,18 @@ class _SequenceChunkedLogProbEntropyFn(torch.autograd.Function):
 
                 m, s, t = _online_logsumexp_and_weighted_update(m, s, t, scaled_logits)
 
-                mask = (labels_chunk >= vocab_start) & (labels_chunk < vocab_end)
-                if torch.any(mask):
-                    idx = (labels_chunk[mask] - vocab_start).to(torch.long)
-                    target_logits[mask] = scaled_logits[mask, idx]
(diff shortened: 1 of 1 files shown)


mikasenghaas (approved):

S1ro1 (commented):
This looks very good to me directionally, will leave on someone with more time to test to approve though!

samsja (approved):

PR #3296, opened 2026-08-18
feat(inference): Dynamo integration

biswapanda:
## Summary

- add a Dynamo inference-pool adapter for worker discovery, health, pause/resume, version fencing, and filesystem, NIXL, and native NCCL updates
- add a Dynamo-native trainer adapter for vLLM stateful weight transfer
- keep Dynamo and the vLLM sidecar agnostic to routed-expert serialization
- normalize native vLLM routed-expert data at the Prime-RL renderer boundary and install that behavior in every env-server worker
- retain Prime-RL existing inference and broadcast implementations for non-Dynamo configurations

## Stack

This branch is the Dynamo integration layer on top of #3443, the current-main rebase of #3260. It contains #3443 at `576ed1e4f`; the Dynamo layer begins at `6004a5660` and currently ends at `bf42c4c35`.

GitHub does not allow an upstream PR to use a branch in a fork as its base, so this PR temporarily targets `main` and includes the #3443 commits in its displayed diff. Once #3443 merges, this PR automatically reduces to the Dynamo-only layer.

## Scope

Dynamo behavior remains concentrated in the Dynamo adapter and weight-transfer implementation. Dynamo forwards routed data as opaque bytes without validating the framework-specific payload. Prime-RL performs any native-vLLM-to-Prime normalization at its own boundary.

## Validation

- Ruff passed for every Python file changed by the Dynamo layer
- `tests/unit/inference/test_dynamo.py`, the new Dynamo configuration tests, and `tests/unit/orchestrator/test_clients.py`: 20 passed
- recent Kubernetes Qwen3.5 multimodal smoke test exercised this restacked Prime-RL branch with Dynamo and the vLLM-Rust sidecar on the vLLM 0.28.0 runtime
- prior Qwen3-MoE two-step math GRPO test completed successfully and exercised opaque routed-expert transport and the Prime-RL routed-expert consumption path
- prior NIXL test initialized the Prime transfer group, published the model tensors, and applied the updated policy in vLLM

The local lightweight environment does not include Torch or optional taskset plugins, so the Torch transport module and repository-wide config sweep were not rerun locally during this restack.

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/rl.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/rl.py
@@ -463,8 +463,9 @@ class RLConfig(BaseConfig):
         filesystem when LoRA is enabled (not yet supported by in-memory transfer) or when no
         inference server is configured.
         """
+        client = self.orchestrator.model.client
         if self.weight_broadcast is None:
-            if self.trainer.model.lora is not None or self.inference is None:
+            if self.trainer.model.lora is not None or (self.inference is None and not client.is_dynamo()):
                 self.weight_broadcast = SharedFileSystemWeightBroadcastConfig()
             else:
                 self.weight_broadcast = SharedNCCLWeightBroadcastConfig()
@@ -475,11 +476,12 @@ class RLConfig(BaseConfig):
                 "have no disk artifact to load from."
             )
         if self.weight_broadcast.type in ("nccl", "nixl"):
-            inference_world_size = (
-                self.inference.vllm.data_parallel_size * self.inference.vllm.tensor_parallel_size
-                if self.inference
-                else 1
-            )
+            if self.inference is not None:
+                inference_world_size = self.inference.vllm.data_parallel_size * self.inference.vllm.tensor_parallel_size
+            elif client.is_dynamo() and self.deployment.type == "single_node":
+                inference_world_size = self.deployment.num_infer_gpus
+            else:
+                inference_world_size = 1
             common_config = dict(
                 host=self.weight_broadcast.host,
                 port=self.weight_broadcast.port,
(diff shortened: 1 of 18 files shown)


packages/prime-rl-configs/src/prime_rl/configs/rl.py
```diff
@@ -458,13 +470,26 @@ def auto_setup_weight_broadcast(self):
...
+                    trainer_config = {
+                        **common_config,
+                        "dynamo": {
+                            "discovery_url": client.dynamo.discovery_url,
+                            "model_name": self.trainer.model.name,
+                            "headers": client.headers,
+                            "headers_from_env": client.headers_from_env,
+                            "api_key_var": client.api_key_var,
+                        },
+                    }
```
samsja, line 506: should we just do common_config["dynamo"] = {...} and not change the trainer config from common config in the line below ?

packages/prime-rl-configs/src/prime_rl/configs/shared.py
```diff
@@ -178,6 +183,12 @@ class ClientConfig(BaseConfig):
     admin_base_url: list[str] | None = None
     """Separate base URLs for admin operations (weight updates, health checks). When set, admin clients bypass routers and hit each server directly — used in multi-replica or disaggregated P/D deployments where the router must not handle admin traffic."""
 
+    dynamo: DynamoConfig | None = None
+    """Dynamo inference discovery configuration."""
+
+    def is_dynamo(self) -> bool:
+        return self.dynamo is not None
```
samsja, line 196: lest remove this and rather use self.dynamo is None instead

src/prime_rl/entrypoints/env_server.py
```diff
@@ -15,6 +16,7 @@
 def setup_worker(log_level: str | None, json_logging: bool, sandbox_labels: list[str]) -> None:
     setup_env_server_logging(log_level, json_logging)
     set_base_sandbox_labels(sandbox_labels)
+    install_native_routed_experts_normalizer()
```
samsja, line 20: we should probably remove this for now until we do a pr to move to vllm format for r3

src/prime_rl/inference/vllm/server.py
```diff
@@ -79,7 +83,23 @@ async def resume(request: Request):
 @router.post("/update_weights")
 async def update_weights(request: Request):
     data = await request.json()
-    await engine_client(request).collective_rpc("update_weights_from_path", args=(data.get("weight_dir"),))
+    update_info = data.get("update_info")
+    if update_info is not None:
```
samsja, line 87: not sure to follow what is this update info

src/prime_rl/entrypoints/env_server.py
```diff
@@ -15,6 +16,7 @@
...
```
samsja, line 19: would remove this for now

PR #3309, opened 2026-08-19
feat: adaptive concurrency for online evals

mikasenghaas:
## Summary

Adaptive concurrency (#3285) for online evals. The evals process drops its fixed `max_inflight_episodes` semaphore and schedules through the orchestrator's pipeline 1:1:

- An eval-only `Dispatcher` (its train side is now optional) admits episodes under the `ConcurrencyController`, fed by the `InferenceMetricsCollector`'s `/metrics` polls — the same KV-pressure-driven sizing, growth pacing, and burst-capped admission the orchestrator uses.
- Configure the band via `[eval.concurrency]` (`min_inflight` / `max_inflight` / `initial_inflight`). Eval episodes are measurements: an overload cut only blocks admission (no `on_overload` cancellation), and the in-flight pool drains through natural completions.
- Fail fast when adaptivity has no signal: external inference APIs (e.g. Prime Inference) expose no vLLM `/metrics`, so the controller would never bootstrap and the cap would silently sit at `min_inflight`. The evals process probes `/metrics` once at startup: no engine metrics + an adaptive band → hard error telling you to pin the band; a pinned band (`min_inflight = max_inflight`) runs fixed concurrency with a warning. A successful probe also bootstraps the controller before the first dispatch.
- Periodic pipeline view (per-env epoch progress + in-flight pool vs. controller cap) and the same `concurrency/*` W&B gauges as the orchestrator.

Secondary: evals is promoted to a first-class entrypoint (`uv run evals`, `src/prime_rl/evals/`), usable beyond the SFT launcher:

- Standalone (no `[online]` block): run one epoch of every configured eval source against the weights the inference server currently serves, then exit. The evals process spawns one env server per source without an explicit `serve.address`, so a standalone eval is a single `uv run evals @ eval.toml` against a separately started inference server.
- Online (`[online]` block with `weights_dir` / `max_steps` / `resume_step`): watch the weights directory for stable `step_{n}` HF checkpoints and evaluate each eligible one — the mode the `sft` launcher configures. The launcher keeps spawning its own env servers and stamps the derived addresses into the evals config (marking them externally managed).
- `examples/evals/swe.toml`: SWE-bench Verified + Terminal-Bench 2 against `deepseek/deepseek-v4-flash` on Prime Inference, 1h agent cap (`agent.timeout.rollout = 3600`), pinned band. Config-load CI covers it (`EvalsConfig` added to `CONFIG_CLASSES`).
- The `start-run` skill covers the new entrypoint (user-facing docs for standalone evals are deferred for now; the Online Evals docs are updated for the rename and `[eval.concurrency]`).

## Breaking

- `eval.max_inflight_episodes` is removed. Concurrency is adaptive; bound it with `[eval.concurrency]` `min_inflight` / `max_inflight` (set them equal to recover fixed concurrency).
- The entrypoint command is renamed: `uv run evaluator` → `uv run evals`.
- `weights_dir`, `max_steps`, and `resume_step` move into the optional `[online]` block. A config without `[online]` now runs standalone (one epoch, then exit) instead of watching `<output_dir>/weights`.
- Config classes renamed: `EvaluatorConfig` → `EvalsConfig`, `OnlineEvalConfig` → `EvalsEvalConfig`, both in `prime_rl.configs.evals` (TOML shape unchanged; `[eval]` in SFT configs works as before). The implementation moves to `prime_rl.evals` (class `Evals` — no `Evaluator` identifier remains); the launcher writes `evals.json` and logs to `evals.log`.

## Verification

Adaptive path — standalone e2e on 2×RTX PRO 6000: dp=2 `Qwen/Qwen3-4B` inference started separately (`uv run inference --vllm.data-parallel-size 2 --vllm.max-model-len 16384`), then one `uv run evals @ eval.toml` with `aime25` (30 examples × 4) + `gpqa` (48 × 2), band `min_inflight=8` / `max_inflight=128`, `max_completion_tokens=8192` ([W&B run](https://wandb.ai/primeintellect/prime-rl/runs/3lsvdbqi)):

- The evals process spawned both env servers, derived the initial cap from measured engine KV (`Derived initial max inflight 66 - 1.1M KV cache tokens / 16.4K tokens per episode`), and grew it turnover-paced 66 → 95 with the signal `clear` throughout (no overload at this scale, ceiling never hit).
- Both epochs finalized and the process exited cleanly in ~9 min: `aime2025` 120/120 (pass@1 0.183, pass@4 0.233, error 0%; truncation 82.5% is the 8k completion cap, not a pipeline issue) and `gpqa-diamond` 96/96.
- gpqa's 49% error rate is a judge artifact of that run's setup: the judge fallback was pointed at the local Qwen3-4B, which often fails the judge output format. Errors surfaced as error rollouts and count-based finalization was unaffected.
- A follow-up mini-run (4 examples, `max_inflight=8`) confirmed the ceiling clamps the derived cap and teardown is clean.

Adaptive path — SFT online evals (`uv run sft @ examples/basic/reverse-text/sft.toml --max-steps 25` + W&B): the launcher wrote `evals.json` and spawned the `evals` process, which ran the base-model eval at step 0 (adaptive cap derived from the local engine: `Derived initial max inflight 18`), watched the weights dir, reloaded step 25, evaluated it (`eval/reverse-text/policy_version` 0 and 25 both logged), and finalized the shared W&B run.

Fail-fast paths, verified against a mock external endpoint (serves `/v1/models`, 404s `/health` and `/metrics` — Prime Inference's shape): an adaptive band aborts at startup with `No engine metrics at <url> - adaptive concurrency has no load signal ... pin the concurrency by setting eval.concurrency.min_inflight = max_inflight`; a pinned band logs `No engine metrics - running with concurrency pinned at N` and runs fixed.

Standalone entrypoint — live run of `examples/evals/swe.toml` against Prime Inference (`deepseek/deepseek-v4-flash`), capped for cost (`--eval.num-examples 8`, `agent.max_turns = 10`): the evals process spawned both harbor env servers, ran 16 episodes in prime sandboxes at the pinned band, and exited cleanly — swebench-verified 8/8 (2m23s), terminal-bench-2 8/8 (5m55s, reward 0.25), 0% errors (100% truncation is the 10-turn smoke cap).

Also verified without GPUs: config round-trip through the `sft` launcher's `build_evals_config` (address stamping, `[online]` block), `evals --help`, ruff, and the config-load tests (121 passed).

## Follow-up

Consolidate the strengths of `uv run eval` (vf) and `uv run evals` into one path — env-server process isolation, a live TUI, and adaptive concurrency together:

1. Pipeline TUI in `evals` (prl-only): the process already holds per-env epoch progress, in-flight vs. cap, the overload signal, and each completed episode's reward/errors client-side — render them as a rich `Live` view gated on `stdout.isatty()`, keeping the current log lines otherwise (SFT online mode redirects to `evals.log`, so it stays log-based automatically).
2. Per-rollout live slots (verifiers): vf's dashboard reads in-process `RunSlot`s, which is why `--rich --server` is refused — the serve protocol only ships finished episodes. Stream episode progress from workers (e.g. turn/usage deltas on the existing heartbeat or `/state` channels); that unlocks live slots for both `eval --server --rich` and `evals`.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit a9846f5aa4663ee8123ce73d2cc82814b48051cc. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/training.md
```diff
@@ -37,6 +38,7 @@ This page covers everything you need to launch, observe, checkpoint, and recover
 | `uv run inference` | vLLM server. | Always use this entrypoint over `vllm serve` — it adds `/update_weights`, `/load_lora_adapter`, and `/init_broadcaster`. |
 | `uv run trainer` | Standalone trainer process group. | Use only when launching the trainer separately from the orchestrator (e.g. multi-node RL without the `rl` wrapper). |
 | `uv run orchestrator` | Standalone orchestrator process. | Pair with a separately-launched trainer, inference, and one `env-server` per source. |
+| `uv run evaluator` | Multi-env evals against a live inference server. | Standalone one-shot evals (see [Standalone Evals](#standalone-evals)), or checkpoint-watching online evals with an `[online]` block (the `sft` launcher configures this). Spawns its own env servers. |
```
mikasenghaas: call this `evals`

docs/training.md
```diff
@@ -17,6 +17,7 @@ This page covers everything you need to launch, observe, checkpoint, and recover
   - [Launch](#launch-1)
   - [SFT-Specific Knobs](#sft-specific-knobs)
   - [Important Metrics](#important-metrics-1)
+- [Standalone Evals](#standalone-evals)
```
mikasenghaas: just call it `Evals`

docs/training.md
```diff
@@ -263,6 +265,45 @@ Pulled from the console log and mirrored to W&B.
 | `perf/peak_memory` | peak GPU memory (GiB) |
 | `time/step`, `time/forward_backward`, `time/save_ckpt` | step breakdown |
 
+## Standalone Evals
```
mikasenghaas: just Evals

mikasenghaas (commented):
add evals config check to config ci

samsja (approved):

PR #3312, opened 2026-08-20
feat: rank-parallel HF weight save

mikasenghaas:
## Summary

- Replace the master-only HF weight save (gather everything to rank 0, rank 0 writes all safetensors shards) with a rank-parallel save used by both the filesystem weight broadcast and `WeightCheckpointManager`.
- `gather_weights_parallel`: all ranks still participate in the per-tensor `full_tensor()` all-gathers (they are collective and materialize the full tensor on every rank anyway), but each rank now keeps a layer-aligned slice on CPU instead of throwing it away. D2H copies go through persistent pinned buffers with non-blocking copies that overlap subsequent gathers.
- Slices are layer-aligned so each rank can run the prime→HF conversion on its own slice (conversion ops only stack/unstack tensors within one layer, and all ops tolerate missing keys).
- `save_state_dict_parallel`: each rank splits its slice into ≤5GB shards, writes them under temporary names, shard maps are exchanged via `all_gather_object`, every rank renames its own files into the global `model-XXXXX-of-XXXXX.safetensors` numbering (non-POSIX shared filesystems can't reliably rename another rank's unpublished writes), and master writes the index after the barrier — so a complete index still implies complete shards, and `STABLE` semantics are unchanged.
- `convert_state_dict_to_hf` unifies the prime/HF conversion branch that was duplicated across the broadcast and weight-checkpoint call sites; format detection uses the model's full key set so rank-partial dicts convert like full ones.
- The layout is inferred from the run: full runs save sharded safetensors through the parallel path; LoRA runs save only the adapter (peft layout: `adapter_model.safetensors` + `adapter_config.json` at the step dir root, plus config/tokenizer), both for the weight checkpoint and the broadcast. The torch serialization format is removed.
- Weight checkpoints save at eval steps by default — the SFT/RL entrypoint validators hand the trainer the eval intervals via a private field, so the HF snapshots line up with eval scores. `ckpt.weights.interval` adds extra steps on top; `ckpt.weights = "None"` disables HF weight checkpoints entirely (RL only — the broadcast, not the weight checkpoint, feeds the engines; SFT online evals consume the weight checkpoints, so the config fails there). Runs with no eval steps and no interval write no HF weights.
- Online evals detect adapter checkpoints and load them via `/load_lora_adapter` under a fixed adapter name (mirrors the orchestrator's `WeightWatcher`); previously the evals reload assumed full-model checkpoints.
- The merged-LoRA gather-to-master machinery is deleted (`gather_weights_on_master`, `WeightCheckpointManager.save_to_path`, `clean_lora_state_dict`); the benchmark script vendors the master gather as its baseline.

## Breaking

- `ckpt.weights` changes meaning: `save_sharded`, `save_format`, and `save_adapter_separately` are removed; `interval` is the only field. Weight checkpoints land at eval steps plus every `interval` steps — no longer at every `ckpt.interval` step, and runs without either trigger write none (offline DCP→HF conversion is a follow-up). Migration: delete the removed fields; set `ckpt.weights.interval` to keep interval-based snapshots.
- `ckpt.skip_gather_master_weights` is removed. Disable weight checkpoints with `ckpt.weights = "None"` (RL only).
- `weight_broadcast.save_sharded` and `weight_broadcast.save_format` (`FileSystemWeightBroadcastConfig`) are removed. Migration: delete the fields from TOMLs.
- Weight checkpoints of LoRA runs contain only the adapter — merged full weights are no longer written. Online evals load the adapter automatically; merge with the base model via `peft` for external use.
- The torch (`.bin`) weight serialization format is gone from `save_state_dict`.

## Benchmark

Laguna-XS.2 (33.4B, 68.1GB bf16), random init, 2x RTX PRO 6000 Blackwell, single local NVMe:

| impl | gather | convert | write | fsync | total |
|---|---|---|---|---|---|
| master (main) | 25.1–25.5s | 0.1s | 47.3–49.2s | 8.7–10.3s | ~83.1s |
| parallel | 2.7s (9.9s on first save, incl. one-time pinning) | 0.1s | 42.6–42.9s | 5.4–6.2s | ~51.9s |

The gather+D2H phase is ~9x faster steady-state. The write phase only gains ~10% here because a single local NVMe (~1.6GB/s) is the bottleneck for any number of writers; on multi-node shared filesystems the parallel writers scale with per-node bandwidth.

The benchmark script (`benchmarks/scripts/bench_weight_save.py`) is not part of this diff — it lives at 929e43c37 for anyone reproducing the numbers (poolside/Laguna-S-2.1 at 117.6B needs an 8-GPU node).

## Verification

- The benchmark's `--check` mode passes on a tiny 4-layer laguna config with EP=2 (exercises stacked-expert unstacking and expert DTensors), on `Qwen/Qwen3-0.6B` (exercises the `revert_weight_conversion` path and tied embeddings), and on `Qwen/Qwen3-8B` (16.4GB, 399 tensors): both implementations produce bitwise-identical tensors, and the index covers every tensor with all referenced shards present.
- A parallel-saved tiny-laguna checkpoint loads via `LagunaForCausalLM.from_pretrained` with the expected HF-format (per-expert) key layout.
- Full e2e: `tests/integration/test_reverse_text.py` and `tests/integration/test_reverse_text_lora.py` pass (7 passed in 11m) — FFT and LoRA RL runs incl. resume, with the expected checkpoint layouts (FFT: safetensors + assets; LoRA: `adapter_model.safetensors` + `adapter_config.json` + assets).

---

> [!NOTE]
> **High Risk**
> Changes core checkpoint and filesystem weight-broadcast I/O, including when HF snapshots are written and LoRA layout. Mis-saves or incomplete shards would break resume, evals, and serving.
> 
> **Overview**
> HF weight checkpoints and filesystem broadcasts no longer gather the full model onto rank 0. Each rank keeps a **layer-aligned CPU slice**, converts it to HF format locally, and writes safetensors shards in parallel (`gather_weights_parallel` / `save_state_dict_parallel`). Layout is inferred: full runs always write sharded safetensors; LoRA runs write only the adapter (`adapter_model.safetensors` + `adapter_config.json`). Torch `.bin` and knobs like `save_sharded`, `save_format`, `save_adapter_separately`, and `skip_gather_master_weights` are gone.
> 
> Weight snapshots are **decoupled from DCP interval**. They land at eval steps (intervals copied onto a private `_eval_intervals` field) plus optional `ckpt.weights.interval`. No evals and no interval means no HF weights; `ckpt.weights_only` without a trigger now errors. Online evals load LoRA adapters under a fixed `eval-adapter` name.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a533c01ad900a327bbeaabaea4760f0352ef063c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -441,14 +441,8 @@ class SignSGDConfig(BaseOptimizerConfig):
 
 
 class WeightCheckpointConfig(BaseConfig):
```
mikasenghaas, line 443: can we not completely remove this?

mikasenghaas: Restored `save_sharded` in 049fffea8 — `False` gathers on the master (resurrected `gather_weights_on_master`) and writes a single safetensors file; the default stays the rank-parallel sharded writer. Left the other two removals as-is: the torch format had no consumers and `save_adapter_separately` is superseded by the LoRA layout inference — say the word if you want either back too. The filesystem broadcast transport stays always-sharded (its only consumer is the engines' reload, where sharded is strictly better).

src/prime_rl/trainer/ckpt.py
```diff
@@ -445,43 +413,35 @@ def save(
...
+                save_state_dict(lora_state_dict, step_path, save_sharded=False, adapter=True)
+                if self.lora_config:
+                    save_lora_config(
+                        model,
+                        step_path,
+                        rank=self.lora_config.rank,
+                        alpha=self.lora_config.alpha,
+                        dropout=self.lora_config.dropout,
+                    )
+                self._save_model_assets(step_path, model, tokenizer, processor)
```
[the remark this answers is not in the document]
mikasenghaas: Fixed in d8ca1883a: `maybe_run_evals` now detects the adapter layout (`adapter_config.json` in the step dir), loads it via `update_weights(..., lora_name=...)` under a fixed adapter name, and switches the pool + policy model name to it — mirroring the orchestrator's `WeightWatcher`. The reload-with-same-name-per-step path is the one the server already supports (forced in-place reload in `/load_lora_adapter`). Caveat: no integration test covers SFT-LoRA + online evals, so this is logic-mirrored from the watcher rather than e2e-verified.

src/prime_rl/trainer/weights.py
```diff
@@ -92,43 +83,148 @@ def save_state_dict(
...
+            if value.is_cuda:
+                buffer = _staging_buffers.get(fqn)
+                if buffer is None or buffer.shape != value.shape or buffer.dtype != value.dtype:
+                    buffer = torch.empty_like(value, device="cpu", pin_memory=True)
+                    _staging_buffers[fqn] = buffer
+                buffer.copy_(value, non_blocking=True)
+                partial[fqn] = buffer
+            else:
+                partial[fqn] = value
+        torch.cuda.synchronize()
```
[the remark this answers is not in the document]
mikasenghaas: False positive — this is the standard safe pattern under CUDA stream semantics. The pinned `copy_` is enqueued on the current stream before the loop drops the reference to the GPU temp. Freeing a CUDA tensor is host-side only: the caching allocator returns the block to its pool tagged with the allocation stream, and a later allocation that reuses it is issued on the same stream, so any kernel writing to it is stream-ordered *after* the already-enqueued D2H read. A cross-stream write without sync would be the only hazard, and `full_tensor` goes through functional collectives, which order the NCCL stream against the current stream on both sides (and ProcessGroupNCCL protects the buffers it touches). Holding refs until the `synchronize()` would keep every gathered full tensor resident on GPU at once, which is exactly the peak this loop avoids.

Empirically the `--check` mode compares against the blocking master gather bitwise: identical for tiny-laguna (EP=2), Qwen3-0.6B, and Qwen3-8B (16.4GB, 399 tensors — thousands of allocator-block reuses across the loop).

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
mikasenghaas: Misread your comment the first time (restored the knob — reverted in fd0793166). Now done as intended in 148bdceab: `WeightCheckpointConfig` and the `ckpt.weights` field are gone entirely. Weight checkpoints are written whenever checkpointing is on; `ckpt.skip_gather_master_weights` remains the opt-out, and layout stays fully inferred (full runs → rank-parallel sharded safetensors, LoRA → adapter only). The SFT online-eval validator now only checks skip_gather_master_weights. Rebased on your adapter-loading fix.

garrett361:
> The torch (.bin) weight serialization format is gone from save_state_dict.

❤️

garrett361:
>  Laguna-XS.2 (33.4B, 68.1GB bf16), random init, 2x RTX PRO 6000 Blackwell

> The gather+D2H phase is ~9x faster steady-state

I was confused how you could get a ~9x speedup via parallelization when only using 2 gpus (since parallel writes are only ~2x max, naively), but it seems that it's the non-blocking writes to pinned memory that are the primary factor here. Avoiding all the cuda syncs from non-pinned transfers is the big win.

docs/training.md
```diff
@@ -204,7 +204,7 @@ num_train_gpus = 1  # trainer
...
+The launcher starts the inference server, one env server per eval source, and an `evals` process next to the trainer. The handoff is the filesystem, not NCCL: the trainer writes an HF weight checkpoint at every step an eval env is due (the only trigger for weight saves), and the evals process watches `weights/step_{n}`, points the inference server at each stable checkpoint (`/update_weights` reload from disk), and runs the due envs against it — sequentially per checkpoint, so every epoch measures exactly one policy version. The base model is evaluated before the first step (disable with `eval.skip_first_step`), and the final checkpoint always fires every env. In-flight eval episodes are sized by the same adaptive concurrency controller as the orchestrator; bound it with `[eval.concurrency]` (`min_inflight` / `max_inflight`; set them equal for fixed concurrency).
```
garrett361, line 207: > the only trigger for weight saves

Is this desirable behavior for everyone? I could imagine users wanting decoupled checkpointing and eval intervals. `ckpt_interval < eval_interval` to protect against losing progress but not needing to pay the eval time cost at each save point

wdyt @samsja ?

garrett361: nvm, I think I misunderstood this and we're still saving dcp ckpts at a decoupled rate. This PR touches a separate code path for HF ckpts for eval, IIUC

mikasenghaas:
Superseded by the stacked pair #3315 (rank-parallel sharded save for the filesystem broadcast) + #3316 (remove weight checkpointing entirely: DCP-only + offline `scripts/dcp_to_hf.py` + SFT weight broadcasts). This branch stays as reference — the benchmark script lives at 929e43c37.

Issue #3310, opened 2026-08-19
Consider removing TARGET_LAG now that off-policyness includes in-queue staleness


mikasenghaas:
## Context

`TARGET_LAG = 1` is hardcoded in the orchestrator. It drives two mechanisms:

1. **Dispatch gate** (`update_dispatch_gate`): pause dispatching new rollouts when the in-flight batch leads the trainer by more than `TARGET_LAG`.
2. **Ship-time hold**: ship batch `N` only after the trainer has published `v{N-1-TARGET_LAG}`.

See #2812 for the discussion on why the value is hardcoded and not user-exposed.

## Why `TARGET_LAG` existed

`TARGET_LAG` (formerly `async_level`) protects against the degenerate case where inference step time << trainer step time. Without the gate, the orchestrator would step arbitrarily far away from the trainer.

That protection was necessary because the old off-policy definition only measured **in-flight** staleness: how many weight updates happened while a rollout was generating. A rollout that finished quickly and then sat in the sink buffer accumulated no measured staleness. So nothing bounded the runaway.

## What changed

Since #3050, `off_policy_steps` measures true **consumption lag**: the version the batch trains on minus the version that generated the rollout, queue time included. In other words, we now capture both in-flight and in-queue staleness (borrowing Applied Compute's framing).

With this definition the orchestrator can no longer silently run away. If it steps ahead of the trainer, in-queue staleness grows and the `max_off_policy_steps` sweep drops the stale groups. The staleness bound now binds where `TARGET_LAG` used to.

## Proposal

Investigate removing `TARGET_LAG`. The orchestrator would constantly try to saturate `max_inflight_rollouts`, and `max_off_policy_steps` (plus the other effective-batch filters) would bound staleness after the fact.

## Open questions

- **Wasted compute vs. gate stalls**: pausing at lag 1 is cheaper than generating rollouts that the staleness sweep later drops. Removal trades gate stalls for oversampling + drops. We should measure which wins in practice.
- **Liveness**: the ship-time hold has a second job — it keeps the orchestrator (and its weight watcher) alive for every pending NCCL/NIXL broadcast the trainer will make (see the comment at the hold, added in #3050). Removing `TARGET_LAG` must preserve this, e.g. by deriving the hold from broadcast bookkeeping instead of a lag constant.
- **Backpressure**: how this interacts with the `out_q` backpressure mechanism needs to be understood.

Credit to Garrett Goon for raising this in a Slack discussion.

PR #3297, opened 2026-08-18
refactor(orchestrator): score native episodes

hallerite:
## Summary

Companion: PrimeIntellect-ai/verifiers#2409, merged and published in `verifiers==0.3.1.dev59`; the submodule is pinned to release commit `4bcb48e55`.

Makes credit assignment and admission operate on native `vf.Episode` objects. The verifier graph is the scored artifact; `TrainingSample` is now only the final trainer transport compiled after admission.

- Algorithms receive episodes, annotating verifier nodes with advantages, reference logprobs, and named loss weights.
- Task samplers observe every finalized episode group; admission gates inspect the same scored episodes and compose with AND.
- The train sink scores the episode, scores the group, evaluates admission, and only then compiles admitted traces into trainer samples. `PreparedTrace` and `PreparedGroup` are no longer needed.
- The dispatcher stamps the full task, environment, structured `GroupInfo`, actual dispatch step, train/eval work kind, and policy span directly on the episode.
- Returned task key and content hash are validated against the dispatched task; class names are not compared because the request protocol transports task data, not its Python class.
- Unexpected dispatcher or watcher termination is propagated to the orchestrator instead of leaving the main loop waiting on an output queue that can no longer advance.
- Prime requires environment, group, and work provenance instead of repairing or defaulting malformed episodes.
- Curriculum checkpoints must exactly match the sampler, gates, and configured environments; incompatible state fails immediately.
- Exact sampled trace ids drive batching and metrics, including partial groups and post-batch zero-advantage filtering.
- Rollout JSONL is exactly one native episode per line, preserving multi-trace and trace-less failure episodes.
- Difficulty-pool sampling continues to use the sampler's `observe()` hook, while `AdvRangeGate` reads the native advantages written by the algorithm.

## Public hooks

```python
class Algorithm:
    async def score_episode(self, episode: vf.Episode) -> None: ...
    async def score_group(self, episodes: list[vf.Episode]) -> None: ...

class TaskSampler(Iterator[vf.Task]):
    def __next__(self) -> vf.Task: ...
    def observe(self, group: list[vf.Episode]) -> None: ...
    def state_dict(self) -> dict[str, Any]: ...
    def load_state_dict(self, state_dict: dict[str, Any]) -> None: ...

class AdmissionGate:
    def admit(self, group: list[vf.Episode]) -> bool: ...
    def state_dict(self) -> dict[str, Any]: ...
    def load_state_dict(self, state_dict: dict[str, Any]) -> None: ...
```

## Validation

[...]

PR #3313, opened 2026-08-20
feat: drop on in-queue staleness + first-class cancellation type

mikasenghaas:
## Summary

Refactors staleness handling around the exact consumption-lag definition (#3050's framing), toward #3310.

- **Exact measurement**: `episode_staleness()` decomposes each trained episode's staleness from its `PolicySpan` into the in-flight share (`end - start`, weight updates during generation) and the in-queue remainder (buffered between completion and ship). New per-step metrics: `off_policy/{mean,max}`, `off_policy/{in_flight,in_queue}/{mean,max}`, `off_policy/dropped`; the step log line's `Max Off-Policy` uses the exact definition. Every quality metric in the step log and the wandb `effective` subset is computed over exactly the traces shipped to the trainer this step (`batch.cohort`); only `Error`/`Cancelled`/`Ratio` are arrival-window rates, since over the shipped set they are 0/0/share-of-shipped by construction.
- **In-queue enforcement**: previously only in-flight episodes were aged and dropped — an episode that finished quickly and sat in the sink accumulated staleness that was measured at ship but never enforced. Now the sink sweeps `pending_batch` before every readiness check and voids anything past `max_off_policy_steps`, so the bound is a hard guarantee on trained data. The dispatcher's version-boundary sweep remains as a compute saver (it must run before the engine pause for KV-abort safety) and uses the same exact step-based formula.
- **Tail semantics**: the trainer now broadcasts every version except `v{max_steps}` (the only one nothing can sample from), instead of skipping the final `TARGET_LAG+1`. The orchestrator stays alive for the last rendezvous (`wait_for_final_broadcast`), so `max_off_policy_steps` holds exactly on every step including the final batch, and the tail special-casing this used to need is gone (the dispatch gate's final-batch exception and the watcher's shortened ModelExpress window). Draining now starts right after shipping the final batch — the old trigger needed a further batch to *fill*, which burned a batch of inference and, under a tight `max_off_policy_steps`, never filled at all (end-of-run livelock).
- **First-class cancellation**: dropped groups emit one typed `Cancellation(group_id, count, reason)` on `out_q` instead of per-episode synthetic `Cancelled`-error episodes; queue-voided episodes are tracked in a `cancelled` id-set on the episode collections, alongside `sampled`/`admitted`. Cancellations no longer confound error metrics: `has_error` stays pure (overload cuts and staleness drops are no longer counted as errors), the effective subset excludes cancelled episodes explicitly, and the step log / wandb gain separate `Cancelled` rates. A stale-voided group also bypasses curriculum admission — a pipeline decision is not a task result.

## Breaking

- **wandb**: new `train/*/all/**/cancelled/mean` and per-step `off_policy/*` keys; cancellations no longer appear under `*/error/Cancelled` or inflate `has_error/mean`. `dispatcher/off_policy_level/{mean,max}` → `dispatcher/off_policy/{mean,max}`, now reporting the exact step-based staleness of in-flight work instead of the aged in-flight counter.
- **Zero-output abort accounting**: cancelled groups still advance the zero-output budget (by their full owed episode count), but stale-voided groups no longer feed the curriculum via `on_result`.
- **Trainer broadcast schedule**: in-memory transports now broadcast `v{max_steps-1}` (previously skipped); the orchestrator waits for it before tearing down. The final drain no longer requires (or generates) a throwaway post-final batch.
- **The bound's meaning tightened**: staleness now counts queue time, not just weight updates during generation. A config tuned under the old in-flight-only definition drops more aggressively at the same numeric value; `max_off_policy_steps = 0` now means strictly synchronous (heavy oversampling drops), where the old `0` only meant "no update mid-generation".

## Verification

- Behavioral smoke (stubbed envs, no GPU): stale `Cancellation` voids arrived episodes without reading as error; a fully-cancelled group still advances the zero-output budget by its owed count; overload cancellation finalizes the partial group and its survivors still train; the queue sweep un-fills a batch, marks the episode cancelled, and counts drops; the dispatcher drops exactly the groups past the bound and emits one typed `Cancellation`; `effective` excludes cancelled while `has_error`/`cancelled` stay disjoint.
- `tests/unit/orchestrator`, `tests/unit/test_configs.py`, `tests/unit/transports`: 201 passed (the `test_qwen3_vl_e2e` failure is pre-existing on `main`).
- `ruff check` and `ruff format --check` clean.
- E2E: 15-step reverse-text run with `--orchestrator.max-off-policy-steps 0` (true on-policy), clean exit in 3m 21s. **All 15 step lines report `Max Off-Policy 0` and `Error 0.0%`** — including the final batch, which now trains on the freshly-broadcast `v{max_steps-1}`. Both enforcement layers fired throughout (45 queue-sweep drops, 14 in-flight cancel sweeps) with per-step `Cancelled` up to 70% and zero genuine rollout failures — e.g. `Step 2 | … | Max Off-Policy 0 | Error 0.0% | Cancelled 70.3% | …`; under the old error-marker representation those cancellations would have read as `Error ~70%`. Draining started immediately at the final ship (`Shipped the final batch — draining pipeline`). Reward climbed 0.14 → ~0.77.
- E2E (filesystem broadcast): 8-step reverse-text run with `--weight-broadcast.type filesystem --orchestrator.max-off-policy-steps 0`, clean exit for both processes. All 8 steps report `Max Off-Policy 0` and `Error 0.0%` — including the final batch, since filesystem writes every version (no skip path) and the final-broadcast wait is correctly inert (0 occurrences). Drain fired at the final ship; the trainer wrote all broadcast dirs and finished cleanly. NIXL is safe by construction (trainer skip window and the watcher's ModelExpress guard both derive from `max_steps` and moved in lockstep) but has not been exercised end-to-end.


mikasenghaas:
On hold until #3297 (preserve native episodes) lands — it reworks the same sink/dispatcher paths, so this will be rebased on top of it rather than merged first. Keeping as draft in the meantime.

docs/training.md
```diff
@@ -59,7 +59,7 @@ A condensed view of the knobs you'll most often tune. For trainer-side paralleli
 |---|---|
 | `orchestrator.batch_size` | Tasks per trainer step. |
 | `orchestrator.group_size` | Rollouts generated per task. |
-| `orchestrator.max_off_policy_steps` | How many distinct policies may have contributed to one rollout before it's discarded (default 8). The main off-policy dial on long agentic rollouts — bump for throughput, lower for tighter on-policyness. Watch `errored_rollouts` and `mismatch_kl/all/mean` when tuning. |
+| `orchestrator.max_off_policy_steps` | Maximum staleness of a trained rollout (default 8): the version a batch trains on minus the version that generated the rollout, queue time included. Groups past the bound are dropped. The main off-policy dial on long agentic rollouts — bump for throughput, lower for tighter on-policyness. Watch `off_policy/*` and `mismatch_kl/all/mean` when tuning. |
```
mikasenghaas: >Groups past the bound are dropped

only episodes past max staleness should drop. not the whole group, so we still get partial groups 

oh acc is this the same, given that finished episodes get stale anywas bc of in-queue now?

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -530,7 +530,7 @@ class OrchestratorConfig(BaseConfig):
     """Maximum training steps. If None, runs indefinitely."""
 
     max_off_policy_steps: int = Field(8, ge=0)
-    """Maximum policies allowed to generate a single rollout. Rollouts generated more than ``max_off_policy_steps`` ahead of training are discarded. Higher values yield better throughput at the cost of off-policy noise."""
+    """Maximum staleness of a trained rollout: the version a batch trains on (v{step-1}) minus the version that generated the rollout, queue time included. Groups past the bound are dropped, in-flight and queued. Higher values yield better throughput at the cost of off-policy noise."""
```
mikasenghaas: add "the _oldest_ version that generated the rollout" since it can be multiple

src/prime_rl/orchestrator/dispatcher.py
```diff
@@ -199,9 +201,9 @@ def __init__(
         # Bounded so the dispatcher backpressures on a slow sink (unbounded
         # when no hard ceiling is configured — the dynamic cap still bounds
         # in-flight work). One entry per episode — the sinks count episodes,
-        # never loose traces.
+        # never loose traces — plus one ``Cancellation`` per dropped group.
         maxsize = max(8, max_inflight_ceiling) if max_inflight_ceiling is not None else 0
-        self.out_q: asyncio.Queue[vf.Episode[Any, Any, Any]] = asyncio.Queue(maxsize=maxsize)
+        self.out_q: asyncio.Queue[vf.Episode[Any, Any, Any] | Cancellation] = asyncio.Queue(maxsize=maxsize)
```
mikasenghaas: can we do `vf.Episode[Any, Any, Any]` -> `vf.WireEpisode`?

src/prime_rl/orchestrator/dispatcher.py
```diff
@@ -777,11 +739,12 @@ async def cancel_inflight_train_episodes(self) -> int:
...
         """Instantaneous, read-only gauges sampled by the periodic logger."""
+        staleness = self.inflight_staleness()
         return {
             "dispatcher/inflight/train": float(self.inflight_train_count),
             "dispatcher/inflight/eval": float(self.inflight_eval_count),
             "dispatcher/queued/eval": float(self.queued_eval_examples),
             "dispatcher/mode": float(self.mode == DispatcherMode.PREFER_EVAL),
-            "dispatcher/off_policy_level/max": float(self.max_off_policy_level),
-            "dispatcher/off_policy_level/mean": self.mean_off_policy_level,
+            "dispatcher/off_policy_level/max": float(max(staleness, default=0)),
```
mikasenghaas: drop the _level suffix here

src/prime_rl/orchestrator/utils.py
```diff
@@ -48,6 +48,27 @@ def eval_work(episode: vf.Episode[Any, Any, Any]) -> vf.EvalWorkInfo:
...
+
+def episode_staleness(episode: vf.Episode[Any, Any, Any], training_step: int) -> tuple[int, int, int]:
+    """``(total, in_flight, in_queue)`` staleness of one train episode when
+    consumed by batch ``training_step``: the version the batch trains on
+    (v{step-1}) minus the version that generated the episode. ``in_flight``
+    is the span's share (weight updates during generation); ``in_queue`` is
+    time spent buffered between completion and ship. Frozen-sourced episodes
+    (no policy span) are never stale."""
+    policy = train_work(episode).policy
+    if policy is None:
```
mikasenghaas, line 65: when would this fire? dont we set this?

src/prime_rl/orchestrator/utils.py
```diff
@@ -48,6 +48,27 @@ def eval_work(episode: vf.Episode[Any, Any, Any]) -> vf.EvalWorkInfo:
...
+        return 0, 0, 0
+    total = max(0, (training_step - 1) - policy.start)
+    in_flight = min(total, max(0, policy.end - policy.start))
```
mikasenghaas, line 68: lets cmopute in_queue explicitly so its clear what this returns

docs/training.md
mikasenghaas: Yes — it's the same thing. Every episode of a group shares one dispatch version (`GroupState.policy_version_at_start` stamps all members' `PolicySpan.start`), and staleness is `(step-1) - start`, so a group's episodes cross the bound at the same instant: episode-level and group-level dropping coincide for staleness. Partial groups still happen, just not from staleness — overload cuts finalize the arrived survivors, which do train. Reworded the docstring and docs row to say episodes are what drop and why groups age out together (c39c5e0d0).

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: Done in c39c5e0d0 — "minus the oldest version that generated the rollout (a rollout can span several weight updates)". That's also literally what the code reads: `PolicySpan.start`.

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: Yes — `env.run` returns `vf.WireEpisode` and the synthetic error episodes are `WireEpisode` too, so the whole dispatcher path is now typed as `vf.WireEpisode` (out_q, handle_completed_episode, emit_episode, _validate_episode_task). Done in c39c5e0d0.

src/prime_rl/orchestrator/dispatcher.py
mikasenghaas: Done in c39c5e0d0 — `dispatcher/off_policy/{mean,max}`; skill table and the PR's Breaking bullet updated.

src/prime_rl/orchestrator/utils.py
mikasenghaas: It fires for frozen-sourced train episodes: when an env's sampler doesn't follow the live policy, `emit_episode` stamps `policy=None` (dispatcher.py — `vf.PolicySpan(...) if live_policy else None`), since a frozen sampler never goes stale. Eval episodes always carry a span, and live train episodes always do too — so None is exactly the frozen case the docstring's last sentence describes.

src/prime_rl/orchestrator/utils.py
mikasenghaas: Done in c39c5e0d0 — `in_queue = total - in_flight` computed explicitly before the return.

samsja (approved):

PR #3315, opened 2026-08-20
feat: rank-parallel sharded save for filesystem broadcast

mikasenghaas:
## Summary

- Replace the master-only save in the filesystem weight broadcast (gather everything to rank 0, rank 0 writes all safetensors shards) with a rank-parallel save.
- `gather_weights_parallel`: all ranks still participate in the per-tensor `full_tensor()` all-gathers (they are collective and materialize the full tensor on every rank anyway), but each rank now keeps a layer-aligned slice on CPU instead of throwing it away. The copies are plain blocking D2H transfers — no pinned staging, no stream-ordering assumptions; the speedup comes from splitting the D2H traffic and, above all, the shard writes across ranks.
- Slices are layer-aligned so each rank can run the prime→HF conversion on its own slice (conversion ops only stack/unstack tensors within one layer, and all ops tolerate missing keys). `convert_state_dict_to_hf` detects the format from the model's full key set so partial dicts convert like full ones.
- `save_state_dict_parallel`: each rank splits its slice into ≤5GB shards, writes them under temporary names, shard maps are exchanged via `all_gather_object`, every rank renames its own files into the global `model-XXXXX-of-XXXXX.safetensors` numbering (non-POSIX shared filesystems can't reliably rename another rank's unpublished writes), and master writes the index after the barrier — so a complete index still implies complete shards, and `STABLE` semantics are unchanged.
- The broadcast is always sharded safetensors: the `save_sharded` / `save_format` fields on `FileSystemWeightBroadcastConfig` are removed. The adapter-only LoRA broadcast is unchanged.
- Weight checkpoints (`WeightCheckpointManager`) are untouched here; a stacked PR removes them entirely.

## Breaking

- `weight_broadcast.save_sharded` and `weight_broadcast.save_format` (`FileSystemWeightBroadcastConfig`) are removed. The filesystem broadcast always writes sharded safetensors. Migration: delete the fields from TOMLs.

## Benchmark

Laguna-XS.2 (33.4B, 68.1GB bf16), random init, 2x RTX PRO 6000 Blackwell, single local NVMe:

| impl | gather | convert | write | fsync | total |
|---|---|---|---|---|---|
| master (main) | 20.5–21.5s | 0.1s | 46.7–48.8s | 6.9–9.3s | ~75–79s |
| parallel | 20.8–21.2s | 0.1s | 20.7s | 8.1–8.7s | ~50s |

With blocking copies the gather is a wash on 2 GPUs (each rank copies half the bytes, serialized with the all-gathers); the win is the parallel write. An earlier revision staged the copies through pinned buffers with non-blocking transfers (gather 25s → 2.7s), but its safety rested on same-stream allocator-reuse ordering — dropped in favor of the obviously-correct blocking copies.

The same benchmark with `--nproc-per-node 8` on an H200 node, Laguna-S-2.1 (117.6B, 235.1GB bf16) — one save, 2 timed reps, `--fsync`. Measured on the earlier pinned-staging revision; the write path (the dominant term) is unchanged, the gather column will now sit between 5.4s and the master's 139s (each rank copies 1/8th of the bytes, blocking):

| target FS | master (main path) | parallel (this PR) | speedup |
|---|---|---|---|
| NFS | 660.6s (gather 139s, write 522s @ 0.45 GB/s) | 89.4s (gather 5.4s, write 84s @ 2.80 GB/s) | 7.4× |
| beegfs | 462.1s (gather 139s, write 323s @ 0.73 GB/s) | 75.4s (gather 5.4s, write 70s @ 3.4 GB/s) | 6.1× |

At this scale the parallel writers turn the shared-FS write from the dominant cost into ~70–85s; peak GPU memory is identical between impls (56.5 GiB). With 8 same-node writers the beegfs edge over NFS is modest — multi-node saves should widen it since writers scale with per-node bandwidth.

The benchmark script is not part of this diff — it lives at 929e43c37 (on the reference branch of #3312) for anyone reproducing the numbers: run it with `--nproc-per-node <n> --fsync`.

## Verification

- The benchmark's `--check` mode passes on a tiny 4-layer laguna config with EP=2 (exercises stacked-expert unstacking and expert DTensors), on `Qwen/Qwen3-0.6B` (exercises the `revert_weight_conversion` path and tied embeddings), and on `Qwen/Qwen3-8B` (16.4GB, 399 tensors): the parallel path produces bitwise-identical tensors to the master-gather path, and the index covers every tensor with all referenced shards present.
- `tests/integration/test_reverse_text.py` and `test_reverse_text_lora.py` pass on the equivalent change (vLLM consumes every per-step broadcast, including the LoRA adapter path).

Supersedes #3312 (re-derived from its first commits as the base of a 2-PR stack).

---

> [!NOTE]
> **Medium Risk**
> Changes a distributed gather/write path on shared filesystems (barriers, tmp-rename, index completeness) and drops broadcast config fields. Weight checkpoints and LoRA adapter-only saves are unchanged.
> 
> **Overview**
> Filesystem weight broadcast no longer gathers the full HF checkpoint on rank 0 and writes all shards there. Each rank keeps a **layer-aligned** CPU slice, converts it to HF format, and writes its own safetensors shards.
> 
> `gather_weights_parallel` still runs collective `full_tensor()` all-gathers, then copies only owned tensors to CPU. `save_state_dict_parallel` writes temp shards, `all_gather`s the map, and each rank renames **its own** files into global `model-XXXXX-of-XXXXX.safetensors` numbering so non-POSIX shared FS stays reliable. Master writes the index after a barrier; `STABLE` semantics are unchanged.
> 
> **Breaking:** `FileSystemWeightBroadcastConfig.save_sharded` / `save_format` are removed — broadcast is always sharded safetensors. Adapter-only LoRA broadcast still uses the master-only path. Weight checkpoints are not changed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db4ba53a943f4d77483851f5abb357d5133444a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/trainer/weights.py
+++ b/src/prime_rl/trainer/weights.py
@@ -1,5 +1,4 @@
 import json
-import re
 import warnings
 from pathlib import Path
 from typing import Literal, cast
@@ -141,20 +140,28 @@ def resolve_fqn(model: nn.Module, key: str) -> str:
     return next(iter(fqns))
 
 
-_LAYER_RE = re.compile(r"^(.*?\blayers\.\d+)\.")
+def _layer_prefixes(model: nn.Module) -> list[str]:
+    """FQN prefixes of the model's decoder blocks: the children of every ``layers``
+    ModuleList. state-dict keys and module names come from the same tree traversal,
+    so the prefixes match keys verbatim, wrapper components and all."""
+    prefixes: list[str] = []
+    for name, module in model.named_modules():
+        if isinstance(module, nn.ModuleList) and name.rsplit(".", 1)[-1] == "layers":
+            prefixes.extend(f"{name}.{i}." for i in range(len(module)))
+    return prefixes
 
 
-def partition_weights(state_dict: dict[str, Tensor], world_size: int, dtype: torch.dtype) -> dict[str, int]:
+def partition_weights(model: nn.Module, world_size: int, dtype: torch.dtype) -> dict[str, int]:
     """Assign each state-dict key an owner rank, balanced by byte size.
 
     All keys of one decoder layer share an owner, so per-rank prime->HF conversion
     (which stacks/unstacks tensors within a layer) always sees complete layers.
     """
+    prefixes = _layer_prefixes(model)
     key_to_unit: dict[str, str] = {}
     unit_bytes: dict[str, int] = {}
-    for key, value in state_dict.items():
-        match = _LAYER_RE.match(key)
-        unit = match.group(1) if match else ""
+    for key, value in model.state_dict().items():
+        unit = next((prefix for prefix in prefixes if key.startswith(prefix)), "")
         key_to_unit[key] = unit
         itemsize = dtype.itemsize if isinstance(value, DTensor) else value.element_size()
         unit_bytes[unit] = unit_bytes.get(unit, 0) + value.numel() * itemsize
@@ -177,7 +184,7 @@ def gather_weights_parallel(model: nn.Module, dtype: torch.dtype = torch.bfloat1
     plain blocking transfers; no stream-ordering assumptions.
     """
     world = get_world()
-    owners = partition_weights(model.state_dict(), world.world_size, dtype)
+    owners = partition_weights(model, world.world_size, dtype)
     partial: dict[str, Tensor] = {}
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", category=FutureWarning, module="torch.distributed")
--- a/src/prime_rl/trainer/weights.py
+++ b/src/prime_rl/trainer/weights.py
@@ -212,6 +212,12 @@ def save_state_dict_parallel(state_dict: dict[str, Tensor], save_dir: Path) -> N
     logger = get_logger()
     world = get_world()
 
+    # Master-only mkdir + barrier: concurrent mkdir from every rank can re-raise
+    # FileExistsError on a parallel FS (EEXIST + stale is_dir()).
+    if world.is_master:
+        save_dir.mkdir(parents=True, exist_ok=True)
+    dist.barrier()
+
     tmp_pattern = f"tmp-rank{world.rank}{{suffix}}.safetensors"
     split = split_torch_state_dict_into_shards(state_dict, filename_pattern=tmp_pattern)
     local_shards: list[tuple[str, dict[str, int]]] = []
--- a/src/prime_rl/trainer/weights.py
+++ b/src/prime_rl/trainer/weights.py
@@ -117,13 +117,25 @@ def convert_state_dict_to_hf(model: nn.Module, state_dict: dict[str, Tensor]) ->
     """
     full_keys = dict.fromkeys(resolve_fqn(model, key) for key in model.state_dict().keys())
     if isinstance(model, PreTrainedModelPrimeRL) and model.is_prime_state_dict(full_keys):
+        # PrimeRL custom model holding weights in prime format: apply the model's
+        # declarative prime->HF conversion chain (renames, expert stack/unstack).
         return model.convert_to_hf(state_dict)
-    from transformers.core_model_loading import revert_weight_conversion
+    else:
+        # Plain transformers model: undo the key renames transformers applied when
+        # it loaded the HF checkpoint.
+        from transformers.core_model_loading import revert_weight_conversion
 
-    return revert_weight_conversion(model, state_dict)
+        return revert_weight_conversion(model, state_dict)
 
 
 def resolve_fqn(model: nn.Module, key: str) -> str:
+    """Resolve a state-dict key to the parameter's canonical fully-qualified name.
+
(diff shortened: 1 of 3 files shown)


src/prime_rl/trainer/weights.py
```diff
@@ -105,6 +109,150 @@ def save_state_dict(
...
+            if owners[key] != world.rank:
+                continue
+            fqn = resolve_fqn(model, key)
+            if value.is_cuda:
+                buffer = _staging_buffers.get(fqn)
+                if buffer is None or buffer.shape != value.shape or buffer.dtype != value.dtype:
+                    buffer = torch.empty_like(value, device="cpu", pin_memory=True)
+                    _staging_buffers[fqn] = buffer
+                buffer.copy_(value, non_blocking=True)
+                partial[fqn] = buffer
```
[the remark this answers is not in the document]
mikasenghaas: nop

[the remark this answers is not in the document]
mikasenghaas: False positive — this is the standard safe pattern under CUDA stream semantics. The pinned `copy_` is enqueued on the current stream before the loop drops the GPU temp. Freeing a CUDA tensor is host-side only: the caching allocator returns the block to its pool tagged with the allocation stream, and only a later allocation on that same stream can reuse it — any kernel writing into the reused block is stream-ordered *after* the already-enqueued D2H read. The only real hazard would be a cross-stream write without sync (what `record_stream` exists for), and `full_tensor` goes through functional collectives, which order the NCCL stream against the current stream on both sides (ProcessGroupNCCL additionally protects the buffers it touches). The host side never reads the pinned buffers before the `torch.cuda.synchronize()`.

Empirically: the benchmark's `--check` mode compares this gather bitwise against the fully blocking master gather — identical on tiny-laguna (EP=2), Qwen3-0.6B, and Qwen3-8B (399 tensors, thousands of allocator-block reuses) — and 18 integration tests had vLLM consume every broadcast with rewards in range.

Caveat worth preserving: the safety depends on everything staying on the current stream. If the copies ever move to a side stream, they need `record_stream`/ref-holding — noting that invariant in a comment on the copy site.

src/prime_rl/trainer/weights.py
```diff
@@ -105,6 +109,156 @@ def save_state_dict(
...
+
+    Format detection uses the model's full key set, so a partial dict (one rank's
+    slice) converts the same way as the full state dict would.
+    """
+    full_keys = dict.fromkeys(resolve_fqn(model, key) for key in model.state_dict().keys())
+    if isinstance(model, PreTrainedModelPrimeRL) and model.is_prime_state_dict(full_keys):
+        return model.convert_to_hf(state_dict)
+    from transformers.core_model_loading import revert_weight_conversion
+
+    return revert_weight_conversion(model, state_dict)
```
samsja: not super clear to me here what are the two if statement doing, we should have a clear 

if a
else B

and have comment what A and B are (A:  prime custom model, B transfrmer ?)

src/prime_rl/trainer/weights.py
```diff
@@ -105,6 +109,156 @@ def save_state_dict(
...
+
+
+def resolve_fqn(model: nn.Module, key: str) -> str:
```
samsja, line 131: need docstring for this type of function with example of what its doing

src/prime_rl/trainer/weights.py
mikasenghaas: Agreed — restructured in ccd1a6be0 into an explicit if/else with a comment per branch: the first is a PrimeRL custom model holding weights in prime format (declarative prime→HF conversion chain: renames, expert stack/unstack), the else is a plain transformers model (undo the key renames transformers applied when loading the HF checkpoint).

src/prime_rl/trainer/weights.py
mikasenghaas: Added in ccd1a6be0 — docstring with an example: `model._orig_mod.layers.0.self_attn.q_proj.weight` (torch.compile wrapper) resolves to `model.layers.0.self_attn.q_proj.weight`, the name the tensor has in the HF checkpoint.

src/prime_rl/trainer/weights.py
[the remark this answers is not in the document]
mikasenghaas: Now moot: 9d9f7467d drops the pinned non-blocking staging entirely — the gather does plain blocking `value.to("cpu")` per owned slice, so there is no window between the copy and the GPU temp being freed and no stream-ordering reliance. (For the record, the previous pattern was safe under same-stream allocator-reuse semantics and passed bitwise `--check` at three model scales, but the subtlety wasn't worth carrying — the speedup at scale comes from the parallel shard writes, which are unchanged.)

samsja (approved):

PR #3322, opened 2026-08-20
fix: guard kernel wheels on published releases

mikasenghaas:
## Summary

- `Build Kernels`' release upload refuses to replace a wheel that already exists on a **published** release. Kernel builds are not byte-reproducible, so a re-upload changes the asset's sha256 and breaks every `uv.lock` that pins it — every `uv sync --locked` (all CI jobs and users) fails with a hash mismatch until main is re-locked.
- Deliberate replacement is opt-in via a new `overwrite` dispatch input, and the job summary now prints each wheel's sha256 alongside the `[tool.uv.sources]` snippet so re-locking in the same motion is easy.
- Draft releases stay clobber-safe: nothing can pin their assets yet, and `tag-and-release.yaml` uploads before promoting the draft, so the release flow is unaffected.
- `skills/kernels/SKILL.md`'s backfill section documents the guard and the `overwrite=true` escape hatch.

## Verification

Not exercised in CI (the guard only triggers on a `workflow_dispatch` against a published release with pre-existing assets). The failure mode it prevents happened today: a manual v0.8.0 backfill re-uploaded `prime_kernels-0.1.0+cu128torch2.11.0-cp312-cp312-linux_x86_64.whl` minutes after #3154 merged with `uv.lock` pinning the previous upload's hash (`aab0776e…` pinned vs `6ce9fc5a…` served), breaking `uv sync --locked` everywhere.

### DIFF
--- a/.github/workflows/build_kernels.yaml
+++ b/.github/workflows/build_kernels.yaml
@@ -33,6 +33,11 @@ on:
         description: "Ref to build the kernels from. Empty = the ref the workflow was started on."
         required: false
         type: string
+      overwrite:
+        description: "Replace wheels that already exist on a published release. Breaks every uv.lock pinning their hash — re-lock in the same motion."
+        required: false
+        type: boolean
+        default: false
   workflow_call:
     inputs:
       release_tag:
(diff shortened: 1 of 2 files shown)


mikasenghaas:
Closing — the immediate breakage was re-locked on main (#3321) and the overwrite-guard approach will be handled separately.

PR #3326, opened 2026-08-21
Don't compute the logprobs and their gradients for tokens no loss com…

MarioSieg:
The loss math for itself it uncganed, default loss fn still runs elementwise of the full packed sequence and mult with a mask:

pg_loss = keep_mask * advantages * importance
kl_loss = loss_mask*log_importanance^2
loss = sum(per_token_loss)
^^^^^ those are elementwise ops on a vector with len seq with no vocab dim, skipping them doesn't seem to save much
what this PR skips is the matmul
when we do
hidden_i = XV -> logits -> logprob_i -> loss
the matmul XV  is of dims   [N, 4096] @ [4096, 151936] which is big
so we don't compute the logprobs and their grads for tokens no loss component reads since the loss can't run a token without its logprob - it's basically same as <<don't compute the loss on masked tokens>> but we save where the FLOPS actually are spent which is the matmul not the elementwise ops

bench results on qwen 30b:
  ┌────────────────────┬───────┬───────┬───────┐                                                                                                                                                                                                                  
  │                    │  on   │  off  │   Δ   │                                                                                                                                                                                                                  
  ├────────────────────┼───────┼───────┼───────┤                                                                                                                                                                                                                  
  │ throughput (tok/s) │ 5,462 │ 5,178 │ +5.5% │                                                                                                                                                                                                                  
  ├────────────────────┼───────┼───────┼───────┤                                                                                                                                                                                                                  
  │ MFU (%)            │ 0.7   │ 0.7   │ +5.5% │                                                                                                                                                                                                                  
  ├────────────────────┼───────┼───────┼───────┤                                                                                                                                                                                                                  
  │ peak memory (GiB)  │ 58.5  │ 58.5  │ +0.0% │                                                                                                                                                                                                                  
  └────────────────────┴───────┴───────┴───────┘
  Step-1 loss/grad_norm match within noise (verified against an on-vs-on and off-vs-off control: the on-vs-off gap sits inside the natural run-to-run spread, not outside it — EP all-to-all reduction makes grad_norm inherently noisier   
  than on dense models, ~5e-3 vs ~3e-5 relative).       

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Changes default training behavior and LM-head numerics path (mask compaction + zero fill on skipped tokens); correctness is covered by unit tests but distributed RL/SFT parity should be watched on mixed-mask batches.
> 
> **Overview**
> Adds **skip masked LM head tokens** (default **on**) so the fused LM head runs the expensive vocab matmul only on positions any loss component actually uses—RL `loss_mask`, CE weights, and ref-KL weights—after the usual next-token shift. The backbone still processes the full sequence; skipped positions get zero logprobs/entropy so importance-ratio math stays stable.
> 
> Implements this via a `keep_mask` on `forward` and fused/Gemma heads: compact → chunked logprob/entropy → expand back. Vanilla LM head ignores the mask; RL disables skipping when the fused head is off or `enable_token_export` needs every token. **Default `fused_lm_head_token_chunk_size` is raised to 8192**; docs describe the new knob and **`perf/lm_head_token_fraction`** logging in RL and SFT.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d1b2e56a181c6ab911f2000b743506cb7bb886ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


docs/scaling.md
```diff
@@ -143,15 +143,28 @@ State-only optimizer offload remains enabled by default with `model.optim_cpu_of
...
+fused_lm_head_token_chunk_size = 8192       # default
 # fused_lm_head_token_chunk_size = "disabled"  # vanilla LM head
 ```
 
-Drop the chunk size further when peak memory is still tight (e.g. with very long sequences); raise it to amortize kernel-launch overhead. SFT training silently disables this (not supported yet). Only available with `model.impl = "custom"`.
+Drop the chunk size when peak memory is still tight (e.g. with very long sequences); raise it to amortize kernel-launch overhead. Only available with `model.impl = "custom"`.
+
+### Skipping Masked Tokens in the LM Head
+
+Most of a packed batch is tokens no loss component reads: prompts, environment observations, and pack padding. The backbone still has to run on them — attention needs them in context — but the LM head does not, because it scores each token independently of the others. `skip_masked_lm_head_tokens` (on by default) gathers the hidden states of the tokens some component actually reads, runs the head only on those, and scatters the results back:
```
garrett361: nit: phrasing is a little confusing here, imo

* `Most of a packed batch is tokens no loss component reads:`. ~The loss computations typically only depends on a small subset of tokens. 
* `backbone` not defined, and also maybe an unnecessary detail to get into this
* Not sure the gather/scatter technical detail is needed, also.

Your call, though.

docs/scaling.md
```diff
@@ -143,15 +143,28 @@ State-only optimizer offload remains enabled by default with `model.optim_cpu_of
...
+
+```toml
+[trainer.model]
+skip_masked_lm_head_tokens = true   # default
+```
+
+Head time and its activation memory then scale with the kept fraction, which `perf/lm_head_token_fraction` reports each step. The head is roughly 10% of a dense 8B step and a much larger share when the vocabulary is big relative to the backbone, so the end-to-end win is largest for small models, long prompts, and observation-heavy agentic rollouts.
+
+Every value the loss reads is unchanged — a token's logprob does not depend on which other tokens the head was given. Gradients move by the same amount that changing `fused_lm_head_token_chunk_size` does, since both reshuffle the order of the same bf16 chunk accumulation. Requires the fused head; ignored with `fused_lm_head_token_chunk_size = "disabled"`, and with `enable_token_export`, which records logprobs for every token of a sequence.
```
garrett361: another nit: I think you could cut the first sentences in this paragraph entirely and move the cfg conditions to the first paragraph, since they're good to know up front

docs/scaling.md
```diff
@@ -143,15 +143,28 @@ State-only optimizer offload remains enabled by default with `model.optim_cpu_of
...
+### Skipping Masked Tokens in the LM Head
+
+Most of a packed batch is tokens no loss component reads: prompts, environment observations, and pack padding. The backbone still has to run on them — attention needs them in context — but the LM head does not, because it scores each token independently of the others. `skip_masked_lm_head_tokens` (on by default) gathers the hidden states of the tokens some component actually reads, runs the head only on those, and scatters the results back:
+
+```toml
+[trainer.model]
+skip_masked_lm_head_tokens = true   # default
+```
+
+Head time and its activation memory then scale with the kept fraction, which `perf/lm_head_token_fraction` reports each step. The head is roughly 10% of a dense 8B step and a much larger share when the vocabulary is big relative to the backbone, so the end-to-end win is largest for small models, long prompts, and observation-heavy agentic rollouts.
```
garrett361, line 165: Nice, but I'm surprised we don't already report the percent of tokens which are masked out of the loss. That's something we should be generally reporting, IMO, regardless of whether `skip_masked_lm_head_tokens` is true or not.

I can add that at some point. CC @mikasenghaas

src/prime_rl/trainer/models/layers/lm_head.py
```diff
@@ -34,6 +34,28 @@ def _float_and_contiguous(tensor: Tensor | None) -> Tensor | None:
     )
 
 
+def lm_head_keep_index(keep_mask: Tensor) -> Tensor | None:
+    """Flattened indices of the LM head tokens which must actually be used"""
+    index = keep_mask.reshape(-1).nonzero(as_tuple=True)[0]
```
garrett361: The `.nonzero` induces a CUDA sync, right? As a future optimization, I think we should be able to avoid this since the number of non-zero elements should be known CPU side already from the dataloader, IIUC. Can revisit in a future PR.

src/prime_rl/trainer/rl/loss.py
```diff
@@ -63,12 +63,13 @@ def compute_entropy(shifted_logits: Float[Tensor, "batch seq vocab"]) -> Float[T
     return entropy
 
 
-def shift_tensor_left(t: Float[Tensor, "batch seq"]) -> Float[Tensor, "batch seq"]:
+def shift_tensor_left(t: Tensor) -> Tensor:
```
garrett361: Curious why the typing changed?

docs/scaling.md
garrett361: Oh, nvm, we're reporting this metric unconditionally here

bench.sh
garrett361, line 1: I'm assuming this will be removed before merging.

garrett361 (dismissed):
Very good. Mostly small comments, which you can address or not per your judgement. But before merging, please:
* Fix CI
* Remove `bench.sh`

PR #3332, opened 2026-08-21
fix(sft): resumed runs restart the data stream with num_workers > 0

mikasenghaas:
## Summary

Since #3299 made `data.num_workers` default to 1, **every resumed SFT run silently restarts the training data from the beginning** and freezes the data position it writes into subsequent checkpoints.

Root cause: `train()` called `iter(dataloader)` before the resume path's `ckpt_manager.load(..., dataloader=dataloader)`. With worker processes, `iter()` forks the worker with a copy of the dataset in its *initial* state; the later `load_state_dict` only stashes state for a future iterator that is never created. Two symptoms:

- the resumed run consumes the dataset from row 0 again (visible as a loss-slope kink at every resume — repeated data);
- `state_dict()` keeps returning the stashed loaded state, so every post-resume checkpoint stores the resume-point position instead of the live one, compounding across restarts.

Fix: create the iterator only after the potential resume load, with a comment stating the constraint.

Also: `get_dataset_state` unpacked exactly one worker snapshot, crashing **any** run with `num_workers > 1` at the startup log line (start or resume) — it now returns positions keyed per worker, and stops dumping `pending_sample` token buffers into the log. It keeps parsing the loader's stashed state rather than the dataset object: after a resume the restored position exists only there (the main-process dataset object stays at position 0; workers receive the state on fork). This is now the single code path: `num_workers = 0` (the only case with a different, inline state layout) is no longer supported.

## Breaking

- `data.num_workers = 0` is rejected at config time (`ge=1`). Inline main-process data loading is gone; use `num_workers = 1` (the default).

## Verification

A/B on `configs/ci/integration/reverse-text-sft/{start,resume}.toml` (Qwen3-0.6B, 1 node, 2 GPUs, 5 steps + resume to 10), comparing `checkpoints/step_N/trainer/dataloader/rank_0.pt` worker-snapshot dataset state:

| variant | ckpt step_5 | ckpt step_10 (post-resume) |
|---|---|---|
| main, `num_workers=1` | `{'step': 213}` | `{'step': 213}` — frozen, data restarted from row 0 |
| main, `num_workers=2` | — | crashes at startup (`ValueError: too many values to unpack` in `get_dataset_state`) |
| this PR, `num_workers=1` | `{'step': 213}` | `{'step': 407}` — advanced correctly |
| this PR, `num_workers=2` | `{'worker_0': 111, 'worker_1': 111}` | `{'worker_0': 213, 'worker_1': 213}` — per-worker state restored and advanced; workers consume disjoint shards |

Discovered on a production 500-step GLM-4.5-Air SFT run (india): after two crash-resumes, both post-resume checkpoints carried the identical dataset position (`{'step': 6657}` at step 150 *and* step 300) and the loss slope steepened at each resume boundary from re-trained data.

### DIFF
--- a/src/prime_rl/trainer/sft/data.py
+++ b/src/prime_rl/trainer/sft/data.py
@@ -692,5 +692,7 @@ def get_dataset_state(dataloader: StatefulDataLoader) -> dict:
     if "dataset_state" in state:
         return state["dataset_state"]
     worker_snapshots = (state.get("_snapshot") or {}).get("_worker_snapshots") or {}
-    (worker_state,) = worker_snapshots.values()
-    return worker_state["dataset_state"]
+    states = {worker_id: snapshot["dataset_state"] for worker_id, snapshot in sorted(worker_snapshots.items())}
+    if len(states) == 1:
+        return next(iter(states.values()))
+    return states
--- a/src/prime_rl/trainer/sft/data.py
+++ b/src/prime_rl/trainer/sft/data.py
@@ -683,21 +683,17 @@ def setup_dataloader(dataset: StatefulIterableDataset, config: DataConfig) -> St
         batch_size=1,
         collate_fn=cat_collate,
         num_workers=config.num_workers,
-        pin_memory=config.num_workers > 0,
+        pin_memory=True,
     )
 
 
 def get_dataset_state(dataloader: StatefulDataLoader) -> dict:
-    """Dataset position(s) for the startup log, parsed from ``StatefulDataLoader.state_dict()``.
+    """Dataset position per worker for the startup log, parsed from ``StatefulDataLoader.state_dict()``.
 
     The loader is the only source that is correct in every case: after a resume's
     ``load_state_dict`` the restored position exists solely in the loader's stashed
     state (it reaches the dataset copies inside workers when the iterator forks them;
-    the main-process dataset object stays at position zero). The stashed layout uses
-    torchdata's private keys: inline for a fresh loader, per-worker snapshots for a
-    loaded or live ``num_workers > 0`` loader."""
-    state = dataloader.state_dict()
-    if "dataset_state" in state:
-        return state["dataset_state"]["dataset"]
-    snapshots = (state.get("_snapshot") or {}).get("_worker_snapshots") or {}
+    the main-process dataset object stays at position zero). The keys are torchdata's
+    private worker-snapshot layout."""
+    snapshots = dataloader.state_dict()["_snapshot"]["_worker_snapshots"]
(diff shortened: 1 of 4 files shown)


src/prime_rl/trainer/sft/data.py
```diff
@@ -692,5 +692,7 @@ def get_dataset_state(dataloader: StatefulDataLoader) -> dict:
     if "dataset_state" in state:
         return state["dataset_state"]
     worker_snapshots = (state.get("_snapshot") or {}).get("_worker_snapshots") or {}
-    (worker_state,) = worker_snapshots.values()
-    return worker_state["dataset_state"]
+    states = {worker_id: snapshot["dataset_state"] for worker_id, snapshot in sorted(worker_snapshots.items())}
+    if len(states) == 1:
+        return next(iter(states.values()))
+    return states
```
garrett361: Why is this needed?

samsja (dismissed):

src/prime_rl/trainer/sft/data.py
mikasenghaas: good q, checking

garrett361: Actually unclear to me why we need to be accessing these private `_snapshot` and `_worker_snapshot` entries at all?  Beyond your changes

garrett361: nvm, just seems to be for logging and testing, with the fix needed to handle generic `num_workers`

garrett361 (dismissed):
LGTM, great catch. Confirmed with claude that this is a somewhat subtle ordering bug and reloading state before creating the iterator is indeed the key.

garrett361:
I also didn't see a great way to add a unit test for this since the bug is fairly subtle and you'd need to test how and in what order the checkpointing and iterator creation occurs. So you need to either call the whole main training method or pull out that logic but always keep it in sync with the main trainer. Neither of which is great.

src/prime_rl/trainer/sft/data.py
mikasenghaas: Good instinct — I tried exactly that (read `dataloader.dataset.state_dict()` directly, no private keys) and the e2e disproved it: after a resume's `load_state_dict`, the restored position exists only in the loader's stashed state — torchdata applies it to the dataset copies inside workers when the iterator forks them, while the main-process dataset object stays at position 0. The log line would report `step: 0` at every resume while training correctly continues mid-stream (verified: dataset object says 0, checkpoint advances 213→407).

So the parse of the (private-schema) loader state is the only source that's correct in all cases. 4787dec7c keeps it but drops the single-worker special case (always keyed per worker now), stops dumping `pending_sample` tokens into the log, and documents the constraint.

garrett361 (approved):

PR #3333, opened 2026-08-21
fix: meta load order

garrett361:
Fixes the dcp-load/post-meta init order, which could previously clobber persistent buffer values which get stored in the HF checkpoint.

Correct order:
1. meta init
2. move to device w/ `to_empty`
3. init buffers (often zeroing)
4. Load in weights, potentially overwriting buffer values, too

Previously 3 and 4 were reversed, so HF weights were loaded in, potentially including non-trivial buffer values, which could then get zeroed out by the subsequent `_init_buffers_post_meta` call. 

Specifically, this was happening for the Laguna `expert_bias` buffers; see https://github.com/PrimeIntellect-ai/prime-rl/pull/3333/commits/b26b316633aea3bc32fd3bbc2441b455e2654855


S1ro1:
Would be quite interested in seeing if this affects KL mismatch in any case for laguna, do you maybe have a wandb screenshot or smth? But really nice find

garrett361:
> Would be quite interested in seeing if this affects KL mismatch in any case for laguna, do you maybe have a wandb screenshot or smth

I think router replay actually would mask this, right? Because the `expert_bias` only affects routing decisions, and not the actual weights applied to expert outputs.  In which case, it would be more like silently (hopefully subtly) incorrect routing decisions, but no additional KL mismatch.  

But to answer your question @S1ro1 I have not done any real runs of anything yet, so no wandb. Just dinky reverse-text and wordle with ~O(1B) models on my dev machine

S1ro1:
> I think router replay actually would mask this, right? Because the expert_bias only affects routing decisions, and not the actual weights applied to expert outputs. In which case, it would be more like silently (hopefully subtly) incorrect routing decisions, but no additional KL mismatch.

Makes sense, we usually don't run with R3 by default to measure mismatch (to isolate issues), so this should not be a factor

S1ro1 (approved):
lgtm

garrett361:
> we usually don't run with R3 by default to measure mismatch (to isolate issues)

also make sense

samsja (approved):

PR #3334, opened 2026-08-21
refactor(orchestrator): make dispatch contracts explicit

hallerite:
## Summary

- replace the train/eval sources' dictionary payloads with an immutable `TaskRequest` returned by `next_task()`
- require every request's run step and pin it once per group, removing the duplicate optional `eval_step` fields and per-episode step resolution
- rename the group-scoped terminal event to `GroupCancellation` and name the dispatcher queue union `DispatchResult`
- keep source requests separate from mutable `GroupState`, and store eval examples as native `vf.Task` objects end to end

## API

```python
@dataclass(frozen=True)
class TaskRequest:
    env_name: str
    task: vf.Task
    step: int


@dataclass
class GroupCancellation:
    kind: WorkKind
    env_name: str
    group_id: str
    count: int
    reason: CancelReason


DispatchResult: TypeAlias = vf.WireEpisode | GroupCancellation
```

The zero-output accounting fix from #3313 remains unchanged: a freshly inserted group whose traces are all dropped as stale advances the watchdog instead of resetting it.

## Verification

- Ruff format and lint passed, including the commit hooks.
- `tests/unit/orchestrator/test_curriculum.py`: 5 passed.
- Eval-source round-robin, required-step propagation, and changed-module import smoke checks passed.
- `tests/unit/orchestrator`, excluding the unrelated Qwen3-VL renderer-fixture mismatch: 79 passed.

---

> [!NOTE]
> **Medium Risk**
> Touches orchestrator scheduling and step pinning, so a typing/rename mistake could mis-tag train vs eval work. Runtime policy is unchanged.
> 
> **Overview**
> Makes train/eval dispatch payloads typed instead of dicts, and pins the run **step** on the request rather than reconstructing it later.
> 
> Sources now return a frozen `TaskRequest` from `next_task()`. Eval examples stay as native `vf.Task` objects. `GroupState` / `InflightEpisode` carry a single `step` field, dropping the eval-only `eval_step` split and `_work_step`.
> 
> The dispatcher queue union is named `DispatchResult`, and the group-drop marker is renamed to `GroupCancellation`. Behavior is otherwise the same: failed weight reloads still drain the eval queue so leftover work cannot leak into a later epoch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4294fc9703e66b6f0d78447c56aa609bb2684df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/orchestrator/types.py
+++ b/src/prime_rl/orchestrator/types.py
@@ -57,11 +57,11 @@ DispatchResult: TypeAlias = vf.WireEpisode | GroupCancellation
 
 @dataclass(frozen=True)
 class TaskRequest:
-    """A task selected by a train or eval source for dispatch."""
+    """A task selected by a train or eval source with its pinned run step."""
 
     env_name: str
     task: vf.Task
-    eval_step: int | None = None
+    step: int
 
 
 @dataclass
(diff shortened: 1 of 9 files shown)


src/prime_rl/orchestrator/types.py
```diff
@@ -52,6 +52,18 @@ class Cancellation:
...
+DispatchResult: TypeAlias = vf.WireEpisode | GroupCancellation
+
+
+@dataclass(frozen=True)
+class TaskRequest:
+    """A task selected by a train or eval source for dispatch."""
+
+    env_name: str
+    task: vf.Task
+    eval_step: int | None = None
```
mikasenghaas: just step?

mikasenghaas (approved):

PR #3335, opened 2026-08-21
chore: shared weight broadcasts protocol

mikasenghaas:
## Summary

Refactor of the weight-broadcast subsystem. The three transports (filesystem, NCCL, NIXL) now share one protocol instead of three private ones.

- **One handshake for every transport**: `WeightSender.broadcast` is a template method running a four-stage sentinel handshake per version — the trainer resets the step dir and offers (`.sender_ready`), blocks until the consumer acknowledges (`.receiver_ready`), transfers (`.started`), commits (`.finished`), and cleans old dirs. Filesystem does the same handshake as the in-memory transports. The per-attempt dir reset kills the stale-marker resume hazard.
- **`WeightReceiver`** (new): consumer-side counterpart. The weight watcher, the orchestrator startup rendezvous, and the evals process drive the same object instead of hand-rolling per-transport branches. Discovery is uniform (`.sender_ready`), and every offered version must be received — a failed receive fails the run instead of being skipped. This also deletes the NIXL version-cap/cycle-counting hack (markers are versioned on every transport). Each transport file colocates its sender and receiver (`transports/weights/{base,filesystem,nccl,nixl}`); the engine-side byte receiver stays in `inference/vllm/worker`.
- **Consumption-driven sequencing**: broadcast `v{N}` iff something consumes it. `broadcast_final` lives once on a shared `BaseWeightBroadcastConfig`, stamped by the launchers, never user-set (RL: `orchestrator.eval is not None`; SFT: `True`). With evals, the final eval now measures `v{max_steps}` on every transport (previously `v{max_steps-1}` on in-memory but a race on filesystem). Without evals, the final broadcast is skipped on every transport (a handshake nobody joins would block).
- **Fail-fast startup everywhere**: RL and SFT both broadcast the incoming policy before the first step. The transport keeps a single `broadcast` method and makes no assumptions about prior runs' filesystem state; resume hygiene (`prune_broadcasts_beyond`) lives at the trainer call sites.
- **Unified cleaning**: the base class keeps `{step, step-1}`, nothing else — broadcasts are purely transitive, `checkpoints/` owns persistence. Replaces RL's `maybe_clean` and SFT's `clean_older`, and fixes NCCL/NIXL dirs leaking forever.
- **LoRA served under the base model name**: the single adapter shadows the base model (vLLM resolves `lora_requests` before the base-model match). The per-step model-name rewrite, `orchestrator.model.lora.name`, and the evals adapter-sniff dance are gone. The filesystem receiver is artifact-driven: a step dir with `adapter_config.json` hot-swaps the adapter without an engine pause.
- **`InferencePool` split**: `InferenceClient` (data plane: train/eval client, `score`) and `AdminClients` (admin plane: one client per engine, `wait_for_ready`), both under `prime_rl.orchestrator.clients`; data-plane variables are named `clients` everywhere. Fixes the httpx-client leak (`stop()` closed only the scorer) and deletes the `pinned_client` no-op.
- **Bounded handshake on every transport**: the trainer's wait for the receiver raises `TimeoutError` after `weight_broadcast.timeout` (now on the shared config base, so filesystem has it too) instead of hanging forever, with progress warnings every 60s.

- **Explicit transport naming**: `transports.rollouts` → `transports.batch` with `BatchSender`/`BatchReceiver` (the `MicroBatch` payload type, all signatures, and the `rollout_transport` config field are unchanged).

## Breaking

Config and on-disk artifacts only (internal APIs have no external consumers):

- On-disk sentinels replaced: `STARTED`/`STABLE`/`NCCL_READY` → `.sender_ready`/`.receiver_ready`/`.started`/`.finished`, one uniform handshake per version on every transport. Trainer and orchestrator must ship together. A resume over old broadcast dirs re-broadcasts the startup version.
- Filesystem runs in lockstep with its consumer: the trainer blocks until each version is acknowledged (bounded by `weight_broadcast.timeout`), no longer writes versions nobody consumes, and the consumer no longer skips to the newest version.
- The LoRA adapter is served under the **base model name**; the `orchestrator.model.lora.name` config field is removed. External clients that addressed the adapter name must use the base name.
- RL with evals broadcasts `v{max_steps}` and the final eval measures it (metric shift). Without evals, `v{max_steps}` is still skipped on live transports.
- Broadcast-dir retention: exactly the last 2 dirs on every transport — no `keep_interval` multiples are kept (broadcasts are transitive; `checkpoints/` owns persistence). NCCL/NIXL dirs were previously leaked; SFT kept newest-only.
- `client.extra_headers_from_state` config field removed (no readers since v1 clients send `X-Session-ID` themselves).
- Inference `max_loras` default 8 → 1; `max_cpu_loras` override removed.

## Verification

E2E, re-run on the final branch state (post review feedback), all with resume mid-run:

- **FFT + NCCL** (reverse-text, 10 steps → resume to 20): startup broadcast of v0 and v10, all four handshake markers in each step dir, dirs pruned to the last 2, v20 skipped (no evals), reward climbing, clean exits.
- **FFT + filesystem** (10 → 20): full handshake per version, v10/v20 skipped (no evals — uniform with NCCL), resume re-broadcasts v10, last-2 retention.
- **LoRA + filesystem** (CI configs, 15 → 20): reward 0.14 → 0.78 across the resume — the adapter is applied under the base model name; PEFT step dirs hot-swapped without engine pause.
- **LoRA + NCCL**: rejected at config time with the new error message.

Unit suites (`tests/unit/{transports,orchestrator,utils}`, `test_configs.py`) and ruff pass.

---

> [!NOTE]
> **High Risk**
> Touches the trainer–inference weight path, eval rendezvous, and LoRA serving. On-disk markers, broadcast sequencing, and adapter naming are breaking; trainer and orchestrator must ship together.
> 
> **Overview**
> Unifies weight broadcast so filesystem, NCCL, and NIXL all use one lockstep protocol: `WeightSender` offers `.sender_ready`, the consumer acks `.receiver_ready`, then `.started` / `.finished`. A new `WeightReceiver` is the single consumer API for the watcher, orchestrator startup, and evals — every offered version must be received or the run fails (bounded by `weight_broadcast.timeout`).
> 
> Broadcasts only versions something consumes (`broadcast_final` when evals are configured). RL/SFT always fail-fast-broadcast the incoming policy; dirs keep only `{step, step-1}`. LoRA is served under the **base model name** (in-place reload); `orchestrator.model.lora.name` is gone. `InferencePool` splits into `InferenceClient` / `AdminClients`. Rollout transport is renamed `transports.batch`.
> 
> **Breaking:** sentinels `STABLE`/`NCCL_READY` → the four-dot markers; filesystem now blocks until ack; adapter clients must use the base name; `extra_headers_from_state` and `max_cpu_loras` removed; `max_loras` default is 1.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d8baca15aeaa0c4286d6102a47f383b61aef3eb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -379,8 +376,18 @@ class CheckpointConfig(BaseConfig):
 class FileSystemWeightBroadcastConfig(BaseConfig):
     type: Literal["filesystem"] = "filesystem"
 
+    broadcast_final: bool = True
```
mikasenghaas: put this on a shared `WeightBroadcastConfig` and make private since this is not user-configured but launcher stamped

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -543,12 +550,6 @@ def auto_setup_tokenizer(self):
             self.tokenizer.trust_remote_code = self.model.trust_remote_code
         return self
 
-    @model_validator(mode="after")
```
mikasenghaas, line 546: are we sure we dont need this anymore? is this set from vf now?

src/prime_rl/orchestrator/algo/__init__.py
```diff
@@ -51,7 +51,7 @@
 }
 
 
-def build_algorithm(config: AlgoConfig, policy_pool: InferencePool) -> Algorithm:
+def build_algorithm(config: AlgoConfig, policy_pool: InferenceClients) -> Algorithm:
```
mikasenghaas: lets call this `policy_pool` -> `clients`

src/prime_rl/orchestrator/orchestrator.py
```diff
@@ -124,7 +122,8 @@ class Orchestrator:
 
     # Always set by ``setup()``
     tokenizer: PreTrainedTokenizer
-    policy_inference: InferencePool
+    policy_inference: InferenceClients
```
mikasenghaas: just clients here as well. unify across the codebase

src/prime_rl/orchestrator/utils.py
```diff
@@ -3,21 +3,14 @@
...
-import time
 from concurrent.futures import ThreadPoolExecutor
-from pathlib import Path
 from typing import Any
 
 import verifiers.v1 as vf
 
 from prime_rl.configs.orchestrator import OrchestratorConfig
-from prime_rl.utils.client import InferencePool
+from prime_rl.utils.client import EngineAdmin, InferenceClients
```
mikasenghaas: can we refactor this into explicit `InferenceClient` and `AdminClients`?

mikasenghaas: and put it under `prime_rl.orchestrator.clients`

src/prime_rl/transports/weights/nixl/nixl.py
```diff
@@ -68,11 +67,18 @@ class TransferGroupIndex:
 
 
 class NIXLWeightBroadcast(WeightBroadcast):
-    def __init__(self, output_dir: Path, config: NIXLWeightBroadcastConfig, parallel_dims: ParallelDims) -> None:
-        super().__init__(output_dir)
+    requires_live_consumer = True
```
mikasenghaas: make this sentinels all caps

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: Yes - v1 clients send it themselves: `SESSION_ID_HEADER = "X-Session-ID"` in `verifiers/v1/clients/client.py`, attached per rollout from the trace id in `train.py:429` and `eval.py:116`. On main the field has zero readers: the only hits are its definition (`configs/shared.py`) and the `auto_setup_session_headers` validator that populates it. The v0 reader lived in `verifiers/legacy/scripts/eval.py`, outside the orchestrator path.

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
mikasenghaas: Done - single `BaseWeightBroadcastConfig` in `configs.shared`, docstring marks it launcher-stamped/never user-set. True pydantic-private isn't possible here: the field must survive the launcher->component config dump.

src/prime_rl/orchestrator/algo/__init__.py
mikasenghaas: Done - `clients` in the algo layer (ctor param + attribute).

src/prime_rl/orchestrator/orchestrator.py
mikasenghaas: Done - data-plane variables are `clients` everywhere (`policy_pool`/`pool`/`policy_inference` all renamed; dispatcher uses `policy_clients` to distinguish from frozen-source clients; admin plane is `admin_clients`).

src/prime_rl/orchestrator/utils.py
mikasenghaas: Done - `InferenceClient` + `AdminClients` under `prime_rl.orchestrator.clients` (whole module moved out of `utils.client`).

src/prime_rl/transports/weights/nixl/nixl.py
mikasenghaas: Done - `REQUIRES_LIVE_CONSUMER`, and the receiver traits too (`DISCOVERY_MARKER`, `CAN_SKIP_VERSIONS`).

samsja (commented):
this loos very good.

As discuss in dm, I think that we should move away from FS allowing to run ahead while the trainer is writting to disk. We should guarantee fs == nccl behavior, at least for now. This should simplify the code also

src/prime_rl/transports/weights/base.py
```diff
@@ -1,18 +1,90 @@
...
 from prime_rl.utils.logger import get_logger
+from prime_rl.utils.pathing import get_all_ckpt_steps, get_broadcast_dir, get_step_path
+
+# Broadcast-dir sentinels, shared by every transport. ``.started`` marks a
+# broadcast attempt in flight — the consumer of a live transport reacts to it
+# by joining the transfer. ``.finished`` marks a finished broadcast — for
+# filesystem, the weights are fully on disk. ``.receiver_ready`` is the
+# consumer's reply on the NCCL path: the engines are paused and inside the
+# receive RPC, so the trainer may enter the collective.
+STARTED_MARKER = ".started"
```
mikasenghaas, line 25: lets use 
- sender ready
- receiver ready
- started
- finished
each concrete broadcsat should go through all stages (e.g. if fs should do the "handshake" even tho its not strictly necessary)

src/prime_rl/transports/weights/base.py
```diff
@@ -1,18 +1,90 @@
...
-    def __init__(self, output_dir: Path, lora_config: LoRAConfig | None = None):
+    """Trainer-side weight publisher. ``broadcast`` wraps the transport's
+    ``_broadcast`` with the shared sentinel protocol: the master resets the
+    step dir and raises ``.started``, the transport moves the weights, the
+    master raises ``.finished`` and prunes old step dirs."""
+
+    # A live transport transfers directly into a running consumer: the trainer
+    # blocks until the receiver joins, and a version the consumer never
+    # receives strands the trainer inside the transfer.
+    REQUIRES_LIVE_CONSUMER: ClassVar[bool] = False
```
mikasenghaas: as sami said, lets remove thsi, all broadcasts should be the same. its fine for fs to be slow

src/prime_rl/transports/weights/base.py
```diff
@@ -1,18 +1,90 @@
...
+        receiver is ready (see ``NCCLWeightBroadcast._broadcast``)."""
+
+    def _clean(self, step: int) -> None:
+        """Remove old broadcast dirs, keeping ``step``, ``step - 1`` (a lagging
+        consumer may still be reading it) and ``keep_interval`` multiples."""
+        broadcast_dir = get_broadcast_dir(self.output_dir)
+        for old_step in get_all_ckpt_steps(broadcast_dir):
+            if old_step >= step - 1:
+                continue
+            if self.keep_interval and old_step % self.keep_interval == 0:
```
mikasenghaas: why does broadcast have a keep interval? this should not be a thing. we use checkpoints/ to persist run state. broadcast is purely transitive

src/prime_rl/transports/weights/base.py
mikasenghaas: Done - four-stage handshake on every transport: `.sender_ready` (trainer offers, dir reset) → `.receiver_ready` (consumer acks; NCCL with engines paused inside the receive RPC) → `.started` → `.finished`. Filesystem does the full handshake too. E2e-verified on all transports incl. resume.

src/prime_rl/transports/weights/base.py
mikasenghaas: Done - `REQUIRES_LIVE_CONSUMER` is gone, and with it `CAN_SKIP_VERSIONS`, per-transport discovery markers, the fs skip-if-finished at startup, the fs write-every-version exception, and the evals skip-and-drain failure path. Every offered version must be received on every transport; the trainer's wait is bounded by `weight_broadcast.timeout` (now on the shared config base so fs has it too).

src/prime_rl/transports/weights/base.py
mikasenghaas: Done - `keep_interval` removed; `_clean` keeps exactly {step, step-1} on every transport (step-1 because a lagging consumer may still be loading it).

src/prime_rl/transports/weights/__init__.py
```diff
@@ -1,26 +1,47 @@
...
         raise ValueError(f"Invalid weight broadcast type: {config.type}")
+
+
+def setup_weight_receiver(
+    broadcast_dir: Path,
+    config: WeightBroadcastConfig,
+    admin_clients: list[AsyncClient],
+    model_name: str,
+) -> WeightReceiver:
+    receivers = {"filesystem": FileSystemWeightReceiver, "nccl": NCCLWeightReceiver, "nixl": NIXLWeightReceiver}
```
mikasenghaas: use if elif else here

mikasenghaas: Done - if/elif/else, matching setup_weight_sender.

samsja (commented):
the pr looks very good but I don't understand why we do 

 - InferencePool split into InferenceClient (data plane) + AdminClients (per-engine admin plane)  - LoRA adapter served under the base model name 
 - client.extra_headers_from_state sticky-routing config removed (dead)
 - Renames: transports/rollouts → transports/batch, MicroBatchSender/Receiver → BatchSender/Receiver (the batch pipe, unrelated to weights)
 - Small fixes: is_last_step == → >=


can we do this in seperate pr ? I was confused as of why it belongs here

mikasenghaas:
Broken up into a stack of 6 per @samsja's review — each individually reviewable and mergeable bottom-up (merge one, retarget the next to main):

1. #3352 fix: treat a resume past max_steps as the last step
2. #3353 chore: remove dead extra_headers_from_state config
3. #3354 refactor: rename transports.rollouts to transports.batch
4. #3355 refactor: split InferencePool into InferenceClient and AdminClients
5. #3356 feat: serve the LoRA adapter under the base model name
6. #3357 feat: uniform four-stage handshake for weight broadcasts (the core)

The composed stack is tree-identical to this branch (verified via tree hash), so the e2e matrix results here carry over: FFT NCCL 10→20 + resume, FFT filesystem 10→20 + resume, LoRA filesystem 15→20 + resume, LoRA+NCCL config rejection — all green. Keeping this PR open as the integration reference until the stack lands; happy to close it now instead.

mikasenghaas:
Landed as the stack #3352-#3355, #3358 (re-land of #3356), #3357 — main now matches this branch's e2e-verified tree.

PR #3338, opened 2026-08-22
fix: adapt tmux script to auto-generated run dirs

mikasenghaas:
## Summary

- `scripts/tmux.sh` still tailed `<output_dir>/logs/latest/*.log`, but runs now write logs to `<output_dir>/<run_name>/logs/latest` — and run names are auto-generated (`<envs>--<model>--<short-id>`) unless set via `--run.name`. Every pane tailed paths that no longer exist.
- Since the run directory is unknown when the session starts, each pane now polls a glob (`<dir>/logs/latest` and `<dir>/*/logs/latest`), picks the newest match, and starts tailing once a run appears. This makes `-o` accept either an output directory (default `outputs`, auto-detects the newest run) or a specific run directory.
- The pane loops use `command ls` so interactive shell aliases (e.g. `ls=exa`) don't break the newest-run lookup.
- The agent prompt now describes the run-dir layout and gives the discovery command instead of hardcoded log paths.
- Docs (`docs/training.md`) and the `monitor-run` skill are updated to match.

## Verification

Before: all four log panes tailed `outputs/logs/latest/*.log`, which no longer exists — panes stayed empty.

After: launched the session against a live run with an auto-generated run name; all four `Logs` panes and the `SUCCESS` filter pane picked up the run's `logs/latest` and streamed output.

### DIFF
--- a/docs/training.md
+++ b/docs/training.md
@@ -335,15 +335,15 @@ tail -F <run_dir>/logs/latest/inference/router.log # multi-node only
 
 ### Console Output
 
-`scripts/tmux.sh` opens a 4-pane tmux session that follows `trainer.log`, `orchestrator.log`, `inference.log`, and the union of env worker logs. Start it before launching:
+`scripts/tmux.sh` opens a 4-pane tmux session that follows `trainer.log`, `orchestrator.log`, `inference.log`, and the union of env worker logs. The panes wait for the newest run under the output directory (`-o`, default `outputs`) and tail its `logs/latest`, so auto-generated run names work out of the box. Start it before launching:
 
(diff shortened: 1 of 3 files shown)


samsja:
I am not sure to follow the design here, we grep all the uuid ? `outputs/*/logs/latest` why not maitaining a `outputs/latest` instead, or do I missunderstand how its working

mikasenghaas:
i dont like it either but unfo we cannot grep `outputs/latest` bc we write to `outputs/<run-name>/latest` and run name is created auto created by default so we cannot know what to grep upfront

PR #3340, opened 2026-08-22
feat(slurm): launch ModelExpress for NIXL transfers

S1ro1:
## Summary

- start a job-scoped Redis and ModelExpress service when `slurm.launch_modelexpress = true`
- inject the trainer-head service address into trainer, inference, and orchestrator processes, or use the configured host when `slurm.launch_modelexpress = false`
- pin every NIXL process to the active InfiniBand HCA nearest its GPU, so an eight-GPU node uses all eight rails
- add a pinned, idempotent installer for ModelExpress v0.3.0 and Redis 7.4.2
- remove stale job-scoped ModelExpress and Redis processes before managed-service startup

This is the launch prerequisite for the stacked transfer optimization in #3341. It does not change the NIXL transfer protocol itself.

## Design

With `slurm.launch_modelexpress = true`, the generated job removes stale ModelExpress and Redis processes from an interrupted managed launch, then starts both services on the trainer head before launching the three prime-rl components. The resolved ModelExpress address is passed to every component, and service logs stay in the run directory. With `slurm.launch_modelexpress = false`, the launcher performs no service cleanup, starts no service, and uses `weight_broadcast.host` and `weight_broadcast.port` unchanged.

Rail selection remains local to each NIXL process. Unless `UCX_NET_DEVICES` is explicitly configured, startup resolves the process GPU's PCI path, selects the nearest active InfiniBand port, and constrains UCX to that port. TP8 therefore uses one distinct local rail per GPU instead of relying on UCX's cluster-dependent default choice. `UCX_NET_DEVICES` remains an explicit site override.

The installer pins both service versions, verifies their checksums, and reuses valid existing installations. Generated jobs fail directly if a required binary or UCX capability is missing.

## Scope

- ModelExpress provides startup discovery for a fixed Slurm allocation.
- Elastic discovery, autoscaling, and fault tolerance are intentionally out of scope.
- The local service adds two small CPU processes on the trainer head and no GPU allocation.

## Validation

- `bash -n scripts/install_modelexpress.sh`
- ran `scripts/install_modelexpress.sh` twice and verified that the pinned binaries are reused
- rebased onto `main` at `21a4a3245`; rendered and shell-checked managed and external ModelExpress modes for single-node and multi-node NIXL Slurm jobs, including main’s trainer/orchestrator completion markers
- verified stale-service cleanup is present only in managed ModelExpress renders and absent in external-service renders
- Ruff 0.15.14: `uv run ruff format --check` and `uv run ruff check`
- `uv run python -m compileall` on changed Python modules
- `uv run pytest -q tests/unit/test_configs.py` (122 passed)

---

> [!NOTE]
> **Medium Risk**
> Changes multi-node SLURM launch and NIXL UCX/IB selection for weight sync; misconfiguration or rail selection bugs could break or degrade policy updates, but scope is limited to NIXL SLURM paths.
> 
> **Overview**
> Adds **job-scoped ModelExpress + Redis** startup for SLURM RL when `[weight_broadcast] type = "nixl"` and **`slurm.launch_modelexpress`** is true (default). Single- and multi-node sbatch templates verify the **`third_party/ucx`** build, optionally start pinned **modelexpress-server** / **redis-server** (new **`scripts/install_modelexpress.sh`**), wait for readiness, and pass **`--weight-broadcast.host`** into trainer, orchestrator, and the local `rl` launcher. With **`launch_modelexpress = false`**, jobs use configured **`weight_broadcast.host`** / **`port`** and skip managed-service cleanup.
> 
> **NIXL networking** now picks **`UCX_NET_DEVICES`** per process from the **InfiniBand port nearest each GPU** (unless already set), and inference workers **rotate which trainer ranks they pull from** so concurrent pulls spread across rails.
> 
> Docs and the start-run skill describe install steps and the new SLURM flag. **Does not change the NIXL transfer protocol** (prerequisite for stacked transfer work).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d809b1e4d24121690faab3a22b1f7f832de4f53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/entrypoints/rl.py
+++ b/src/prime_rl/entrypoints/rl.py
@@ -464,11 +464,7 @@ def write_slurm_script(config: RLConfig, config_dir: Path, script_path: Path) ->
         if config.weight_broadcast is not None and config.weight_broadcast.type == "nixl"
         else None
     )
-    launch_modelexpress = nixl_broadcast is not None and nixl_broadcast.host in {
-        "localhost",
-        "127.0.0.1",
-        "0.0.0.0",
-    }
+    launch_modelexpress = nixl_broadcast is not None and config.slurm.launch_modelexpress
     modelexpress_vars = {
         "use_nixl_broadcast": nixl_broadcast is not None,
         "launch_modelexpress": launch_modelexpress,
--- a/src/prime_rl/entrypoints/rl.py
+++ b/src/prime_rl/entrypoints/rl.py
@@ -459,9 +459,28 @@ def write_slurm_script(config: RLConfig, config_dir: Path, script_path: Path) ->
     train_env_names = env_server_names(config, "train")
     eval_env_names = env_server_names(config, "eval")
 
+    nixl_broadcast = (
+        config.weight_broadcast
+        if config.weight_broadcast is not None and config.weight_broadcast.type == "nixl"
+        else None
+    )
+    launch_modelexpress = nixl_broadcast is not None and nixl_broadcast.host in {
+        "localhost",
+        "127.0.0.1",
+        "0.0.0.0",
+    }
+    modelexpress_vars = {
+        "use_nixl_broadcast": nixl_broadcast is not None,
+        "launch_modelexpress": launch_modelexpress,
+        "modelexpress_host": nixl_broadcast.host if nixl_broadcast is not None else "",
+        "modelexpress_port": nixl_broadcast.port if nixl_broadcast is not None else 0,
+        "modelexpress_redis_port": 6380 if nixl_broadcast is not None and nixl_broadcast.port == 6379 else 6379,
+    }
+
     if config.deployment.type == "single_node":
         script = template.render(
             **config.slurm.template_vars,
+            **modelexpress_vars,
             config_path=config_dir / RL_CONFIG,
             output_dir=config.run_dir,
             launcher_dir=get_launcher_dir(config.run_dir),
@@ -512,6 +531,7 @@ def write_slurm_script(config: RLConfig, config_dir: Path, script_path: Path) ->
             orchestrator_on_inference=config.deployment.orchestrator_on_inference,
             train_env_names=train_env_names,
             eval_env_names=eval_env_names,
+            **modelexpress_vars,
         )
     else:
         script = template.render(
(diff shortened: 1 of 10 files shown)


src/prime_rl/templates/multi_node_rl.sbatch.j2
```diff
@@ -236,6 +247,83 @@ srun --kill-on-bad-exit=1 bash -s <<'LAUNCH_SH'
...
+
+        MX_METADATA_BACKEND=redis \
+        REDIS_URL="redis://127.0.0.1:$MODEL_EXPRESS_REDIS_PORT" \
+        MX_HEARTBEAT_TIMEOUT_SECS=31536000 \
+            "$MX_BIN/modelexpress-server" \
+            --port "$MODEL_EXPRESS_PORT" \
+            --database-path "$MX_STATE_DIR/models.db" \
+            --cache-directory "$MX_STATE_DIR/cache" \
+            > "$MX_STATE_DIR/server.log" 2>&1 &
+        MODEL_EXPRESS_PID=$!
```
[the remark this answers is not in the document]
S1ro1: Fixed in abb98e2b2. Managed NIXL Slurm launches now remove stale modelexpress-server and redis-server processes before binding the job-scoped ports in both single-node and multi-node templates. External ModelExpress mode does not perform this cleanup.

src/prime_rl/entrypoints/rl.py
```diff
@@ -459,9 +459,24 @@ def write_slurm_script(config: RLConfig, config_dir: Path, script_path: Path) ->
...
+        if config.weight_broadcast is not None and config.weight_broadcast.type == "nixl"
+        else None
+    )
+    launch_modelexpress = nixl_broadcast is not None and config.slurm.launch_modelexpress
+    modelexpress_vars = {
+        "use_nixl_broadcast": nixl_broadcast is not None,
+        "launch_modelexpress": launch_modelexpress,
+        "modelexpress_host": nixl_broadcast.host if nixl_broadcast is not None else "",
+        "modelexpress_port": nixl_broadcast.port if nixl_broadcast is not None else 0,
+        "modelexpress_redis_port": 6380 if nixl_broadcast is not None and nixl_broadcast.port == 6379 else 6379,
```
garrett361, line 473: What's up with these specific `modelexpress_redis_port` ports?

garrett361 (approved):
This looks good, though I should say I'm still learning about NIXL, `UCX*`, etc. so I don't claim the sharpest understanding here

src/prime_rl/transports/weights/nixl/agent.py
vipulSharma18, line 146: interesting. sort of makes sense given this is a training codebase with large size messages. Hm, but yeah, very specific choice. Someone knew what they were doing...

PR #3350, opened 2026-08-23
fix(trainer): distribute routing replay padding across experts

EazyReal:
## Summary

Sequence packing appends loss-masked tokens to align training batches. With router replay enabled, those tokens still pass through MoE dispatch, but their replay payload was filled with zero bytes. Every synthetic route therefore selected expert 0, concentrating padding-only forward work on one expert and potentially one expert-parallel rank.

## Fix

- Read the model's routed-expert count from the same configuration already loaded by the batch packer, including composite text configs.
- Distribute each layer's complete padding route set as full expert cycles plus an evenly spaced remainder, then rotate assignments between layers. The cycle order keeps every row's top-k distinct across cycle boundaries, bounds per-expert count skew to one assignment, and spreads the unavoidable remainder optimally across any contiguous expert-parallel partition.
- Widen compact replay payloads when the model's expert range requires it, preserving all sampled route values. For example, a `uint8` payload for a 512-expert model becomes `uint16` before padding is appended.
- Fail closed before mutating the batch when routing padding is required but the expert count is unavailable.

Non-padding replay routes and loss-mask behavior are unchanged.

## Verification

- `uv run pytest -q tests/unit/orchestrator/test_batch.py`: 33 passed on Python 3.12.
- A 256-expert, top-k 8, one-row padding case assigns one route to each of eight contiguous EP ranks per layer.
- A 256-expert, top-k 2, seven-row padding case produces 14 distinct routes per layer; per-expert counts differ by at most one and the eight-rank histogram is `[1, 1, 2, 2, 2, 2, 2, 2]` up to rank ordering.
- A 256-expert, top-k 8, 33-row padding case crosses a full expert cycle while retaining distinct routes per row, per-expert skew of one, and exactly 33 assignments per EP rank.
- A 512-expert case verifies `uint8` to `uint16` widening, exact preservation of sampled routes, row/layer rotation, and balanced aggregate EP assignment.
- Ruff check, Ruff format check, and `git diff --check` pass for the changed files.

No multi-GPU or DeepEP end-to-end training run was performed; the regression coverage validates the production packing payload and its dispatch-facing invariants on CPU.

### DIFF
--- a/src/prime_rl/orchestrator/packing.py
+++ b/src/prime_rl/orchestrator/packing.py
@@ -1,7 +1,7 @@
 from transformers import AutoConfig
 
 from prime_rl.configs.orchestrator import OrchestratorConfig
-from prime_rl.trainer.batch import build_bin_cost, prepare_batch
+from prime_rl.trainer.batch import build_bin_cost, get_model_num_experts, prepare_batch
 from prime_rl.transports.batch.types import MicroBatch, TrainingSample
 from prime_rl.utils.logger import get_logger
 
@@ -24,6 +24,7 @@ class BatchPacker:
             )
             model_config = None
         self.bin_cost = build_bin_cost(model_config)
+        self.num_experts = get_model_num_experts(model_config)
 
     def pack(self, samples: list[TrainingSample]) -> list[list[MicroBatch]]:
         return prepare_batch(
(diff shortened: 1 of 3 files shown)


mikasenghaas:
uh this is cool, we will check!

MarioSieg:
Amazing - looks good so far :)
Can you rebase onto main and fix the conflicts?

EazyReal:
Done!

PR #3361, opened 2026-08-23
fix(nccl): preserve checkpoint chunk boundaries during vLLM reload

rasdani:
## Summary

- preserve trainer-broadcast state-dict chunk boundaries during checkpoint-format vLLM reloads
- load layer chunks within one layerwise reload lifecycle while retaining the flat stream for kernel-format loading
- avoid duplicating received tensors by yielding views into the NCCL receive buffer
- add regression coverage for framing, reload ordering, and receive-buffer storage sharing

## Failure behavior

Laguna-S-2.1 reaches the initial trainer-to-vLLM policy update, but every TP8 receiver fails while reconstructing state-dict chunk 3. The first failure is the extra tensor copy in the receive path:

```text
tensor = concatenated[offset : offset + numel].view(shape).clone()
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 20.00 MiB. GPU 2 has a total capacity of 79.18 GiB of which 11.94 MiB is free.
```

The update endpoint then reports:

```text
"POST /update_weights HTTP/1.1" 500 Internal Server Error
```

That exception aborts consumption of the current NCCL frame. The automatic retry reuses the already partially consumed stream, so payload bytes are interpreted as the next frame header. The first visible symptom is:

```text
Receiving 54639 layer state dicts
```

Subsequent impossible metadata lengths and very large allocation attempts are secondary stream-desynchronization symptoms, not the initiating error.

## Root cause

The NCCL sender broadcasts checkpoint weights as independently framed state-dict chunks, but the vLLM receiver flattened them into one stream. The receiver also cloned every split tensor, requiring an additional device allocation while the static model and KV cache occupied nearly all GPU memory. That first allocation failure interrupted a frame; retries then consumed payload bytes as metadata headers.

The fix preserves the sender's chunk boundaries through the checkpoint reload path and yields views into the contiguous NCCL receive buffer instead of cloning each tensor. vLLM is initialized once, loads each received chunk within that lifecycle, and finalizes once.

## Testing

- `uv run pytest -q tests/unit/inference tests/unit/transports/test_nccl_broadcast.py` — 17 passed, 1 skipped
- Ruff check and format checks passed
- controlled Laguna-S-2.1 BF16 GPU A/B with one trainer node and one TP8 inference node:
  - pristine `origin/main` at `f53e4d9c`: all ranks failed on the first additional 20 MiB receive clone; `/update_weights` returned HTTP 500 and a retry read the desynchronized `54639` chunk header
  - this PR at `8d2caf78`: all eight ranks received 49/49 chunks and `/update_weights` returned HTTP 200
- post-update Laguna-S-2.1 chat completion stopped normally and returned a coherent, factually correct response

### DIFF
--- a/src/prime_rl/inference/vllm/worker/nccl.py
+++ b/src/prime_rl/inference/vllm/worker/nccl.py
@@ -8,7 +8,7 @@ from vllm.distributed.utils import StatelessProcessGroup
 from vllm.logger import init_logger
 
 from prime_rl.inference.vllm.worker.weight_transfer import (
-    load_weights_checkpoint_layerwise,
+    load_weight_groups_checkpoint_layerwise,
     load_weights_kernel,
     update_mla_absorbed_weights,
 )
@@ -51,7 +51,7 @@ def receive_state_dict(communicator: PyNcclCommunicator) -> Generator[tuple[str,
         # Split concatenated tensor back into individual tensors
         offset = 0
         for key, shape, numel in tensor_info_list:
-            tensor = concatenated[offset : offset + numel].view(shape).clone()
+            tensor = concatenated[offset : offset + numel].view(shape)
             offset += numel
             try:
                 yield key, tensor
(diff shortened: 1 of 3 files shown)


rasdani:
superseded by https://github.com/PrimeIntellect-ai/prime-rl/pull/3365

PR #3373, opened 2026-08-24
fix: stream-ordering crash in SignSGD under FSDP2

mikasenghaas:
## Summary

Fixes the deterministic `CUDA error: an illegal memory access` that killed every SignSGD run without full optimizer offload (crash right after step 1, all attempts, plain and `optim_cpu_offload` paths alike).

- **Root cause**: the pre-broadcast `torch.cuda.empty_cache()` (added in #3234 for FP8 gather headroom) runs immediately after `optimizer.step()` and returns cached blocks to the driver. A kernel still in flight that references a cached block faults once the block is freed under it. SignSGD's per-param loop launches thousands of tiny kernels whose async tail is still running at that point; AdamW's compact foreach step happens to finish in time, which is why only SignSGD crashed. `CUDA_LAUNCH_BLOCKING=1` makes the crash vanish (ordering, not indexing), and the failing stack points at `empty_cache` itself.
- **Fix**: `torch.cuda.synchronize()` before that `empty_cache()` — drain pending work before releasing blocks. Costs ~ms once per step, protects every optimizer.
- **Hardening**: rewrite `SignSGD.step` with foreach ops over local shards, bucketed per (device, dtype) — a few fused launches instead of thousands (`aten._foreach_sign` has no DTensor sharding strategy, hence `to_local()`; every op is pointwise and grads share the param's placement at step time, so the update commutes with sharding). Update semantics are unchanged and verified against the previous formula (decoupled decay `p *= 1 - lr*wd`, then `p -= lr * sign(g)`).

## Verification

Before/after on GLM-4.5-Air scaleswe, 2 trainer + 2 inference nodes, `sign_sgd` + `optim_cpu_offload`, no full offload:

| state | outcome |
|---|---|
| main (per-param SignSGD, no sync) | illegal memory access right after step 1, 2/2 slurm attempts |
| foreach SignSGD alone, no sync | same crash after step 3 (narrowed, not fixed) |
| main + `CUDA_LAUNCH_BLOCKING=1` | no crash — confirms stream-ordering race |
| **this PR (sync + foreach)** | [runs clean](https://wandb.ai/primeintellect/swe-ablations?nw=nwuser&search=sgd-nooffload): 6+ steps and counting through the previously fatal step-1→2 boundary, healthy grad norms and KL |

Also verified on the fake-data debug config (`torchrun --nproc-per-node=2 -m prime_rl.entrypoints.trainer @ configs/debug/fake/rl.toml --optim.type sign_sgd`): 5/5 steps.

### DIFF
--- a/src/prime_rl/trainer/rl/train.py
+++ b/src/prime_rl/trainer/rl/train.py
@@ -84,7 +84,11 @@ def train(config: TrainerConfig):
     # Setup the monitors
     asyncio.run(
         monitors.setup(
-            wandb=config.monitors.wandb, file=config.monitors.file, output_dir=config.output_dir, run_config=config
+            producer="trainer",
+            wandb=config.monitors.wandb,
+            file=config.monitors.file,
+            output_dir=config.output_dir,
+            run_config=config,
         )
     )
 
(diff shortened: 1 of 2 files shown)


mikasenghaas:
Correction: the superseding PR is #3383 (not #3382).

PR #3378, opened 2026-08-24
feat(trainer): push post-meta buffer reinit down to owning layers

garrett361:
Pushes down `init_buffers_post_meta` onto whatever module actually owns the buffer. The top-level `PreTrainedModelPrimeRL.init_buffers_post_meta` method on model classes then recursively calls this method on all buffer-owning submodules, raising an error if any such modules do not implement this method. Effectively introduces a contract that all buffer-owning modules in `prime-rl` should at least attempt to properly initialize their own buffers, making it harder to accidentally leave buffers improperly initialized, which was a historical issue.

- There's an exemption escape hatch (`_init_buffers_post_meta_exempt`) for buffer-owning modules the codebase doesn't control and can't add the hook to. Currently only needed by `qwen3_5`/`qwen3_5_moe`, whose vision-tower rope is an upstream `transformers` class.
- Added a completeness test per architecture, in each model's own test file, that constructs on meta, `to_empty()`s, and asserts every buffer comes out finite — previously only 2 of 12 architectures were ever exercised through this path in CI.

## Verification

- `uv run pytest tests/unit/train/ -q` green (excluding the pre-existing, unrelated `test_nemotron_h*` failures, confirmed via `git stash` comparison against `main`).
- Manually confirmed the completeness check fails loudly (`TypeError`) when a layer's hook is removed, and that the new `random_init` regression test fails before the fix and passes after.


tests/unit/train/models/test_glm4_moe.py
```diff
@@ -124,5 +124,35 @@ def test_glm4_moe() -> None:
...
+        vocab_size=64,
+    )
+    with torch.device("meta"):
+        model = PrimeRLGlm4MoeForCausalLM(config)
+    model.to_empty(device="cuda")
+
+    model.init_buffers_post_meta()
+
+    for name, buffer in model.named_buffers():
+        assert torch.isfinite(buffer).all(), f"buffer {name} is not finite after init_buffers_post_meta"
```
garrett361, line 154: These `torch.isfinite` tests are super weak since in practice meta-init usually populates zeros into the tensor, but they're cheap

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -128,6 +128,9 @@ class Qwen3_5PreTrainedModel(PreTrainedModelPrimeRL, HFQwen3_5PreTrainedModel):
     config_class = Qwen3_5TextConfig
     base_model_prefix = "model"
     supports_gradient_checkpointing = True
+    # Vision-tower rope belongs to upstream transformers' Qwen3_5VisionModel; we can't add
+    # init_buffers_post_meta to it, so it's exempted and reinitialized explicitly below instead.
+    _init_buffers_post_meta_exempt = (Qwen3_5VisionRotaryEmbedding,)
```
samsja: wondering if we shouldn't patch the rotary embedding class instead of rather inherit from it somewhere to override it ? 

or rather can't we just patch the end model that inherit from this one ?

src/prime_rl/trainer/models/qwen3_5_moe/modeling_qwen3_5_moe.py
```diff
@@ -35,6 +38,19 @@
...
+        self.theta ** (torch.arange(0, self.dim, 2, dtype=torch.float32, device=self.inv_freq.device) / self.dim)
+    )
+    self.inv_freq.copy_(inv_freq)
+
+
+# Qwen3_5MoeVisionRotaryEmbedding is upstream transformers code; Qwen3_5MoeVisionModel.__init__
+# hardcodes its construction, so we can't subclass-and-inject like elsewhere. Attach the
+# buffer-init hook to the class directly instead.
+Qwen3_5MoeVisionRotaryEmbedding.init_buffers_post_meta = _init_vision_rope_buffers_post_meta
+
```
garrett361, line 53: CC @samsja ended up just patching in this method on the HF class, as here

src/prime_rl/trainer/models/laguna/modeling_laguna.py
```diff
@@ -82,6 +82,20 @@ def forward(
             sin = emb.sin() * attention_scaling
         return cos.to(dtype=x.dtype), sin.to(dtype=x.dtype)
 
+    def init_buffers_post_meta(self) -> None:
```
S1ro1, line 85: I'm honestly not a big fan of having this model specific code here - would try to think of a way to make this owned by the rope layer itself as well

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -32,6 +32,19 @@
...
+    inv_freq = 1.0 / (
+        self.theta ** (torch.arange(0, self.dim, 2, dtype=torch.float32, device=self.inv_freq.device) / self.dim)
+    )
+    self.inv_freq.copy_(inv_freq)
+
+
+# Qwen3_5VisionRotaryEmbedding is upstream transformers code; Qwen3_5VisionModel.__init__
+# hardcodes its construction, so we can't subclass-and-inject like elsewhere. Attach the
+# buffer-init hook to the class directly instead.
+Qwen3_5VisionRotaryEmbedding.init_buffers_post_meta = _init_vision_rope_buffers_post_meta
```
S1ro1, line 45: We can probably vendor the rotary embeddings ourselves, would do it anyway given the aim to remove transformers so we can do it now instead of this ugly patch

src/prime_rl/trainer/models/qwen3_5_moe/modeling_qwen3_5_moe.py
```diff
@@ -35,6 +38,19 @@
...
+    inv_freq = 1.0 / (
+        self.theta ** (torch.arange(0, self.dim, 2, dtype=torch.float32, device=self.inv_freq.device) / self.dim)
+    )
+    self.inv_freq.copy_(inv_freq)
+
+
+# Qwen3_5MoeVisionRotaryEmbedding is upstream transformers code; Qwen3_5MoeVisionModel.__init__
+# hardcodes its construction, so we can't subclass-and-inject like elsewhere. Attach the
+# buffer-init hook to the class directly instead.
+Qwen3_5MoeVisionRotaryEmbedding.init_buffers_post_meta = _init_vision_rope_buffers_post_meta
```
S1ro1, line 52: dito here, maybe let's vendor

src/prime_rl/trainer/models/base.py
```diff
@@ -1,7 +1,37 @@
+from typing import Protocol, runtime_checkable
+
+import torch.nn as nn
 from torch import Tensor
 from transformers.modeling_utils import PreTrainedModel
 
 
+@runtime_checkable
+class _PostMetaBufferInitModule(Protocol):
```
S1ro1: let's make this PostMetaBufferInitModule, I hate this smell of ai code where it does private classes

src/prime_rl/trainer/models/base.py
```diff
@@ -1,7 +1,37 @@
...
+    def init_buffers_post_meta(self) -> None: ...
+
+
+def _run_init_buffers_post_meta(module: nn.Module, exempt: tuple[type[nn.Module], ...] = ()) -> None:
```
S1ro1: dito here

S1ro1 (commented):
In general I like the idea, some minor nits on cleanup - would prefer to spend some time on this to future proof it, we're in no rush to get this in, so instead make it clean without patches etc

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
garrett361: Agreed, but didn't want to mix concerns here

src/prime_rl/trainer/models/laguna/modeling_laguna.py
garrett361: @S1ro1 this is a method on `LagunaRotaryEmbedding` so this is owned by the rope layer, no?

src/prime_rl/trainer/models/base.py
garrett361: Lol, I'll change it, but pretty sure I asked for private classes and fns in these spots since they're not things I'd expect users to need to use or know about

src/prime_rl/trainer/models/laguna/modeling_laguna.py
S1ro1: Oh i see, i thought it's the parent model, then all good

src/prime_rl/trainer/models/base.py
S1ro1: makes sense, just my aversion, can keep if you think so

garrett361: already changed, no strong feelings either way

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
garrett361: patched, but a better solution is to vendor ourselves, which I'll leave to a separate PR.

PR #3387, opened 2026-08-25
feat(dashboard): agent view commands + report tab with verified citations

eligotts:
## What

Lets a coding agent on the same machine drive the open dashboard and show grounded, clickable trace evidence. The dashboard remains a filesystem reader: commands carry addresses, never trace or report content.
Screenshot 2026-08-25 at 12 08 04 PM
Screenshot 2026-08-25 at 12 07 37 PM
Screenshot 2026-08-25 at 12 07 53 PM

### View-command control plane

- `POST /api/view` accepts `{run, tab, step, kind, subset, episode, line, trace, branch, report, highlight}` and validates supplied addresses against the filesystem. Episode ids resolve to lines; malformed values return 400 and missing resources or indices return 404.
- `GET /api/view/events` fans commands to connected tabs over SSE. Server and browser queues are bounded and commands are applied serially, preserving partial-command semantics without unbounded memory growth.
- `409` means no tab was connected. The command is retained for a newly opened bare dashboard; deep links and reloads keep their own target.
- Commands use the same navigation functions as clicks and install a requested trace address before initial loading, avoiding a needless fetch of the latest step.
- The **follow** toggle turns automatic navigation into a toast with a **go** action.

SSE is intentional: this is a one-way local control channel, `EventSource` reconnects natively, and no WebSocket dependency is needed.

### Report tab and verified citations

- Renders `<run>/reports/*.md`, newest first, using a small Markdown subset over escaped text. Raw HTML is never interpreted; the open report refreshes every five seconds in live mode.
- The dashboard skill creates a report only when the user explicitly requests one. Ordinary trace questions are answered normally, with `/api/view` used when showing evidence is useful.
- Citations are inline `[^id]` markers backed by one-line JSON definitions. Each requires a stable episode id, trace address, verbatim quote, and short note; positional `line` citations are rejected.
- Verification fetches the episode directly by id, independent of the normal 5,000-row list window. Matching is case-sensitive and whitespace-insensitive; repeated quotes remain broken unless node/prefix/suffix/field identifies one occurrence.
- Optional `field: "content" | "reasoning"` disambiguates message fields. Citation navigation defaults to the all-branches view, while an explicit branch must contain the cited node.
- Verified chips navigate to the exact trace and paint the quote. Quote-bearing navigation switches the new Rendered trace view to Messages, where highlights attach to structured message text. Broken or missing citations refuse navigation with an explanation. Citation caches are scoped to one report render so changed reports and rewritten traces cannot reuse stale results.

### Large-run filesystem behavior

- Episode lookup supports a direct `episode` filter, so citation correctness does not depend on pagination or run size.
- Trace-summary sidecars store file size plus a constant-size checkpoint immediately before the observed EOF. Live caches and sidecars invalidate together when a trace file is rewritten larger at the same path, without hashing large files.
- Concurrent scans cannot replace a farther-ahead offsets cache.
- `free_port` uses `SO_REUSEADDR`, matching the server listener so rapid restarts do not drift past a reusable port.

The localhost-default security posture, flat report files, escaped Markdown subset, and SSE transport are retained deliberately. Auth, report-write APIs, a Markdown dependency, or WebSockets would add surface area without improving this local workflow.

### Agent contract and docs

- `skills/dashboard/SKILL.md` keeps only the command and citation details an agent needs: grouped trace addresses, 409 handling, stable episode ids, explicit opt-in report creation, and multiple citations only when a claim genuinely depends on distinct passages.
- `src/prime_rl/dashboard/README.md` and `docs/training.md` describe the fifth dashboard view and local agent workflow.

An intermediate commit added report threads/drafts/save and the following commit reverted it as unnecessary scope. Squash merge is recommended to omit that net-zero experiment from history.

## Verification

- Ruff check and format check on the changed Python files
- JavaScript syntax check
- Dashboard skill validation
- Existing dashboard browser smoke against fresh, uniquely named seeded artifacts
- Browser exercises covering stable-id enforcement, positional-citation rejection, content/reasoning and branch navigation, SSE view commands, highlights, follow mode, large episode sets, live rewrites, and zero console errors
- Direct server probes for rewrite invalidation, 5,001-episode lookup, and view-command validation

---

> [!NOTE]
> **Medium Risk**
> New local APIs and complex trace/citation resolution touch large-run file caching and navigation; scope is localhost filesystem reads only, with validation before navigation, but bugs could mis-route viewers or show false verification on edge-case trace shapes.
> 
> **Overview**
> Adds a **Report** tab (fifth dashboard view) that lists and renders markdown from `<run>/reports/`, with inline `[^id]` citations backed by JSON definitions. The UI verifies each quote against on-disk traces (episode id lookup, whitespace-insensitive matching) and lets verified chips jump into the trace viewer with the passage highlighted.
> 
> Introduces a local **agent control plane**: `POST /api/view` validates on-disk addresses (run, tab, step/kind/subset, episode id → line, trace/branch, report slug, highlight anchors) and fans commands to all tabs via `GET /api/view/events` (SSE). A **follow** toggle applies commands automatically; when off, a toast offers **go**. `409` is returned when no tab is connected but the command is stored for a later bare open.
> 
> Trace infrastructure changes support this workflow: episode list filtering by stable `episode` id (not dependent on the 5k list window), offset/sidecar caches invalidated on file **rewrites** via a small EOF **checkpoint** (not full-file hashing), and `SO_REUSEADDR` when picking a free port for quicker restarts.
> 
> Docs and `skills/dashboard/SKILL.md` document `/api/view`, report authoring (only when the user asks), and citation rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 994d61d0a70353504274d0a9070986486f477442. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


skills/dashboard/SKILL.md
```diff
@@ -51,3 +51,60 @@ A clean exit releases `daemon.json`; a stale file from a dead process is taken
...
+`title` and one-line JSON citation definitions:
+
+```markdown
+---
+title: Why does reward dip at step 4?
+---
+
+The dip is provider errors, not policy regression [^err].
+
+[^err]: {"step": 4, "kind": "train", "subset": "all", "episode": "ep-...", "node": 0, "quote": "engine overloaded", "note": "The failed call that emptied this step's batch."}
```
mikasenghaas, line 93: how are those citations persisted? afaiu the model is asked to produce a .md report + then post a citation, wondering if citations should be a native trace field (e.g. on message node, trace in general, etc.) as well which the models/humans can use so annotations persist outside of the dashboard world?

mikasenghaas (approved):

PR #3411, opened 2026-08-27
refactor(moe): unify runtime and checkpoint layout

S1ro1:
### Model and checkpoint contract

- Replace the separate standard, GPT-OSS, and Nemotron expert/router implementations with one canonical MoE tree.
- Store routed expert weights as separate `gate_proj`, `up_proj`, and `down_proj` parameters. Non-gated experts omit `gate_proj`.
- Nest every custom router at `mlp.router.gate`, routed experts at `mlp.experts`, and optional shared experts inside `mlp`.
- Store the single selection-only router bias at `mlp.router.selection_bias`; model-specific checkpoint names are converted only at the HF boundary.
- Keep shared expert projections canonical. Qwen3.5 alone nests its scalar sigmoid gate at `mlp.shared_expert.output_gate`, with conversion mapping the standalone Hugging Face key at the boundary.
- Use the same activation protocol and `FeedForward` implementation for dense MLPs, shared experts, and grouped experts.
- Make shared-expert construction exclusively parent-owned: model layers pass an explicit `FeedForward`/subclass or `None`, while `MoEArgs` and `MoE` never infer or construct one.
- Keep existing HF source layouts at the conversion boundary. GPT-OSS alone adds the structural split/join required for its interleaved source tensor.
- Require `.prime-v1` in reusable `<conversion_dir>/prime` caches so legacy Prime layouts fail before loading.

### MoE runtime contract

- Use one routed execution sequence for every backend: dispatch tokens, run canonical grouped experts, combine tokens.
- Select grouped expert compute through explicit BF16, DeepGEMM FP8, or MXFP8 implementations. `MoEArgs` remains architecture-only.
- Select local/Torch or DeepEP token dispatch once during model setup. Dispatchers own routing metadata, communication, padding, chunking, and synchronization.
- Keep `TokenDispatcher` as the structural interface while a shared base owns the canonical run/synchronize sequence; local, Torch, and DeepEP implementations own only their dispatch/combine details.
- Keep DeepEP combine events, concatenate streams, and flattened dispatch state owned by each dispatcher; chunking uses the equal input shape shared by EP ranks and does not synthesize empty chunks.
- Ignore dispatch configuration when expert parallelism resolves to one rank; local dispatch remains active and DeepEP has no optimizer side effects.
- Keep grouped expert execution directly in `GroupedExperts.forward` and router confidence accounting directly in `TokenChoiceTopKRouter.forward`.
- Shard every canonical expert parameter with one backend-independent `Shard(0)` placement.
- Keep dense linear quantization, routed-expert compute, and EP transport independent.
- Remove the TorchTitan module hooks, expert decorator, dynamically attached process group, backend attributes, tensor-subclass interception, and process-global MXFP8 patches from the reachable path.
- Vendor the narrow MXFP8 grouped-GEMM and Torch EP transport implementation in [prime-kernels PR #1](https://github.com/PrimeIntellect-ai/prime-kernels/pull/1), preserving the pinned torchao revision and BSD license.

## Public contracts

Checkpoint tree:

- `mlp.router.gate`: `nn.Linear`
- `mlp.router.selection_bias`: optional persistent selection-only buffer
- `mlp.experts.gate_proj`: stacked `nn.Parameter` for gated experts, absent for non-gated experts
- `mlp.experts.up_proj`: stacked `nn.Parameter`
- `mlp.experts.down_proj`: stacked `nn.Parameter`
- `mlp.shared_expert.{gate_proj,up_proj,down_proj}`: `FeedForward` projections when present
- `mlp.shared_expert.output_gate`: Qwen3.5-only scalar `nn.Linear`
- expert matrices: `[num_experts, output_features, input_features]`

Runtime configuration:

```toml
[trainer.model.moe.compute]
type = "bf16"  # bf16 | deepgemm_fp8 | mxfp8

[trainer.model.moe.dispatch]
type = "torch"  # torch | deepep
transport = "bf16"  # bf16 | mxfp8 for torch dispatch
```

DeepEP owns its tuning fields:

```toml
[trainer.model.moe.dispatch]
type = "deepep"
num_sms = 20
token_chunk_size = 4096
```

This is intentionally breaking. `enable_grouped_gemm`, `enable_a2a`, `ep_comm_backend`, `deepep_num_sms`, and `deepep_token_chunk_size` are removed without aliases.

## Validation completed

- Foundation conversion, detector, marker, state-loading, checkpoint, NIXL graph, and wire-format tests passed.
- Focused CPU configuration and MoE contracts: 143 passed.
- Focused MoE runtime configuration parametrizations after the EP=1 no-op change: 6 passed.
- Affected custom-model CUDA suite: 24 passed, 3 skipped across AFMoE, GLM, GPT-OSS, Nemotron, Qwen3/Qwen3.5, and state loading.
- Final H200 model suite after execution-path cleanup: 81 passed, 3 skipped (Slurm 2597).
- Torch EP on two H200 GPUs passed with BF16 and DeepGEMM FP8, top-k 2, both score locations, shared experts, gated/non-gated/bias variants, uneven routing, a rank receiving zero routed tokens, and input/router/expert gradients.
- Final Torch EP BF16 and DeepGEMM FP8 matrix passed after the cleanup (Slurm 2596).
- DeepEP on two H200 GPUs passed with BF16 and DeepGEMM FP8, chunking on/off, both score locations, shared experts, a rank receiving zero routed tokens, and gradients.
- Final DeepEP BF16 and DeepGEMM FP8 matrix passed after the cleanup (Slurm 2595).
- Parent-owned shared-expert construction: 16 focused CPU tests and 64 H200 model tests passed, with 3 expected skips (Slurm 2600).
- Full-graph MoE and shared `FeedForward` compile validation passed after the ownership cleanup (Slurm 2601).
- Local DeepGEMM FP8 passed across all supported topology, activation, and bias combinations.
- `torch.compile(fullgraph=True)` forward/backward passed for the full MoE and shared `FeedForward` across all supported topology/activation combinations.
- Single-rank DCP and Torch EP + FSDP DCP model/optimizer resume passed.
- The `mxfp8_moe` wheel build includes its public module, manifest, and license; the module imports against the pinned torchao build.
- Unsupported H200 MXFP8 selection fails before model mutation.
- Review audit: every queued DeepEP combine event is now synchronized, including chunked execution (two-H200 Slurm 2604).
- Dispatcher review cleanup passed on two H200s: DeepEP BF16/DeepGEMM with chunking and a zero-receive rank (Slurm 2847), plus Torch EP topology/bias/zero-receive coverage (Slurm 2848).
- The exact `reverse_text_moe` integration passed end-to-end on two H200s with an isolated conversion cache (Slurm 2607).
- Conversion ranks now make the create decision on master and validate `.prime-v1` only after the final all-rank barrier.
- Canonical router selection bias: 12 CPU MoE/conversion tests passed; CUDA checkpoint loading, Nemotron, and GLM model tests passed, with the existing GLM BF16 gradient comparison passing on isolated rerun after one threshold-edge miss (Slurm 2851/2852).
- Ruff, Ruff formatting, compileall, and `git diff --check` passed.

## Deliberate exclusions

- MultiLoRA is not adapted to the new runtime and may fail.
- Fused-MoE trainer integration is removed and will be reconsidered only after the canonical runtime is finalized.
- Routed-expert selective activation checkpointing is removed; the remaining selective targets are unchanged.

## Remaining before marking ready

- Run numerical MXFP8 grouped-GEMM and MXFP8 Torch EP transport forward/backward on SM100 hardware. The available Slurm fleet currently has only H200 nodes.
- Merge prime-kernels PR #1, publish wheels containing `mxfp8_moe`, and update the `pyproject.toml`/lock wheel pins. The currently pinned v0.8.0 wheel does not contain this module.

---

> [!NOTE]
> **High Risk**
> This changes the MoE forward/backward path, expert-parallel dispatch, checkpoint key layout, and trainer TOML in breaking ways across all custom MoE models.
> 
> **Overview**
> **Breaking trainer config:** MoE is configured via `[trainer.model.moe.compute]` (`bf16`, `deepgemm_fp8`, `mxfp8`) and `[trainer.model.moe.dispatch]` (`torch` with `transport`, or `deepep` with `num_sms` / `token_chunk_size`). Dense `[trainer.model.quantization]` no longer drives expert GEMMs or EP transport. Removed flags include `ep_comm_backend`, `deepep_*`, `moe_use_grouped_mm`, `moe_fused_kernel`, and quantization `enable_grouped_gemm` / `enable_a2a`.
> 
> **Unified MoE execution:** All custom models share one `MoE` path—`TokenChoiceTopKRouter`, stacked `GroupedExperts` (`gate_proj` / `up_proj` / `down_proj`), and a pluggable `TokenDispatcher` (`LocalTokenDispatcher`, `TorchTokenDispatcher`, `MXFP8TorchTokenDispatcher`, `DeepEPTokenDispatcher`). `configure_moe_runtime` wires grouped GEMM backends and dispatch at setup; EP weight sharding is only `ExpertWeightParallel`. DeepEP dispatch/combine, grouped-GEMM permutation, and chunk pipelining live in the new dispatcher layer instead of ad hoc `MoE` hooks.
> 
> **Checkpoint and model surface:** PrimeRL checkpoints rename routed weights from `w1/w2/w3` to `gate/down/up_proj`, move selection bias to `mlp.router.selection_bias`, and add GPT-OSS HF↔prime conversion. `LatentMoE` / Nemotron-specific expert classes are folded into the canonical tree; dense layers use `FeedForward`. Reusable `prime` conversion caches must include a `.prime-v1` marker.
> 
> **Removed / docs:** Fused `flash_moe` trainer integration and `bench_fused_moe.py` are dropped; `routed_experts` selective activation checkpointing is removed. Docs and examples (e.g. GLM-5) are updated for the new tables.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ffd5be7d48d28b889f9d73c8815ab0eb618b9dc9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/trainer/distributed/deepep.py
```diff
@@ -330,14 +359,137 @@ def combine_tokens(hidden_states: torch.Tensor, state: _DispatchState) -> torch.
...
+        routed_outputs.append(run_pending_chunk(pending_state))
+        if len(routed_outputs) == 1:
+            return routed_outputs[0]
+
+        concatenate_stream = _get_concatenate_stream()
+        with torch.cuda.stream(concatenate_stream):
+            sync_combine()
+            output = torch.cat(routed_outputs, dim=0)
+            self._output_event = concatenate_stream.record_event()
+        return output
```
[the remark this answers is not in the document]
S1ro1: Fixed in 5f6dde00d. DeepEP currently serializes work on its communication stream, so corruption was not established from the overwritten scalar alone, but retaining only one event made that ordering assumption implicit. The runtime now retains and waits every outstanding combine event. Chunked BF16 and DeepGEMM execution passed on two H200s (Slurm 2604).

src/prime_rl/trainer/model.py
```diff
@@ -1002,13 +954,18 @@ def load_dcp_from_hf(model: nn.Module, config: ModelConfig, parallel_dims: Paral
...
+            if cache_exists:
+                _validate_prime_conversion_cache(snapshot_path)
+            elif get_world().is_master:
                 logger.debug(
                     f"Converting snapshot state dict to PrimeRL format and saving to {snapshot_path} on master rank. This is a one-time operation."
                 )
                 snapshot_state_dict = load_state_dict(source_path)
                 model.convert_to_prime(snapshot_state_dict)
                 save_state_dict(snapshot_state_dict, snapshot_path)
+                (snapshot_path / _PRIME_CONVERSION_MARKER).touch()
```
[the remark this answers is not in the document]
S1ro1: Fixed in 5f6dde00d. The existing final all-rank barrier and marker validation already covered the normal conversion path, so the reported deadlock was broader than the code implied. The pre-barrier rank-local exists state was still redundant and potentially inconsistent: only master now creates a missing cache, then every rank validates `.prime-v1` after the final barrier. The exact reverse_text_moe integration passed end-to-end on two H200s (Slurm 2607).

docs/advanced.md
```diff
@@ -35,31 +35,50 @@ impl = "custom"        # or "hf" to force the HF path
...
 ```
 
 GLM-5.2 adds IndexShare: the DSA sparse-attention indexer runs only on a subset of layers and the remaining layers reuse the cached top-k indices. The trainer reads this schedule from the model's `indexer_types` config field and enables the index cache automatically, so no extra config is needed. To override the schedule manually, set `[trainer.model.index_cache]` (`topk_freq` or `topk_pattern`).
 
 ### Expert Parallelism Backends
 
-`model.ep_comm_backend` picks the all-to-all kernel used for EP dispatch/combine:
+`[trainer.model.moe.dispatch]` selects how routed tokens are dispatched and combined:
+
+- **`torch`** (default): torch all-to-all with `transport = "bf16"` or, when MXFP8 expert compute is selected, `transport = "mxfp8"` on SM100.
```
garrett361, line 73: Do we want to be more opinionated and set the `moe.dispatch.type` based on the `moe.compute.type`? Like if we're choosing mxfp8 compute, is there a reason to not choose mxfp8 transport as well?

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -161,21 +160,80 @@ class DebugModelConfig(BaseConfig):
...
+
+
+class DeepEPMoEDispatchConfig(BaseConfig):
+    """Dispatch and combine routed tokens with DeepEP."""
+
+    type: Literal["deepep"] = "deepep"
+    num_sms: int = Field(20, ge=1)
+    """SMs allocated to DeepEP communication kernels."""
+
+    token_chunk_size: int | None = Field(None, ge=1)
```
garrett361, line 220: Choose a reasonable default here?

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -332,19 +378,19 @@ def quantization_only_with_custom_impl(self):
...
+        if isinstance(dispatch, DeepEPMoEDispatchConfig):
+            if isinstance(self.ep, int) and self.ep <= 1:
+                raise ValueError("model.moe.dispatch.type='deepep' requires model.ep > 1.")
+            if isinstance(compute, MXFP8MoEComputeConfig):
+                raise ValueError("MXFP8 expert compute does not support DeepEP dispatch.")
+        elif dispatch.transport == "mxfp8":
+            if not isinstance(compute, MXFP8MoEComputeConfig):
+                raise ValueError("MXFP8 transport requires model.moe.compute.type='mxfp8'.")
+            if isinstance(self.ep, int) and self.ep <= 1:
+                raise ValueError("MXFP8 transport requires model.ep > 1.")
```
garrett361: Would have expected this config to be a no-op when model.ep = 1 rather than erroring when set?

src/prime_rl/trainer/distributed/deepep.py
```diff
@@ -159,11 +172,9 @@ def backward(ctx, grad_combined: torch.Tensor) -> tuple[torch.Tensor, None]:
...
 @torch.compiler.disable()
 def sync_combine() -> None:
-    global _pending_combine_event
-
-    if _pending_combine_event is not None:
-        _pending_combine_event.current_stream_wait()
-        _pending_combine_event = None
+    for event in _pending_combine_events:
+        event.current_stream_wait()
+    _pending_combine_events.clear()
```
garrett361: For my own understanding, why did we move from syncing via a single global event to a list of events?

src/prime_rl/trainer/distributed/deepep.py
```diff
@@ -262,6 +275,16 @@ def dispatch_tokens_async(
     *,
     score_before_experts: bool = True,
 ) -> _PendingDispatchState:
+    num_input_tokens = hidden_states.shape[0]
+    if num_input_tokens == 0:
```
garrett361: when are we dispatching zero tokens? Looks like a debugging path?

src/prime_rl/trainer/distributed/deepep.py
```diff
@@ -330,14 +355,137 @@ def combine_tokens(hidden_states: torch.Tensor, state: _DispatchState) -> torch.
             hidden_states.dtype
         )
     hidden_states = _unpermute_tokens(hidden_states, state.permuted_indices, state.num_recv_tokens)
-    return _DeepEPCombine.apply(hidden_states, state.handle_id)
+    combined = _DeepEPCombine.apply(hidden_states, state.handle_id)
+    return combined[: state.num_input_tokens]
```
garrett361: Why is the slice needed? Would have expected `num_input_tokens` are automatically returned from the `combine`

src/prime_rl/trainer/distributed/deepep.py
```diff
@@ -330,14 +355,137 @@ def combine_tokens(hidden_states: torch.Tensor, state: _DispatchState) -> torch.
...
+
+
+@dataclass(frozen=True)
+class DeepEPDispatchState:
+    backend: _DispatchState
```
garrett361: I find this config a bit confusing. DeepEPDispatchState owns a _DispatchState attr (rather than being a subclass of _DispatchState?) and the owned state is called `backend`?

docs/advanced.md
S1ro1: sometimes it performs bad for some reason, wouldn't be here

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
S1ro1: the default is not-chunked, which performs the best without tuning

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
S1ro1: agreed, will fix

src/prime_rl/trainer/distributed/deepep.py
```diff
@@ -330,14 +355,137 @@ def combine_tokens(hidden_states: torch.Tensor, state: _DispatchState) -> torch.
...
+            torch.distributed.all_reduce(max_tokens, op=torch.distributed.ReduceOp.MAX, group=self.group)
+            max_tokens_per_rank = max_tokens.item()
+            chunk_ranges = [
+                (min(start, x.shape[0]), min(start + self.token_chunk_size, x.shape[0]))
+                for start in range(0, max_tokens_per_rank, self.token_chunk_size)
+            ]
+            if not chunk_ranges:
+                chunk_ranges = [(0, 0)]
+
+        def dispatch_chunk(start: int, end: int) -> _PendingDispatchState:
```
garrett361: We are redefining this `dispatch_chunk` function and `run_pending_chunk` below every time `run` is called. Refactor to avoid this?

src/prime_rl/trainer/distributed/deepep.py
garrett361: Also a little unclear to me why we have a global `_pending_combine_events` list rather than it being an attribute on each `DeepEPTokenDispatcher` instance. I see we export `sync_combine`, but don't seem to actually consume it outside of this file, so I'm missing why it's so decoupled from the dispatcher class

src/prime_rl/trainer/distributed/token_dispatcher.py
```diff
@@ -0,0 +1,341 @@
...
+    permutation: PermutationState
+
+
+@dataclass(frozen=True)
+class TorchDispatchState(LocalDispatchState):
+    input_splits: list[int]
+    output_splits: list[int]
+
+
+class _AllToAll(torch.autograd.Function):
```
garrett361, line 92: Why do we prefer this to the functional collectives version of all-to-all `all_to_all_single_autograd`? Vaguely recall there maybe have been some time where funcol a2a didn't support autograd, but I thought functional collectives are preferred these days

src/prime_rl/trainer/distributed/token_dispatcher.py
```diff
@@ -0,0 +1,341 @@
...
+class PermutationState:
+    input_shape: torch.Size
+    permuted_indices: torch.Tensor
+
+
+@dataclass(frozen=True)
+class LocalDispatchState:
+    num_tokens: int
+    token_indices_experts_sorted: torch.Tensor
+    scores_after_experts: torch.Tensor | None
```
garrett361, line 82: In future work, we can probably simplify this to `scores` and drop all `score_before_experts = True` support. More of a note for myself.

src/prime_rl/trainer/distributed/token_dispatcher.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+def _scatter_routed_output(
+    routed_output: torch.Tensor,
+    *,
+    num_tokens: int,
+    token_indices_experts_sorted: torch.Tensor,
+    scores_after_experts: torch.Tensor | None,
+) -> torch.Tensor:
+    if scores_after_experts is not None:
```
garrett361, line 192: Another optimization note for future work: would like to see if dispatching the scores and fusing the scoring into the swiglu is more benficial than just doing the scoring locally in the combine.

src/prime_rl/trainer/distributed/token_dispatcher.py
```diff
@@ -0,0 +1,341 @@
...
+            score_before_experts=score_before_experts,
+        )
+        routed_output = experts(routed_input, num_tokens_per_expert)
+        return self.combine(routed_output, state)
+
+    def synchronize(self) -> None:
+        return None
+
+
+class TorchTokenDispatcher(LocalTokenDispatcher):
```
garrett361: nit: some odd inheritance patterns here. True comms-dispatching classes being subclasses of `LocalTokenDispatcher` is a bit off semantically, but then also we end up overriding many of the base class's methods entirely in the subclasses and have trivial methods elsewhere (e.g. `LocalTokenDispatcher.sync`. Seems like more of an ABC pattern is expected here

src/prime_rl/trainer/models/layers/grouped_gemm.py
```diff
@@ -0,0 +1,68 @@
+from dataclasses import dataclass
+from types import ModuleType
+from typing import Protocol
+
+import torch
+
+
+class GroupedGemm(Protocol):
```
garrett361, line 8: just noting that I like the use of Protocols here and throughout, think it's a nice pattern

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
S1ro1: Fixed in 505ea8eb0. When EP is disabled or auto-resolves to one rank, dispatch configuration is ignored and runtime setup uses LocalTokenDispatcher. An unused DeepEP setting also no longer disables gradient clipping; the obsolete error assertions were removed.

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -4,428 +4,53 @@
...
-                selected_experts_indices,
-                leaves[4],
-                ctx.num_experts,
-                align_m=_load_fused_moe_kernel().MXFP8_SCALE_BLOCK,
-            )
-        wanted = [leaf for leaf in leaves if leaf.requires_grad]
-        computed = iter(torch.autograd.grad(out, wanted, grad_out) if wanted else ())
-        grads = {p: (next(computed) if leaf.requires_grad else None) for leaf, p in zip(leaves, grad_poses)}
-        return grads[0], grads[1], grads[2], grads[3], None, grads[5], None, None
+    bias = torch.repeat_interleave(bias, num_tokens_per_expert.to(torch.int64), dim=0)
```
garrett361: `repeat_interleave` will cause a cuda sync here, which we should be able to avoid by specifying the `output_size`. Think that should all be known CPU side at launch time

src/prime_rl/trainer/distributed/token_dispatcher.py
S1ro1: This actually gets refactored again in the followup PR, as we need to make it an op so selective AC can see it

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -722,13 +230,29 @@ def forward(
...
+                selection_scores = selection_scores + self.selection_bias
+            if expert_bias is not None:
+                selection_scores = selection_scores + expert_bias
+            _, selected_experts_indices = torch.topk(
+                selection_scores,
+                k=self.top_k,
+                dim=1,
+                sorted=self.topk_sorted,
+            )
+            top_scores = scores.gather(dim=1, index=selected_experts_indices)
```
garrett361, line 250: This all looks a little weird to me. There are scenarios where we maybe adding both an `expert_bias` and a `self.selection_bias` buffer to the `selection_scores`?

Also, I think `selection_bias` is an excellent semantic name, but am concerned that this will mess up checkpoint loading for checkpoints which persist the expert bias buffer. Usually the buffer has a different name than this, and I'm not seeing any code that handles a possible necessary fqn conversion.

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -737,241 +261,105 @@ def forward(
...
-        routed_outputs.append(run_pending_chunk(pending_state))
+    def prepare_expert_input(self, x: torch.Tensor) -> torch.Tensor:
+        return x
 
-        shared_output = self.shared_expert(x) if self.shared_expert is not None else None
-        sync_combine()
-        routed_output = routed_outputs[0] if len(routed_outputs) == 1 else torch.cat(routed_outputs, dim=0)
-        return routed_output if shared_output is None else shared_output + routed_output
+    def prepare_expert_output(self, x: torch.Tensor) -> torch.Tensor:
+        return x
```
garrett361, line 361: This is purely for Nemotron to override?

S1ro1: yes, but I think it'a fairly nice "API surface" to expose, so wouldn't mind it

garrett361 (dismissed):

garrett361 (approved):

PR #3413, opened 2026-08-27
feat(evals): add resumable standalone runs

samsja:
## Summary

- add top-level `[ckpt]` and `[resume]` support to standalone `uv run evals`
- write atomic cursor checkpoints at `<output_dir>/checkpoints/step_N/evals/progress.pt`, where N is the durable eval-source task cursor
- store only the cursor in eval progress checkpoints; generated episode records are not copied into checkpoint payloads
- resume with `--resume`, `--resume.step N`, or `--resume.dir path/to/checkpoints/step_N` by skipping the durable cursor prefix
- retry partial groups and out-of-order completions beyond the durable cursor
- size resumed eval batches to the remaining queued task suffix
- reject checkpointing in online mode, where progress is coupled to the trainer weight-broadcast lifecycle
- keep the training start-run skill current with the new workflow
- clarify that checkpoint writes use temp-file replacement so resume never observes a partial pickle

## Verification

- `uv run pytest tests/unit/evals/test_ckpt.py` — 3 passed
- focused `ruff check` — passed
- focused Python `ruff format --check` — passed
- `git diff --check` — passed

Local note: `uv run evals --help` is currently blocked before CLI loading by the locked `prime-sandboxes==0.2.39` / `connectrpc==0.11.1` import mismatch (`connectrpc.client` is missing).

---

> [!NOTE]
> **Medium Risk**
> Shared orchestrator types (`Dispatcher`, `EvalSource`, `EvalSink`) change, but checkpoint/resume logic is gated to standalone evals with config validation; unit tests cover cursor and I/O semantics.
> 
> **Overview**
> Standalone `uv run evals` can now be made **interruptible** via optional `[ckpt]` and `[resume]` on `EvalsConfig` (rejected when `[online]` is set).
> 
> Progress is a **task cursor** over round-robin eval tasks: `EvalSource` assigns `source_index`, advances the durable cursor only over a fully completed prefix (`mark_completed`), and on resume skips tasks below the loaded cursor. `CheckpointManager` writes only `{"cursor": N}` to `<output_dir>/checkpoints/step_N/evals/progress.pt` using atomic temp-file replace; saves run every `ckpt.interval` completed groups in standalone mode.
> 
> The evals loop loads resume state at setup, sizes epochs with `EvalSink.set_batch_size` from remaining queued tasks, and on each finished eval group calls `dispatcher.pop_source_index` to drive checkpointing. Online eval rollout totals use the triggered queue instead of full env example counts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4dc69fac552a55cb6c7c412a0d4e1b194a39c03. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/evals.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/evals.py
@@ -4,7 +4,7 @@ from pydantic import Field, model_validator
 
 from prime_rl.configs.monitors import MonitorsConfig
 from prime_rl.configs.orchestrator import ConcurrencyConfig, EvalConfig
-from prime_rl.configs.shared import ClientConfig, LogConfig
+from prime_rl.configs.shared import ClientConfig, LogConfig, ResumeConfig
 from prime_rl.configs.trainer import WeightBroadcastConfig
 from prime_rl.utils.config import BaseConfig, default_output_dir
 
@@ -66,6 +66,13 @@ class OnlineConfig(BaseConfig):
     this step."""
 
 
+class CheckpointConfig(BaseConfig):
+    """Checkpoint progress for an interruptible standalone eval run."""
+
+    interval: int = Field(1, ge=1)
+    """Save after every N completed task groups."""
+
+
 class EvalsConfig(BaseConfig):
     """``uv run evals``: run the configured evals against a live inference server.
     Standalone (no ``[online]``), one epoch of every eval source runs against the
@@ -95,6 +102,12 @@ class EvalsConfig(BaseConfig):
     """Directory to write outputs to — rollout traces and logs are written as
     subdirectories. Shared with the trainer for online evals. Defaults to ``$PRL_OUTPUT_DIR`` if set, else ``outputs``."""
 
+    ckpt: CheckpointConfig | None = None
+    """Checkpoint standalone eval progress. Checkpoint steps count completed task groups."""
+
+    resume: ResumeConfig | None = None
+    """Resume a standalone eval run from a group checkpoint. A bare block loads the latest checkpoint."""
+
     log: LogConfig = LogConfig()
 
     monitors: MonitorsConfig = MonitorsConfig()
@@ -115,5 +128,14 @@ class EvalsConfig(BaseConfig):
             raise ValueError(
                 "eval.skip_first_step only applies to online evals - a standalone run "
                 "would skip its only eval epoch. Remove it or add an [online] block."
+            )
+        return self
+
+    @model_validator(mode="after")
+    def validate_checkpointing_is_standalone_only(self):
+        if self.online is not None and (self.ckpt is not None or self.resume is not None):
+            raise ValueError(
+                "evals ckpt/resume only supports standalone evals - online evals are coordinated "
+                "with the trainer's live weight-broadcast lifecycle"
             )
         return self
--- a/packages/prime-rl-configs/src/prime_rl/configs/evals.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/evals.py
@@ -70,7 +70,7 @@ class CheckpointConfig(BaseConfig):
     """Checkpoint progress for an interruptible standalone eval run."""
 
     interval: int = Field(1, ge=1)
-    """Save after every N completed task groups."""
+    """Save after the task cursor advances by N completed groups."""
 
 
 class EvalsConfig(BaseConfig):
@@ -103,10 +103,10 @@ class EvalsConfig(BaseConfig):
     subdirectories. Shared with the trainer for online evals. Defaults to ``$PRL_OUTPUT_DIR`` if set, else ``outputs``."""
 
     ckpt: CheckpointConfig | None = None
-    """Checkpoint standalone eval progress. Checkpoint steps count completed task groups."""
+    """Checkpoint standalone eval progress. Checkpoint steps are task cursor positions."""
 
     resume: ResumeConfig | None = None
-    """Resume a standalone eval run from a group checkpoint. A bare block loads the latest checkpoint."""
+    """Resume a standalone eval run from a cursor checkpoint. A bare block loads the latest checkpoint."""
 
     log: LogConfig = LogConfig()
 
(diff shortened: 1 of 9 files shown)


docs/training.md
```diff
@@ -176,6 +176,18 @@ uv run sft @ examples/basic/reverse-text/sft.toml --monitors.wandb
...
+
+`uv run evals` can checkpoint a one-shot eval after completed task groups, so a long eval can continue after interruption without rerunning completed groups. Enable `[ckpt]` and choose the save cadence with `interval`:
+
+```toml
+[ckpt]
+interval = 10
+keep_last = 2
+```
+
+Resume from the latest saved group count with `--resume`, select one with `--resume.step N`, or fork from another output directory with `--resume.dir path/to/checkpoints/step_N`. A checkpoint preserves completed-group results required for the final aggregate; any group that was only partially complete when the checkpoint was written is retried. Resuming without `[ckpt]` loads progress but writes no new checkpoints. Evals checkpoint/resume is only available without an `[online]` block because online eval progress is coupled to the trainer's live weight-broadcast lifecycle.
```
samsja: remove

packages/prime-rl-configs/src/prime_rl/configs/evals.py
```diff
@@ -66,6 +66,19 @@ class OnlineConfig(BaseConfig):
...
+    """Checkpoint progress for an interruptible standalone eval run."""
+
+    interval: int = Field(1, ge=1)
+    """Save after every N completed task groups."""
+
+    keep_last: int | None = Field(None, ge=1)
+    """Keep at most this many recent eval checkpoints. If None, keep all."""
+
+    keep_interval: int | None = Field(None, ge=1)
+    """Keep checkpoints at every N completed task groups permanently."""
```
samsja: doesnt really matter her we can keep them all as its small

tests/unit/evals/test_ckpt.py
```diff
@@ -0,0 +1,60 @@
...
+
+    source = EvalSource(envs, config)
+    source.mark_completed("env", task.key, task.hash, step=0)
+    state = source.state_dict()
+
+    resumed = EvalSource(envs, config)
+    resumed.load_state_dict(state)
+    assert resumed.trigger(0) == ["env"]
+    assert resumed.next_task().task is task
+    assert resumed.next_task() is None
```
samsja: not needed

src/prime_rl/evals/ckpt.py
```diff
@@ -0,0 +1,109 @@
...
+if TYPE_CHECKING:
+    from prime_rl.orchestrator.eval_sink import EvalSink
+    from prime_rl.orchestrator.eval_source import EvalSource
+
+
+class CheckpointManager:
+    def __init__(self, output_dir: Path, config: CheckpointConfig | None) -> None:
+        self.config = config or CheckpointConfig()
+        self.ckpt_dir = get_ckpt_dir(output_dir)
+
```
samsja, line 20: can sipligy the whoe calss and remove the deletig logic etcc

src/prime_rl/evals/ckpt.py
```diff
@@ -0,0 +1,77 @@
...
+                        "completed_groups": step,
+                        "eval_source": eval_source.state_dict(),
+                        "eval_sink": eval_sink.state_dict(),
+                    },
+                    f,
+                )
+            os.replace(tmp_name, ckpt_path / "progress.pt")
+        except BaseException:
+            with contextlib.suppress(OSError):
+                os.unlink(tmp_name)
```
garrett361, line 45: Why this write-to-temp-then-move pattern? Some possible race condition or something?

src/prime_rl/evals/evals.py
```diff
@@ -253,6 +277,9 @@ async def run(self) -> None:
         else:
             await self.watch()
 
+        if self.config.ckpt is not None and self.completed_groups > 0:
+            self.save_checkpoint()
+
```
garrett361: Why is there this unconditional initial save? As opposed to the later saves keyed off of interval comparison `self.completed_groups % self.config.ckpt.interval == 0`

src/prime_rl/evals/ckpt.py
```diff
@@ -0,0 +1,77 @@
...
+        path: Path | None = None,
+    ) -> None:
+        ckpt_path = path if path is not None else self.get_ckpt_path(step)
+        state_file = ckpt_path / "progress.pt"
+        if not state_file.is_file():
+            raise FileNotFoundError(f"Evals checkpoint not found at {state_file}")
+        get_logger().info(f"Loading evals checkpoint from {state_file}")
+        with open(state_file, "rb") as f:
+            state = pickle.load(f)
+        if state["completed_groups"] != step:
```
garrett361: Looks like we check consistency, but don't restore the `completed_groups` value anywhere. Do we need to?

garrett361: nvm, looks like we do it down here https://github.com/PrimeIntellect-ai/prime-rl/pull/3413/changes#diff-6ab100d417f56ce26a4b5e88e4a39fef72e128584823c5f579163bb101b5dd6dR172

src/prime_rl/evals/ckpt.py
samsja: Yep, the intent is crash safety rather than inter-process coordination: write the full pickle to a temp file in the target checkpoint dir, then `os.replace` it into `progress.pt` atomically. That way a resume path either sees the previous complete checkpoint or the new complete checkpoint, never a truncated pickle if the process dies mid-write. I clarified that inline.

src/prime_rl/evals/evals.py
samsja: This thread is outdated now. The old unconditional save path was replaced with `maybe_save_checkpoint(force=True)`, which still exits unless checkpointing is enabled and the durable cursor advanced since the last save. So it only flushes a final cursor checkpoint when there is new completed progress to persist.

src/prime_rl/evals/ckpt.py
samsja: This is outdated after the cursor-only checkpoint rewrite. The checkpoint payload is now just `{ "cursor": N }`; load validates that the payload cursor matches the checkpoint step and then restores it via `EvalSource.load_state_dict`.

mikasenghaas (approved):

PR #3417, opened 2026-08-27
fix: shard SFT data and report worker progress

mikasenghaas:
## Summary

- Resolve SFT data shard identities after worker processes start.
- Shard fake data by DP rank while replicating it across CP ranks.
- Read epoch and sample progress from the furthest worker cursor.
- Aggregate per-source sample and token counters across workers.
- Preserve worker counters across dataloader checkpoint resume.
- Extend the existing CPU dataloader suite with tiny in-memory SFT data across worker, DP, and CP layouts.
- Clarify the dummy renderer token decode in the sharding test.

## Verification

A one-GPU SFT run used two local dataset splits, two data workers, and four logged steps.

- Before: sample progress was `5, 5, 9, 9`, and source ratio metrics were absent.
- After: sample progress was `9, 10, 17, 18`, and source sample and token ratios were present at every step.
- Before the sharding fix, two workers yielded `0, 0, 1, 1, 2, 2, 3, 3` from increasing fake data.
- The CPU regression matrix verifies unique DP shards, CP replicas, checkpoint resume, counters, and the `0 → 1` epoch transition.
- The fake-data test verifies distinct samples across DP ranks and identical samples across CP ranks.
- The uneven-packing regression verifies that worker progress never moves backward.
- `uv run pytest tests/unit/train/sft/test_dataloader.py -q` passed with 16 tests.
- `uv run ruff check tests/unit/train/sft/test_dataloader.py src/prime_rl/trainer/sft/data.py src/prime_rl/trainer/sft/train.py` passed.

---

> [!NOTE]
> **Medium Risk**
> Changes distributed data partitioning and checkpointed iterator state, which can silently skew training if wrong; behavior is covered by expanded unit tests but affects core SFT loops.
> 
> **Overview**
> Fixes **incorrect SFT data sharding and stale progress metrics** when using multiple DataLoader workers and/or context-parallel (CP) ranks.
> 
> **Sharding:** `data_rank` / `data_world_size` are now derived from **DP rank only** (`rank // non_dp_size`), so CP ranks see the same samples while DP ranks stay distinct. Rank/worker identity is resolved in **`__iter__` via `_setup_world_info()`** instead of at dataset construction, so worker processes get the right shard after fork.
> 
> **Progress & checkpoints:** `CatDataset` state now stores and restores per-source **`num_samples` / `num_tokens`**. New **`get_dataset_progress()`** reads worker snapshots, uses the **furthest worker** for `step`/`epoch`, and **sums** counters across workers. SFT training logs **`progress/*`** from the dataloader instead of the main-process dataset object (which stayed at zero after resume).
> 
> **Tests:** Dataloader tests drop the heavy tokenizer/GPU mark, add a shared dummy renderer, and cover multi-rank/worker sharding, resume, CP replication, and monotonic progress under uneven packing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30f7de1d229f8e88e0f334be6b8baaa7d2dccf14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/unit/train/sft/test_dataloader.py
```diff
@@ -130,12 +123,117 @@ def test_fake_dataset_single_rank_state_with_packing():
...
     torch.random.set_rng_state(rng_state)
     resumed_batch = next(resumed_dataiter)
 
     for key in ("input_ids", "position_ids", "target_ids", "loss_mask", "seq_lens"):
         torch.testing.assert_close(resumed_batch[key], expected_batch[key])
+
+
+@pytest.mark.parametrize("num_workers", [1, 2])
+@pytest.mark.parametrize("world_size", [1, 2])
+def test_dataloader_shards_across_ranks_and_workers(num_workers: int, world_size: int, dummy_renderer):
```
garrett361: Let's please also add another parameterize here which also covers `non_dp_size in [1, 2]` with proper assertions. Looks like we have various combos of `{num_workers, world_size, non_dp_size}` but never simultaneously all > 1

src/prime_rl/trainer/sft/data.py
```diff
@@ -697,3 +703,24 @@ def get_dataset_state(dataloader: StatefulDataLoader) -> dict:
     private worker-snapshot layout."""
     snapshots = dataloader.state_dict()["_snapshot"]["_worker_snapshots"]
     return {wid: snap["dataset_state"]["dataset"] for wid, snap in sorted(snapshots.items())}
+
+
+def get_dataset_progress(dataloader: StatefulDataLoader) -> dict:
+    """Dataset position and aggregate counters from dataloader workers."""
+    snapshot = dataloader.state_dict()["_snapshot"]
+    worker_snapshots = snapshot["_worker_snapshots"]
```
garrett361, line 711: Similar discomfort to previous SFT dataloader PRs regarding needing to access private dict keys. 

I don't see a good solution for getting this info from the `StatefulDataloader` itself.  A more robust impl could be to derive the `num_{samples,tokens}` etc from the `Batch` items as they are consumed

tests/unit/train/sft/test_dataloader.py
mikasenghaas: done

tests/unit/train/sft/test_dataloader.py
```diff
@@ -130,12 +123,135 @@ def test_fake_dataset_single_rank_state_with_packing():
...
+            )
+            return setup_dataloader(dataset, config)
+
+        dataloader = setup_epoch_dataloader()
+        dataiter = iter(dataloader)
+        epochs = set()
+
+        def next_sample(dataiter, dataloader, index: int) -> int:
+            micro_batch = next(dataiter)
+            sample = micro_batch["target_ids"][0, 0].item() - ord("0") - 2
```
garrett361, line 198: Seems hard to maintain as written. What is `ord("0") - 2` about? Need to at least document the intention/need here.

garrett361 (approved):

PR #3419, opened 2026-08-27
refactor(trainer): make activation checkpointing operator-based

S1ro1:
## Summary

- Replace method- and attribute-based selective activation checkpointing with one non-reentrant checkpoint region around each selected transformer block.
- Keep `mode = "full"` as the default. Full mode retains only mandatory non-replayable operations and recomputes everything else; selective mode additionally retains configured expensive operations through the same Prime-owned policy.
- Allow selective mode to replace its public default targets with explicit operator names or namespaces. Correctness-required targets remain active for every selective configuration.
- Register real runtime boundaries for Torch EP/CP collectives, DeepEP combine, ring attention, sparse MLA, dense/grouped DeepGEMM, MXFP8 transport, and MoE routing telemetry so checkpointing, autograd, and compile observe the same operations.
- Keep sparse MLA dispatcher, fake, and autograd registration beside the vendored forward/backward kernel entrypoints rather than in GLM model code.
- Collapse ring attention into one backend-neutral forward/backward custom-op pair. Its adapters call only the pinned FA2, FA3, and FA4 APIs explicitly, without signature probing or generated per-backend registrations.
- Remove checkpoint-only attention helpers and the model-specific checkpointing module. Generic attention, AFMoE, Qwen3.5, and Laguna keep their projection/output logic directly in `forward`; GLM DSA keeps only its substantive `mla_latents` and `mla_up_proj` helpers.
- Rebase onto the Torch 2.13, CUDA 13, and vLLM 0.28 runtime from current `main`, and update FlashAttention 4 to `0251105` with the shared TVM FFI version required by vLLM.
- Ignore FSDP replay-only profiler markers when selective checkpointing validates the saved forward operation sequence.
- Prefetch the final MoE block, experts, and separately wrapped router before backward replay so FSDP collectives occur in the same order as the saved forward.

## Runtime contract

- Model implementations do not expose checkpoint target names or checkpoint-only methods.
- `freq` selects whole transformer blocks; the dispatched operator and selected target set determine whether selective AC saves or recomputes its result.
- Both modes retain mandatory MoE routing, telemetry, non-replayable DeepEP, and CUDA-to-CPU metadata operations so they execute once. Full mode recomputes every other operation; selective mode additionally retains its configured targets.
- The default selective targets retain supported attention, matrix multiplication, grouped GEMM, and distributed communication operators. Elementwise operations, norms, reshapes, and RoPE replay.
- Custom targets replace the selective defaults and match either an exact operator name or an operator namespace. They cannot exclude mandatory routing, telemetry, non-replayable communication, or CUDA-to-CPU metadata operations.
- Unsupported or missing backend operators fail at their normal call site; there is no checkpoint fallback layer.
- MXFP8 transport resolves `mxfp8_moe` inside its custom-op implementation. MXFP8 grouped compute still resolves it during runtime setup because the exported grouped GEMM is already an operator boundary.

## Configuration

```toml
[trainer.model.ac]
mode = "selective" # full | selective
freq = 1
# targets = ["aten::mm", "prime_rl_collectives"]
```

Omitting `targets` uses the public Prime defaults. Any explicit list replaces those defaults; an empty list retains only the mandatory operations.

## Validation

- Ruff, Ruff formatting, `uv lock --check`, and `git diff --check` passed.
- Focused config, checkpointing, and Qwen3.5 CP tests on the rebased dependency stack: 138 passed.
- A6000 CI regression suite passed all eight checks: four-GPU peak memory was 6.2467 GiB against the 6.2988 GiB baseline (down from 6.6185 GiB before the fix); one-GPU peak memory was 10.6738 GiB.
- Compiled selective AC with DeepEP completed a two-GPU, two-step `samsja/mini-glm-moe` training smoke after rebasing onto the merged MoE runtime.
- The pinned FA2, FA3, and FA4 varlen backends all use the `[heads, total_tokens]` LSE layout; the ring fake and backward paths enforce that single contract.
- Two-H200 compiled ring validation matched each backend's direct attention reference exactly for FA2, FA3, and FA4 output and `dq`, `dk`, and `dv`.
- Full H200 unit suite: 574 passed, 9 skipped.
- Eager and compiled output/gradient/replay validation passed for local MoE, Torch EP, DeepEP with and without chunking, CP, FA2/FA3 ring attention, sparse MLA, dense DeepGEMM, grouped DeepGEMM, and the production MXFP8 transport boundary with a fake numerical kernel.
- FSDP parameter/state-dict parity and selective AC plus activation offloading passed.
- Whole-model parity passed for Llama, Nemotron-H, GPT-OSS, AFMoE, Laguna, and Qwen3.5 full-attention layers. Sparse MLA parity also passed after restoring its substantive helper structure.
- On the stacked Qwen3.8 Flash Next implementation, selective and full activation checkpointing each completed 20 RL steps on 24 H200s with compile, FSDP, expert parallelism, SignSGD, NIXL policy reload, and CPU optimizer-state offload. Steps 2-20 averaged 8,008 tokens/s for selective and 7,211 tokens/s for full; both peaked at 130.4 GiB and held mismatch KL at 0.0014-0.0021 with a final value of 0.0016.

## Hardware boundary

- Numerical MXFP8 grouped GEMM and transport still require SM100 hardware. The H200 validation exercised the production custom-op/autograd/compile boundary with a BF16 fake transport kernel.
- FA4 forward/backward numerical validation passed on H200 (SM90); Blackwell remains untested in this PR.
- Qwen3.5 GatedDeltaNet whole-model parity did not reach execution because the no-AC reference forward remained in FLA kernel compilation until the diagnostic jobs were cancelled. Qwen3.5 full-attention parity passed, and the GatedDeltaNet runtime code itself is unchanged.

---

> [!NOTE]
> **High Risk**
> Touches core training memory/throughput (activation checkpointing), distributed MoE/DeepEP/ring attention autograd, and compile boundaries—incorrect replay or double-counted routing would show up as silent wrong grads or NCCL deadlocks.
> 
> **Overview**
> **Replaces layer-specific selective activation checkpointing** with whole-block non-reentrant checkpointing driven by a PyTorch selective operator policy (`activation_checkpointing.py`). Semantic targets like `norm` / `attn_proj` are gone; `model.ac.targets` is now optional operator names or namespaces that replace Prime’s defaults, while mandatory saves still cover MoE routing/top-k, DeepEP, and CUDA→CPU copies.
> 
> **Registers distributed and kernel paths as `torch.library` custom ops** so checkpoint replay, autograd, and compile see the same boundaries: new `prime_rl_collectives` for EP/CP all-to-all and gather/reduce-scatter, `deepep::combine`, unified `prime_rl_ring` ring attention (FA2/FA3/FA4), `prime_rl::sparse_mla`, FP8/grouped FP8 GEMMs, and `record_moe_routing_statistics` so routing counters are not double-applied on recompute. Inline `autograd.Function` wrappers in token dispatch, Mamba CP, and ring attention are removed in favor of these ops.
> 
> **Simplifies trainer wiring and models**: `apply_ac` always wraps whole transformer blocks; the old `layers/checkpointing.py` hook layer is deleted and attention modules no longer expose checkpoint-only `attn_projections` / `output_proj` splits. FSDP backward prefetch now includes the last block’s MoE experts/router. Docs, examples, and benchmarks drop legacy selective target lists; **flash-attn-4** is bumped to `0251105` with a pinned **apache-tvm-ffi** override.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3a51c78f1ed46b0603ec42c1011c36e14533b710. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/trainer/models/layers/ring_attn.py
```diff
@@ -96,19 +192,260 @@ def _fa3_varlen_backward(
...
+        max_seqlen_k: int,
+        local_k_slice_start: int,
+        local_k_slice_stop: int,
+        heads_k_stride: int,
+        causal: bool,
+        group_name: str,
+        window_size_left: int,
+        window_size_right: int,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        return torch.empty_like(q), q.new_empty((q.shape[1], q.shape[0]), dtype=torch.float32)
```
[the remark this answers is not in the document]
S1ro1: The reported mismatch does not occur with the backends pinned by this repository. FA2 2.8.3, FA3 3.0.0, and the pinned FA4 varlen APIs all return LSE as `[heads, total_tokens]`. The three-dimensional branch was stale compatibility code, so 7e0ac18 removes it and makes the backward path use the same explicit two-dimensional contract as the fake implementation.

src/prime_rl/trainer/activation_checkpointing.py
```diff
@@ -0,0 +1,117 @@
...
+        "aten::bmm",
+        "aten::convolution",
+        "aten::linear",
+        "aten::mm",
+        "prime_rl::fp8_blockwise_mm",
+        "prime_rl::grouped_fp8_gemm",
+        "prime_rl::sparse_mla",
+    }
+)
+DEFAULT_SELECTIVE_TARGETS = DEFAULT_SELECTIVE_SAVE_NAMESPACES | DEFAULT_SELECTIVE_SAVE_OPERATIONS
```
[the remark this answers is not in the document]
S1ro1: This one is valid. A dispatch trace confirms that FLA GatedDeltaNet exposes only surrounding `aten` operations: its core Triton launch is hidden inside a plain `torch.autograd.Function`, and `mamba_chunk_scan_combined` has the same contract. Adding an `fla`/`mamba_ssm` target would therefore do nothing. I am leaving this thread open until the stacked native Qwen/Nemotron implementations give those forward/backward kernels real `torch.library` boundaries; that preserves the operator-based SAC design without restoring model-attribute hooks.

src/prime_rl/trainer/distributed/deepep.py
```diff
@@ -15,6 +17,8 @@
 
 _buffer: Buffer | None = None
 _handle_cache: dict[int, object] = {}
+_combine_backward_handles: dict[int, object] = {}
+_combine_dispatchers: WeakValueDictionary[int, object] = WeakValueDictionary()
```
samsja: can we add comment why we use weak value here

src/prime_rl/trainer/distributed/deepep.py
```diff
@@ -120,51 +127,53 @@ def _dispatch_backward(
...
-        return grad_x, None, None
+    previous_event = _new_event_overlap()
+    combined, _, after_event = _buffer.combine(
+        x=x,
+        handle=handle,
+        previous_event=previous_event,
+        async_finish=True,
+        allocate_on_comm_stream=True,
+    )
+    dispatcher = cast("DeepEPTokenDispatcher | None", _combine_dispatchers.get(dispatcher_id))
```
samsja: lets remove this cast no ? i hate cast, even more when its using "string" and can be None

src/prime_rl/trainer/activation_checkpointing.py
samsja, line 74: can we add in coment what is the difference between and operator and a namespace here

src/prime_rl/trainer/activation_checkpointing.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def _mandatory_checkpoint_policy(
+    _context: SelectiveCheckpointContext,
+    operation: torch._ops.OpOverload | torch._ops.HigherOrderOperator,
+    *args,
+    **kwargs,
```
samsja, line 83: not sure to follow why we have args and kwrgs, is this because its the way the function is being applied by torch ?

pyproject.toml
```diff
@@ -250,7 +252,7 @@ torchao = [
     { url = "https://github.com/PrimeIntellect-ai/prime-kernels/releases/download/v0.1.1/torchao-0.18.0+git5f2baf9d5-cp310-abi3-linux_aarch64.whl", marker = "platform_machine == 'aarch64'" },
 ]
 dion = { git = "https://github.com/samsja/dion.git", rev = "d891eeb" }
-flash-attn-4 = { git = "https://github.com/Dao-AILab/flash-attention.git", subdirectory = "flash_attn/cute", rev = "82d6441" }
+flash-attn-4 = { git = "https://github.com/Dao-AILab/flash-attention.git", subdirectory = "flash_attn/cute", rev = "0251105" }
```
samsja, line 255: why changed here ?

src/prime_rl/trainer/distributed/deepep.py
S1ro1: Added in 3a51c78f1. The custom-op schema can carry only the integer dispatcher ID; the weak map resolves it without keeping dispatcher instances alive after their owning runtime is released.

src/prime_rl/trainer/distributed/deepep.py
S1ro1: Removed in 3a51c78f1. The weak map is now typed as  with postponed annotations, so  has the correct optional type without a cast.

src/prime_rl/trainer/activation_checkpointing.py
S1ro1: Added in 3a51c78f1. An operation target matches one qualified , while a namespace target matches every operation whose  has that value.

src/prime_rl/trainer/activation_checkpointing.py
S1ro1: Yes. PyTorch calls the policy as . We need  here to inspect  for  and retain only CUDA-to-CPU metadata copies. Added that explanation in 3a51c78f1.

pyproject.toml
S1ro1: This aligns the bottom of the stack with the FA4 runtime required by the GPT-OSS work:  provides the learned-sink API, and FA4, vLLM, and TileLang must share the pinned TVM FFI runtime. Keeping the pin in this lower PR avoids carrying a second dependency contract in the stacked model PR.

samsja (approved):

PR #3429, opened 2026-08-28
feat(models): add native Qwen3.5 family implementation

S1ro1:
## Summary

Unify Qwen3.5 dense, MoE, and VLM training under a native implementation, including the Qwen3.8 checkpoint that uses this architecture. Remove the separate MoE package and global Transformers model patches.

- Keep `Qwen3_5PreTrainedModel(PreTrainedModelPrimeRL)` shared by the backbone and causal LM; remove unused Transformers TP/PP plans, split-module metadata, cache arguments, and redundant initialization. Modules own their initialization, including zero-centered RMSNorm.
- Use Prime's shared attention and MoE runtime. Gated full attention consumes the fused QKV projection with the correct doubled query width; configured MoE load balancing allocates its router selection bias.
- Own packed DeltaNet, zero-centered RMSNorm, MRoPE, and the vision encoder. Hybrid layers support Ulysses CP through FLA boundary-state exchange; full attention uses common attention dispatch.
- Convert official fused MoE expert weights to Prime's split projections and back, excluding MTP weights.
- Keep the narrow compiler boundary for FLA CP convolution, whose collective layout fails Dynamo tracing. Vision FlashAttention has no compiler disable and uses each backend's declared return format directly.

Qwen tests live under `tests/unit/train/models/qwen3_5/`: `test_model.py` parametrizes initialization, context-parallel setup, and forward/backward/packing checks over dense and MoE configs; `test_vlm.py` groups vision tests with CPU MRoPE checks. MoE router replay remains a dedicated check. Tests use native configs and models without Transformers reference models, HF weight comparisons, or Hub downloads. Public model helpers use `AutoModelForCausalLMPrimeRL.from_config` with the trainer's shared auto-attention resolver, without hardcoded backends or private attention config writes. Remove the class/source-introspection test and duplicate shared-attention checks. Remove the blanket BF16 optimizer/reduction restriction for RL VLMs; FP32 defaults remain unchanged and FSDP still computes in BF16.

Text and vision consume Prime's resolved `model.attn` configuration directly, including under Ulysses CP. Remove Qwen constructor fallback/propagation logic, the forced vision SDPA override, and its unused vision branch.

MTP and video training are outside this runtime's scope. `scripts/mini_moe.py` is unchanged from main and is outside this PR's validation.

## Math KL validation

Both checkpoints completed 20 steps on `math-env` (`PrimeIntellect/Hendrycks-Math`, train split), with **batch size 64**. Every step-mean KL is below **0.015**.

| Checkpoint | Steps | Mean KL over 20 steps | Highest step-mean KL |
|---|---:|---:|---:|
| `Qwen/Qwen3.8-27B` | 20 | 0.000359685 | 0.000953884 |
| `Qwen/Qwen3.5-35B-A3B` | 20 | 0.000208713 | 0.000462048 |

Runs executed concurrently on separate H200 nodes using `uv run rl @ <config>`, with four trainer and four inference GPUs per model. Both Slurm jobs completed with exit code 0: dense job 387, MoE job 386. Tested commit: `9983e7bed`.

Settings: main's `debug` algorithm with constant per-token advantage 1.0; original task sampler; math-verify scoring without the optional LLM judge; Prime VM runtime enforcing the task network policy; sequence length 4096, maximum completion length 2048, group size 8, learning rate 1e-6, compilation, selective activation checkpointing, FlashAttention 3, and NCCL weight updates. MoE additionally uses EP=4 and router replay. Inference uses `max_num_seqs=64`. Optimization and reduction dtypes retain their FP32 defaults.

All 20 batches per model contained 64 trainable episodes, and all gradient norms were finite. These are asynchronous E2E measurements of `mismatch_kl/all/mean`; the summary averages the 20 step means and includes normal policy staleness. The debug algorithm exercises training with synthetic advantages, so this is a correctness check, not a learning-quality benchmark. These 20-step runs use text inputs; packed image-conditioned validation is recorded below.

<details>
<summary>All 20 step-mean KL values</summary>

| Step | Dense mean KL | MoE mean KL |
|---:|---:|---:|
| 1 | 0.000087817 | 0.000103068 |
| 2 | 0.000133938 | 0.000132430 |
| 3 | 0.000112734 | 0.000115124 |
| 4 | 0.000193951 | 0.000099179 |
| 5 | 0.000205668 | 0.000214333 |
| 6 | 0.000359477 | 0.000348763 |
| 7 | 0.000838096 | 0.000161281 |
| 8 | 0.000150568 | 0.000207813 |
| 9 | 0.000190916 | 0.000158351 |
| 10 | 0.000246897 | 0.000258552 |
| 11 | 0.000492465 | 0.000462048 |
| 12 | 0.000236148 | 0.000177551 |
| 13 | 0.000311170 | 0.000316216 |
| 14 | 0.000953884 | 0.000286797 |
| 15 | 0.000817523 | 0.000233271 |
| 16 | 0.000320892 | 0.000184788 |
| 17 | 0.000567697 | 0.000175972 |
| 18 | 0.000280278 | 0.000164684 |
| 19 | 0.000264224 | 0.000196673 |
| 20 | 0.000429352 | 0.000177372 |

</details>

## Packed VLM validation

Both models completed five packed image RL updates on MMMU Math dev, with batch size 32 and an 8192-token trainer packing limit. All 160 samples per model included images.

| Model | Steps | Mean KL | Maximum step-mean KL | Images per packed microbatch |
|---|---:|---:|---:|---:|
| Qwen/Qwen3.8-27B | 5 | 0.000501153 | 0.000729994 | 5–6 |
| Qwen/Qwen3.5-35B-A3B | 5 | 0.000346154 | 0.000458347 | 5–6 |

| Step | Dense KL | MoE KL |
|---:|---:|---:|
| 1 | 0.000231250 | 0.000167199 |
| 2 | 0.000387219 | 0.000343682 |
| 3 | 0.000553381 | 0.000360291 |
| 4 | 0.000729994 | 0.000401253 |
| 5 | 0.000603919 | 0.000458347 |

Native models; trainable vision encoder; FA3; compiled decoder blocks and selective activation checkpointing; FP32 optimization and gradient reduction; CP1; MoE EP4 and router replay; debug advantages; group size 8; maximum completion length 1024; inference context 4096; NCCL weight broadcasts. These are correctness smokes on four repeated diagram questions, not learning-quality benchmarks.

Packed MRoPE exactly matched individual-document MRoPE for all 320 training samples, and image-grid/pixel-patch counts matched each document’s image-token spans. Both Slurm jobs exited 0 (dense 400, MoE 399). MoE needed allocator retries at the H200 memory limit.

Image+CP+compile remains separately unresolved; these E2E runs use CP1.

Tested runtime contents: `dcdf770df` (subsequent commits only change tests). Run configs and detailed batch/position evidence are under `/home/matej/prime-rl-verify/pr3429-20260910/`.

## Other validation

- Attention cleanup at `a2173601f`: H200 Qwen suite 24 passed, one skipped; official dense/MoE model loading preserves configured attention for text and vision at CP sizes 1 and 2. An additional compiled image/CP backward check exposed `Autograd not support dtype: Int` with selective activation checkpointing; the same error reproduces using the previous SDPA vision path. Compiled image/CP validation remains unresolved. The completed math runs above use CP=1.

- Native-only H200 Qwen suite: **20 passed, one expected single-rank CP skip**, with `HF_HUB_OFFLINE=1` and `TRANSFORMERS_OFFLINE=1` (Slurm 407; consolidated dense/MoE suite). Existing config suite: **175 passed**. Ruff check/format pass. Both dense and MoE run the shared packed-versus-unpacked comparison. MoE replays identical expert selections to isolate document boundaries; ordinary routing forward/backward remains covered.
- CI at `9983e7bed`: CPU unit tests, Ruff, slim install, and CodeQL pass. GPU CI passes the Qwen tests; its five failures are in unchanged DeepSeek V4 tests (three packed-batch assertions reproduced on unmodified main `bad80b1a3`, two FP8 architecture failures). GPU CI: 66 passed, 82 skipped, 599 deselected, five failed.
- Two-rank H200 compiled packed DeltaNet output and input-gradient checks passed with the CP convolution compiler boundary.
- Exact official checkpoint key/shape conversion and reverse conversion passed: dense 1,199 source keys to 1,184 runtime keys; MoE 1,811 source keys to 1,066 runtime keys after excluding MTP.
- Full H200 unit suite: 737 passed, 12 skipped, three DeepSeek V4 failures also reproduced on unmodified main. The weight-transfer test imports the unified model.

---

> [!NOTE]
> **High Risk**
> Large change to core training models (hybrid attention, MoE weight conversion, packed multimodal RL) with removed Transformers workarounds; regressions would affect Qwen3.5/3.8 RL and VLM correctness rather than peripheral code.
> 
> **Overview**
> Replaces the split **Qwen3.5 dense + MoE packages** and **Transformers runtime monkey-patches** with one **native PrimeRL implementation** under `models/qwen3_5/`. Dense text, MoE text, and composite VLMs all dispatch to **`Qwen3_5ForCausalLM`**, driven by local configs (`Qwen3_5TextConfig`, `Qwen3_5MoeTextConfig`, vision/composite configs) and updated **`AutoConfig` / VLM registration**.
> 
> The rewrite drops the HF-backed text stack in favor of **Prime `FlashAttention`**, **packed varlen GatedDeltaNet** (FLA + Ulysses CP), **zero-centered RMSNorm**, **MRoPE**, a **custom vision encoder**, and **MoE layers** with HF↔Prime **expert weight conversion** (MTP dropped). VLM prep keeps dummy vision runs for FSDP symmetry and shards **`routed_experts`** under CP. **`model.py`** no longer applies Qwen3.5 checkpoint/MRoPE/varlen patches or forces vision SDPA under CP.
> 
> **Trainer/config cleanup:** removes the **VLM must use bfloat16** validator and the duplicate check at model load; **`parallel_dims`** treats MoE on nested **`text_config`**. Ring/Ulysses patching no longer lists Qwen-specific attention classes because **`Qwen3_5Attention` subclasses `FlashAttention`**.
> 
> **Tests/docs:** deletes HF-comparison tests and the old `qwen3_5_moe` tree; adds **`tests/unit/train/models/qwen3_5/`** (forward/pack/CP/VLM/MRoPE). **`development.md`** points VLM authors at the unified package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9162ce07425b56e01f6b62d3b24b141a8ff4a51. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/trainer/models/qwen3_5/attention.py
```diff
@@ -0,0 +1,86 @@
...
+                output_bias=config.attention_bias,
+                use_qk_norm=False,
+                rms_norm_eps=config.rms_norm_eps,
+            ),
+            flash_attn_version=flash_attn_version,
+        )
+        self.num_heads = config.num_attention_heads
+        self.num_key_value_heads = config.num_key_value_heads
+        self.attn_output_gate = config.attn_output_gate
+        q_output_size = self.num_heads * self.head_dim * (2 if self.attn_output_gate else 1)
```
garrett361, line 34: Little awkward with the overrides and redefinitions of sub modules

src/prime_rl/trainer/models/qwen3_5/attention.py
```diff
@@ -0,0 +1,86 @@
...
+        self.qkv_sizes = (q_output_size, self.qkv_sizes[1], self.qkv_sizes[2])
+        self.q_proj = nn.Linear(config.hidden_size, q_output_size, bias=config.attention_bias)
+        self.q_norm = Qwen3_5RMSNorm(self.head_dim, config.rms_norm_eps)
+        self.k_norm = Qwen3_5RMSNorm(self.head_dim, config.rms_norm_eps)
+
+    def forward(
```
garrett361, line 40: Similar nit: we are overriding so much of this that it’s barely a subclass any more

garrett361: For me, the smell here is that I'd expect the FlashAttention class to be less responsible.I’d expect it to consume qkv with rope already applied as necessary as inputs, and not be responsible for creating these.

Wdyt? Not a change for this PR.

src/prime_rl/trainer/models/qwen3_5/gated_delta_net.py
```diff
@@ -0,0 +1,114 @@
...
+import torch.nn.functional as F
+from fla.modules import FusedRMSNormGated
+from fla.modules.conv import causal_conv1d
+from fla.ops.cp import build_cp_context
+from fla.ops.gated_delta_rule import chunk_gated_delta_rule
+from torch import nn
+
+from prime_rl.trainer.models.qwen3_5.configuration_qwen3_5 import Qwen3_5TextConfig
+
+# FLA's CP convolution uses an all-gather layout that Dynamo cannot trace.
```
garrett361, line 11: Can you explain why we need this? Though we'd just graph break. Would like to understand why we need to disable

src/prime_rl/trainer/models/qwen3_5/gated_delta_net.py
```diff
@@ -0,0 +1,114 @@
...
+        mixed_qkv = self.in_proj_qkv(hidden_states)
+        output_gate = self.in_proj_z(hidden_states).reshape(
+            batch_size, sequence_length, self.num_value_heads, self.value_head_dim
+        )
+        beta = self.in_proj_b(hidden_states).sigmoid()
+        decay = -self.A_log.float().exp() * F.softplus(self.in_proj_a(hidden_states).float() + self.dt_bias)
+
+        context = None
+        if self.context_parallel_group is not None:
+            context = build_cp_context(
```
garrett361, line 70: Future optimization: can probably just build once per model, not once per GDN layer

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -144,197 +126,153 @@ def cp_support(cls, config) -> CPSupport:
...
+        self.norm = Qwen3_5RMSNorm(config.hidden_size, config.rms_norm_eps)
+        self.rotary_emb = Qwen3_5RotaryEmbedding(config)
 
-    def get_input_embeddings(self):
+    def get_input_embeddings(self) -> nn.Embedding:
         return self.embed_tokens
 
-    def set_input_embeddings(self, value):
-        self.embed_tokens = value
+    def set_input_embeddings(self, embeddings: nn.Embedding) -> None:
```
garrett361, line 163: Can we delete all of the `{get,set}_input_embeddings` methods everywhere? They seem to be a `transformers` holdover that we don't use, IIUC.

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -144,197 +126,153 @@ def cp_support(cls, config) -> CPSupport:
...
+        inputs_embeds: torch.Tensor | None = None,
+        routed_experts: torch.LongTensor | None = None,
         *,
         seq_lens: torch.LongTensor,
         seq_lens_are_pre_shard: bool = False,
-    ) -> BaseModelOutputWithPast:
-        if (input_ids is None) ^ (inputs_embeds is not None):
-            raise ValueError("You must specify exactly one of input_ids or inputs_embeds")
-
+    ) -> BaseModelOutput:
```
garrett361: Why did the returned type change

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -144,197 +126,153 @@ def cp_support(cls, config) -> CPSupport:
...
-        hidden_states = inputs_embeds
-        position_embeddings = self.rotary_emb(hidden_states, position_ids)
-
-        cu_seqlens_are_pre_shard = seq_lens_are_pre_shard
+        position_embeddings = self.rotary_emb(inputs_embeds, position_ids)
 
-        for decoder_layer in self.layers:
+        hidden_states = inputs_embeds
+        for layer_index, decoder_layer in enumerate(self.layers):
+            layer_routed_experts = routed_experts[:, :, layer_index] if routed_experts is not None else None
```
garrett361: Understand this

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -144,197 +126,153 @@ def cp_support(cls, config) -> CPSupport:
...
-    def __init__(self, config: PretrainedConfig):
+    def __init__(self, config) -> None:
         super().__init__()
         self.config = config
-        self.visual = Qwen3_5VisionModel._from_config(config.vision_config)
+        self.visual = Qwen3_5VisionModel(config.vision_config)
         self.language_model = Qwen3_5Model(config.text_config)
 
-    def get_input_embeddings(self):
+    def get_input_embeddings(self) -> nn.Embedding:
```
garrett361: Understand this

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -144,197 +126,153 @@ def cp_support(cls, config) -> CPSupport:
...
+            vision_grid = torch.tensor([[1, merge_size, merge_size]], device=inputs_embeds.device)
 
+        image_embeds = self.visual(pixel_values, vision_grid).pooler_output.to(inputs_embeds.dtype)
         if has_images:
-            image_mask = input_ids == self.config.image_token_id
-            image_mask = image_mask.unsqueeze(-1).expand_as(inputs_embeds).to(inputs_embeds.device)
+            image_mask = (input_ids == self.config.image_token_id).unsqueeze(-1).expand_as(inputs_embeds)
             inputs_embeds = inputs_embeds.masked_scatter(image_mask, image_embeds)
         else:
+            # Every rank must retain the vision graph so FSDP collectives stay symmetric.
```
garrett361: Understand this

src/prime_rl/trainer/models/qwen3_5/norm.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def reset_parameters(self) -> None:
+        nn.init.zeros_(self.weight)
+
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        input_dtype = hidden_states.dtype
+        hidden_states = hidden_states.float()
+        variance = hidden_states.square().mean(dim=-1, keepdim=True)
+        hidden_states = hidden_states * torch.rsqrt(variance + self.variance_epsilon)
+        return ((1.0 + self.weight.float()) * hidden_states).to(input_dtype)
```
garrett361, line 22: weird norm layer

src/prime_rl/trainer/models/qwen3_5/rotary_embedding.py
```diff
@@ -0,0 +1,172 @@
...
+            section[index] += 1
+        return section
+
+    @staticmethod
+    def compute_default_rope_parameters(
+        config: Qwen3_5TextConfig,
+        device: torch.device | None = None,
+        seq_len: int | None = None,
+    ) -> tuple[torch.Tensor, float]:
+        del seq_len
```
garrett361, line 51: Why del?

src/prime_rl/trainer/models/qwen3_5/rotary_embedding.py
```diff
@@ -0,0 +1,172 @@
...
+
+        inv_freq, self.attention_scaling = self.rope_init_fn(config, device)
+        self.register_buffer("inv_freq", inv_freq, persistent=False)
+        self.register_buffer("original_inv_freq", inv_freq.clone(), persistent=False)
+
+        self.mrope_section = config.rope_parameters.get("mrope_section")
+        if self.mrope_section is None:
+            self.mrope_section = self.scaled_mrope_section(inv_freq.numel())
+
+    @staticmethod
```
garrett361: Any particular reason to make this static?

src/prime_rl/trainer/models/qwen3_5/rotary_embedding.py
```diff
@@ -0,0 +1,172 @@
...
+        positions = torch.arange(0, rotary_dim, 2, dtype=torch.int64, device=device).float()
+        return 1.0 / (rope_parameters["rope_theta"] ** (positions / rotary_dim)), 1.0
+
+    def reset_parameters(self) -> None:
+        inv_freq, self.attention_scaling = self.rope_init_fn(self.config, self.inv_freq.device)
+        self.inv_freq.copy_(inv_freq)
+        self.original_inv_freq.copy_(inv_freq)
+
+    @torch.no_grad()
+    @dynamic_rope_update
```
garrett361: Understand decorator

src/prime_rl/trainer/models/__init__.py
```diff
@@ -36,8 +41,11 @@
 AutoConfig.register("minimax_m2", MiniMaxM2Config, exist_ok=True)
 AutoConfig.register("nemotron_h", NemotronHConfig, exist_ok=True)
 AutoConfig.register("qwen3_moe", Qwen3MoeConfig, exist_ok=True)
+AutoConfig.register("qwen3_5", Qwen3_5Config, exist_ok=True)
```
garrett361: Understand why we need this any more

src/prime_rl/trainer/models/base.py
```diff
@@ -44,8 +44,9 @@ def keep_in_fp32_for_weight_transfer(cls, name: str) -> bool:
         return False
 
     @classmethod
-    def from_config(cls, config, **kwargs):
+    def from_config(cls, config, trust_remote_code: bool = False, **kwargs):
         """Public from_config that mirrors the Auto class API."""
+        del trust_remote_code
```
garrett361, line 49: Why del?

src/prime_rl/trainer/model.py
```diff
@@ -88,277 +87,6 @@ def pre_download_model(model_name: str, *, skip_weights: bool = False) -> None:
     )
 
 
-def _patch_qwen3_5_moe_conversion_mapping():
```
garrett361, line 91: Great. Love deleting patches

src/prime_rl/trainer/models/__init__.py
```diff
@@ -95,7 +105,7 @@ def supports_custom_impl(model_config: PretrainedConfig) -> bool:
 # Points to the same unified class — the config drives text-only vs VLM behavior.
 _CUSTOM_VLM_MAPPING: dict[str, type] = {
     "qwen3_5": Qwen3_5ForCausalLM,
-    "qwen3_5_moe": Qwen3_5MoeForCausalLM,
+    "qwen3_5_moe": Qwen3_5ForCausalLM,
```
garrett361, line 108: Just verifying the class change here was intended

src/prime_rl/trainer/parallel_dims.py
```diff
@@ -280,6 +280,7 @@ def _is_moe_model(config: ModelConfig) -> bool:
     from transformers import AutoConfig
 
     model_config = AutoConfig.from_pretrained(config.name, trust_remote_code=config.trust_remote_code)
+    model_config = getattr(model_config, "text_config", model_config)
```
garrett361, line 283: Looks like we could use a similar fix in `is_tt_moe_model`

tests/unit/train/models/qwen3_5/test_model.py
```diff
@@ -0,0 +1,282 @@
...
+    with torch.device(device):
+        model = AutoModelForCausalLMPrimeRL.from_config(
+            config, attn_implementation=runtime_config.attn, dtype=torch.bfloat16
+        )
+    inject_prime_lm_head(model, chunk_size=None)
+    return model
+
+
+@pytest.mark.gpu
+def test_norms_remain_zero_centered_after_model_init(text_config):
```
garrett361, line 91: can probably drop this test?

tests/unit/train/models/qwen3_5/test_model.py
```diff
@@ -0,0 +1,282 @@
...
+    vlm_model = get_model(get_vlm_config(text_config), device="meta")
+    setup_model_cp(vlm_model, cp_group, cp_rank=0, cp_world_size=2)
+    assert vlm_model.model.language_model.context_parallel_group is cp_group
+    assert vlm_model.model.language_model.layers[0].linear_attn.context_parallel_world_size == 2
+
+
+@pytest.mark.gpu
+def test_qwen3_5_gated_delta_net_context_parallel():
+    if int(os.environ.get("WORLD_SIZE", 1)) != 2:
+        pytest.skip("run with torchrun --nproc-per-node=2")
```
garrett361, line 122: I think this always skips in CI, then? Maybe can delete if so, good enough it currently runs and passes, and can keep it in this branch's commit history

garrett361 (approved):
LGTM, various small comments/questions

src/prime_rl/trainer/models/qwen3_5/gated_delta_net.py
S1ro1: this was here before the PR, it's on my TODO to explore this more

src/prime_rl/trainer/models/qwen3_5/modeling_qwen3_5.py
S1ro1: I think we use those somewhere, noted though to research it

src/prime_rl/trainer/models/__init__.py
S1ro1: Yeah now both MoE and dense are a single class, which is imo a bit cleaner

src/prime_rl/trainer/models/base.py
S1ro1: this is like such an ugly thing - hf keeps random things in the config sometimes and they break downstream from here...

src/prime_rl/trainer/models/qwen3_5/attention.py
S1ro1: agree ye, i think we might actually want to move attn to not even be subclassable at some point

src/prime_rl/trainer/models/base.py
garrett361: I actually just saw this pattern in torchtitan as well, I think some particular agent has developed a preference for this. Apparently the intent is to show that the arg (inherited from some parent class) is unused in this particular instance. And it quiets type checkers. 

But yeah, seems dangerous if anything else ever references the same object.

PR #3433, opened 2026-08-30
feat: admit episodes at their declared fan-out cost

hallerite:
Step 1 of the fan-out scheduling discussion (Slack thread): admit episodes at their **declared fan-out cost** instead of one permit each.

## Problem
The concurrency controller already thinks in **concurrent model streams** — its cap is sized from engine KV capacity / `max_model_len`, and its KV/queue feedback reacts to real request load. But the dispatcher admits every episode as **one** permit. A fan-out harness (rlm sub-agents at `max_concurrent_subagents=2`; more with deeper trees) runs several concurrent streams inside one episode, so `current_inflight` undercounts real streams by the fan-out factor: the pool overshoots the sized capacity and the controller only finds out via KV pressure/queue cuts after the fact.

## Change
- The harness declares its per-episode stream bound as `max_concurrent_model_calls` (resolved via `getattr`, default **1** — non-fan-out envs and older harness configs are completely unchanged).
- `Dispatcher` resolves a per-`(kind, env)` cost once, prechecks `available_permits >= cost`, acquires/releases in cost units, and records `cost` on `InflightEpisode` so every completion/cancel/drop path refunds exactly what was reserved. Overload sheds count permits, not episodes.
- `concurrency.max_inflight` docstring updated: the unit is concurrent model streams (an episode admits at its fan-out).

## Notes
- **Deliberately static** (declared max, not measured concurrency): it reserves for the worst case, which costs some utilization when bursts are brief — measured fan-out in our sub-agent eval runs averaged ~1.1–1.4 vs a declared 2. Follow-ups discussed in the thread: feed measured inflight requests from the interception layer into the controller (step 2), router-side backpressure (step 3).
- One semantic shift to be aware of: where `max_inflight` was used to bound per-episode external resources (e.g. sandboxes), a fan-out env now stays below the same ceiling by its fan-out factor (conservative direction).
- Verifiers side (separate PR): `HarnessConfig` grows the `max_concurrent_model_calls` declaration, with the rlm harness deriving it from `max_depth`/`max_concurrent_subagents`.

### DIFF
--- a/packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
+++ b/packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
@@ -430,7 +430,10 @@ class ConcurrencyConfig(BaseConfig):
     """Minimum number of in-flight episodes. Set ``min_inflight = max_inflight`` to recover fixed concurrency."""
 
     max_inflight: int | None = Field(1024, ge=1)
-    """Maximum number of in-flight episodes. Set it to avoid runaway concurrency, especially to limit other external resources (e.g. sandboxes). None removes the ceiling."""
+    """Maximum in-flight concurrency, in concurrent model streams: an episode is admitted at its
+    harness's declared fan-out (``max_concurrent_model_calls``, default 1), so with fan-out harnesses
+    the episode count stays below this by the fan-out factor. Set it to avoid runaway concurrency,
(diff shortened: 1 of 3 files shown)


hallerite:
Closing: the adaptive concurrency controller's KV/queue feedback already absorbs harness fan-out in practice, so per-episode static cost accounting isn't needed. (Discussion consensus.)

PR #3442, opened 2026-08-31
chore(nixl): use device-agnostic torch calls in weight transfer

yafshar:
## Change

Replace the hardcoded `torch.cuda` calls in the NIXL weight-transfer path with their
`torch.accelerator` equivalents, the API this repo already uses in
`src/prime_rl/utils/act_offloading.py`:

| before | after |
| --- | --- |
| `torch.cuda.synchronize(device)` | `torch.accelerator.synchronize(device)` |
| `torch.cuda.set_device(device)` | `torch.accelerator.set_device_index(device)` |
| `torch.cuda.current_device()` | `torch.accelerator.current_device_index()` |

Seven call sites across the vLLM update worker and the trainer-side sender. On a CUDA
accelerator each pair is the same underlying call, so this is a no-op: no new
abstraction, no config, no dispatch on device type.

`torch.accelerator.current_device_index()` requires torch >= 2.7; the project pins
torch >= 2.13.

## Verification

On an L40S/B200 (torch 2.13.0+cu130), `torch.accelerator.current_device_index()` returns the
same index as `torch.cuda.current_device()`, and both `synchronize(device)` and
`set_device_index(device)` accept the worker's `torch.device` and succeed.

### DIFF
--- a/src/prime_rl/inference/vllm/worker/nixl.py
+++ b/src/prime_rl/inference/vllm/worker/nixl.py
@@ -482,7 +482,7 @@ class NIXLWeightUpdateWorker(Worker):
         started = time.perf_counter()
         self.apply_transfer_plan(plan)
         update_mla_absorbed_weights(self.raw_model)
-        torch.cuda.synchronize(self.device)
+        torch.accelerator.synchronize(self.device)
         logger.info(
             "Applied NIXL policy update on rank %d in %.2fs",
             self.model_express.rank,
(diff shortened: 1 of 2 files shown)


yafshar:
@S1ro1 @mikasenghaas mind a look? +7/-7, all one-line swaps: `torch.cuda.{synchronize,set_device,current_device}` →
the `torch.accelerator` equivalents already used in `utils/act_offloading.py`. On CUDA each pair is the same
underlying call, so it's a no-op — no new abstraction, no config, no dispatch on device type.

Verified on an L40S and a B200 (torch 2.13.0+cu130): `current_device_index()` returns the same index as
`torch.cuda.current_device()`, and `synchronize(device)` / `set_device_index(device)` take the worker's
`torch.device`. Checked with a non-zero device index and inside the prefetch thread, since that's where a
set-device regression would hide.

Non-CUDA support is separate PRs and not part of this one.

PR #3444, opened 2026-08-31
feat: trainer trace annotations + dashboard overlays

mikasenghaas:
## Summary

Every rollout episode is written to disk exactly once, as an arrival-ordered stream, and everything learned about it afterwards is an append-only annotation folded back on read. The trainer's recomputed per-token streams ride that mechanism, and the trace viewer renders them as overlays (entropy, the trainer↔sampling mismatch, the IPO stable mask) next to the loss mask.

**Layout** (replaces `rollouts/step_N/{train,eval}/{all,effective}/traces.jsonl`):

```
<run_dir>/monitors/file/metrics.jsonl                              # every metric row, tagged by producer
<run_dir>/monitors/file/traces/stream.jsonl                        # every episode, appended as it arrives
<run_dir>/monitors/file/traces/stream.index.jsonl                  # one compact row per episode, with its byte offset
<run_dir>/monitors/file/traces/annotations/<producer>.jsonl        # trace updates: orch ship-time facts, trainer per-token streams
<run_dir>/monitors/file/traces/annotations/<producer>.index.jsonl  # each update's scalars and where its record sits
```

Everything the file monitor dumps lives under its own directory, and nothing is written there when the monitor is off. The traces and everything written about them sit under `traces/`, and every index is named for the stream it indexes and sits beside it. Each file is append-only with exactly one writer, so there are no locks and live tailing keeps working.

**Steps are events, not locations.** A trace is dispatched at one step, arrives at another and is trained at a third, so a step directory could only ever be honest about one of them — and picking arrival forced the arrival step onto the batch wire just so the trainer could find the right directory. Instead each step is stamped as its own event on the record. `orchestrator/annotations.py` owns both events: `stamp_arrival` puts `info.kind` plus `info.dispatch` / `info.arrival` (`{step, time}`) on a trace as its episode lands, and `stamp_ship` produces the update carrying `info.effective` and `info.ship` — the orchestrator step whose batch shipped the cohort, or for eval the policy version the epoch measured. Staleness is `ship.step - dispatch.step`; the trace viewer shows all three for a train trace, and an eval trace shows the step it measured instead (it never ships in a batch and never goes stale).

- **Annotations**: one record format and one fold, in `prime_rl.monitors.trace_updates`, used by the file monitor, the trainer writer and the dashboard. An update names a trace and carries an `info` dict merged into the trace's own plus per-branch token streams, full-length over the branch's token prefix (nulls mark unknown positions) so folding never depends on the producer's loss mask.
- **Indexes**: the file monitor summarises each episode into `stream.index.jsonl` as it writes it, and each update's scalars into the producer's `.index.jsonl`, both with byte offsets. A reader browses the indexes and seeks for the one episode it opens, so the cost of listing a run stops scaling with the token streams — which outweigh the episodes themselves.
- **Branch identity on the batch wire**: `TrainingSample` gains `trace_id` / `branch_index`; `MicroBatch` carries them per sequence, parallel to `sequence_lengths`, preserved through packing/padding and nulled on dummy micro batches.
- **Trainer**: accumulates one update per trained sequence per forward (packing-boundary first value and non-finite values are null), gathers them to rank 0 at the end of each step, and logs them through its own file monitor — the same monitor it already logs metrics through — rather than writing to disk behind it. Always on; the token-export path is gone.
- **Dashboard**: same HTTP API shape. The traces tab opens on the whole stream — newest first, paged as you scroll, with a histogram of episodes finished per interval you can click to narrow to a moment — and switches to step mode for the cohort that shipped at one step (arrow keys walk the step bar). Episode reads fold every annotation onto the record, so the overlays work in both views. Token signals are Advantage, Entropy, Mismatch (`r − log r − 1`, computed client-side from the trainer and sampling logprobs), Stable mask (`|Δp|` vs the run's IPO `eps`), Loss Mask and Content Mask; an overlay the open episode cannot colour is disabled in the selector and names the input it is missing, rather than silently painting nothing. In a token overlay the reasoning renders inside the token spans that already cover it instead of in its own box.
- Verifiers gained only the two per-token node fields the records land on (`MessageNode.trainer_logprobs` / `entropies`, with matching `Branch` spread properties) — [verifiers#2481](https://github.com/PrimeIntellect-ai/verifiers/pull/2481), merged; `deps/verifiers` tracks verifiers main.

## Verification

Reverse-text, 4 steps: 645 episodes on one stream, 496 annotated by both producers, 55,694 masked tokens folded with every node's compact arrays equal to its masked-token count. Mean KL mismatch 0.0014, mean entropy 0.61. Staleness came out as 336 traces shipped at their dispatch step and 160 a step later — the lag the old layout could not express. Re-run end to end on a 20-step reverse-text run after the layout change. SFT-online-eval and verifiers `uv run eval` run shapes checked through the real API with a FastAPI test client; the dashboard's browser smoke suite covers the stream/step views, the histogram, the step bar and every token signal.

## Breaking

- The `rollouts/` trace tree is gone. Episodes now live in one `monitors/file/traces/stream.jsonl` stream, written once at arrival, with the effective cohort expressed as `monitors/file/traces/annotations/<producer>.jsonl` records instead of duplicate episode copies (per-token advantages and the scalar advantage now arrive via annotations). The dashboard reads only the new layout; runs recorded before this change need the old dashboard.
- `metrics.jsonl` moves from the run root to `monitors/file/metrics.jsonl`.
- `trainer.enable_token_export` is removed and `token_exports/` is no longer written; the trainer's per-token streams are always logged as annotations.
- The filesystem batch transport writes packed micro batches to `batches/step_<n>/rank_<r>.bin` (previously `rollouts/step_<n>/`).

## 20-step GLM-Air ScaleSWE comparison

- The baseline ran `main`; the comparison ran this PR. The code change was the only behavioral difference. Run names, output directories, and conversion cache paths were isolated.
- Trainer step 1: before 874.0 s; after 822.0 s (-5.9%).
- Trainer steps 1–20: before 3,179.2 s; after 2,995.2 s (-5.8%).
- Steady trainer steps 2–20: before 121.3 s mean / 120.0 s median; after 114.4 s mean / 99.0 s median. The after run was 5.7% faster by mean step time.
- Trace data at each run's trainer step-20 cutoff: before 3.516 GB across duplicated `all` and `effective` traces; after 3.476 GB across the arrival stream plus orchestrator and trainer annotations. The after layout was 1.1% smaller despite 864 MB of per-token trainer logprob and entropy annotations.
- Dashboard on the after run at step 20: a 128-episode page took 15.7 ms, a 2.22 MB annotated episode took 42.7 ms, and its timeline took 9.7 ms.

The after job was cancelled after step 24 at the researcher's request. The baseline completed trainer step 20.

---

> [!NOTE]
> **Medium Risk**
> Large observability and on-disk contract change (breaking paths and removed `trainer.enable_token_export`); training/orchestration logic is mostly additive but any tool still expecting `rollouts/` or root `metrics.jsonl` will break.
> 
> **Overview**
> Replaces the per-step `rollouts/.../traces.jsonl` tree with a **file monitor layout** under `monitors/file/`: metrics at `metrics.jsonl`, episodes in one append-only `traces/stream.jsonl` plus sidecar indexes, and **append-only trace annotations** per producer (orchestrator ship-time facts, trainer per-token streams).
> 
> **Orchestrator/evals** stamp arrival metadata on episodes at log time and log **effective cohort** membership, advantages, and ship step via `log_annotations` instead of duplicating episode files. **Trainer** drops `enable_token_export` / `token_exports/`; per-token logprobs and entropies always go through the same annotation path. **Dashboard** reads the new layout: stream vs step cohort views, paged `/api/runs/.../episodes`, arrival histogram, and token overlays (entropy, mismatch KL, IPO stable mask, masks) folded from annotations.
> 
> Docs, skills, configs, and `run_single_benchmark.py` are updated for the new paths and for skipping non-step metric rows. Filesystem batch transport writes to `batches/step_*` instead of under `rollouts/`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16bf663b7805ecfe7d993ff6b6d34c8b7bc47308. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -681,8 +681,8 @@ class TrainerConfig(BaseConfig):
     metrics_server: MetricsServerConfig | None = None
     """Prometheus metrics server configuration. If set, exposes a ``/metrics`` endpoint for scraping."""
 
-    enable_token_export: bool = False
-    """Opt-in per-token JSONL export for rollout debugging. When enabled, writes token ids and aligned trainer metrics after each forward pass."""
+    enable_trace_annotations: bool = True
```
mikasenghaas: always do this, not configurable

mikasenghaas (commented):
i dont like that we write to trace_annotations. i think we should be writing to 

```
<run-dir>/traces/step_{n}/traces.jsonl # all episodes from env server
<run-dir>/traces/step_{n}/annotations.jsonl # all/effective metadata + run info + trainer data
```

packages/prime-rl-configs/src/prime_rl/configs/trainer.py
mikasenghaas: Done — the flag is gone; the writeback always runs (41ad201).

mikasenghaas:
Restructured in 41ad201 per the review + our discussion: episodes now live in `traces/step_<n>/{train,eval}.jsonl` (written once at arrival, per-kind so step availability stays stat-based) with per-producer annotation files `traces/step_<n>/annotations/{orchestrator,trainer}.jsonl` — the all/effective duplication is gone, everything is append-only with one writer per file, and the `rollouts/` dir is retired (filesystem batch transport moved to `batches/`).

docs/configuration.md
```diff
@@ -2,7 +2,7 @@
 
 Every `prime-rl` entrypoint uses [`pydantic-config`](https://github.com/PrimeIntellect-ai/pydantic-config): TOML files for reproducible base configs, CLI flags for one-off overrides.
 
-> **AI agents working in this repo:** the equivalent runbook is at [`skills/configs/SKILL.md`](https://github.com/PrimeIntellect-ai/prime-rl/blob/main/skills/configs/SKILL.md), with extra runtime hints (where config classes live, validator conventions, the trainer-side `enable_token_export` flag) that aren't surfaced here.
+> **AI agents working in this repo:** the equivalent runbook is at [`skills/configs/SKILL.md`](https://github.com/PrimeIntellect-ai/prime-rl/blob/main/skills/configs/SKILL.md), with extra runtime hints (where config classes live, validator conventions, the trainer-side `enable_trace_annotations` flag) that aren't surfaced here.
```
mikasenghaas: `enable_trace_annotations` is outdated no?

skills/configs/SKILL.md
```diff
@@ -86,15 +86,9 @@ The `sft` entrypoint takes the same eval shape at the top level for online evals
...
-
-```toml
-enable_token_export = true
-```
-
-Leave it unset for normal training. When enabled, it exports every sequence from each exporting rank.
+The RL trainer writes its per-token streams (recomputed logprobs, entropies) as trace-update JSONL to `<run_dir>/traces/step_<n>/annotations/trainer.jsonl`, one record per trained sequence keyed by `trace_id` and `branch_index`, placed next to the trace's arrival file (records land in the step the trace was logged at, which under off-policy lag differs from the trained step). Rank 0 is the only writer: every rank gathers its records to it at the end of each step. The dashboard folds these onto rollout traces at read time to power the trainer-logprob, entropy, mismatch-KL, and stable-mask token overlays. Always on; the trainer does not decode token text.
```
mikasenghaas: ```(records land in the step the trace was logged at, which under off-policy lag differs from the trained step). Rank 0 is the only writer: every rank gathers its records to it at the end of each step. The dashboard folds these onto rollout traces at read time to power the trainer-logprob, entropy, mismatch-KL, and stable-mask token overlays. Always on; the trainer does not decode token text.``` too much detail -- remove

skills/dashboard/SKILL.md
```diff
@@ -78,6 +78,15 @@ for multi-agent traces and `highlight` entries shaped as `{node, quote,
 reason, field?}`. On `409`, tell the user to open the returned `url`; the stored
 command applies when the tab connects.
 
+The trace viewer's token-signal selector colors recorded tokens by advantage,
```
mikasenghaas: remove this

tools/convert_traces_to_hf_dataset.py
```diff
@@ -146,7 +146,9 @@ def register_in_dataset_card(root: Path, subset: str, split: str, rel_path: str)
 
 def main() -> None:
     parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
-    parser.add_argument("traces", type=Path, help="a run's traces.jsonl (one episode per line)")
+    parser.add_argument(
```
mikasenghaas: revert this

mikasenghaas:
Reworked per the discussion — traces are now a stream, not a step tree.

- `monitors/file/{traces,metrics}.jsonl` + `monitors/file/annotations/<producer>.jsonl`; the trainer logs its annotations through its own file monitor instead of writing behind it, so one component owns the layout.
- Steps are stamped as events on the record (`info.dispatch` / `info.arrival` / `info.ship`, each `{step, time}`) rather than implied by a directory that could only be honest about one of them. Only `effective` ties to a step — the orch ship step, or the policy version for eval — and `all` is the whole arrival-ordered stream. `logged_at_step` is off the batch wire entirely.
- The verifiers trace-update API is dropped: nothing imported it (both writers emitted plain dicts, the dashboard read them with orjson), so the format and the fold now live once in `prime_rl.utils.trace_updates`. vf#2481 is down to the two node fields.
- `branch_index` restored; the four review comments are addressed.

E2E on reverse-text: 645 episodes, 496 annotated, staleness reads 336 traces at lag 0 and 160 at lag 1.

Still open: the pageable index sidecar you picked — the dashboard currently holds one summary per episode in memory for the stream (fine at this size, not at 1M episodes). Happy to do it here or as a follow-up.

examples/basic/reverse-text/rl.toml
```diff
@@ -21,14 +21,32 @@ env.taskset.id = "reverse-text"
...
 [trainer.optim]
 lr = 3e-6
 
 [ckpt] # Checkpoint at the end of training
 
 [inference]
 
-# PrimeIntellect Qwen3 fine-tune - select its renderer explicitly (the -SFT
-# suffix misses the MODEL_RENDERER_MAP exact match).
+# PrimeIntellect Qwen3 fine-tune - select its renderer and reasoning parser
```
mikasenghaas: remove comment

skills/configs/SKILL.md
```diff
@@ -86,15 +86,9 @@ The `sft` entrypoint takes the same eval shape at the top level for online evals
 
 In TOML, an empty section header (`[ckpt]`) does the same.
 
-## RL trainer token exports
+## RL trainer trace annotations
```
mikasenghaas: remove this section entirely

src/prime_rl/monitors/file.py
```diff
@@ -10,11 +10,54 @@
...
+    one is stamped as its own event rather than implied by where the record is stored."""
+    work = getattr(episode.run, "work", None)
+    for trace in episode.traces:
+        trace.info["kind"] = kind
+        if work is not None:
+            trace.info["dispatch"] = {"step": work.step, "time": trace.timing.start}
+        trace.info["arrival"] = {"step": step, "time": now}
+
+
+def _cohort_step(step: int, kind: Kind) -> int:
```
mikasenghaas: this is non-sense. we shouldn't max cap at 1 for evals bc you can eval the base policy. remove this

src/prime_rl/monitors/file.py
```diff
@@ -10,11 +10,54 @@
...
+from prime_rl.utils.trace_updates import index_suffix, make_update, update_index_row
 from prime_rl.utils.utils import sanitize
 
 if TYPE_CHECKING:
     import verifiers.v1 as vf
 
+OPTS = orjson.OPT_APPEND_NEWLINE | orjson.OPT_SERIALIZE_NUMPY
+
+
+def _stamp_arrival(episode: vf.Episode, kind: Kind, step: int, now: float) -> None:
```
mikasenghaas: this shouldn't be in this file. put it into `orchestrator/annotations.py` as a public functions. it should be
- `stamp_arrival` -> at episode arrival
- `stamp_ship` -> at ship batch time

src/prime_rl/monitors/file.py
```diff
@@ -46,22 +92,63 @@ async def log_metrics(self, metrics: dict[str, Any], step: int | None) -> None:
...
+        """``all`` appends each episode to the trace stream as it completes — every
+        episode is serialized exactly once, in arrival order, whatever its kind, so an
+        in-progress run can be tailed. ``effective`` writes no second copy: it records
+        what the ship-time cohort learned (see ``_effective_update``) as annotations
+        that readers fold back onto the stream. Episode-level failures are preserved
+        even when no trace was produced."""
+        if subset == "effective":
+            now = time.time()
+            await self.log_annotations(
+                [_effective_update(trace, step, kind, now) for episode in episodes for trace in episode.traces]
```
mikasenghaas: can we make the util over episode directly?

src/prime_rl/trainer/rl/annotations.py
mikasenghaas, line 1: rename to annotations.py

src/prime_rl/utils/pathing.py
```diff
@@ -204,8 +204,14 @@ def get_ckpt_dir(output_dir: Path) -> Path:
     return output_dir / "checkpoints"
 
 
-def get_rollout_dir(output_dir: Path) -> Path:
-    return output_dir / "rollouts"
+def get_batch_dir(output_dir: Path) -> Path:
+    return output_dir / "batches"
+
+
+def get_file_monitor_dir(output_dir: Path) -> Path:
```
mikasenghaas, line 211: just do helpers for monitor dir. the file monitor should live in the file monitor directly

src/prime_rl/monitors/file/traces/index.py
mikasenghaas, line 1: isn't this dashboard specific? if so, let's move there

tests/unit/train/rl/test_trace_annotations.py
mikasenghaas, line 1: remove tests

docs/configuration.md
mikasenghaas: Gone — the docs no longer mention it (the flag itself was removed in 41ad201).

skills/configs/SKILL.md
mikasenghaas: Removed — see below, the whole section is gone.

skills/dashboard/SKILL.md
mikasenghaas: Removed. The paragraph is down to one sentence on what `subset` selects, which is the part that affects how you address an episode through `/api/view`.

tools/convert_traces_to_hf_dataset.py
mikasenghaas: Already reverted — `git diff main...HEAD` on this file is empty.

examples/basic/reverse-text/rl.toml
mikasenghaas: Removed (96e0cf4).

skills/configs/SKILL.md
mikasenghaas: Section deleted (96e0cf4). With the flag gone there is nothing configurable left to document.

src/prime_rl/monitors/file.py
mikasenghaas: Removed (96e0cf4) — the cohort keys to its own trigger with no floor.

Worth flagging the consequence, since it is the thing you asked me to fix two messages ago: evals fire from `WeightWatcher.on_update`, so an eval epoch's step *is* the policy version it measured, and policy versions are 0-indexed. With no floor, a run that evaluates the base policy gets a step 0 on the bar again. That is the honest reading of "you can eval the base policy", so I have left it — say the word if you want the bar to render a 0 eval-only entry differently.

src/prime_rl/monitors/file.py
mikasenghaas: Done (96e0cf4): new `orchestrator/annotations.py` with public `stamp_arrival` (called at the arrival log sites, orchestrator + evals) and `stamp_ship` (called at ship time, returning the updates). The file monitor is back to writing what it is handed.

src/prime_rl/monitors/file.py
mikasenghaas: Done — both take `list[vf.Episode]` and do the trace walk internally, so no caller repeats the double comprehension.

src/prime_rl/trainer/rl/annotations.py
mikasenghaas: Renamed to `trainer/rl/annotations.py`, and the class to `AnnotationWriter` (96e0cf4).

src/prime_rl/utils/pathing.py
mikasenghaas: Done — `pathing` keeps only `get_file_monitor_dir`; `get_trace_dir` / `get_trace_stream` / `get_annotations_dir` / `get_index_path` now live in `monitors/file.py` and the dashboard imports them from there.

tests/unit/train/rl/test_trace_annotations.py
mikasenghaas: Removed (96e0cf4).

src/prime_rl/monitors/file/traces/index.py
mikasenghaas: Not quite — the file monitor *writes* the index (`index_row` on every episode as it lands), the dashboard only reads it, so moving it under `dashboard/` would point the writer at an optional-extra package.

Moved it to `monitors/trace_index.py` instead, next to the writer, and took `trace_updates.py` with it for the same reason — they are the two on-disk formats the file monitor owns. Both are out of `utils/` either way.

src/prime_rl/evals/evals.py
```diff
@@ -508,6 +510,7 @@ async def finalize_eval_batch(self, batch: EvalBatch) -> None:
 
         if batch.episodes.effective:
             await monitors.log(batch.episodes.effective.vf_episodes, batch.step, "eval", "effective")
+            await monitors.log_annotations(stamp_ship(batch.episodes.effective.vf_episodes, batch.step))
```
hallerite: stamp_ship reads a bit weird

mikasenghaas: stamp_batch?

hallerite (dismissed):
LGTM. I think the files can get very big, so we def need a solution for this, but fine to get this version in imo

mikasenghaas:
Review triage — all three addressed.

**`stamp_ship` reads weird** (@hallerite, and the `stamp_batch?` draft): renamed to `stamp_batch` in 961683a7, alongside `stamp_arrival`. The record key it writes is still `info.ship`, so the name and the field have drifted apart — renaming the key too means updating the dashboard readers and migrating existing runs, so I left it out of the rename. Say the word if you'd rather they matched.

**Unparseable rows use 0-based lines** (Bugbot): real, fixed in 22fe72ad. The summarizer's success path counts from 1 and the decode-failure path did not, so a torn row pointed at the episode before it.

Reproduced on an index-less stream (verifiers `uv run eval` shape) with a deliberately torn second line: rows now report 1..4 and the torn one reports 2. That surfaced a second problem the report did not mention — opening it raised `orjson.JSONDecodeError` straight through the endpoint as a 500 — so `read_episode_at` now answers `422 episode 2 is unparseable`. A stream tailed mid-append always has one of those as its last line.

**Live poll wipes scrolled pages** (Bugbot): real, fixed in 22fe72ad. While the reader is scrolled, a poll now takes only the new count and leaves the loaded pages and the scroll position alone. The etag is deliberately not stored in that case, so the first poll after they scroll back to the top refreshes for real.

Measured in the browser: scrolled to 21765px with 896 rows loaded, a poll with a stale etag keeps both; at the top the same poll rebuilds to the first 128 as before.

494 unit tests and the dashboard browser suite pass on 22fe72ad.

mikasenghaas:
**Scrolled poll breaks episode pagination** — true, both halves. Fixed in 399e4559.

**Duplicates.** A newest-first page is addressed by offset, so a page fetched against a longer stream than the one already loaded repeats its tail. Reproduced end to end: 200 episodes loaded, 20 arrive, pull the next page → 20 duplicate rows (lines 20…1). Lines are stable, so an appended page now drops what is already held; same scenario after the fix adds no duplicates.

Worth noting the offset/growth mismatch predates this branch — a poll used to paper over it by rebuilding from page 1 every few seconds. Holding the table made it persist, so this is the fix that was owed either way.

**Stale cohort on a query change.** Also real: the hold did not check that the query still matched, so a filter or step change landing while the reader was scrolled was swallowed until something else reloaded. The hold is now conditional on the query being unchanged. Verified: with 512 rows loaded, flipping the kind filter to train-only and letting a poll land rebuilds to 128 rows, all of them train.

One thing the report gets slightly wrong: the duplicates do not come from the poll itself — the poll only advances the count. They appear on the *next append* after it, which is why the fix is in the merge rather than in the hold.

hallerite (dismissed):

PR #3445, opened 2026-08-31
feat: carry SFT progress on samples and batches

mikasenghaas:
## Summary

- `Sample` is now a dataclass that carries its own metadata: `source` (the subset/split it came from), `step` (the dataset position when it was emitted), and `epoch`.
- `Batch` is a dataclass that keeps the list of `samples` it packs. Progress is exposed as computed properties over those samples: `step`, `epoch`, `num_samples_by_source`, `num_tokens_by_source`, and `num_padding_tokens`.
- `CatDataset` only groups samples into packs and checkpoints the pending sample. `cat_collate` concatenates, truncates, pads, and tensorizes a pack into a `Batch` (with a `pin_memory` method, since torch pins custom batch types through it).
- The trainer accumulates progress from the batches it consumes into `SFTProgress` (epoch and per-source counters on top of `Progress`), which is checkpointed with the app state. `get_dataset_progress` and `get_dataset_state`, which parsed torchdata's private `_snapshot._worker_snapshots` layout, are removed.
- New `progress/ratio_padding_tokens` metric: the padded fraction of each step's batch tokens.

Semantics note: `progress/num_samples` and `progress/epoch` now reflect samples the trainer consumed, not the furthest worker cursor, so they no longer include worker prefetch.

## Breaking

- SFT trainer checkpoints written before this PR fail to resume with `Missing key in checkpoint state_dict: app.progress.epoch` (the checkpointed progress gained fields). Resume such runs with `ckpt.skip_progress = true`, or restart.
- The SFT dataloader checkpoint no longer stores per-source counters, and stores the pending sample as a `Sample` instead of a dict. A pre-PR dataloader checkpoint that held a pending sample cannot restore it.

## Verification

- `uv run pytest tests/unit/train/sft/ -q` passed with 47 tests.
- A 2-worker `StatefulDataLoader` smoke on this box confirmed batch tensors are pinned through `Batch.pin_memory` and metadata properties are populated.
- The old-checkpoint resume failure mode was reproduced with a standalone DCP save/load round trip.

### DIFF
--- a/docs/training.md
+++ b/docs/training.md
@@ -240,6 +240,7 @@ Pulled from the console log and mirrored to W&B.
 - `val/loss`, `val/perplexity` — validation metrics when `[val]` is set, logged every `val.interval` steps.
 - `eval/{env}/...` — online eval metrics when `[eval]` is set, logged at each evaluated checkpoint step.
 - `progress/epoch`, `progress/num_samples`, `progress/num_tokens` — dataset progress.
+- `progress/ratio_padding_tokens` — fraction of the step's batch tokens that are padding (packing efficiency).
 - `progress/<subset>/ratio_{samples,tokens}` — when training on multiple HF subsets/splits, the realized mixing ratio.
 
 **Stability and optimization:**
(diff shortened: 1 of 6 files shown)


mikasenghaas:
Closing — approach not needed.

PR #3451, opened 2026-08-31
feat(trainer): add runtime gate/up and qkv fusions

S1ro1:
## Summary

Adds runtime parameter fusions, on by default: `model.fusions` packs parameters that are always computed together into one tensor, turning several GEMMs into one. Fusions are runtime-only — checkpoints keep the canonical parameter names and shapes, so a run can enable or disable one at any point and still load its own checkpoints.

```toml
[trainer.model.fusions]
enabled = ["gate_up", "qkv"]   # default; [] disables
raise_on_fail = false          # default: a fusion no module supports logs a warning
shard_fused_on_dim1 = false    # default; experimental zero-copy checkpointing, see Notes
```

- **`gate_up`** — each gated MoE expert's `gate_proj` and `up_proj` become one `[num_experts, 2 * intermediate_size, hidden_size]` weight. The forward is a single grouped GEMM whose transpose and chunk are views; there is no weight reshape.
- **`qkv`** — an attention layer's `q_proj`, `k_proj` and `v_proj` (and their biases) become one `nn.Linear`. `FlashAttention.project_qkv` is shared with `LagunaFlashAttention` so both support it.

## Review fixes and verification (2026-09-08)

Review found and fixed three previously unchecked checkpoint cases:

- **Exports and model-only resume:** DCP returns `{}` when no optimizers are supplied. Both fusion optimizer-state translators now preserve that empty dictionary instead of raising `KeyError: 'state'`.
- **AdamW resume with optimizer-state CPU offload:** packed parameters restored their moments but lost scalar state, including AdamW's step counter. The write-back now restores scalar tensors and non-tensor state as well as packed moments.
- **Full CPU offload checkpoint save:** splitting packed CPU optimizer tensors requires CPU collectives. Both training entrypoints now enable Gloo alongside NCCL when full offload is enabled.

Verified on H200s through Slurm job **232**:

- The complete existing unit suite and Ruff checks passed.
- All **32 distributed checkpoint combinations** passed: AdamW/Muon, fused/unfused save and load, default/dimension-1 sharding, and optimizer-state CPU offload on/off. Canonical weights, optimizer state, and the next update with identical gradients were bitwise exact.
- Full CPU offload completed model-only resume/save and optimizer resume/save. All **311 model tensors** were exact after model-only resume; **931 model/master/step tensors** were checked across optimizer resume, with unchanged weights and correctly incremented counters.
- Expert outputs and matrix-weight gradients, and QKV projections with/without bias, were exact. BF16 input-gradient relative differences were about 0.29%; bias-reduction variation also occurred in repeated unfused calculations.
- All **11 RL/SFT training, resume, and export checks** passed (6 SFT, 5 RL). BF16 and FP8 exports reload in HF, and direct DCP-to-FP8 matches chained conversion byte-for-byte. RL completed 20 steps with final reward **0.8362** and passed its mismatch-KL checks.

The dashboard browser checks cannot install OS dependencies on this cluster because they require sudo. The prior A6000 benchmark CI job was cancelled without executing any steps; its hardware-specific baseline was not validated on H200. Blackwell-only MXFP8 was not exercised.

Verification artifacts: `/home/matej/prime-rl-verify/pr3451-20260908/`. This review changes only the fusion helpers and RL/SFT distributed initialization; it adds no tests or skill changes.

## Design

A fusion is a function that replaces some of a module's parameters with one `PackedParameterSpec`, which owns the split sizes, the packing dimension, and the canonical names it stands in for. Everything else falls out of that description, so a new fusion adds no machinery:

- **Checkpoints** — state-dict hooks registered alongside the packing swap the packed entry for its canonical entries on the way out and back on the way in. The canonical entries are views of the packed storage when the packing dimension is unsharded (`gate_up`: experts shard on dim 0, pack on dim 1); a DTensor sharded along the packing dimension (`qkv`: FSDP shards dim 0) is unsharded by the split, so those entries are replicated copies. Anything that loads into the state dict in place therefore packs the loaded entries back explicitly (`write_back_loaded_packed_parameters` after the HF weight load, `write_back_loaded_packed_optimizer_state` / `join_loaded_optimizer_state_for_runtime` on resume). DCP therefore writes `gate_proj`/`up_proj` and `q_proj`/`k_proj`/`v_proj` and never sees the packed layout. The optimizer has no equivalent hook (its `state_dict()` hooks fire while state is still keyed by integer param ids, before DCP maps them to FQNs), so the same translation is applied explicitly around `get_state_dict` in `AppState`.
- **Muon** — `PackedParameterSpec.muon_matrix_partitions` yields the split sizes when the packing is on the output dimension. Those go to Dion's new `matrix_partitions` argument, which orthogonalizes each logical matrix separately on the one physical parameter, with a per-partition learning-rate adjustment. A packed parameter therefore trains exactly as the parameters it replaces would — including per-projection scaling for grouped-query attention — while keeping a single momentum tensor. Uses the merged matrix-partition support in the pinned `samsja/dion@99e5f7e` (samsja/dion#3).

Modules advertise fusions through `supported_fusions`; requesting one that no module supports logs a warning, or fails at startup with `raise_on_fail = true`. Fusions are skipped when LoRA is enabled, since it targets the unfused projections.

## Performance

Qwen3-30B-A3B, 8×H200, EP=8, `seq_len=8192`, torch EP dispatch, selective AC, compile, no offload. Mean throughput over steps 3–6:

| precision | unfused | fused | |
|---|---|---|---|
| BF16 | 37636 tok/s (MFU 13.3%) | 37352 tok/s (MFU 13.2%) | −0.8% |
| FP8 (`quantization.type=fp8` + `moe.compute.type=deepgemm_fp8`) | 30594 tok/s (MFU 10.8%) | 34062 tok/s (MFU 12.0%) | **+11.3%** |

BF16 is unchanged, as expected: the separate GEMMs already read the same activation, so there is nothing to save. The gain is under FP8, where each GEMM quantizes its activation independently — unfused casts the same tensor 3× for q/k/v and 2× for gate/up, fused casts it once. Peak memory is unchanged (101.8 → 102.0 GiB).

## Validation

Correctness, on GPU:

- Forward output and all weight gradients of the fused grouped experts are **bitwise identical** to the unfused module; the packed q/k/v projections are bitwise identical to the separate ones. Only the MoE input gradient differs (max abs `1.5e-5` on a `2.4e-3` scale), by one bf16 accumulation rounding from doing one `[T, 2I]` dgrad instead of two `[T, I]`.
- Muon updates on a packed parameter are **bitwise identical** to updating the parameters it replaces, over multiple steps, for both fusions.
- All four fused/unfused optimizer checkpoint round trips are exact, including the next optimizer update after each. For `gate_up` the checkpoint entries alias the packed storage; for `qkv` under FSDP they are replicated copies of the q/k/v weights and their optimizer state, materialized transiently on every rank at load and checkpoint time (~1 GB bf16 weights + ~2 GB fp32 Muon momentum for Qwen3-30B-A3B). `shard_fused_on_dim1 = true` (experimental, default off) shards fused 2-D parameters along dim 1 via FSDP2's `shard_placement_fn`, so the split stays a local view: the q/k/v weight entries alias the packed storage and `model.state_dict()` holds 0.1 MiB extra instead of 72 MiB on the mini model (0 / 144 → 72 / 72 weight entries aliasing; biases stay copies, having no second dimension). It is not a correctness fix on its own — biases would still load as zero and the optimizer-offload resume path replaces its state tensors after the template is built, so the explicit write-back stays in place with the flag on or off. Constraints: `hidden_size % dp_shard == 0` (FSDP2 raises otherwise) and an extra chunk-cat per all-gather of dim-1 shards. #3486 measured the placement alone and is closed in favour of this flag.
- DCP metadata for a fused EP=2 checkpoint contains 94 `gate_proj` and 94 `up_proj` entries and zero `gate_up_proj` entries.

End to end on SLURM (2 GPUs, EP=2, compiled, Muon, both fusions unless noted):

| run | result |
|---|---|
| BF16 fake-data training + checkpoint save | ✅ |
| resume: fused ckpt → fused runtime | ✅ |
| resume: fused ckpt → **unfused** runtime | ✅ |
| resume: unfused ckpt → **fused** runtime | ✅ |
| optimizer-state CPU offload, save + resume | ✅ |
| DeepGEMM FP8 expert compute | ✅ |
| pretrained `mini-glm-moe` RL run with vLLM weight reload | ✅ mismatch KL `0.0001` at both steps |

Plus `ruff format`/`ruff check`, and `pytest tests/unit/train/models tests/unit/test_configs.py` (`215 passed, 3 skipped`), including a new `tests/unit/train/models/test_fusions.py`.

## Notes

- `pyproject.toml` and `uv.lock` pin `samsja/dion@99e5f7e`, which contains the merged matrix-partition support from samsja/dion#3.

## Verification on merged `main` (2026-09-04)

Branch merged with `main` (no conflicts) and re-verified on 8×H200 nodes via SLURM. `ruff check`/`ruff format --check` clean; `pytest tests/unit -m "not gpu"`: 536 passed, 1 skipped; `pytest tests/unit -m gpu`: 81 passed, 8 skipped.

**Numerics against an fp32 reference** (`samsja/mini-glm-moe`, single GPU, no FSDP, 512 random tokens; reference is HF `Glm4MoeForCausalLM` in fp32 eager). Whole-model outputs are not bitwise identical between fused and unfused bf16 — the packed GEMMs take a different accumulation path — but the fused model is exactly as close to fp32 as the unfused one:

| model (bf16) | logits mean abs diff vs fp32 | KL(fp32 ‖ model) | grad rel. err vs fp32: q/k/v · experts gate/up · rest |
|---|---|---|---|
| unfused | 7.64e-3 | 3.97e-5 | 4.52e-2 · 4.96e-2 · 4.48e-2 |
| fused `qkv` | 7.64e-3 | 3.82e-5 | 4.14e-2 · 4.81e-2 · 4.12e-2 |
| fused `gate_up` | 7.22e-3 | 3.83e-5 | 4.52e-2 · 5.19e-2 · 4.51e-2 |
| fused both | 7.27e-3 | 3.88e-5 | 4.31e-2 · 4.87e-2 · 4.27e-2 |
| HF bf16 sdpa (noise floor) | 7.72e-3 | 4.48e-5 | 4.48e-2 · 5.19e-2 · 4.46e-2 |

Fused vs unfused state dicts are bitwise identical after loading, and each packed parameter equals the concatenation of the parameters it replaces.

**Fake-data training matrix** (`mini-glm-moe`, 2 GPUs, EP=2, Muon, compile + AC on, `ckpt.interval=5`, 10 steps), all runs complete without errors:

| run | result |
|---|---|
| fused / unfused, optimizer offload on / off | ✅ losses agree to ≤ 5e-4 relative on every step |
| `shard_fused_on_dim1 = true`: load, resume (both offload paths), fused ↔ unfused resume, 10-step loss vs unfused | ✅ (verified with this code before the flag existed; see #3486 for the measurements) |
| fused ckpt → fused resume, fused ckpt → unfused resume | ✅ post-resume losses agree to ≤ 1e-3 relative |
| unfused ckpt → fused resume vs unfused ckpt → unfused resume | ✅ |
| the same three resumes with optimizer-state CPU offload | ✅ |
| DeepGEMM FP8 (`quantization.type=fp8`, `moe.compute.type=deepgemm_fp8`), fused and unfused | ✅ |

(Resumed runs deliberately diverge from the un-interrupted run: `FakeDataLoader` reseeds from a batch counter that restarts at 0, so a resume replays the first batches. Fused and unfused resumes see identical data and are compared against each other.)

**RL end to end** (`configs/ci/integration/reverse-text-moe/start.toml`, custom impl, EP=2, 2 trainer GPUs + 1 vLLM GPU, 20 steps, filesystem weight broadcast): fused `mismatch_kl/all/mean` 1.8e-4–2.2e-4 on every step vs 1.9e-4–2.1e-4 unfused.

**Fix included in this branch.** Before `fix(trainer): load packed parameters and their optimizer state explicitly`, the same runs showed the `qkv` fusion training from zero attention weights: the HF weight load wrote into the replicated q/k/v copies, every q/k/v tensor in a fused checkpoint was exactly 0, the fused/unfused step-1 loss differed by 20% reproducibly under both bf16 and FP8, and the RL run aborted with no trainable rollouts. A fused resume also lost the Muon momentum of every packed parameter (verified with lr=0: momentum after resume was 1.00× instead of 1.95× the saved value; both fusions with optimizer CPU offload, `qkv` without). With the fix, loaded weights are bitwise identical to the source and momentum is restored on both resume paths. Full optimizer offload (`model.full_offload`) with fusions was not exercised. MXFP8 expert compute needs Blackwell and could not be tested on H200.

## Cluster verification with merged Dion (2026-09-05)

Pinned `samsja/dion@99e5f7e` (merged samsja/dion#3) and refreshed `uv.lock`; its tree is identical to the prior fork pin. Synced with `uv sync --all-extras --inexact`, then ran GPU verification through Slurm on H200s:

- 10 prime-rl fusion/activation-checkpointing tests, 13 Dion single-GPU tests, and 4 Dion distributed tests passed.
- All **16 distributed checkpoint combinations** passed: fused/unfused save and load, default/dimension-1 sharding, and optimizer-state offload on/off. Canonical weights, optimizer state, and the next Muon update with identical supplied gradients were **bitwise exact**.
- **13 compiled EP=2 training/resume variants** completed on `samsja/mini-glm-moe` (512-token fake batches): both fusion modes, all four resume combinations, state offload, dimension-1 sharding, and DeepGEMM FP8. Fused/unfused BF16 losses differed by at most **0.031%** across the three initial steps.
- All **406 pretrained model tensors** were bitwise exact in the zero-LR checkpoint and after its offloaded resume. All 13 variants saved canonical model/optimizer projection names, with no packed `qkv_proj` or `gate_up_proj` entries.
- The real **2-step RL run with vLLM weight reloads** completed with mismatch KL **0.00019046 / 0.00019005**. Fake-data runs use random inference logprobs, so their mismatch KL is not a trainer/inference agreement metric.
- H200 module checks: gate/up outputs and matrix-weight gradients, plus Q/K/V outputs with/without bias, were exact. BF16 MoE input-gradient relative differences were ~0.29%. Bias reductions in an all-BF16 module varied ~1.5%, also reproduced by repeating the unfused calculation itself.

Slurm evidence: 147/148 (test suites), 149 (RL), 150/158 (checkpoint equivalence), 153 (training matrix), 157 (saved checkpoint inspection). Job 148's test suites passed before an overly strict ad-hoc BF16 bias assertion stopped that script; the numerical diagnostic was resolved separately. Full optimizer offload and Blackwell-only MXFP8 were not exercised.

## CI collection fix (2026-09-05)

The `reverse_text` integration job failed before running any tests because importing `prime_rl.utils.weights` first triggered a circular import through the model registry and `fusions`. Import `PreTrainedModelPrimeRL` inside `convert_state_dict_to_hf`, its only use, so loading weight utilities no longer eagerly imports the model registry.

Reproduced the failing import locally before the fix. After the fix, standalone collection of `test_reverse_text.py` succeeds (6 tests), the complete integration suite collects (38 tests), and all 6 fusion unit tests pass. Ruff lint/format checks and `git diff --check` pass.

---

> [!NOTE]
> **Medium Risk**
> Changes core training paths (FSDP sharding, weight load, DCP optimizer state, Muon) and defaults fusions on; incorrect packing/load logic would corrupt weights or momentum on resume, though extensive checkpoint/resume testing is described in the PR.
> 
> **Overview**
> Adds **runtime parameter fusions** (on by default via `[trainer.model.fusions]`) that pack related weights into single tensors so MoE **gate/up** and attention **q/k/v** run as one GEMM each, while **checkpoints and exports still use canonical names** (`gate_proj`, `q_proj`, etc.).
> 
> A new **`fusions` module** implements packing (`PackedParameterSpec`), state-dict hooks, HF/DCP load **write-back** for packed weights, and explicit **optimizer-state split/join** around distributed checkpoint save/load (including CPU-offload resume). Optional **`shard_fused_on_dim1`** wires FSDP2 **`shard_placement_fn`** for experimental zero-copy fused weight/optimizer I/O.
> 
> **Attention and MoE layers** advertise fusions through `supported_fusions`; forwards use **`project_qkv`** / a single grouped **`gate_up`** GEMM when fused. Fusions are **skipped with LoRA**. **Muon** gets **`matrix_partitions`** from packed params via an updated **dion** pin. Docs and unit tests cover config and canonical state-dict behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52e736c685485fdbca109c86721017ab8065f65d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MarioSieg:
LGTM! What about MXFP8

docs/advanced.md
```diff
@@ -82,6 +83,24 @@ token_chunk_size = 4096
 
 With DeepEP, gradient clipping is currently not supported. (`optim.max_norm` is set to `None` automatically.)
 
+### Runtime Fusions
+
+`model.fusions` packs parameters that are always computed together into one tensor, turning several GEMMs into one. Both fusions are on by default:
+
+- `gate_up` — each gated MoE expert's `gate_proj` and `up_proj` become one `[num_experts, 2 * intermediate_size, hidden_size]` weight, halving the routed-expert grouped GEMMs.
```
garrett361, line 90: Also apply to `FeedForward`? EDIT: reading further, I see various changes are needed if we were to apply fusion to `FeedForward`, since up/gate fusion relies of various internals of the moe layer that aren't satisfied by FeedForward

docs/advanced.md
```diff
@@ -82,6 +83,24 @@ token_chunk_size = 4096
...
+```toml
+[trainer.model.fusions]
+enabled = ["gate_up", "qkv"]   # [] disables
+```
+
+Fusions are runtime-only. Checkpoints keep the canonical parameter names and shapes, so a run can turn a fusion on or off at any point and still load its own checkpoints, and exported weights are unaffected. Only modules that support a fusion are packed; a requested fusion that no module supports logs a warning, or fails at startup with `raise_on_fail = true`. Fusions are skipped when LoRA is enabled.
+
+Muon receives the packed layout as matrix partitions and orthogonalizes each logical matrix on its own, so a packed parameter trains exactly as the parameters it replaces would — including per-projection learning-rate scaling for grouped-query attention — while keeping a single momentum tensor.
+
+The experimental `shard_fused_on_dim1 = true` shards fused 2-D weights along dim 1 under FSDP, which makes weight loading and checkpointing zero-copy: fused weights and their optimizer state are read and written in place rather than assembled into a full copy on each rank first. It requires `hidden_size` to be divisible by the FSDP shard mesh size.
```
garrett361: Why does sharding on dim1 achieve this? Avoids some conversion via `full_tensor` to  different  sharding pattern?

docs/advanced.md
```diff
@@ -82,6 +83,24 @@ token_chunk_size = 4096
 
 With DeepEP, gradient clipping is currently not supported. (`optim.max_norm` is set to `None` automatically.)
 
+### Runtime Fusions
+
+`model.fusions` packs parameters that are always computed together into one tensor, turning several GEMMs into one. Both fusions are on by default:
+
+- `gate_up` — each gated MoE expert's `gate_proj` and `up_proj` become one `[num_experts, 2 * intermediate_size, hidden_size]` weight, halving the routed-expert grouped GEMMs.
+- `qkv` — an attention layer's `q_proj`, `k_proj` and `v_proj` (and their biases) become one linear layer.
```
garrett361, line 91: Just an FYI that if we want this for deepseek v4 we'll need some special handling, as keys and value get locked together (single `Linear` for both) in that model.

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -128,19 +134,26 @@ def to_local(tensor: torch.Tensor) -> torch.Tensor:
         offsets = torch.cumsum(num_tokens_per_expert, dim=0, dtype=torch.int32)
         x_bf16 = x.bfloat16()
 
-        up_proj = to_local(self.up_proj).transpose(-2, -1)
-        up = self.grouped_gemm(x_bf16, up_proj.bfloat16(), offs=offsets)
+        if self.gate_up_proj is None:
+            up_proj = to_local(self.up_proj).transpose(-2, -1)
```
garrett361, line 138: Side note: I noticed we're redefining the `to_local` function on every MoE fwd pass. Should move it. Will do in a separate PR.

src/prime_rl/trainer/models/fusions.py
```diff
@@ -0,0 +1,304 @@
...
+        )
+        tensors = [getattr(projection, parameter_name) for projection in projections.values()]
+        setattr(
+            qkv_proj,
+            parameter_name,
+            nn.Parameter(spec.pack_logical_tensors(tensors), requires_grad=tensors[0].requires_grad),
+        )
+        register_packed_parameter_state_dict_hooks(module, spec)
+
+    module.qkv_proj = qkv_proj
```
garrett361, line 113: Checking: this only works because you already called `self.register_module("qkv_proj", None)` on the attn class, right?

src/prime_rl/trainer/models/fusions.py
```diff
@@ -0,0 +1,304 @@
...
+
+def apply_model_fusions(
+    model: nn.Module, requested_fusions: Sequence[str], raise_on_fail: bool = True
+) -> dict[str, int]:
+    """Apply each requested fusion to every module that supports it; returns the module count per fusion."""
+    fused_module_counts: dict[str, int] = {}
+    for fusion_name in requested_fusions:
+        num_fused_modules = 0
+        for module in model.modules():
+            fusion: RuntimeFusion | None = getattr(module, "supported_fusions", {}).get(fusion_name)
```
garrett361, line 129: nit: this is fine for an initial PR, but I don't love `getattr` checking for this kind of thing.  Prefer `isinstance` checking, but I also don't think too much class or protocol churn is warranted at this stage

garrett361: Similar comments for how we look for hard coded `"gate_proj"` and `"up_proj"` FQN strings

garrett361: I see this bites a little in that modifying this code to apply to `FeedForward` requires a decent number of changes.

src/prime_rl/utils/weights.py
```diff
@@ -98,6 +97,8 @@ def convert_state_dict_to_hf(model: nn.Module, state_dict: dict[str, Tensor]) ->
     Format detection uses the model's full key set, so a partial dict (one rank's
     slice) converts the same way as the full state dict would.
     """
+    from prime_rl.trainer.models.base import PreTrainedModelPrimeRL
```
garrett361, line 100: why?

tests/unit/train/models/test_fusions.py
garrett361, line 1: no tests for the optimizer code needed?

src/prime_rl/utils/weights.py
S1ro1: Introduced a circular import somewhere along the way

docs/advanced.md
S1ro1: Oh that should be easy, we can just add `supported_fusions = {}` on the module, if it's already fused we don't need to support it custom way I think

garrett361: yeah, shouldn't be too hard, but might also need to moves some `Linear`s to bare `nn.Parameter`s. Similar to what's needed for `FeedForward` support. Not hard, just noting it's not automatic.

garrett361: oh, nvm, it's the moe handling that assumes no Linears. Not attn. Anyway, details not worth discussing more here

garrett361 (approved):

PR #3457, opened 2026-09-01
Mario/lowprec interface

MarioSieg:


src/prime_rl/trainer/models/layers/lowprecision.py
```diff
@@ -0,0 +1,56 @@
...
+    @abstractmethod
+    def is_linear_shape_supported(self, in_features: int, out_features: int) -> bool: ...
+
+    @abstractmethod
+    def convert_linear(self, mod: nn.Linear) -> nn.Linear: ...
+
+    """Convert a high-precision linear to this recipe's low-precision linear."""
+
+
+def replace_all_linear_with_low_precision_linear(
```
garrett361, line 26: I think this `Linear` replacement recipe is a bit overly narrow for the direction prime-rl is headed.  E.g. the `GroupedExperts` doesn't use `Linear` instances at all, and I think other `Linear`s may be removed in the future, too (given @S1ro1's dislike of `Linear`)

https://github.com/PrimeIntellect-ai/prime-rl/blob/ffb845c13c44154503be73bdb881885d5044487b/src/prime_rl/trainer/models/layers/moe.py?plain=1#L84-L161

So I think we need to target some more fine grained GEMM/GroupedGEMM abstractions?

MarioSieg: Good point, I had a splitted version originally but I removed it after Mateij's MoE cleanup made it more clean generally, I'll make it more fine grained again.

PR #3461, opened 2026-09-01
fix(evals): show resumed cursor progress

samsja:
## Summary

- resumed standalone evals show progress against the full epoch instead of only the remaining suffix: the pipeline view derives shown totals from the full batch size (`EvalSink.batch_size_for`), so a run resumed at cursor 80/100 prints `80/100` instead of `0/20`; sink completion thresholds stay suffix-only so resumed epochs finish without re-waiting on skipped tasks
- `EvalSink.batch_progress` also reports expected-but-empty batches, so an env shows `0/N (0.0%)` immediately after its epoch fires, before any result arrives

Checkpoint format is unchanged — it still saves only the cursor.

---

> [!NOTE]
> **Low Risk**
> Display and progress-reporting only; eval batch completion thresholds and checkpoint format are unchanged.
> 
> **Overview**
> **Resumed standalone evals** now report epoch progress against the **full** batch size in the periodic pipeline view, not only the remaining suffix after a checkpoint cursor. `collect_pipeline_view` maps sink tuples (`arrived` / `expected` for the outstanding work) into `shown_arrived/shown_expected` using `EvalSink.batch_size_for`, so a resume at cursor 80/100 logs **80/100** instead of **0/20**. Completion logic in the sink is unchanged—it still waits only on the suffix.
> 
> **`EvalSink.batch_progress`** also unions in `expected_batch_sizes`, so an env appears as **0/N (0.0%)** as soon as its epoch is triggered, before any episodes land. A unit test covers that expected-but-empty batch case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4db44cc600b01df18a118da34e23a381f6829577. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/prime_rl/evals/evals.py
+++ b/src/prime_rl/evals/evals.py
@@ -551,10 +551,11 @@ class Evals:
         disp_drain = self.dispatcher.metrics.drained(train_envs=set(), eval_envs={env.name for env in self.eval_envs})
 
         parts = []
-        for env_name, step, arrived, expected, buffered in sorted(self.eval_sink.batch_progress()):
-            skipped = self.eval_source.skipped_task_count(env_name, step) * self.eval_sink.group_size_for(env_name)
-            shown_arrived = skipped + arrived
-            shown_expected = skipped + expected
+        for env_name, _step, arrived, expected, buffered in sorted(self.eval_sink.batch_progress()):
+            # ``expected`` is only the remaining suffix of a resumed epoch; add the
+            # cursor-skipped prefix back so progress shows against the full epoch.
+            total = self.eval_sink.batch_size_for(env_name)
+            shown_arrived, shown_expected = total - expected + arrived, total
             part = (
                 f"{env_name} {shown_arrived}/{shown_expected} ({shown_arrived / shown_expected:.1%})"
                 if shown_expected
--- a/src/prime_rl/evals/evals.py
+++ b/src/prime_rl/evals/evals.py
@@ -552,8 +552,6 @@ class Evals:
 
         parts = []
         for env_name, _step, arrived, expected, buffered in sorted(self.eval_sink.batch_progress()):
-            # ``expected`` is only the remaining suffix of a resumed epoch; add the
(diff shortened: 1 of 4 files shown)


src/prime_rl/evals/evals.py
```diff
@@ -551,8 +551,15 @@ def collect_pipeline_view(self) -> tuple[str, dict[str, float]]:
...
-        for env_name, _step, arrived, expected, buffered in sorted(self.eval_sink.batch_progress()):
-            part = f"{env_name} {arrived}/{expected} ({arrived / expected:.1%})" if expected else env_name
+        for env_name, step, arrived, expected, buffered in sorted(self.eval_sink.batch_progress()):
+            skipped = self.eval_source.skipped_task_count(env_name, step) * self.eval_sink.group_size_for(env_name)
+            shown_arrived = skipped + arrived
+            shown_expected = skipped + expected
+            part = (
+                f"{env_name} {shown_arrived}/{shown_expected} ({shown_arrived / shown_expected:.1%})"
+                if shown_expected
+                else env_name
```
samsja: bot sure about this part cant we just keep a total request that we need to process the same across turns and in the ckpot just save cursor whcih is also the current data amout we have process ?

src/prime_rl/evals/evals.py
```diff
@@ -552,7 +552,15 @@ def collect_pipeline_view(self) -> tuple[str, dict[str, float]]:
 
         parts = []
         for env_name, _step, arrived, expected, buffered in sorted(self.eval_sink.batch_progress()):
-            part = f"{env_name} {arrived}/{expected} ({arrived / expected:.1%})" if expected else env_name
+            # ``expected`` is only the remaining suffix of a resumed epoch; add the
```
hallerite: can we remove the comment

samsja: done — removed in 4db44cc60

hallerite (approved):

PR #3463, opened 2026-09-01
feat: constant trainer batch size

mikasenghaas:
## Summary

- Add `orchestrator.constant_trainer_batch_size` with a default of `true`.
- Backfill filtered samples when constant batches are enabled.
- Keep faster, variable-size batches available when the option is disabled.
- Always filter samples with no training signal before trainer delivery.
- Warn about sustained trainer waits and orchestrator waits.
- Warn about high episode discard rates without counting buffered work.
- Report stale, errored, and no-signal discard counts.
- Update the training docs and run-monitoring skill.

## Breaking

- Remove `orchestrator.train.filter_zero_advantages`. Delete this field from existing configs. The orchestrator now always filters zero-advantage RL tokens.

## Verification

Before, zero-advantage filtering ran after batch selection. The orchestrator did not replace filtered samples. The trainer could receive fewer samples than `orchestrator.batch_size`.

Paired reverse-text runs used five steps, batch size 64, group size 4, sequence length 512, and 128 maximum completion tokens.

| `constant_trainer_batch_size` | Effective episodes by step | Mean orchestrator step | Total orchestrator time |
|---|---|---:|---:|
| `true` | 64, 64, 64, 64, 64 | 8.76s | 43.79s |
| `false` | 40, 40, 36, 44, 52 | 6.49s | 32.45s |

The default constant mode produced exactly 64 effective episodes on every step. It was 35% slower because it generated replacement episodes.

Additional five-step reverse-text runs exercised the warning paths:

- An 8-second local orchestrator delay triggered the trainer-wait warning.
- An 8-second local trainer delay triggered the orchestrator-wait warning.
- `max_off_policy_steps = 0` triggered discard warnings. One warning reported 48/56 discarded episodes: 40 stale, 0 errored, and 8 with no signal.

The local delays were temporary verification changes. They are not in this diff.

`uv run pytest tests/unit/test_configs.py tests/unit/orchestrator/test_batch.py tests/unit/orchestrator/test_metrics.py -q` passed with 172 tests.

`uv run pytest tests/unit/test_configs.py tests/unit/orchestrator/test_batch.py tests/unit/orchestrator/test_metrics.py tests/unit/orchestrator/test_advantage.py` passed with 182 tests.

`uv run ruff check` and `uv run ruff format --check` passed for all changed Python files.

---

> [!NOTE]
> **Medium Risk**
> Changes default batch assembly and removes a config knob, which can shift rollout compute vs trainer step size; behavior is documented but existing configs must drop `filter_zero_advantages`.
> 
> **Overview**
> Adds **`orchestrator.constant_trainer_batch_size`** (default **`true`**) so rollout batches still reach **`orchestrator.batch_size`** effective traces after dropping samples with no training signal (e.g. all-zero RL advantages). With constant mode, zero-advantage pruning happens **before** traces enter the pending batch so the sink keeps collecting replacements; with **`false`**, pruning runs at ship time for faster collection but **variable** trainer batch sizes.
> 
> **Breaking:** removes **`orchestrator.train.filter_zero_advantages`** — zero-advantage RL token filtering is always applied; only the timing/backfill behavior is controlled by the new flag.
> 
> **`TrainBatch`** now carries **`buffered_episode_ids`**, **`cancelled_attempts`**, and **`stale_attempts`** so the orchestrator can warn when more than half of a window’s attempts are discarded, split into **stale**, **errored**, and **no_signal**. The trainer and orchestrator also log imbalance warnings when **`time/wait_for_batch`** or **`time/wait_for_policy`** dominates active step time. Docs and the monitor-run skill document the knob and **`time/wait_for_policy`** (replacing **`time/wait_for_ckpt`** in guidance).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2befb8878975ab6408793840cf8dd6e4ae9a0f1c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/training.md
+++ b/docs/training.md
@@ -133,7 +133,9 @@ Pulled from the console logs and mirrored to W&B.
 | Source | Metric | Reading |
 |---|---|---|
 | trainer | `time/wait_for_batch` | **high → orchestrator bottleneck** |
-| orchestrator | `time/wait_for_ckpt` | **high → trainer bottleneck** |
+| orchestrator | `time/wait_for_policy` | **high → trainer bottleneck** |
+
+The trainer warns when batch wait time exceeds active trainer time. Add inference nodes when this warning persists. The orchestrator warns when policy wait time exceeds active orchestrator time. Add trainer nodes when this warning persists. The orchestrator also warns when it discards more than half of an episode window. This warning reports stale, errored, and no-signal counts.
 
 ## SFT Trainer
 
--- a/docs/training.md
+++ b/docs/training.md
@@ -58,7 +58,7 @@ A condensed view of the knobs you'll most often tune. For trainer-side paralleli
 | Knob | What it does |
 |---|---|
 | `orchestrator.batch_size` | Tasks per trainer step. |
-| `orchestrator.constant_trainer_batch_size` | Collect replacement samples after zero-advantage filtering. Enabled by default. Disable it for faster collection with variable trainer batch sizes. |
+| `orchestrator.constant_trainer_batch_size` | Keep trainer batches constant when samples have no training signal, such as zero advantage on all tokens. Enabled by default. Disable it for faster collection with variable trainer batch sizes. |
 | `orchestrator.group_size` | Rollouts generated per task. |
 | `orchestrator.max_off_policy_steps` | Maximum staleness of a trained rollout (default 8): the version a batch trains on minus the oldest version that generated the rollout, queue time included. Episodes past the bound are dropped; a group shares one dispatch version, so its episodes age out together. The main off-policy dial on long agentic rollouts — bump for throughput, lower for tighter on-policyness. Watch `off_policy/*` and `mismatch_kl/all/mean` when tuning. |
 | `[orchestrator.algo]` | Training algorithm — its `type` names it (`grpo` default, `max_rl`, `rae`, `hierarchical_grpo`, `opd`, `opsd`, `sft`, `echo`). See [Algorithms](#algorithms). |
--- a/docs/training.md
+++ b/docs/training.md
@@ -58,6 +58,7 @@ A condensed view of the knobs you'll most often tune. For trainer-side paralleli
 | Knob | What it does |
 |---|---|
 | `orchestrator.batch_size` | Tasks per trainer step. |
+| `orchestrator.constant_trainer_batch_size` | Collect replacement samples after zero-advantage filtering. Enabled by default. Disable it for faster collection with variable trainer batch sizes. |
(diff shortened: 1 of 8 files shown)


docs/training.md
```diff
@@ -58,6 +58,7 @@ A condensed view of the knobs you'll most often tune. For trainer-side paralleli
 | Knob | What it does |
 |---|---|
 | `orchestrator.batch_size` | Tasks per trainer step. |
+| `orchestrator.constant_trainer_batch_size` | Collect replacement samples after zero-advantage filtering. Enabled by default. Disable it for faster collection with variable trainer batch sizes. |
```
mikasenghaas: > Collect replacement samples after zero-advantage filtering

this is not intuitive. say smth like "Filters out samples which carry no training signal (e.g. zero advantage on all tokens)"

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -305,7 +305,7 @@ class TrainConfig(BaseConfig):
     """Shared training sampling configuration."""
 
     filter_zero_advantages: bool = True
```
mikasenghaas: this is redundant now no? i think this should be default and should always be on -- if we don't have constant trainer batch size we should still filter zero adv samples before the batch is shipped so we dont waste fwd/bwd passes on it

skills/training/monitor-run/SKILL.md
```diff
@@ -152,6 +152,10 @@ All metrics print to the console log (and W&B when configured).
 | orchestrator | `off_policy/{mean,max}`, `off_policy/{in_flight,in_queue}/{mean,max}`, `off_policy/dropped` | per-step staleness of trained rollouts |
 | env server | event loop lag (min/mean/p90/p99/max), active task distribution | periodic |
 
+The trainer warns when batch wait time exceeds active trainer time. Add inference nodes when this warning persists. The orchestrator warns when policy wait time exceeds active orchestrator time. Add trainer nodes when this warning persists. The orchestrator also warns when it discards more than half of an episode window and reports stale, errored, and no-signal counts.
+
+`orchestrator.constant_trainer_batch_size` defaults to `true`. It replaces samples removed by zero-advantage filtering. Set it to `false` for faster collection with variable trainer batch sizes.
```
mikasenghaas: > It replaces samples removed by zero-advantage filtering. 

dont narrate deprecated things

samsja (approved):

PR #3466, opened 2026-09-02
fix(trainer): zero-index position_ids fallback in LM-head wrapper

garrett361:
Use the standard zero-indexing convention for `position_ids` in patched LM head forward,  (`_patch_model_forward`), rather than one-indexing. 

A uniform shift is mathematically invisible to RoPE, which sees only query-key angle differences, but it is wrong for attention that keys off absolute token indices (as is important for deepseek v4 where compression boundaries are tied to absolute positions).


S1ro1 (approved):
lgtm

PR #3468, opened 2026-09-02
Overlap shared expert FFN with Communication

MarioSieg:
This PR improves performance by overlapping EP A2A with the shared-expert FFN computatione by using a side CUDA stream.
MoE.forward runs the shared-expert FFN after token_dispatcher.run even tough there is no data dependency directly.

This optimization only works on models which have a shared-expert like (GLM4/5, Qwen 3.5 MoE) but not on Qwen3 or GPT OSS.
In a real SFT run with QWen3.5 MoE 35B with full 40 layers (seq len =8k) the average step time of 15 benchmarked is reduced by 7%.
We go from 58 s to 54 s and from 1110 tps to 1260 tps.

So + 7% faster but the peak memory also went up +4.9GB.

---

> [!NOTE]
> **Medium Risk**
> Changes the MoE forward execution model on the EP hot path (CUDA streams and tensor lifetime), though parity is tested; enabling it can also increase peak GPU memory.
> 
> **Overview**
> Adds an optional **`overlap_shared_expert`** flag on **`[trainer.model.moe.dispatch]`** when `type = "torch"`. The trainer wires it into custom MoE layers so, on CUDA with a non-local token dispatcher (expert-parallel all-to-all), the **shared expert FFN** can run on the default stream while **dispatch → routed experts → combine** runs on a dedicated side CUDA stream instead of running the shared expert only after `token_dispatcher.run`.
> 
> The overlapped path splits `TokenDispatcher` into explicit `dispatch` / `combine` steps with stream events and `record_stream` bookkeeping; it is disabled for `LocalTokenDispatcher` (no comms to hide) and is a no-op without a shared expert. Docs describe the knob; a GPU unit test checks forward outputs and gradients match the sequential path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5277219a8233c64c62bfe1f98fa2ee07cbc9ca4c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/prime-rl-configs/src/prime_rl/configs/trainer.py
```diff
@@ -202,6 +202,9 @@ class TorchMoEDispatchConfig(BaseConfig):
     transport: Literal["bf16", "mxfp8"] = "bf16"
     """Wire format for routed activations and their reverse-path gradients."""
 
+    overlap_shared_expert: bool = False
+    """Run the shared expert FFN on a different CUDA steam than the reouted dispatch/combine a2a. Requires a modle with shared experts, no-op otherwise."""
```
garrett361: nit: modle -> model

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -19,6 +19,18 @@
 
 ScoreFuncType = Literal["softmax", "sigmoid", "topk_softmax"]
 
+_comm_stream: torch.cuda.Stream | None = None
```
garrett361: nit: since we might have more streams in the future, let's use a more descriptive name. `_shared_expert_overlap_stream` or similar

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -424,19 +442,22 @@ def forward(
...
             )
 
-        routed_output = self.token_dispatcher.run(
-            self.prepare_expert_input(x),
-            top_scores,
-            selected_experts_indices,
-            self.experts,
-            score_before_experts=self.score_before_experts,
-        )
+        if self._overlap_shared_expert and self.shared_expert is not None and x.is_cuda:
```
garrett361: I think we should also have a check that we're not using `LocalTokenDispatcher`, since there are no comms in that case (no EP).

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -445,6 +466,40 @@ def forward(
...
+        master_stream = torch.cuda.current_stream()
+        comm_stream.wait_stream(master_stream)
+        with torch.cuda.stream(comm_stream):
+            routed_input, num_tokens_per_expert, state = self.token_dispatcher.dispatch(
+                self.prepare_expert_input(x),
+                top_scores,
+                selected_experts_indices,
+                score_before_experts=self.score_before_experts,
+            )
+            dispatch_done = comm_stream.record_event()
```
garrett361, line 494: I think this section is missing several `record_stream` calls that are required for generic correctness. Any tensor `t` that's created on a non-`comm_stream` stream needs `t.record_stream(comm_stream`) called on it so that the caching allocator knows not to deallocate the underlying memory before it gets used.  So, this applies to all of `{x,top_scores,selected_expert_indices}`


```suggestion
            x.record_stream(comm_stream)
            top_scores.record_stream(comm_stream)
            selected_experts_indices.record_stream(comm_stream)
            routed_input, num_tokens_per_expert, state = self.token_dispatcher.dispatch(
                self.prepare_expert_input(x),
                top_scores,
                selected_experts_indices,
                score_before_experts=self.score_before_experts,
            )
            dispatch_done = comm_stream.record_event()
```

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -445,6 +466,40 @@ def forward(
...
+        shared_output = self.shared_expert(x)
+        master_stream.wait_event(dispatch_done)
+        routed_input.record_stream(master_stream)
```
garrett361, line 497: Similarly `record_stream` is needed on `num_tokens_per_expert` before consuming on the master stream:

```suggestion
        routed_input.record_stream(master_stream)
        num_tokens_per_expert.record_stream(master_stream)
```

Tensors inside `state` don't need `record_stream` in this impl because it's only ever consumed on the same comms stream that created it.

src/prime_rl/trainer/models/layers/moe.py
```diff
@@ -445,6 +466,40 @@ def forward(
...
+        routed_output = self.experts(routed_input, num_tokens_per_expert)
+        comm_stream.wait_stream(master_stream)
+        with torch.cuda.stream(comm_stream):
```
garrett361, line 501: I think this can all be simplified down, also, to use a single stream ctx manager. Basically
1. Sync streams
2. Launch shared expert compute on main
3. create ctx manager and do all routed expert comms/compute there
4. sync back up

Something like:

```py
@torch.compiler.disable()
    def _dispatch_experts_overlapped(
        self,
        x: torch.Tensor,
        top_scores: torch.Tensor,
        selected_experts_indices: torch.Tensor,
    ) -> tuple[torch.Tensor, torch.Tensor]:
        # Overlap the shared expert with the routed dispatch/combine a2a on a different CUDA stream.
        # Autograd replays the op on its own original stream so this overlap works in backward too.
        comm_stream = _get_comm_stream()
        master_stream = torch.cuda.current_stream()
        comm_stream.wait_stream(master_stream)
        shared_output = self.shared_expert(x)
        with torch.cuda.stream(comm_stream):
            x.record_stream(comm_stream)
            top_scores.record_stream(comm_stream)
            selected_experts_indices.record_stream(comm_stream)
            routed_input, num_tokens_per_expert, state = self.token_dispatcher.dispatch(
                self.prepare_expert_input(x),
                top_scores,
                selected_experts_indices,
                score_before_experts=self.score_before_experts,
            )
            routed_output = self.experts(routed_input, num_tokens_per_expert)
            routed_output = self.token_dispatcher.combine(routed_output, state)
            self.token_dispatcher.synchronize() # GG: actually unclear where this is intended to be called, mostly dead code today
        routed_output.record_stream(master_stream)
        master_stream.wait_stream(comm_stream)
        return routed_output, shared_output
```

tests/unit/train/models/test_moe.py
```diff
@@ -162,3 +162,42 @@ def test_expert_type_and_activation_are_independent(expert_type, activation):
...
+        moe = MoE.from_args(
+            MoEArgs(num_experts=4, top_k=2, load_balance_coeff=None),
+            dim=16,
+            hidden_dim=32,
+            shared_expert=FeedForward(dim=16, hidden_dim=32),
+        ).to("cuda")
+        moe.experts.init_weights(0.02)
+        moe.router.init_weights(0.02)
+        moe.shared_expert.init_weights(0.02)
+        moe.set_token_dispatcher(LocalTokenDispatcher(num_experts=4, top_k=2, token_group_alignment=1))
```
garrett361: Kind of confused about the utility of this test where there's no actual comms. I guess w/ multiple streams and no real compute, you could potentially get overlap if neither kernel uses all of the SMs?

garrett361 (commented):
Left some comments on the technical side, haven't gotten to think about the design side much yet.

PR #3471, opened 2026-09-02
feat: consolidate eval cli

mikasenghaas:
## Summary

Migrates the eval CLI from verifiers into the prime-rl ecosystem: `uv run eval` is prime-rl's entrypoint for evaluating any model on any environment, standalone or as the online evals of a training run.

Why:

- consolidation of features: adaptive concurrency, the local dashboard, resume from the trace stream, live rollouts
- easier code sharing between the CLI entrypoints (`rl`, `sft`, `eval`)
- direct access to the pre-installed environments (`prime-envs` submodule)
- one workspace for post-training: training and evals

Important:

- the verifiers eval CLI stays for a transition period as the quick in-process path, renamed `uv run vf-eval` so prime-rl owns `eval` (PrimeIntellect-ai/verifiers#2610 removes it later); env servers and their streamed traces (PrimeIntellect-ai/verifiers#2512, merged; the submodule points at verifiers `main`) are consumed here only
- evals run through env servers, for scalability; the launcher (`eval`, `sft`, `rl`) owns them
- `--resume` continues an interrupted run from its trace stream: landed episodes rejoin the epoch, only the owed rollouts run; `num_examples`/`group_size` may change, the model, sampling and env config may not
- multi-env capable (`[[source]]` blocks); the single-env command shorthands are preserved
- adaptive concurrency only against local deployments (it reads vLLM `/metrics`); against an API, concurrency is pinned via `min_inflight = max_inflight` (the default pins 128)
- prime-rl's core dependencies are plain CPU; install them and the environments with `uv sync --all-packages`
- Prime runs and traces integrate through the single monitor abstraction (`--monitors.prime`, `--monitors.wandb`, the file monitor)

Examples:

```bash
uv run eval gsm8k                                            # default model on Prime Inference
uv run eval gsm8k -n 32 -r 4 -m openai/gpt-5.6-luna           # 32 tasks × 4 rollouts, another model
uv run eval terminal-bench-2 --env.taskset.tasks '["fix-git"]' --env.agent.harness.id bash -n 1 -r 2
uv run eval @ configs/debug/eval/single-turn.toml            # gsm8k, null harness
uv run eval @ configs/debug/eval/multi-turn.toml             # 16 terminal-bench-2 tasks, bash harness
uv run eval @ configs/debug/eval/multi-env.toml              # terminal-bench-2 under bash and rlm side by side
uv run eval @ configs/debug/eval/aime2026.toml               # AIME 2026, avg@16
uv run eval @ configs/debug/eval/tb2.toml                    # Terminal-Bench 2, avg@4
uv run eval @ configs/debug/eval/resume.toml --run.name x    # then: --resume
uv run eval gsm8k --monitors.prime                           # stream the evaluation to the platform
```

## Breaking

- `uv run evals` is `uv run eval` (`EvalsConfig` → `EvalConfig`); config shape: `[eval.client]` → `[client]`, `[eval.concurrency]` → `[concurrency]`, `[[eval.source]]` → `[[source]]`, `eval.num_examples`/`eval.group_size` → top level. `[online]` is gone: online evals are spawned by `sft`.
- `eval` has no `[ckpt]` block; `--resume` is a switch that reads the run's trace stream.
- SFT online evals: no public entrypoint (`uv run sft` with an `[eval]` block spawns it); `logs/attempt_N/evals.log` → `eval.log`, `evals.json` → `eval.json`, W&B label `online-eval`; `EvalsEvalConfig` → `SFTOnlineEvalConfig`.
- `ratio` is accepted on training sources only, `interval` on online eval sources only; a standalone eval's `[[source]]` takes neither.
- `orchestrator.env_server_base_port` and the eval entrypoint's `env_server_base_port` are removed: launcher-managed env servers bind an OS-assigned port and publish it (`configs/attempt_N/resolved/envs/<split>/<name>.address`); `serve.address` pins one. `uv run env-server` without `serve.address` binds an OS-assigned port (`address_file` publishes it).
- `configs/evals/swe.toml` is removed (`uv run eval swebench-verified --env.agent.harness.id bash`).

## Verification

- Unit: `uv run pytest tests/unit` green (configs, orchestrator, eval, monitors incl. the live reader).
- Debug configs: single-turn, multi-turn, multi-env and aime2026 run to completion against Prime Inference. Resume: single-env and multi-env runs interrupted with SIGTERM and resumed hold exactly examples × group_size unique episodes with the plan and epoch metrics over the full set; resuming with a larger `group_size` runs only the extra rollouts; resuming with another model is refused.
- Core install (`uv sync` without extras) runs `uv run eval`; an SFT run with online evals starts one env server per source.
- Live streaming, on the final code: gsm8k 16 single-turn episodes, terminal-bench-2 fix-git with bash (13-14 turns, the pending preview alternating with committed turns in the live files), kuhn-poker 4 two-seat episodes (8 traces), a 5-step reverse-text RL run (768 train episodes with token ids, logprobs and masks); no assembly mismatch.
- Platform: `--monitors.prime` on a two-rollout fix-git eval opened the evaluation 9 s after launch, streamed both rollouts and closed it; the dashboard linked to it throughout.
- Dashboard behaviour checked in headless Chromium: live rows and viewer follow, handover in place, metrics pane per env, filter menus, platform button states.

---

> [!NOTE]
> **Medium Risk**
> Breaking CLI and config renames affect existing scripts; env-server address discovery replaces fixed ports across `rl`, `sft`, and `eval` launch paths.
> 
> **Overview**
> **Replaces `uv run evals` with `uv run eval`** as the standalone evaluation entrypoint, sharing the orchestrator eval pipeline (per-source env servers, adaptive or pinned concurrency, file/W&B/Prime monitors). Config is **flattened** (`[[source]]`, `[client]`, `[concurrency]` at top level); **`--resume`** continues from the trace stream instead of a `[ckpt]` cursor; SFT **online eval** is renamed and typed as `SFTOnlineEvalConfig` (no public `[online]` block).
> 
> **Env server wiring changes:** fixed `env_server_base_port` ranges are removed—launcher-managed servers bind **OS-assigned loopback ports** and publish addresses via **`address_file`** / `configs/.../envs/<split>/<name>.address`. Training **`ratio`** and online-eval **`interval`** are scoped to the right source types only.
> 
> **Docs and examples** add `docs/eval.md`, per-example `eval.toml` files, and `configs/debug/eval/*` smokes; **`configs/evals/swe.toml` is deleted**. CI adds a **gsm8k eval** integration test on VM runners.
> 
> **Dashboard** gains live rollout APIs and eval-centric metrics (plan-driven progress, platform links, eval run status), renames **`evals.log` → `eval.log`**, and tightens static asset revalidation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82f2f0eda7628c874be6dcfcda700cab26fd4442. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/configuration.md
```diff
@@ -100,14 +100,12 @@ Overlay TOMLs **replace** lists wholesale — an overlay that wants to add one i
 CLI takes a JSON literal. TOML uses a table. CLI dicts deep-merge with TOML dicts — CLI keys win on conflict but don't wipe the file's keys:
 
 ```bash
-uv run rl @ rl.toml --orchestrator.train.source.0.args \
-  '{"dataset_name": "openai/gsm8k", "dataset_subset": "main"}'
+uv run rl @ rl.toml --env-vars '{"HF_HUB_OFFLINE": "1"}'
```
mikasenghaas: why are those changes in this pr?

docs/training.md
```diff
@@ -37,6 +38,8 @@ This page covers everything you need to launch, observe, checkpoint, and recover
 | `uv run inference` | vLLM server. | Always use this entrypoint over `vllm serve` — it adds `/update_weights`, `/load_lora_adapter`, and `/init_broadcaster`. |
 | `uv run trainer` | Standalone trainer process group. | Use only when launching the trainer separately from the orchestrator (e.g. multi-node RL without the `rl` wrapper). |
 | `uv run orchestrator` | Standalone orchestrator process. | Pair with a separately-launched trainer, inference, and one `env-server` per source. |
+| `uv run eval` | Multi-env evals against a live inference server. | One epoch per source, adaptive concurrency, cursor checkpoints + `--resume`, dashboard + optional platform upload; see [Evals](#evals). |
+| `uv run online-eval` | Evals per trainer weight broadcast. | Spawned by `uv run sft` for [Online Evals](#online-eval); standalone only with filesystem broadcasts. |
```
mikasenghaas: this should not be a public entrypoint (its interanl to the sft entrypoint)

docs/training.md
```diff
@@ -260,6 +263,40 @@ Pulled from the console log and mirrored to W&B.
 | `perf/peak_memory` | peak GPU memory (GiB) |
 | `time/step`, `time/forward_backward`, `time/save_ckpt` | step breakdown |
 
+## Evals
```
mikasenghaas: make a new .md for evals. it should be called eval.md -- also add a skill which should be similar to the training skill

examples/eval/agentic-judge.toml
mikasenghaas, line 1: remove

examples/eval/best-of-n.toml
mikasenghaas, line 1: remove

examples/eval/rlm-docker.toml
mikasenghaas, line 1: remove

examples/eval/terminal-bench-2.toml
mikasenghaas, line 1: remove for now

examples/eval/agentic-judge.toml
mikasenghaas, line 1: acc lets put all the eval configs into `configs/debug` -- examples should be e2e training examples which include evals. e.g. the basic reverse text would have a eval.toml, sft.toml and rl.toml

packages/prime-rl-configs/src/prime_rl/configs/eval.py
```diff
@@ -0,0 +1,175 @@
...
+        if self.run.dir is None:
+            self.run.dir = self.run.name
+        if self.monitors.wandb is not None and self.monitors.wandb.name is None:
+            self.monitors.wandb.name = self.run.name
+        if self.monitors.prime is not None and self.monitors.prime.name is None:
+            self.monitors.prime.name = self.run.name
+        return self
+
+
+class SFTEvalConfig(ScheduledEvalConfig, ServedEvalConfig):
```
mikasenghaas: why do we need this class, can this not just be in online eval config?

packages/prime-rl-configs/src/prime_rl/configs/monitors.py
```diff
@@ -48,6 +48,11 @@ class PrimeMonitorConfig(BaseConfig):
     """Run name shown on the platform. Inherits ``run.name`` when unset."""
```
mikasenghaas, line 48: lets make the monitor structure 

`TrainMonitorsConfig` and `EvalMonitorsConfig` each having `{Source}{Train,Eval}MonitorConfig`, e.g. `PrimeTrainMonitorConfig` and `PrimeEvalMonitorConfig`

packages/prime-rl-configs/src/prime_rl/configs/orchestrator.py
```diff
@@ -393,6 +380,30 @@ def validate_unique_env_names(self):
         return self
 
 
+class ScheduledEvalConfig(EvalSourcesConfig):
```
mikasenghaas, line 387: can we also call this `RLOnlineEvalConfig`?

mikasenghaas: and then in sft it would be called `SFTOnlineEvalConfig`

skills/training/monitor-run/SKILL.md
```diff
@@ -82,7 +82,8 @@ Verify liveness with `curl -sf <url>/api/runs` and hand the researcher the `url`
 {run_dir}/logs/latest/
 ├── trainer.log                # rank 0 stdout
 ├── orchestrator.log           # orchestrator stdout
-├── evals.log                  # SFT online-eval evals stdout
+├── eval.log                  # `uv run eval` process
+├── online-eval.log           # SFT online-eval process
```
mikasenghaas: both online eval and eval should write to `eval.log`

skills/training/start-run/SKILL.md
```diff
@@ -95,42 +95,60 @@ curl http://localhost:8000/v1/chat/completions \
 - Entrypoint: `src/prime_rl/entrypoints/inference.py`
 - SLURM: single-node, multi-node, and disaggregated deployments
 
-## `evals` — multi-env evals
+## `eval` — multi-env evals
```
mikasenghaas: move this into standalone eval skill

pyproject.toml
```diff
@@ -42,7 +43,8 @@ sft = "prime_rl.entrypoints.sft:main"
 inference = "prime_rl.entrypoints.inference:main"
 trainer = "prime_rl.entrypoints.trainer:main"
 orchestrator = "prime_rl.entrypoints.orchestrator:main"
-evals = "prime_rl.entrypoints.evals:main"
+eval = "prime_rl.entrypoints.eval:main"
+online-eval = "prime_rl.entrypoints.online_eval:main"
```
mikasenghaas: no public entrypoint

tests/integration/test_reverse_text_eval.py
```diff
@@ -0,0 +1,111 @@
...
+            if httpx.get(url, timeout=2.0).status_code == 200:
+                return
+        except (httpx.ConnectError, httpx.ReadTimeout):
+            pass
+        time.sleep(1.0)
+    raise TimeoutError(f"Inference server at {url} did not become ready within {timeout_s}s")
+
+
+@pytest.fixture(scope="module")
+def inference(output_dir: Path) -> Generator[subprocess.Popen, None, None]:
```
mikasenghaas: lets run eval ci against pi inference, not against vllm

mikasenghaas: then it can also be cpu

.github/workflows/cpu_tests.yaml
```diff
@@ -43,6 +43,37 @@ jobs:
       - name: Cleanup output_dir
         run: rm -rf /tmp/outputs
 
+  eval-test:
```
mikasenghaas: move this into regular integration tests, its fine if it runs on gpu

configs/ci/integration/gsm8k-eval.toml
```diff
@@ -0,0 +1,13 @@
...
+# default model on Prime Inference — the eval path end to end without a GPU or sandbox.
+num_examples = 4
+dashboard = false
+
+[sampling]
+max_completion_tokens = 8192
+
+[[source]]
+env.taskset.id = "gsm8k"
+env.agent.harness.id = "bash"
```
mikasenghaas: use null harness, then we dont need turn cap as well

docs/eval.md
```diff
@@ -0,0 +1,125 @@
+# Eval
+
+This page covers `uv run eval` — evaluating one or more environments against a live inference server — and where the online evals of a training run come from. For launching and observing training runs, see [Training](training.md).
```
mikasenghaas: make eval.md not deal with online evals. this should purely be for offline evals.

mikasenghaas: also dont say "live inference server" just say evaluating a model in one or more environments

docs/eval.md
```diff
@@ -0,0 +1,125 @@
...
+- [Online Evals](#online-evals)
+- [Metrics](#metrics)
+
+## Standalone Evals
+
+`uv run eval` evaluates one or more environments against a live inference server and exits after one epoch per source. It reuses the orchestrator's eval pipeline: env servers are spawned per source, episodes are admitted under the concurrency controller (pinned at 128 by default), and every episode streams into the run's trace stream and metrics.
+
+### Launch
+
+The default client is Prime Inference with `deepseek/deepseek-v4.1-flash`. Authenticate with `PRIME_API_KEY` or `prime login`:
```
mikasenghaas: dont mention specific defaults

mikasenghaas: do mention that we default to pi infernece which requires authentication

docs/eval.md
```diff
@@ -0,0 +1,125 @@
...
+To evaluate a model you serve yourself, start a `uv run inference` vLLM server (or any OpenAI-compatible API) and point the client at it:
+
+```bash
+uv run inference --vllm.model Qwen/Qwen3-4B
+uv run eval gsm8k -n 32 -r 4 -m Qwen/Qwen3-4B --client.base_url http://localhost:8000/v1
+```
+
+Single-source shorthands: `<taskset-id>` names the run's only source, `--env.<field> <value>` sets a field of that source's env block (`--env.agent.harness.id bash`, `--env.taskset.tasks '["fix-git"]'`), `-n`/`-r` set `num_examples`/`group_size`, `-m` the model, and `-c N` pins the concurrency band (`concurrency.min_inflight = max_inflight = N`). The shorthands cannot be combined with a TOML that defines `[[source]]` blocks. `uv run eval -h` lists them.
+
+Concurrency is pinned at 128 in-flight episodes by default; `-c N` repins it. External APIs expose no vLLM `/metrics` to adapt to, so the pin is what they run with. Against a vLLM server, set `min_inflight < max_inflight` in `[concurrency]` to let the band adapt to KV usage like the orchestrator's.
```
mikasenghaas: just to  "to dynamically adjust the number of concurrent episodes for maximum throughput"

docs/eval.md
```diff
@@ -0,0 +1,125 @@
...
+
+[[source]]
+env.taskset.id = "aime25"
+env.agent.harness.id = "null"
+env.agent.runtime.type = "subprocess"
+```
+
+Per-source `num_examples`, `group_size` and `sampling` override the top-level defaults. Every source's env server is spawned by the eval process unless the source sets `serve.address`, in which case the server is externally managed. A spawned server binds an OS-assigned loopback port and publishes it to `configs/attempt_N/resolved/envs/eval/<name>.address`, which the eval process reads, so concurrent runs on one host never collide on a port.
+
+The basic examples ship an `eval.toml` next to their `sft.toml` and `rl.toml` (e.g. [`examples/basic/reverse-text/eval.toml`](../examples/basic/reverse-text/eval.toml)) for the baseline and final evals of the walkthrough; override the model with `-m` to evaluate a trained checkpoint. Smoke configs against Prime Inference live in [`configs/debug/eval/`](../configs/debug/eval), one per shape: single turn, multi turn (a sandboxed terminal task), resume, and multi env (the same terminal tasks under the bash and rlm harnesses), plus AIME 2026 at avg@16 and Terminal-Bench 2 at avg@4.
```
mikasenghaas: remove this paragraph

docs/eval.md
```diff
@@ -0,0 +1,125 @@
...
+Like the training launchers, the console shows the start line, log paths and dashboard URL, stays quiet while the eval runs (only errors surface), and ends with the success line and the dashboard URL again. Results and progress are in the dashboard; `eval.log` keeps everything.
+
+The task cursor is checkpointed after every completed group (`[ckpt]`: `interval` counts completed groups, `keep_last` prunes older cursors; disable with `--no-ckpt`). Relaunch with the same `--run.name` and `--resume` (or `--resume.step N`) to skip the completed prefix; partially completed groups are retried:
+
+```bash
+uv run eval @ eval.toml --run.name my-eval
+uv run eval @ eval.toml --run.name my-eval --resume
+```
+
+### Live View
```
mikasenghaas: make a section for monitors with subsection for file (local dashboard) and prime (platform)

docs/eval.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Run Directory and Resume
+
+The run writes to `output_dir / run.name` (auto-generated as `<envs>--<model>--<short-id>`) with the same layout as training runs: `configs/attempt_<n>/` (the launch command, TOML, and resolved `eval.json`), `logs/attempt_<n>/eval.log` plus one `envs/eval/<name>.log` per source, and `monitors/file/` with `metrics.jsonl` and the trace stream. `--clean` wipes a used run directory, `--dry-run` writes the config and exits, `--no-dashboard` skips the dashboard daemon.
+
+Like the training launchers, the console shows the start line, log paths and dashboard URL, stays quiet while the eval runs (only errors surface), and ends with the success line and the dashboard URL again. Results and progress are in the dashboard; `eval.log` keeps everything.
```
mikasenghaas: unnecessary detail, remove

docs/eval.md
```diff
@@ -0,0 +1,125 @@
...
+env.taskset.id = "aime25"
+env.agent.harness.id = "null"
+env.agent.runtime.type = "subprocess"
+```
+
+Per-source `num_examples`, `group_size` and `sampling` override the top-level defaults. Every source's env server is spawned by the eval process unless the source sets `serve.address`, in which case the server is externally managed. A spawned server binds an OS-assigned loopback port and publishes it to `configs/attempt_N/resolved/envs/eval/<name>.address`, which the eval process reads, so concurrent runs on one host never collide on a port.
+
+The basic examples ship an `eval.toml` next to their `sft.toml` and `rl.toml` (e.g. [`examples/basic/reverse-text/eval.toml`](../examples/basic/reverse-text/eval.toml)) for the baseline and final evals of the walkthrough; override the model with `-m` to evaluate a trained checkpoint. Smoke configs against Prime Inference live in [`configs/debug/eval/`](../configs/debug/eval), one per shape: single turn, multi turn (a sandboxed terminal task), resume, and multi env (the same terminal tasks under the bash and rlm harnesses), plus AIME 2026 at avg@16 and Terminal-Bench 2 at avg@4.
+
+### Run Directory and Resume
```
mikasenghaas: rename this section to just `## Checkpointing`

docs/eval.md
```diff
@@ -0,0 +1,125 @@
...
+## Online Evals
+
+Training runs evaluate the live policy on a step interval with the same eval sources:
+
+- **RL** — the `[orchestrator.eval]` block (`RLOnlineEvalConfig`): `interval`, `[[orchestrator.eval.source]]` with per-source overrides, and `skip_first_step` / `retrigger_on_resume`. The orchestrator runs the epochs against its inference pool; see [Training § Useful Knobs](training.md#useful-knobs).
+- **SFT** — the `[eval]` block (`SFTOnlineEvalConfig`) plus an `[inference]` block. The `sft` launcher starts the inference server, one env server per source, and an online-eval process that evaluates every weight broadcast; see [Training § Online Evals](training.md#online-evals).
+
+Both write their eval episodes and metrics into the training run's monitors, so results show up next to the training curves in the dashboard and W&B; the SFT online-eval process logs to `logs/attempt_<n>/eval.log`.
+
+## Metrics
```
mikasenghaas, line 104: remove

skills/dashboard/SKILL.md
```diff
@@ -36,7 +36,7 @@ latest and peak prompt lengths first, followed by cumulative token processing
...
...counted dot plot, else a beeswarm (env metrics included). Phases use a ramp of the accent. A dot is an episode or task: hover for its value, click to open the trace; hover the background for min/p10/median/p90/max (a faint boxplot sits behind the dots). The filter button's `include errors` (default off) adds errored episodes, drawn red. The traces tab shows in-flight rollouts in the episode table (pulsing dot in `#`, phase badge, `dispatched → arrived (elapsed)`, counts that grow per turn, reward empty until the episode lands; errored episodes carry a red left border in place of a status column; status toggles in flight / completed; click one to follow it in the viewer, transcript only, pinned to the newest turn until you scroll up, the request in flight's uncommitted messages dimmed as `awaiting model`), polled once a second with an etag (`?etag=` answers `{unchanged}`), folded from `monitors/file/traces/live/*.jsonl` at `/api/runs/<run>/live` and `/api/runs/<run>/live/<trace_id>`. The logs tab's components menu lists every env server under `all envs`, so one env's log can be toggled into its own pane. Replay defaults to 8× so long coding-agent model waits remain watchable; select
```
mikasenghaas: there is a bbunch of unnecessary info here. cut this down signifincatly. review the whole file. only keep information that is useful for you to know when i ask questions abt the dashboard (which cannot be answered by looking at the code)

skills/eval/SKILL.md
mikasenghaas, line 1: make sure this file is ~same structure as training in terms of style nad brevity

skills/training/monitor-run/SKILL.md
```diff
@@ -100,10 +100,14 @@ Usually tailing `trainer.log`, `orchestrator.log`, and `inference.log` is enough
 Scan for problems:
 
 ```bash
-grep -E "WARNING|ERROR" {run_dir}/logs/latest/{trainer,orchestrator,evals,inference}.log
+grep -E "WARNING|ERROR" {run_dir}/logs/latest/{trainer,orchestrator,eval,inference}.log
 grep -E "WARNING|ERROR" {run_dir}/logs/latest/envs/{train,eval}/*.log
 ```
 
+### Live rollouts
```
mikasenghaas, line 107: i like "live rollouts". use this name across the dash as well

tests/unit/test_configs.py
```diff
@@ -25,7 +25,8 @@
     OrchestratorConfig,
     InferenceConfig,
     EnvServerConfig,
-    EvalsConfig,
+    EvalConfig,
+    SFTOnlineEvalConfig,
```
mikasenghaas: do we need the sft online configs? i dont think we expect to have a online eval standalone configs to test right?

tests/unit/monitors/test_live_folds.py
mikasenghaas, line 1: remove

src/prime_rl/entrypoints/eval.py
```diff
@@ -0,0 +1,165 @@
...
+        components.extend((f" {name}", config_dir / "envs" / "eval" / f"{name}.json") for name in env_names)
+    logger.info(f"Configs:\n{format_config_message(config_dir, 'eval', components)}")
+    if config.dry_run:
+        logger.success("Dry run complete. To start the eval, remove --dry-run from your command.")
+        return
+
+    logger.info(format_log_message(log_dir, eval=True, env_names={"eval": env_names}))
+    dashboard_url = ensure_dashboard(config.output_dir, logger) if config.dashboard else None
+    log_dashboard_url(logger, dashboard_url)
+    from prime_rl.eval.eval import run_eval
```
hallerite, line 163: The advertised core-only install cannot run eval yet. Importing `run_eval` reaches `utils.utils` through the monitors, which imports `torch` unconditionally. Torch isn't in the core dependency closure, so this fails with `ModuleNotFoundError: No module named 'torch'` without the extra dependencies. I reproduced the import failure and checked the lockfile; CI installs all extras, which masks it. Could we separate the eval imports from training-only dependencies or provide an explicit CPU dependency set, and add a smoke check for the documented minimal installation?

src/prime_rl/eval/eval.py
```diff
@@ -0,0 +1,84 @@
...
+            eval_env_names=[source.resolved_name for source in config.source],
+            overview_flavor="eval",
+        )
+        await self.runner.setup()
+        eval_source = self.runner.eval_source
+        if config.resume is not None:
+            if config.resume.dir is not None:
+                self.ckpt_manager.load(config.resume.dir_step, eval_source, path=config.resume.dir / "eval")
+            else:
+                self.ckpt_manager.load(config.resume.step or self.ckpt_manager.latest_step(), eval_source)
```
hallerite: Resume restores scheduling progress, but not the full evaluation results. This loads only the cursor into a fresh evaluation sink, so final logged/platform metrics cover the remaining suffix, and `run_epoch()` overwrites the expected-count plan with that smaller count. Out-of-order completions also get duplicated in the append-only trace stream: I reproduced tasks `0, 2, 3` completing before interruption, leaving cursor `1`; resume then schedules `1, 2, 3` again without removing the earlier results. Could we reconcile persisted episodes with both scheduling and aggregation, so an interrupted/resumed evaluation produces the same result as an uninterrupted one?

src/prime_rl/eval/runner.py
```diff
@@ -0,0 +1,388 @@
...
+        # The launcher-set $PRL_RUN_ID is the run identity; standalone runs mint a local one.
+        self.run_id = os.environ.get("PRL_RUN_ID") or uuid.uuid4().hex
+        self.run_name = os.environ.get("PRL_RUN_NAME")
+        wandb_enabled = monitors.get(monitors.WandbMonitor) is not None
+
+        get_logger().info(f"Initializing inference pool (base_url={config.client.base_url}, model={config.model})")
+        self.clients = InferenceClient(config.client, model_name=config.model)
+        self.admin_plane = AdminPlane(config.client)
+
+        self.spawn_env_servers()
```
hallerite: This also runs for SFT online evals, whose launcher has already started one env server per managed source. `build_online_eval_config()` leaves those sources' `serve.address` as `None`, so this starts a second set of servers. Both sets publish to the same address files and write the same logs, making server selection race-dependent and allocating unnecessary resources. I checked the spawn logic with a single managed source and it schedules an additional server launch. Could we give one process ownership of launching the servers and have the other only wait for their published addresses?

hallerite (commented):
The consolidation makes sense to me, and sharing the eval machinery with training seems like the right direction. I found three correctness issues we should address before landing; I've left the details inline.

and ofc, though already acknowledged, we need hosted evals

src/prime_rl/entrypoints/eval.py
mikasenghaas: Confirmed and fixed in 6250bf0ae. Torch reached the eval through `utils.utils` (process-group teardown, default dtype) and the algorithm package (GRPO/MaxRL group scoring); those three sites now look torch up lazily, so `prime_rl.eval.eval` and the entrypoint import without it. CI gained a `core-eval-install` job: `uv sync --locked --package prime-rl --package gsm8k`, the eval import with `torch` asserted absent from `sys.modules`, and `eval gsm8k -n 1 --dry-run`. Verified locally in a fresh core-only venv, including a real two-example gsm8k eval against Prime Inference.

src/prime_rl/eval/runner.py
mikasenghaas: Fixed in 45f512cfb: the runner never spawns servers anymore. Env servers belong to the launcher, like the orchestrator's belong to `rl`: the `eval` entrypoint writes the env-server configs and spawns one server per source without `serve.address` (cleaned up in a `finally`), `sft` keeps doing the same for its online evals, and the runner only waits for the published addresses. One helper (`write_env_server_config`) writes the config for all three launchers. Verified with an SFT run with online evals: exactly one `PRL::EnvServer` process, one config/address/log per source, evals at steps 0/2/4.

src/prime_rl/eval/eval.py
mikasenghaas: Agreed, and thanks for the repro. The cursor-only checkpoint restores scheduling but not the results, so a resumed run re-runs out-of-order completions (duplicates in the stream), re-logs a smaller plan, and reports metrics over the suffix only. The fix under discussion is to make the trace stream the source of truth on resume: read the landed episodes, keep the good ones (matched to the run's tasks up to group_size, like verifiers' eval CLI did), rebuild the stream from them, replay them into the sink so the epoch's metrics and the platform upload cover the full set, and schedule only what is still owed. Leaving this open until that lands.

src/prime_rl/orchestrator/clients.py
```diff
@@ -18,6 +18,17 @@
...
 
+def resolve_api_key(api_key_var: str) -> str:
+    """The API key named by ``api_key_var``; ``PRIME_API_KEY`` also falls back to the prime
+    CLI config (``prime login``), like the verifiers client does. ``"EMPTY"`` when unset."""
+    api_key = os.environ.get(api_key_var)
+    if not api_key and api_key_var == "PRIME_API_KEY":
+        from prime_cli.core.config import Config as PrimeConfig
+
+        api_key = PrimeConfig().api_key
+    return api_key or "EMPTY"
```
samsja, line 30: I still dont understand why this is here tbh, do we still have this logic right now ?

src/prime_rl/eval/eval.py
mikasenghaas: Fixed in 20fe0d462: the cursor checkpoint is gone and a resume reads the trace stream instead. Ok episodes rejoin the epoch through the monitors (rebuilt stream, full-epoch metrics, platform upload), only the rollouts still owed run, errored and cut-off episodes are owed again. Episodes match their task by `task.key`, so `num_examples`/`group_size` may change between launches; model, sampling and env config are asserted equal against the previous attempt's resolved config. Verified by interrupting single-env and multi-env runs with SIGTERM and resuming: the resumed stream holds exactly examples x group_size unique episodes, the plan and epoch metrics cover the full set.

src/prime_rl/orchestrator/clients.py
mikasenghaas: we need this so that pi inference auth works automatically

src/prime_rl/eval/resume.py
```diff
@@ -0,0 +1,139 @@
...
+                    return
+
+
+def take_landed(run_dir: Path) -> list[dict]:
+    """The ok eval episodes the run has landed. The file monitor's directory is removed
+    with them: the resumed attempt writes a fresh stream, plan and metrics."""
+    stream = get_trace_stream(run_dir)
+    landed = [record for record in read_records(stream) if record.get("ok")] if stream.is_dir() else []
+    shutil.rmtree(get_file_monitor_dir(run_dir), ignore_errors=True)
+    return landed
```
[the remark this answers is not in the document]
mikasenghaas: Fixed in db514fd26: `take_landed` renames `monitors/file` to `monitors/file.previous` instead of deleting it; a later resume reads both streams and keeps each episode once, and the set-aside directory is removed only after the epoch holds the restored episodes (`release_previous`). Covered by a unit test that simulates a resume dying after re-logging one episode, and by an interrupted gsm8k run resumed end to end.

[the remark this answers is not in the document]
mikasenghaas: Simplified in 4e009278d: the previous attempt's `monitors/file` is kept as `monitors/file.attempt_N` (nothing is deleted, no release step); a resume reads every attempt's stream and keeps each episode once.

src/prime_rl/eval/resume.py
```diff
@@ -0,0 +1,160 @@
...
+    landed: dict[str, dict] = {}
+    for directory in (previous, current):
+        if (directory / stream).is_dir():
+            for record in read_records(directory / stream):
+                if record.get("ok"):
+                    landed.setdefault(record["id"], record)
+    shutil.rmtree(previous, ignore_errors=True)
+    if current.is_dir():
+        current.rename(previous)
+    return list(landed.values())
```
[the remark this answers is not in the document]
mikasenghaas: No longer applies as of 4e009278d: nothing is deleted anymore. The directory a resume finds is renamed to `monitors/file.attempt_N` and kept, and `take_landed` reads every archive plus the current stream, so a resume that dies during setup leaves the original episodes in `file.attempt_1` for the next one.

samsja (dismissed):

src/prime_rl/eval/resume.py
```diff
@@ -0,0 +1,153 @@
...
+def previous_config(run_dir: Path) -> dict:
+    """The resolved config of the run's latest earlier attempt."""
+    current = get_config_dir(run_dir)
+    attempts = sorted(
+        (path for path in (run_dir / "configs").glob("attempt_*/resolved/eval.json") if path.parent != current),
+        key=lambda path: int(path.parts[-3].removeprefix("attempt_")),
+    )
+    if not attempts:
+        raise FileNotFoundError(f"Nothing to resume: {run_dir} has no earlier eval attempt")
+    return orjson.loads(attempts[-1].read_bytes())
```
hallerite: A rejected resume becomes the comparison baseline for the next attempt. The launcher writes `configs/attempt_N/resolved/eval.json` before `check_config()` runs, and this selects the latest attempt regardless of whether it passed validation. I reproduced: run with model A; resume with model B correctly fails; repeat the same B command and it succeeds, comparing against the rejected attempt's B config while restoring A's episodes. A dry run with a changed config can likewise become the baseline. Could we validate against the config associated with the saved results, and only advance that baseline after successful validation? Otherwise the guard can silently allow results from different models or sampling settings to be combined.

src/prime_rl/eval/resume.py
```diff
@@ -0,0 +1,153 @@
...
+        for key in targets[env.name]:
+            targets[env.name][key] *= env.config.group_size
+    kept: list[vf.WireEpisode] = []
+    counts: dict[str, Counter[str]] = defaultdict(Counter)
+    for record in landed:
+        env_name = record["env"].get("name") or record["env"]["id"]
+        key = record["task"]["key"]
+        if counts[env_name][key] >= targets.get(env_name, Counter())[key]:
+            continue
+        kept.append(vf.WireEpisode.model_validate(record))
```
hallerite, line 156: Restored episodes keep their original `group.id`, but the dispatcher assigns a fresh UUID to the remaining rollouts of the same task. Both `pass_at_k()` and the dashboard group by that ID, so a partial-group resume splits one task into multiple metric groups. For one task with rewards `[1, 0, 0, 0]`, interrupted after the first rollout, I reproduced pass@1 changing from 0.25 to 0.50, pass@2 from 0.50 to 0.00, and pass@4 disappearing. Increasing `group_size` on resume has the same problem. Could we preserve or reconstruct a consistent evaluation group across restored and newly generated episodes, including the group identity written to the trace stream?

src/prime_rl/monitors/file/traces/live.py
```diff
@@ -0,0 +1,282 @@
...
+    def __init__(self) -> None:
+        self._folds: dict[Path, tuple[int, dict[str, Any], EpisodeAssembly]] = {}
+
+    def read(self, path: Path) -> tuple[dict[str, Any], dict[str, Any]] | None:
+        try:
+            size = path.stat().st_size
+        except FileNotFoundError:
+            self._folds.pop(path, None)
+            return None
+        consumed, dispatch, assembly = self._folds.get(path) or (0, {}, EpisodeAssembly())
```
hallerite: The dashboard's synchronous list and individual-trace endpoints share one `LiveFolds` instance, so overlapping requests can enter here concurrently. Both can read the same `consumed` offset and the same mutable `assembly`, then apply the newly appended delta twice. I reproduced this with two concurrent reads synchronized before folding: a file containing two nodes leaves three nodes in the cache, and subsequent reads retain the duplicate. Could we synchronize the cache's read/fold/update operation and return a stable snapshot for response serialization? This can occur while a live trace is open alongside the rollout list or in multiple dashboard tabs.

hallerite (commented):
Thanks for addressing the first round. I checked the updated stack: the eval runner now imports without torch, server ownership is fixed, and resume restores landed episodes and the full epoch count. I found two remaining resume correctness issues and a concurrency issue in the live dashboard cache, detailed inline.

Validation: 27 targeted prime-rl tests and 99 verifiers non-live tests passed, plus local reproductions of the issues below.

src/prime_rl/eval/resume.py
mikasenghaas: Fixed in ca8063804. Each attempt stamps its resolved config into the file monitor directory once it runs (`monitors/file/eval.json`, beside the episodes it produces, archived with them), and `previous_config` reads that stamp instead of the latest `configs/attempt_N`. A rejected resume and a dry run never stamp, so they cannot become the baseline. Reproduced your sequence: run with model A, resume with B fails, repeat B fails again with the same error.

src/prime_rl/eval/resume.py
mikasenghaas: Fixed in ca8063804. `plan` returns the group id a task's kept episodes carry, `EvalSource.restore` puts it on the owed `TaskRequest`, and the dispatcher opens the group under that id instead of minting one, so restored and new rollouts of a task share one `group.id` in the stream, in `pass_at_k` and in the dashboard. Verified on an interrupted 8×2 run: after the resume every task has exactly one group, including the tasks split across the two attempts.

src/prime_rl/monitors/file/traces/live.py
mikasenghaas: Fixed in ca8063804. `LiveFolds.read` and `forget_missing` run under a lock, and `read` hands back a snapshot (an orjson round-trip) rather than the growing assembly, so a delta folds once and no caller serializes a fold that another request is extending. Verified by hammering the list and single-trace endpoints concurrently during a live run: 258 responses, none with more nodes than the file held.

hallerite (dismissed):
LGTM, great work

hallerite (approved):

PR #3472, opened 2026-09-02
chore(inference): introduce admin plane abstraction

samsja:
## Summary

- consolidate the inference control path into a single `AdminPlane` abstraction over its worker HTTP clients
- make the static plane own readiness, NCCL setup, weight updates, and lifecycle management
- expose one typed `update_weights` operation; each receiver identifies its transport explicitly
- route filesystem, NCCL, and NIXL receivers through the admin plane instead of passing raw HTTP client lists
- remove the separate admin-plane protocol, unused factory argument, and pass-through weight-update wrapper

`InferenceClient` remains the data-plane client for generation and scoring, while `AdminPlane` coordinates control operations against one or more inference workers. Alternate backends can specialize a transport operation without adding backend flags to weight receivers.

## Validation

- `uv run ruff check src/prime_rl/orchestrator/clients.py src/prime_rl/orchestrator/orchestrator.py src/prime_rl/evals/evals.py src/prime_rl/transports/weights tests/unit/orchestrator/test_clients.py`
- `uv run python -m compileall -q src/prime_rl/orchestrator/clients.py src/prime_rl/orchestrator/orchestrator.py src/prime_rl/evals/evals.py src/prime_rl/transports/weights`
- `uv run pytest -q tests/unit/orchestrator/test_clients.py` (10 passed)

The transport test module could not be collected on this host because its optional CUDA import chain requires `flash_attn`.

---

> [!NOTE]
> **Medium Risk**
> Refactors the weight-broadcast control path (pause/resume, NCCL/NIXL/filesystem updates) used during training; behavior should be equivalent but mistakes here can strand engines paused or desync policy versions.
> 
> **Overview**
> Renames **`AdminClients`** to **`AdminPlane`** and makes it the single control-plane entry point for inference workers (readiness, NCCL init, weight updates, LoRA load, teardown).
> 
> **NCCL setup** and **weight updates** move from module-level helpers into **`AdminPlane.initialize_nccl`** and **`AdminPlane.update_weights`**, with callers passing an explicit **`transport`** (`filesystem` | `nccl` | `nixl`). Orchestrator, evals, and all **`WeightReceiver`** implementations now take an **`AdminPlane`** instead of raw **`httpx`** client lists; **`load_lora_adapter`** and **`init_nixl_broadcast`** follow the same pattern.
> 
> Unit tests are updated and **`test_admin_plane_initializes_nccl`** is added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65043370b7a9c857e6e3302f665b0b8c389043f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/prime_rl/orchestrator/clients.py
```diff
@@ -107,11 +108,64 @@ async def wait_for_ready(self, model_name: str) -> None:
...
+        self,
+        weight_dir: Path | None,
+        *,
+        step: int = 0,
+        on_paused: Callable[[], None] | None = None,
+    ) -> None:
+        await update_weights(self.clients, weight_dir, step=step, on_paused=on_paused)
+
+    async def load_lora_adapter(self, lora_name: str, lora_path: Path) -> None:
+        await load_lora_adapter(self.clients, lora_name, lora_path)
```
samsja: this seems off like we just repeat the function call, i think the admin plan should just expose a client or smth or the code that are using client should consume the admin pane

samsja: Agreed. I removed the pass-through methods. The weight-operation functions now consume `AdminPlane` directly and obtain the selected client set from it; receivers pass the plane through without unpacking it. Updated in 45e02a6be.

src/prime_rl/orchestrator/clients.py
```diff
@@ -107,11 +107,67 @@ async def wait_for_ready(self, model_name: str) -> None:
...
+        await _update_weights(self.clients, weight_dir, step=step, on_paused=on_paused)
+
     async def aclose(self) -> None:
         for client in self.clients + self._router_clients:
             await client.aclose()
 
 
+def setup_admin_plane(client_config: ClientConfig, model_name: str) -> AdminPlane:
+    """Build the control plane for a policy inference deployment."""
+    return AdminClients(client_config)
```
samsja: sees a bit useless also drop model name ?

src/prime_rl/orchestrator/clients.py
```diff
@@ -107,11 +107,67 @@ async def wait_for_ready(self, model_name: str) -> None:
...
+        )
+
+    async def update_nccl_weights(
+        self,
+        weight_dir: Path,
+        *,
+        step: int = 0,
+        on_paused: Callable[[], None] | None = None,
+    ) -> None:
+        await _update_weights(self.clients, weight_dir, step=step, on_paused=on_paused)
```
samsja: where is this coming from i dont get it

src/prime_rl/orchestrator/clients.py
```diff
@@ -300,7 +356,7 @@ async def _resume_engines(admin_clients: list[AsyncClient]) -> None:
     logger.debug("All inference engines resumed")
 
 
-async def update_weights(
+async def _update_weights(
     admin_clients: list[AsyncClient],
```
samsja: why rename here ?

src/prime_rl/orchestrator/clients.py
```diff
@@ -338,6 +394,15 @@ async def update_weights(
         await _resume_engines(admin_clients)
 
 
+async def update_weights(
+    admin_plane: AdminPlane,
+    weight_dir: Path | None,
+    step: int = 0,
+    on_paused: Callable[[], None] | None = None,
+) -> None:
+    await _update_weights(admin_plane.clients, weight_dir, step=step, on_paused=on_paused)
```
samsja: this seems just passing info around ? why ?

src/prime_rl/orchestrator/clients.py
```diff
@@ -107,6 +106,84 @@ async def wait_for_ready(self, model_name: str) -> None:
...
+    async def update_nccl_weights(
+        self,
+        weight_dir: Path,
+        *,
+        step: int = 0,
+        on_paused: Callable[[], None] | None = None,
+    ) -> None:
+        await self.update_weights(weight_dir, step=step, on_paused=on_paused)
+
+    async def update_weights(
```
samsja: why do we have both what is the difference ?

src/prime_rl/orchestrator/clients.py
```diff
@@ -107,6 +106,93 @@ async def wait_for_ready(self, model_name: str) -> None:
...
+        """Receive weights through the static NCCL update endpoint."""
+        await self._update_weights(weight_dir, step=step, on_paused=on_paused)
+
+    async def update_filesystem_weights(self, weight_dir: Path, *, step: int = 0) -> None:
+        """Load weights from a checkpoint on the shared filesystem."""
+        await self._update_weights(weight_dir, step=step)
+
+    async def update_nixl_weights(self, *, step: int = 0) -> None:
+        """Trigger a weight pull from the initialized NIXL transport."""
+        await self._update_weights(None, step=step)
```
samsja: fyi here we do this because dynamo has a different way to do nccl broadcast

src/prime_rl/orchestrator/clients.py
```diff
@@ -107,6 +106,93 @@ async def wait_for_ready(self, model_name: str) -> None:
...
+                    )
+
+        await asyncio.gather(
+            *(
+                initialize_client(admin_client, client_num * gpus_per_server)
+                for client_num, admin_client in enumerate(self.clients)
+            )
+        )
+
+    async def update_nccl_weights(
```
mikasenghaas: why do we have a method for each broadcast again? this should be abstracted in the broadcast class no?

samsja: dynamo has a different way to do nccl broadcast hence the method

hallerite (approved):

PR #3475, opened 2026-09-02
feat(inference): add Dynamo admin plane

samsja:
## Summary

- add `DynamoAdminPlane` as a specialization of the static `AdminPlane` introduced by #3472
- discover Dynamo workers and pin a stable topology before constructing worker admin connections
- encapsulate Dynamo rank-offset calculation, topology validation, collective-RPC payloads, retries, and failure behavior in the client
- override the unified weight-update operation only for NCCL; filesystem and NIXL delegate to the default admin-plane route
- select the Dynamo plane through a factory that is introduced only when the second implementation exists

This is a stacked follow-up to #3472 and incorporates the Dynamo behavior from #3469. `InferenceClient` remains responsible for data-plane requests; generic weight receivers call one typed `AdminPlane.update_weights` operation and contain no Dynamo flags or worker metadata.

## Validation

- `uv run ruff check packages/prime-rl-configs/src/prime_rl/configs/rl.py packages/prime-rl-configs/src/prime_rl/configs/shared.py src/prime_rl/evals/evals.py src/prime_rl/inference/dynamo.py src/prime_rl/orchestrator/clients.py src/prime_rl/orchestrator/orchestrator.py tests/unit/inference/test_dynamo.py tests/unit/test_configs.py`
- `uv run python -m compileall -q src/prime_rl/inference/dynamo.py src/prime_rl/orchestrator/clients.py src/prime_rl/orchestrator/orchestrator.py src/prime_rl/evals/evals.py`
- focused Dynamo, admin-plane, and configuration tests (35 passed)

The complete `tests/unit/test_configs.py` suite additionally requires taskset plugins that are not installed in this worktree (`scaleswe`, `reverse-text`, and others).


src/prime_rl/orchestrator/clients.py
```diff
@@ -189,6 +189,14 @@ async def aclose(self) -> None:
             await client.aclose()
 
 
+def setup_admin_client(client_config: ClientConfig, model_name: str) -> AdminClient:
+    if client_config.dynamo is not None:
+        from prime_rl.inference.dynamo import DynamoAdminClient
+
+        return DynamoAdminClient(client_config, model_name)
+    return AdminClient(client_config)
```
samsja: can we do if else here I dont like like this

PR #3479, opened 2026-09-03
feat(deepseek_v4): sparse attention kernels

garrett361:
Adds a TileLang implementation of the flash-attention style kernel for DS V4. Based on the similar kernel used for GLM.  The kernel is wired through the DS V4 model and the attention arguments are refactored to naturally adapt to the shape expected by the efficient kernel (e.g. no more `attention_mask` realization). 

Next kernels:
1. Indexer
2. mHC
3. RoPE


notes/ds-v4-kernels/bench/raw/.gitignore
garrett361, line 1: Everything under `notes/ds-v4-kernels` is scratch dev files and will be removed prior to merge.

src/prime_rl/trainer/models/deepseek_v4/eager_reference.py
```diff
@@ -0,0 +1,94 @@
+"""Naive DeepSeek V4 attention reference.
+
```
samsja, line 2: yeah nice to keep the eager as ref

notes/ds-v4-kernels/bench/raw/.gitignore
garrett361: removed

samsja (approved):

PR #3480, opened 2026-09-04
fix(transports): make NCCL respect keep_in_fp32_for_weight_transfer

garrett361:
The NCCL weight-broadcast transport hardcoded every tensor to `bfloat16`, ignoring `keep_in_fp32_for_weight_transfer`. This PR makes NCCL transport respect the fp32 overrides. This makes NCCL honor the same declaration the other two transports already respect. Note: NIXL does its own thing and does not use this shared helper, yet. 

- Extracted `resolve_wire_dtype()` so the filesystem and NCCL paths share one implementation.
- Lifted `NCCLBroadcaster._resolve_dtensors` to a module-level `resolve_dtensors()`, since a method reachable only through a live `PyNcclCommunicator` can't be unit tested.
- Added `tests/unit/utils/test_weights.py`, the flag's first test coverage, parametrized over all six models that declare it.


src/prime_rl/transports/weights/nccl.py
```diff
@@ -80,6 +81,26 @@ def filter_state_dict_by_layers(
...
+    Only DTensors are touched, since only they need gathering. A buffer is never sharded, so it
+    goes on the wire in whatever dtype it already holds and its fp32 declaration, if it has one,
+    is never consulted. NIXL instead applies the declaration to every floating-point tensor, so
+    the transports still disagree about buffer dtypes.
+    """
+    for key, value in list(state_dict.items()):
+        if isinstance(value, DTensor):
+            # only gather after the downcast as it will be faster
+            target_dtype = resolve_wire_dtype(keep_in_fp32, key, default_dtype)
+            state_dict[key] = cast(DTensor, value.to(target_dtype)).full_tensor()
```
samsja: i would remove the cast here, always skeptical of cast personnaly in python unless we have like mypy enforce in the ci

garrett361: also thought it was a little weird, but carried it over. Has no effect, but will remove it.

garrett361: removed

samsja (dismissed):

samsja (approved):

Issue #749, opened 2025-08-16
Write token loss scale on orchestrator


mikasenghaas:
Similarly to how we write per-token advantages directly from the orchestrator, we should do the same for loss scales. This will make implementing different loss scales much easier and allow us to cleanly implement both GSPO variants.

PR #3487, opened 2026-09-04
feat(transports): add mx_refit ModelExpress weight-broadcast transport

KavinKrishnan:
## Summary

Adds `mx_refit`, a ModelExpress weight-broadcast transport for Prime-RL.

Trainer ranks publish local FSDP shards under one version. Each vLLM worker pulls the tensor slices needed by its TP/EP rank over NIXL/UCX.

## Changes

* Adds the sender, receiver, configuration, and vLLM worker integration.
* Keeps `modelexpress_rl` optional until `mx_refit` is selected.
* Uses a unique ID per publish attempt and bounded discovery/release waits.
* Removes the shared-filesystem receiver acknowledgement from the MX path.
* Pins readers across RDMA rails while preserving explicit `UCX_NET_DEVICES`.
* Uses trainer-only sources and retires failed installs.
* Reports publish, discovery, wire, install, memory reclaim, and ModelExpress metadata timings.
* Supports conditional CUDA cache reclamation. The default remains `always`.
* Preserves model-required FP32 transfer dtypes (see below).
* Verifies that the initial refit actually replaces changed weights.
* Fans out Ray tensor-parallel inference configuration.
* Keeps the FP8 indexer's quantization scalars in FP32.

### FP8 indexer scalars

`eps`, `fp8_min` and `fp8_max` reach the indexer kernel as Python floats, which
Inductor lowers as f64, so quantization arithmetic runs at double precision
against f32 tensor data. On GLM-5 that surfaces as a compile failure in the
indexer path rather than as a numerical difference, which makes it look
unrelated to precision. Casting the three scalars to `tl.float32` runs the math
at the precision the tensors already carry. Thirteen lines, independent of
ModelExpress, included here because GLM-5 is the motivating configuration for
the transport and does not train without it.

## Transfer precision

Some models need specific tensors to cross the wire in FP32 — the GLM-5 router selection bias, for example — while everything else transfers at BF16. `mx_precision` turns a model's `keep_in_fp32_for_weight_transfer()` policy into ModelExpress wire-dtype overrides. Optimizer and reduction dtypes are unaffected; this controls transfer precision only.

The override is named only when the installed client accepts it. `FSDPTrainerContext.wire_dtype_overrides` postdates the released ModelExpress, and naming it unconditionally would make this transport require an unreleased client from every model, including models that declare no overrides at all. A model that *does* declare them fails loudly against a client that cannot carry them, rather than silently transferring at BF16 — a dropped FP32 requirement is invisible until it shows up as a diverged policy many steps later.

So: `mx_refit` works against released ModelExpress. Models with an FP32 policy additionally need a client carrying `wire_dtype_overrides`.

## Warm updates

Registered trainer buffers and receiver transfer plans stay allocated and are reused. Prime-RL waits for inference to consume a version before the trainer overwrites those buffers.

With content digests disabled, ModelExpress also reuses trainer manifests and receiver manifest bytes while registration and tensor structure remain unchanged. With content digests enabled, version manifests are regenerated and fetched for verification, but the transfer plan is still reused.

`release_version()` removes version publication and manifest bookkeeping. It does not free the persistent staging buffers or NIXL registrations.

## Results

Qwen3-30B-A3B on GB200, 4 FSDP trainer GPUs, TP4/EP4 inference, 30 warm refits. Warm medians from the 2026-09-09 run:

* Prime-RL refit: 1.5352 s
* Publish: 0.2986 s
* Discovery: 0.0085 s
* Wire: 0.4109 s
* Install: 0.3488 s
* Receiver E2E: 0.8074 s
* Trainer E2E: 1.6266 s
* Wire payload: 15.8892 GB per inference rank

Every warm manifest and transfer-plan lookup hit the cache, and warm receivers fetched no manifest bytes. Trainer metadata work was 44.7 ms median. A separate digest-enabled run completed four warm refits with content verification and no mismatch, reusing every warm transfer plan.

**A caution on that top-line number.** It does not reproduce across days. Re-running the identical image on 2026-09-14 measured **1.8150 s**, a shift of +0.280 s from node placement and cluster load alone. Refit time here is dominated by cross-rank arrival skew, which moves with the neighbours you land next to, so single-run comparisons against a remembered figure are not meaningful at this resolution.

Measured properly, as paired arms on one clique 90 minutes apart, 27 warm refits each:

| Arm | Median | Mean | SD |
|---|---:|---:|---:|
| Before this PR's later commits | 1.8150 | 1.8436 | 0.2493 |
| This PR | 1.8423 | 1.8299 | 0.1587 |

Mean delta −0.014 s, 95% CI [−0.125, +0.098]. The added verification, precision and handshake work costs nothing measurable, and refit-time variance is lower (SD 0.159 against 0.249).

## Configuration

```
weight_broadcast:
  type: mx_refit
  host: <model-express-host>
  port: <model-express-port>
  timeout: 2700
  reclaim_memory: always # or if_needed
  reclaim_headroom_gb: <headroom>
```

`mx_refit` requires a Gloo CPU backend; the trainer enables one automatically when this transport is selected.

## Restart behavior

A joint trainer, inference, and orchestrator restart recovers. A trainer-only restart can leave roles on different steps and exits after the configured timeout with guidance to restart both sides.

## Dependencies and validation

**This PR is not installable against the pinned ModelExpress and should not be merged before an MX release.** `pyproject.toml` pins `modelexpress==0.3.0`, which provides no `modelexpress_rl` API, so CI can only exercise the `ImportError` guard in `transports/weights/__init__.py`. Two symbols the code uses are unreleased:

| Symbol | Source |
|---|---|
| `modelexpress_rl.*` (`FSDPTrainerContext`, `ModelExpressTrainerClient`, `TrainerStagingMode`, `WeightPayloadFormat`) | MX `main`, unreleased |
| `FSDPTrainerContext(wire_dtype_overrides=...)` | MX [#747](https://github.com/ai-dynamo/modelexpress/pull/747), **unmerged** |
| `generator.apply_weight_streaming(...)`, selected by `MX_REFIT_STAGING_BYTES` | MX [#749](https://github.com/ai-dynamo/modelexpress/pull/749), **unmerged** |

Merge order: land #747 and #749, cut an MX release, then bump the pin and `uv.lock` here and drop the reflection in `mx_precision.build_trainer_context`.

Already merged and required: [#704](https://github.com/ai-dynamo/modelexpress/pull/704) (trainer-only MoE source selection, large manifests) and [#735](https://github.com/ai-dynamo/modelexpress/pull/735) (timing, manifest lifecycle, structural reuse).
* Rebased on `main`; 62 targeted unit tests pass across precision, verification, phase timing, handshake, client setup and memory reclamation.
* 113 targeted ModelExpress lifecycle, invalidation, fallback, digest, and timing tests passed.
* The 30-refit digest-off and four-refit digest-on runs passed, plus the paired 2026-09-14 arms above.

## Scope note

The measured Qwen3-30B evidence above validates this exact tree, image and model. It is not an equivalence test and does not by itself qualify full GLM-5 refits or every cache mode; the GLM-5 campaign evidence is still accumulating on the ModelExpress side.


KavinKrishnan:
@samsja @mikasenghaas The warm-refit manifest work is validated and the PR description now has the final GCP results and exact ModelExpress dependencies. The 30-refit digest-off run and four-refit digest-on run passed. Ready for final review.

src/prime_rl/transports/weights/mx_refit.py
```diff
@@ -0,0 +1,330 @@
...
+                        None,
+                        transport="mx_refit",
+                        step=step,
+                        version_uid=uid,
+                        phase_timer=timer,
+                    )
+            finally:
+                # Retiring releases the trainer even when installation fails.
+                with timer.phase("retire"):
+                    await asyncio.shield(asyncio.to_thread(self._control.delete_weight_version, uid))
```
zhengluo-nv, line 316: [P1] Do not retire the version until every replica update has reached a terminal state. AdminPlane.update_weights uses asyncio.gather without return_exceptions; its first exception propagates while sibling HTTP updates can continue. This finally block can therefore move the version to RELEASING, let the trainer release/reuse its registered buffers, and race those in-flight RDMA reads. Await and aggregate every replica result before retirement, and keep the version alive until all reader leases are gone. Ideally this consensus/retirement operation should be encapsulated by MX rather than reconstructed here.

src/prime_rl/transports/weights/mx_refit.py
```diff
@@ -0,0 +1,330 @@
...
+import os
+import time
+import uuid
+from pathlib import Path
+
+import grpc
+import torch
+import torch.distributed as dist
+import torch.nn as nn
+from modelexpress_rl import (
```
zhengluo-nv, line 15: [P1] This transport is not installable through the repository's declared dependency set. Both this head and current main pin modelexpress==0.3.0, which does not provide modelexpress_rl, so selecting mx_refit fails before this code can run. Please land the minimal adapter against a released MX version and update pyproject.toml/uv.lock in the same PR. Keep wire-dtype overrides and streaming installation in follow-ups after their MX APIs are released instead of requiring source-checkout installation.

src/prime_rl/transports/weights/mx_rdma.py
```diff
@@ -0,0 +1,18 @@
...
+_MIN_GBPS = "MX_RESHARD_MIN_GBPS"
+_DEFAULT_MIN_GBPS = "25"
+
+
+def apply_rdma_defaults() -> None:
+    """Set ModelExpress rail and throughput defaults."""
+    # Enabling automatic pinning would override an explicit UCX device list.
+    if _NIC_PIN not in os.environ and _UCX_NET_DEVICES not in os.environ:
+        os.environ[_NIC_PIN] = "auto"
+    os.environ.setdefault(_MIN_GBPS, _DEFAULT_MIN_GBPS)
```
zhengluo-nv, line 18: [P2] Prime-RL should not silently choose ModelExpress rail-selection and minimum-bandwidth policy. Setting MX_RDMA_NIC_PIN and MX_RESHARD_MIN_GBPS here changes MX behavior for every deployment and can reject a valid slower topology. These defaults belong in ModelExpress or explicit deployment configuration; the framework adapter should preserve caller-selected settings.

src/prime_rl/inference/vllm/worker/mx_refit.py
```diff
@@ -0,0 +1,124 @@
...
+    def prepare_initial_verification(self) -> dict:
+        if os.environ.get("MX_VERIFY_INITIAL_REFIT") != "1":
+            raise RuntimeError("Initial refit verification is not enabled")
+        if getattr(self, "_initial_verification_started", False):
+            raise RuntimeError("Initial refit verification cannot be repeated in the same engine")
+        self._initial_verification_started = True
+        model = self.model_runner.get_model()
+        budget = int(os.environ.get("MX_VERIFY_CPU_BYTES", str(64 * 1024**3)))
+        self._initial_snapshot = snapshot_weights(model, max_bytes=budget)
+        perturb_weights(model, self._initial_snapshot)
```
zhengluo-nv, line 78: [P2] Please move the snapshot/NaN-perturbation/restoration flow into an E2E qualification harness. It is useful evidence, but adding destructive test behavior, MX-specific environment flags, runtime endpoints, and record schemas to Prime-RL's production worker materially enlarges the integration surface. The production adapter should stay focused on stage/apply/release; qualification can exercise that adapter externally.

zhengluo-nv (changes requested):
Requesting changes because this head has a lifecycle race and is not installable through Prime-RL's declared dependency contract. The integration also exposes too much ModelExpress lifecycle and transport policy upstream. Prime-RL should own training-step boundaries, pause/resume policy, and replica-wide success consensus; ModelExpress should own version creation/readiness, leases, retirement, physical planning, and RDMA defaults.

Please split this into: (1) a minimal integration against a released MX client, (2) MX-side lifecycle/facade changes, and (3) separate qualification tooling. The cross-node Ray TP changes, generic memory-reclamation changes, and independent FP8 kernel fix should be separate PRs. The current PR is 2,464 additions across 37 files and is conflict-marked against main.

KavinKrishnan:
Thanks — this is the right call on all four points, and I've agreed to the split. Plan below.

## P1 lifecycle race — fixed in `bcb27a6`

You're right, and it's worse than "the first exception propagates": `gather` doesn't cancel the siblings, so those replicas keep pulling the trainer's registered buffers over RDMA while `update_weights` unwinds.

One refinement, because it changes where the damage shows up. `release_version()` doesn't free the trainer's staging buffers or NIXL registrations — it removes publication and manifest bookkeeping. What actually protects those buffers from being *overwritten* is Prime-RL waiting for inference to consume the version. Premature retirement satisfies that wait, so the corruption lands when the **next** step publishes over memory the failed step's siblings are still reading. A step later than the failure that caused it, which is the part that would have made this miserable to diagnose.

Fixed by gathering with `return_exceptions=True`, requiring every replica to reach a terminal state, logging the additional failures and re-raising the first so callers matching on the exception type are unaffected. Demonstrated directly:

```
plain gather           -> [(w0,start), (w1,start)]              sibling finished: False
_gather_every_replica  -> [(w0,start), (w1,start), (w1,finish)] sibling finished: True
```

with a regression test asserting the slow replica has finished before the failure is allowed out. Applied to both `/update_weights` and `/mx_prepare_initial_refit`.

Agreed the consensus/retirement operation belongs in ModelExpress rather than reconstructed here. I'll raise it on the MX side; this fix closes the race from the framework, it doesn't make the layering right.

## P1 installability — accepted, and your sequencing is better than mine

I'd flagged this in the description but my plan was to land ai-dynamo/modelexpress#747 and #749, cut a release, then bump. Yours gets a reviewable integration in now instead of gating on MX's release cadence, which has been the long pole. Taking it.

## P2 RDMA defaults — agreed, one correction

Moving them. `MX_RDMA_NIC_PIN` and `MX_RESHARD_MIN_GBPS` are policy and a framework adapter shouldn't pick them.

For accuracy though: `MX_RESHARD_MIN_GBPS` warns rather than rejecting. MX's own reference says it "emits a `refit-slow-throughput-v1` JSON warning" and "warns rather than aborting, unlike the ceiling" — `MX_RESHARD_MAX_GBPS` is the one that rejects. So a slower topology isn't refused, it's noisy. Both are also `setdefault`-guarded, so an explicit operator value wins. Doesn't change the conclusion; a 25 Gbps floor shipping as a framework default will still cry wolf on legitimately slower fabrics.

## P2 verification harness — agreed

`perturb_weights` writing NaNs into the live model on a production path isn't defensible, and you're the second reviewer to say the qualification flow doesn't belong in `src/`. Moving it out.

## The split

- **PR 1 — minimal `mx_refit` adapter** against a released MX client, with `pyproject.toml`/`uv.lock` updated in the same PR so it's installable on merge. Stage/apply/release only, including this race fix. RDMA defaults dropped.
- **PR 2 — FP32 wire-dtype overrides**, after ai-dynamo/modelexpress#747 releases. Drops the reflection in `mx_precision.build_trainer_context`.
- **PR 3 — bounded streaming installation**, after ai-dynamo/modelexpress#749 releases.
- **Separate** — Ray TP fan-out; memory-reclamation changes; the FP8 indexer fix, which was its own PR (#3545) before I folded it in and goes back to being independent.
- **Out of `src/`** — the snapshot/perturb/restore qualification flow.

I'd like to keep this PR open until PR 1 is up, purely so the review thread and the measured evidence stay reachable, then close it in favour of the split. Say if you'd rather I close it now.

On the MX-side facade: agreed on the ownership line you drew. Version creation/readiness, leases, retirement, physical planning and RDMA defaults are MX's; step boundaries, pause/resume and replica consensus are Prime-RL's. #749 is where most of that lifecycle currently leaks, so that's the natural place to push the encapsulation.

PR #3504, opened 2026-09-08
perf(deepseek_v4): score the Lightning Indexer with the fused FP8 kernel

garrett361:
Second of three stacked PRs, builds on #3479.

Scores the DeepSeek V4 Lightning Indexer with `fp8_indexer`, the fused Triton kernel GLM DSA already runs, dropping the eager float32 `(batch, seq_len, heads, entries)` intermediate.

The indexer now rejects a batch larger than 1 instead of looping over the batch axis: the entry ranges it hands the kernel are built from sample 0's causal thresholds, so every other sample would have been scored against the wrong range.

### DIFF
--- a/src/prime_rl/trainer/model.py
+++ b/src/prime_rl/trainer/model.py
@@ -42,6 +42,7 @@ from prime_rl.trainer.models import (
     get_custom_vlm_cls,
     supports_custom_impl,
 )
+from prime_rl.trainer.models.deepseek_v4.attention import DeepseekV4Indexer
 from prime_rl.trainer.models.fusions import (
     apply_model_fusions,
     get_fsdp_shard_placement_fn,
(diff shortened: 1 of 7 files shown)


src/prime_rl/trainer/models/deepseek_v4/attention.py
```diff
@@ -547,34 +497,33 @@ def __init__(self, config: DeepseekV4Config):
...
-        # An early query has fewer than `top_k` readable entries, so top-k still hands back
-        # masked-out ones. Mark those `IGNORE_SLOT` (-1) rather than letting them leak into attention.
-        return torch.where(
-            readable.gather(-1, top_k_indices), top_k_indices, torch.full_like(top_k_indices, IGNORE_SLOT)
+        layout = packed.compression_layouts[self.compressor.compress_rate]
+        entry_start = layout.first_entry_of_doc[packed.tok_doc_idx].int()
+        entry_stop = (entry_start + self.compressor.causal_threshold(packed.position_ids)[0]).int()
+
+        # fp8_indexer has no batch axis
+        top_k_indices = torch.stack(
```
garrett361: or I could `assert` batch size = 1 and unsqueeze?

src/prime_rl/trainer/models/deepseek_v4/attention.py
```diff
@@ -547,34 +497,33 @@ def __init__(self, config: DeepseekV4Config):
             config, self.head_dim, config.compress_rates["compressed_sparse_attention"], n_series=2
         )
         self.q_b_proj = nn.Linear(config.q_lora_rank, self.num_heads * self.head_dim, bias=False)
-        self.scorer = DeepseekV4IndexerScorer(config)
+        self.weights_proj = nn.Linear(config.hidden_size, self.num_heads, bias=False)
 
+    @torch.no_grad()  # Returns non-differentiable integer indices.
     def forward(self, hidden_states: torch.Tensor, q_residual: torch.Tensor, packed: PackedContext) -> torch.Tensor:
         batch, seq_len, _ = hidden_states.shape
```
S1ro1, line 504: We can always assume batch is 1, this will also simplify the handling below, we always have packed varlen batches

src/prime_rl/trainer/models/deepseek_v4/attention.py
S1ro1: yes exactly, assert bs==1 and unsqueeze

src/prime_rl/trainer/model.py
```diff
@@ -461,20 +462,22 @@ def get_full_offload_dtype_policy(
...
+    symmetric.
     """
+    # TODO: no model here trains its indexer. DeepSeek's auxiliary KL objective, which supervises
+    # the top-k selection, is unimplemented, so these params are frozen rather than learned.
     logger = get_logger()
     num_frozen = 0
 
     for module in model.modules():
-        if isinstance(module, Indexer):
+        if isinstance(module, (Indexer, DeepseekV4Indexer)):
```
S1ro1, line 480: Will be good at some point to unify Indexer into a common API, such as MoE, but not needed rn I think

S1ro1 (commented):
minor nits, else looks good!

src/prime_rl/trainer/models/deepseek_v4/attention.py
garrett361: done

S1ro1 (approved):
lgtm!

PR #3508, opened 2026-09-08
feat(deepseek_v4): support context parallelism

garrett361:
Adds `ring` style CP support to deepseek v4.

The comms requirements are a bit different than those of other models currently supported. As a point of comparison, for GLM which also has key/value compression, the key/value creation is embarrassingly parallel along the seq dim and the compressed key/values can be created from cp-rank-local data.  Comms are of the form `gather_for_cp(kv_cp_shard)`.

For deepseek v4 attention layers there are two general components: sliding-window parts and compressed attention parts. The SWA treatment is embarrassingly parallel and handled analogously to GLM. The compressor components are not embarrassingly parallel and require different handling. The compression step involves pooling along seq dimension, which is what removes the embarrassingly parallel property and necessitates cross cp-rank communication prior to kv formation. This means that the `gather_for_cp` needs to occur at a comparatively earlier step than in GLM, say: we never create a sharded version of `kv` (for the compressed component) and instead use `gather_for_cp` to gather the tensors required to create the unsharded `kv` directly.

A more memory optimized approach would only pass small windows of tokens across CP ranks and then all-gather the small, post-compression tensors, but we re-use `gather_for_cp` and go for simplicity here.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -35,6 +35,7 @@ impl = "custom"        # or "hf" to force the HF path
 | Trinity (AFMoE) | `arcee-ai/Trinity-Mini`, … | ✅ | ✅ |
 | GLM-4 / GLM-4.5 / INTELLECT-3 | `THUDM/GLM-4-9B-0414`, `zai-org/GLM-4.5`, `PrimeIntellect/INTELLECT-3`, … | ✅ | ✅ |
 | GPT-OSS | `unsloth/gpt-oss-20b-BF16`, … | ✅ | ✅ |
+| DeepSeek V4 | `deepseek-ai/DeepSeek-V4-Flash-0731` | ✅ | `ring` only |
 
 Selective activation checkpointing works with either implementation. The custom path additionally enables EP, CP, low-precision training, and grouped MoE kernels. Forcing `impl = "hf"` is mostly useful when debugging and disables those model-specific runtime features.
 
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -35,7 +35,7 @@ impl = "custom"        # or "hf" to force the HF path
 | Trinity (AFMoE) | `arcee-ai/Trinity-Mini`, … | ✅ | ✅ |
 | GLM-4 / GLM-4.5 / INTELLECT-3 | `THUDM/GLM-4-9B-0414`, `zai-org/GLM-4.5`, `PrimeIntellect/INTELLECT-3`, … | ✅ | ✅ |
 | GPT-OSS | `unsloth/gpt-oss-20b-BF16`, … | ✅ | ✅ |
-| DeepSeek V4 | `deepseek-ai/DeepSeek-V4-Flash-0731` | ✅ | `ring` only |
+| DeepSeek V4 | `deepseek-ai/DeepSeek-V4-Flash-0731` | ✅ | ✅ |
 
 Selective activation checkpointing works with either implementation. The custom path additionally enables EP, CP, low-precision training, and grouped MoE kernels. Forcing `impl = "hf"` is mostly useful when debugging and disables those model-specific runtime features.
 
--- a/scripts/mini_moe.py
+++ b/scripts/mini_moe.py
@@ -21,8 +21,6 @@ from transformers.models.qwen3_5_moe.modeling_qwen3_5_moe import (
     Qwen3_5MoeForConditionalGeneration as HFQwen3_5MoeVLM,
 )
 
-from prime_rl.trainer.models.deepseek_v4 import DeepseekV4Config
-from prime_rl.trainer.models.deepseek_v4 import DeepseekV4ForCausalLM as PrimeRLDeepseekV4ForCausalLM
 from prime_rl.trainer.models.glm4_moe import Glm4MoeConfig
 from prime_rl.trainer.models.glm4_moe import Glm4MoeForCausalLM as PrimeRLGlm4MoeForCausalLM
 from prime_rl.trainer.models.laguna import LagunaConfig
@@ -33,7 +31,6 @@ from prime_rl.trainer.models.minimax_m2 import MiniMaxM2ForCausalLM as PrimeRLMi
 from prime_rl.trainer.models.qwen3_5_moe import Qwen3_5MoeForCausalLM as PrimeRLQwen3_5MoeVLM
 from prime_rl.utils.logger import setup_logger
 from prime_rl.utils.utils import default_dtype
-from prime_rl.utils.weights import load_state_dict, save_state_dict
 
 setup_logger("info")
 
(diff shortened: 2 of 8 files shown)


scripts/mini_moe.py
garrett361, line 1: For testing, will be reverting all edits to this file.

tests/unit/train/models/test_deepseek_v4_kernels.py
```diff
@@ -1134,3 +1147,149 @@ def test_kernel_and_eager_consumers_agree_on_shared_weights(layer_idx, doc_lens)
             assert kernel_grad is None and eager_grad is None, f"{name} trains on one path but not the other"
             continue
         _assert_within_the_bfloat16_floor(kernel_grad, eager_grad, param.grad, f"{name} gradient")
+
+
+# Widths that 4 divides. `(517, 1019)` puts the `cp = 2` cut inside the second document, and
+# `(300,)` gives 75-token shards at `cp = 4`, narrower than `sliding_window = 128`.
```
garrett361, line 1147: LMK if I should drop all of this

src/prime_rl/trainer/models/deepseek_v4/attention.py
```diff
@@ -235,25 +243,33 @@ class PackedContext:
...
-        if self.position_ids.shape[-1] != total_tokens:
-            raise ValueError(
-                f"position_ids covers {self.position_ids.shape[-1]} tokens, but the row has {total_tokens}"
-            )
-        if self.window_indices.shape[0] != total_tokens:
-            raise ValueError(
-                f"window_indices covers {self.window_indices.shape[0]} query rows, but the row has {total_tokens}"
-            )
+        n_queries = self.tok_doc_idx.shape[0]
+        assert self.position_ids.shape[-1] == n_queries, (
```
S1ro1: I probably missed this before, but can these cases ever sensibly even happen? that would point to issues in our code, which would lead to a failure anyway? Wondering even if we need those checks really

src/prime_rl/trainer/models/deepseek_v4/attention.py
```diff
@@ -339,51 +373,64 @@ class SparseAttnInputs:
...
     def __post_init__(self) -> None:
         # Shape invariants only. Asserting on index values would read the device, and this runs
         # once per layer per step.
-        if self.kv_buf.ndim != 4 or self.kv_buf.shape[2] != 1:
-            raise ValueError(f"kv_buf must be (batch, n_positions, 1, head_dim), got {tuple(self.kv_buf.shape)}")
-        if self.indices.ndim != 4 or self.indices.shape[2] != 1:
-            raise ValueError(f"indices must be (batch, seq_len, 1, n_slots), got {tuple(self.indices.shape)}")
-        if self.indices.shape[0] != self.kv_buf.shape[0]:
-            raise ValueError(f"kv_buf covers {self.kv_buf.shape[0]} batch entries and indices {self.indices.shape[0]}")
+        assert self.kv_buf.ndim == 4 and self.kv_buf.shape[2] == 1, (
```
S1ro1: Would say the same here, maybe worth removing?

garrett361: yeah I think just remove, overly defensive/sanity checking during initial testing

garrett361: removed

docs/advanced.md
```diff
@@ -35,6 +35,7 @@ impl = "custom"        # or "hf" to force the HF path
 | Trinity (AFMoE) | `arcee-ai/Trinity-Mini`, … | ✅ | ✅ |
 | GLM-4 / GLM-4.5 / INTELLECT-3 | `THUDM/GLM-4-9B-0414`, `zai-org/GLM-4.5`, `PrimeIntellect/INTELLECT-3`, … | ✅ | ✅ |
 | GPT-OSS | `unsloth/gpt-oss-20b-BF16`, … | ✅ | ✅ |
+| DeepSeek V4 | `deepseek-ai/DeepSeek-V4-Flash-0731` | ✅ | `ring` only |
```
samsja:
```suggestion
| DeepSeek V4 | `deepseek-ai/DeepSeek-V4-Flash-0731` | ✅ | ✅ |
```

src/prime_rl/trainer/models/deepseek_v4/attention.py
```diff
@@ -262,30 +262,48 @@ def build(
...
-        # Document-local by construction: a token's position is its distance from its own
-        # document's start, which is what `causal_threshold` and the entry rotation count in.
-        position_ids = (tok_idx - cu_seqlens[tok_doc_idx])[None]
         compress_rates = {
             config.compress_rates[layer_type]
             for layer_type in set(config.layer_types)
             if layer_type in config.compress_rates
         }
 
+        # These fields have one entry per query token, so they cover this rank's tokens only.
```
S1ro1, line 296: I'm gonna assume this logic is sound, I really don't feel like reviewing indexing/spliting logic, if you want me to just lmk but I assume it just constructs the Context to be CP aware

src/prime_rl/trainer/models/deepseek_v4/attention.py
```diff
@@ -607,6 +661,11 @@ class DeepseekV4Attention(nn.Module):
     Every layer type runs that same core over its local sliding window. The two compressed
     types additionally own a `compressor` whose output is concatenated onto the local KV,
     which is how a layer sees past the window.
+
+    Under context parallelism this rank holds the queries of one contiguous shard of the sequence
+    and all-gathers the key side, which every query needs in full. What crosses the wire is the
+    narrowest thing that suffices: `kv_proj`'s single `head_dim`-wide vector per token, and the
+    compressor's own per-token projections, never the `hidden_size`-wide stream they came from.
```
S1ro1: nit of nits: reads a bit AI sloppy, would just directly comment where the gather happens something along the lines:

```
We gather after down-projecting to minimize the payload size
```

src/prime_rl/trainer/models/deepseek_v4/modeling_deepseek_v4.py
```diff
@@ -194,6 +201,13 @@ def __init__(self, config: DeepseekV4Config):
 
         self.post_init()
 
+    def _cp_rank_and_world_size(self) -> tuple[int, int]:
```
S1ro1: This feels very wrong LOL, I think we have a somewhat ugly design here where the PackedContext is stateful (requires cp information), while its owner shouldn't be. However happy with this if we name it properly `get_cp_rank_and_size` or such. 

Also I think we don't need getattr here, it's always set by default, so would do it directly

src/prime_rl/trainer/models/deepseek_v4/attention.py
```diff
@@ -645,12 +704,26 @@ def __init__(self, config: DeepseekV4Config, layer_idx: int):
         compressor_class = COMPRESSOR_CLASSES[self.layer_type]
         self.compressor = compressor_class(config) if compressor_class is not None else None
 
+        self._cp_group: dist.ProcessGroup | None = None
```
S1ro1, line 693: nit: I'm not sure if we have some consensus on these being private/public, happy to just merge with as is.

tests/unit/train/models/test_deepseek_v4_kernels.py
```diff
@@ -1123,3 +1141,154 @@ def test_kernel_and_eager_consumers_agree_on_shared_weights(layer_idx, doc_lens)
...
+
+@requires_sparse_attn_kernel
+@pytest.mark.parametrize("doc_lens", CP_DOC_LENS, ids=CP_DOC_IDS)
+@pytest.mark.parametrize("cp_world_size", CP_WORLD_SIZES, ids=CP_WORLD_SIZE_IDS)
+@pytest.mark.parametrize(
+    "layer_idx",
+    [pytest.param(V4FLASH_CSA_LAYER, marks=requires_fp8_indexer), V4FLASH_HCA_LAYER, V4FLASH_SLIDING_LAYER],
+    ids=V4FLASH_LAYER_IDS,
+)
+def test_context_parallel_shards_reproduce_the_whole_row(layer_idx, cp_world_size, doc_lens, monkeypatch):
```
S1ro1, line 1226: Feels like the 2 tests test somewhat similar things, also given we'll eventually remove the eager path just for the sake of tests, maybe unify, resp remove this first one and keep the kernel one only (and extend it if needed to cover all things like grads, etc. Again optional

samsja (dismissed):

src/prime_rl/trainer/models/deepseek_v4/attention.py
garrett361: yes, just handles CP aware sharding, with some parts becoming cp-rank-local and others staying global as needed.

src/prime_rl/trainer/models/deepseek_v4/attention.py
garrett361: Agreed, cutting down.

src/prime_rl/trainer/models/deepseek_v4/attention.py
garrett361: I don't like how we set cp attrs anywhere currently, but yes I'm going to defer changing this for now

src/prime_rl/trainer/models/deepseek_v4/modeling_deepseek_v4.py
garrett361: yeah, I saw and hated this but left it for future cleanup. I'll resolve it now

tests/unit/train/models/test_deepseek_v4_kernels.py
garrett361: removed one

src/prime_rl/trainer/models/deepseek_v4/modeling_deepseek_v4.py
garrett361: Removed this specific method and reworked the cp attr setting, but might require more discussion

garrett361: Now I restored `_cp_rank_and_world_size` because my change broke other models. I'd like to rework and unify how we do CP setup across models in a separate PR where this will also get cleaned up, if that's alright by you.

garrett361: Did `_cp_rank_and_world_size` -> `get_cp_rank_and_world_size` now

S1ro1 (approved):
lgtm

PR #3512, opened 2026-09-09
fix(deepseek_v4): hardware guards for unrunnable gpu CI tests

garrett361:
Custom deepseek v4 flash attention requires more SMEM than some gpus (like the current CI runners) have. Skip the tests in those cases.

### DIFF
--- a/tests/unit/train/models/test_deepseek_v4_kernels.py
+++ b/tests/unit/train/models/test_deepseek_v4_kernels.py
@@ -180,52 +180,39 @@ V4FLASH_MODEL = dict(
     },
 )
 
-# The hand-built tensors of the first section describe the same CSA layer `V4FLASH_MODEL` does, so
-# they are read off it rather than written out again: 64 heads over 512 channels, each query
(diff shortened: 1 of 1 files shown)


tests/unit/train/models/test_deepseek_v4_kernels.py
```diff
@@ -172,52 +180,39 @@ def _set_attn_impl(module: nn.Module, impl: str) -> None:
     },
 )
 
-# The hand-built tensors of the first section describe the same CSA layer `V4FLASH_MODEL` does, so
-# they are read off it rather than written out again: 64 heads over 512 channels, each query
-# gathering `sliding_window + index_topk = 128 + 512` slots from a single KV group, in bfloat16.
+# Read off `V4FLASH_MODEL` so the hand-built tensors below describe the same CSA layer it does.
```
garrett361, line 183: just shortening up comments everywhere

S1ro1 (approved):

PR #3533, opened 2026-09-11
chore(trainer): unify context-parallel setup across models

garrett361:
Unifies all context parallel setup so that a single frozen `CPContext` dataclass is shared by all cp-aware modules, stored under a `self.cp_contex: CPContext` attr.


src/prime_rl/trainer/models/base.py
```diff
@@ -169,4 +190,14 @@ def init_buffers_post_meta(self) -> None:
         raise NotImplementedError(f"init_buffers_post_meta is not implemented for {self.__class__.__name__}")
 
 
+def _setup_context_parallel_children(
+    module: nn.Module, cp_group: dist.ProcessGroup, cp_rank: int, cp_world_size: int, cp_style: CPStyle
+) -> None:
+    for child in module.children():
+        if hasattr(child, "setup_context_parallel"):
+            child.setup_context_parallel(cp_group, cp_rank, cp_world_size, cp_style)
+        else:
```
S1ro1: Maybe this shouldn't be else? Or modules should be responsible for children?

garrett361: You should be out celebrating. But I don't have a strong preference here. Went recursive for completeness, but can also just go top-level and delegate submodule cp setup responsibility to the top-level children of `PreTrainedModelPrimeRL`.

src/prime_rl/trainer/models/nemotron_h/mamba.py
```diff
@@ -153,11 +161,11 @@ def forward(self, hidden_states: torch.Tensor, cu_seqlens: torch.Tensor) -> torc
         dt_bias = self.dt_bias
         norm_weight = self.norm.weight
 
-        if hasattr(self, "process_group"):
```
samsja, line 156: can't believe we had hasattr here before, i hate hasattr with a deep pation

samsja (dismissed):
lgtm

src/prime_rl/trainer/models/base.py
garrett361: oh, nvm, misunderstood and I see the gap. Will address

samsja (dismissed):

src/prime_rl/utils/cp.py
```diff
@@ -3,80 +3,71 @@
...
+    cp_context = CPContext(cp_group, cp_rank, parallel_dims.cp, config.cp_style)
+    for module in model.modules():
+        if not hasattr(module, "cp_context"):
+            continue
+        if not isinstance(module.cp_context, CPContext):
+            raise TypeError(
+                f"{type(module).__name__}.cp_context is {type(module.cp_context).__name__}, not a CPContext; "
+                "context-parallel setup claims that attribute name, so rename it"
+            )
+        module.cp_context = cp_context
```
garrett361, line 68: I don't love this keying off `cp_context` attr existence, but it is better than threading a `cp_context` attr through all `__init__` methods. Open to suggestions.

S1ro1: agreed but can't think of better one

S1ro1 (approved):
lgtm

PR #3535, opened 2026-09-12
feat(moe): add SonicMoE BF16 expert compute backend

S1ro1:
Adds `model.moe.compute.backend = "sonicmoe"` under BF16 compute for Qwen3 MoE, using the upstream [SonicMoE package](https://github.com/S1ro1/sonic-moe/tree/0bf06ac7bf5b0800a7202743c1fb5722ed004350).

`GroupedExperts` owns its parameters and delegates execution through one `ExpertCompute` interface:

- `GroupedGemmExpertCompute` implements gate/up → activation → down. `BF16ExpertCompute`, `DeepGemmFP8ExpertCompute`, and `MXFP8ExpertCompute` inherit that forward and supply their GEMM callable and token alignment. MXFP8 binds its gradient-precision recipe with `partial`.
- `SonicMoEExpertCompute` implements the expert pipeline directly with Sonic's fused kernels.

Runtime selects one compute backend. Sonic consumes already-dispatched expert rows and fused gate/up weights; the existing local, torch EP, and DeepEP dispatchers retain routing-score application and combination. Fusion runs in normal model setup.

```toml
[trainer.model]
name = "Qwen/Qwen3-30B-A3B"
ep = 2

[trainer.model.moe.compute]
type = "bf16"
backend = "sonicmoe"

[trainer.model.moe.dispatch]
type = "torch" # or "deepep"
```

Uses the default `gate_up` fusion. The dependency is installed with the `sonic-moe` extra and included in `all`.

Validation:

- MoE, fusion, and config unit suites: **198 passed**.
- H200, local and EP2: outputs, input gradients, router gradients, and expert-weight gradients pass comparisons against the native BF16 compute baseline with a 2% relative L2 tolerance, including uneven routing and unused experts.
- Local and torch EP2 forward/backward also pass `torch.compile(fullgraph=True)`.
- DeepEP eager EP2 passes. Strict fullgraph compilation fails in DeepEP's pybind configuration construction with both Sonic and the default BF16 compute backend.
- H200 torch EP2: five packed math SFT steps through `uv run sft`, with compilation, full activation checkpointing, and activation offloading. Two-layer native Qwen3 MoE checkpoint; loss **11.8892 → 9.0528**, gradient norms **15.0008 → 9.5380**. The first four 1024-token rows packed **131 samples**. Optimizer and reduction dtypes retain their FP32 defaults.
- DeepEP with full activation checkpointing fails during FSDP recomputation with a cached CPU-copy shape mismatch (**1048576 vs 5**) in both the default BF16 compute backend and Sonic. With checkpointing off, compiled DeepEP also fails while tracing buffer initialization (`ProcessGroup._device_types`). Eager DeepEP SFT passes all five packed math steps with `model.compile`, `model.ac`, and `model.ac_offloading` set to `"None"`; loss **11.8892 → 9.0881**.
- Ruff and `git diff --check` pass.

H200 EP=8 benchmark: both native BF16 and Sonic complete 20 packed math SFT steps with sequence length 8192, batch 8/microbatch 1, selective checkpointing, default compilation, SignSGD, FlashAttention 3, and QKV + gate/up fusion. Optimizer/activation offloading is disabled; optimization/reduction dtypes remain FP32.

| Backend | Mean step (6–20) | Tokens/s | Estimated MFU | Rank-0 peak reserved memory |
|---|---:|---:|---:|---:|
| Native BF16 | 1.702 s | 38,502 | 13.58% | 89.34 GiB |
| SonicMoE | 1.743 s | 37,595 | 13.26% | 68.62 GiB |

Both runs have finite losses and zero NaN losses. Across all GPUs, peak physical memory sampled every second was 99.89 GiB native / 74.70 GiB Sonic. MFU uses the trainer's estimator, which counts attention at sequence length 8192 despite packing shorter documents.

Two small package patches make SAC and changing EP token counts practical:

- [QuACK benchmark isolation](https://github.com/Dao-AILab/quack/pull/204), backported to the required 0.6.4 version: isolate autotuning trials from autograd and SAC dispatch modes. Otherwise synthetic warmup matmuls can retain tens of GiB or pollute checkpoint saved-tensor counts. The final kernel call preserves normal model behavior.
- [SonicMoE EP token bucketing](https://github.com/S1ro1/sonic-moe/commit/0bf06ac7bf5b0800a7202743c1fb5722ed004350): include the variable received-token axis in transposed weight-gradient inputs. Compiled SAC checks show no additional tuning or QuACK compilation-cache misses for subsequent token counts in the same bucket, with output and gradient relative L2 differences below 0.6% versus native BF16.

The successful run uses these package pins and no synchronization changes. The upstream QuACK autotuner test file passes all four tests on H200; removing either isolation context reproduces its corresponding checkpoint failure.


garrett361:
> DeepEP with full activation checkpointing fails during FSDP recomputation with a cached CPU-copy shape mismatch (1048576 vs 5) in both the default BF16 compute backend and Sonic

still an issue?

src/prime_rl/trainer/models/layers/expert_compute.py
```diff
@@ -0,0 +1,107 @@
...
+        token_group_alignment: int = 8,
+    ) -> None:
+        self.gemm = gemm
+        self.token_group_alignment = token_group_alignment
+
+    def __call__(self, experts: "GroupedExperts", x: torch.Tensor, num_tokens_per_expert: torch.Tensor) -> torch.Tensor:
+        assert x.dim() == 2
+
+        def to_local(tensor: torch.Tensor) -> torch.Tensor:
+            return tensor.to_local() if isinstance(tensor, DTensor) else tensor
```
garrett361, line 55: noting we're redefining this helper on every `__call__`

src/prime_rl/trainer/models/layers/expert_compute.py
```diff
@@ -0,0 +1,107 @@
...
+class GroupedGemmExpertCompute(ExpertCompute):
+    def __init__(
+        self,
+        gemm: Callable[[torch.Tensor, torch.Tensor, torch.Tensor], torch.Tensor],
+        token_group_alignment: int = 8,
+    ) -> None:
+        self.gemm = gemm
+        self.token_group_alignment = token_group_alignment
+
+    def __call__(self, experts: "GroupedExperts", x: torch.Tensor, num_tokens_per_expert: torch.Tensor) -> torch.Tensor:
```
garrett361, line 51: initial reaction is that this is an annoyingly logic-heavy base class and the fact that the primary/only subclass of it `SonicMoEExpertCompute(ExpertCompute)` needs to completely rewrite the only method the parent class provides is a bit of a smell. Not sure how much this abstraction is gaining us. wdyt?

PR #3538, opened 2026-09-13
fix(inference):Fix native disk KV cache offload tier type

Alex-Shan:
## Summary

Align the disk-backed KV cache secondary tier with vLLM's registered filesystem tier name: `fs`, not `fs_python`.[#43600](https://github.com/vllm-project/vllm/pull/43600)

## Problem

The inference configuration emitted a secondary tier of the form

```python
{"type": "fs_python", "root_dir": <configured disk path>}
```

but the tier factory only registers the filesystem secondary tier under the name `fs`:

```text
example, fs, p2p, obj
```

Because the identifier did not match, vLLM aborted engine initialization with:

```text
ValueError: Unknown secondary tier type: 'fs_python'
```

## Changes

- Rename the filesystem secondary tier identifier from `fs_python` to `fs` everywhere it is constructed or referenced — configuration generation, tier-factory registration, and log messages.
- Keep `root_dir` as the configured disk path; no behavioural change to the tier itself, only its name.
- Update the affected tests to use `fs`.
- No changes to the TOML configuration format.

## Validation

With CPU offload and disk offload both enabled, on vLLM `vLLM 0.28.0` and a Qwen3-4B inference configuration:

- the engine initializes without the `Unknown secondary tier type` error;
- `fs_python` no longer appears in generated configs or logs.


S1ro1 (approved):
lgtm, thanks!

PR #3539, opened 2026-09-13
Fix Qwen3.5 NIXL weight-loading composition

S1ro1:
Qwen3.5-35B-A3B fails on its first NIXL update because vLLM's packed Q/K/V convolution loader reaches native `aten.copy_` through indexed assignment. Record that terminal write through the existing recorder. Use the full checkpoint config and text `model_type` so MoE export retains the correct prefixes and conversion chain with upstream Transformers configs.

Record `cat` and its aliases as children in `LazyWeight._ops`, preserving each input's preceding graph. Every terminal write uses the same graph lowerer: it starts with one `RecordedCopy`, adds source contributions at cat, and translates subsequent operations into source chains and destination layouts. `apply_chain` evaluates tensor operations; the existing per-source transfer planner and replay engine consume the resulting copies unchanged.

For example, `cat([a, b])[3:6]` with four rows per input becomes `a[3:4] -> destination[0:1]` and `b[0:2] -> destination[1:3]`. Transpose can preserve two copies; reshape can split them further. Nested concatenations and BF16/FP32 promotion follow the same lowering path. Casts retain their original order and use the existing replay rules.

Validation:
- `uv run --no-sync pytest tests/unit/transports/test_nixl_graph.py -q`: 209 passed. Ruff lint/format and whitespace checks passed.
- Graph tests cover copy recording through method/native/indexed writes, strided destinations, concatenation with one to three inputs, subsequent transformations, nested concatenations, aliases, and cast order. Model-specific export/loader tests and unrelated validation checks are excluded from this unit suite.
- Earlier GPU validation completed 20 steps in job 572. Job 576 reached step 9 on `f708d540d` before cancellation. Job 577 was launched from `6614dc272`; those runs predate this test cleanup.
- GPU setup: Qwen/Qwen3.5-35B-A3B, custom trainer EP=8, inference TP=8, SAC, full CPU SignSGD offload, math-env on Prime sandboxes, 2048 completion tokens.

---

> [!NOTE]
> **Medium Risk**
> Changes core NIXL weight trace/lowering used on every inference update; mistakes could corrupt sharded loads, though behavior is heavily covered by new unit tests.
> 
> **Overview**
> Fixes the first NIXL policy update for **Qwen3.5-35B-A3B** by making lazy weight tracing handle **concatenation**, **indexed assignment**, and **post-cat layout ops** the way vLLM’s packed Q/K/V loaders actually write weights.
> 
> **NIXL graph (`graph.py`)** now records `torch.cat` / aliases on `LazyWeight`, routes terminal writes through **`lower_graph`** so each trainer source becomes one or more `RecordedCopy` entries (no `cat` left at replay time), and hooks **`aten.copy_.default`** so slice assignment is traced like `copy_`. **`apply_chain`** still plans/replays per-source ops; cat is metadata-only until lowering splits it.
> 
> **Config wiring**: vLLM tracing passes **`hf_config`** (not `hf_text_config`), and Qwen3.5 MoE conversion gates on **`text_config.model_type == "qwen3_5_moe_text"`** instead of a custom config class—keeping expert rename/split-concat prefixes aligned with upstream Transformers.
> 
> **Tests**: new **`tests/unit/transports/test_nixl_graph.py`** covers strided destinations, concat + narrow/chunk/split/reshape/transpose/casts, nested cats, and copy entrypoints (method, native ATen, assignment).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e26b0e736eac207062ce50e9357e5ec9a9473436. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/unit/transports/test_nixl_graph.py
```diff
@@ -0,0 +1,199 @@
...
+    recorder.register_destination_storage(owner)
+    lazy = LazyWeight("weight", source.shape, source.dtype, source.device, recorder)[::2, ::2].float()
+    view = destination[1:5:2, 1:7:2]
+
+    if entrypoint == "method":
+        assert view.copy_(lazy) is view
+    elif entrypoint == "native":
+        assert torch.ops.aten.copy_.default(self=view, src=lazy) is view
+    else:
+        destination[1:5:2, 1:7:2] = lazy
```
garrett361, line 28: @S1ro1 TBH I'm trying to review this but am struggling even to understand the tests. Like what are we verifying here exactly? What do the different `entrypoint` values mean semantically?

garrett361 (commented):
Sorry this is taking forever, getting to it now.

tests/unit/transports/test_nixl_graph.py
garrett361: nvm, I get what the `entrypoint` names mean now. Are the asserts meaningful, though? Not asking for a change, just wanted to understand. Don't see how they could be violated.

garrett361 (approved):
Not my strongest review, but in my understanding of the tests, they look good

PR #3541, opened 2026-09-14
fix(orchestrator): stop solve_rates counting a zero-sum group as solved none

erkancamli:
### Problem

\`solve_rates\` classifies each group with two different kinds of predicate (\`metrics.py:250-251\`):

\`\`\`python
solved_none = sum(sum(trace.reward for trace in group) == 0 for group in groups.values())
solved_all  = sum(all(trace.reward == 1.0 for trace in group) for group in groups.values())
\`\`\`

\`solved_all\` is a per-trace predicate; \`solved_none\` sums the group's rewards. That is fine for the binary 0/1 rewards the metric was written for, but rewards can go negative here, and then the sum is the wrong shape in two ways:

- **A group whose rewards cancel** reads as "solved none" even though a trace solved the task. \`[1.0, -1.0]\` sums to 0.
- **A group where every trace scored below zero** is not counted at all. A single trace with reward \`-1.0\` sums to \`-1\`, so it falls through to \`solved_some\`, which is the one thing it certainly is not.

The second one is the case that actually shows up. \`configs/debug/algo/rae.toml\` ships \`group_size = 1\` on \`kuhn-poker\`, and \`solve_rates\` groups per agent via \`by_agent()\` (\`metrics.py:288-292\`), so a lost game is a one-trace group with reward \`-1.0\`. On such a run \`solved_none\` can read 0% while the agent is losing half its games, and every one of those losses is reported as \`solved_some\`.

\`RAEAlgorithm\`'s docstring says why negative rewards are first class here rather than a corner case:

> in a zero-sum game the group mean is ~0 whatever the policy does

### Fix

State the predicate directly: a group solved none when no trace in it scored above zero.

\`\`\`python
solved_none = sum(max(trace.reward for trace in group) <= 0 for group in groups.values())
\`\`\`

\`solved_none\` and \`solved_all\` stay mutually exclusive (\`max <= 0\` versus \`all == 1.0\`), so \`solved_some\` cannot go negative. Groups are built with \`setdefault(...).append(...)\`, so none is empty and \`max()\` is always defined. For binary 0/1 rewards nothing changes.

### Test

\`test_solve_rates_with_negative_rewards\` in \`tests/unit/orchestrator/test_metrics.py\`.

Before the fix:

\`\`\`
E       assert (0.0, 1.0, 0.0) == (0.0, 0.0, 1.0)
E         At index 1 diff: 1.0 != 0.0
\`\`\`

After: \`14 passed\`.

### Environment note

\`uv sync\` does not run in my sandbox (\`deep-ep = false\` is rejected by the older uv available to me, and the \`deps/verifiers\` submodule comes up empty), so I built an ad-hoc venv and installed \`verifiers\` from PyPI instead. In that venv 21 other tests under \`tests/unit/orchestrator\` fail for environment reasons. I verified they are pre-existing by stashing this change and re-running: the same 21 failures appear, with 69 passing instead of 70. Only \`test_metrics.py\` is affected by this patch, and it is fully green.

---

Correcting my own first revision: it quoted \`solved_all\` as \`all(trace.reward > 0 ...)\` and argued the fix was to mirror it. The real line is \`all(trace.reward == 1.0 ...)\`, so \`max(...) <= 0\` is not its mirror and that argument did not hold. The predicate is still the one I would pick, but on its own terms, and the all-negative case above is the reason rather than the zero-sum one I led with.


erkancamli:
No checks have run on this one yet, so there is nothing for a reviewer to go on. Could someone approve the workflow runs?

The change is one line plus a regression test: a group whose rewards sum to zero, like `[1.0, -1.0]`, was counted as `solved_none` because that branch summed the group rewards while `solved_all` is computed per trace. The test fails on `main` and passes with the fix.

mikasenghaas (approved):
eventually, we prob want a task to expose a`correctness_threshold: float | None` which we can use to classify each sample and then derive solve none/all rates. but this pr seems like a strict improvement to what we currently have so gtg from my side

erkancamli:
Friendly ping on this one: it was approved on 17 Sept and all checks are green (the benchmark_regression run was cancelled rather than failed). It is 53 commits behind main but only touches `solve_rates` and its test; happy to rebase if you'd prefer. The `correctness_threshold` follow-up you suggested is tracked in #3566.

PR #3542, opened 2026-09-14
exp: adaptive concurrency improvs

mikasenghaas:
## Summary

- Add 50-step Terminal Lego runs for prefill/decode disaggregation, Mooncake KV offload, and Qwen3.5 hybrid linear attention, composed from a shared `base.toml` and experiment overlays.
- Raise the adaptive-concurrency ceiling to 2,048 so the controller can expose an operating band instead of remaining capped at 1,024.
- Run Mooncake and linear attention with one inference node and one replica; retain the required one-prefill plus one-decode node layout for P/D disaggregation.
- Use Qwen3-30B-A3B or Qwen3.5-35B-A3B at a 65,536-token context on one trainer node.
- Use group size 8, EP=8 for training, and expert parallelism for every inference run. Use CP=8 for the Qwen3 runs and CP=1 for Qwen3.5 to avoid its unsupported integer-autograd path during activation-checkpoint recomputation.
- Use activation offloading, state-only optimizer offload, and SignSGD.
- Keep router replay, reward length penalties, evals, and checkpoints disabled so the runs isolate inference behavior.
- Give each `-v3` variant its own run, W&B, SLURM, and Prime sandbox label.
- Fix single-rank P/D roles by omitting vLLM external-load-balancer rendezvous arguments when data parallelism is one.
- Select vLLM's V1 runner from the persisted P/D KV-transfer flag in per-rank configs so NIXL P/D transfer starts correctly.
- Document the metric mapping and evaluation criteria, and add a JSONL analyzer for concurrency, queue, cache, throughput, and P/D-stage metrics.

## Verification

- Ran the focused runner-selection tests in `tests/unit/test_configs.py` (2 passed).
- Loaded the Terminal Lego CLI schema with authenticated Hugging Face access.
- Dry-ran all three stacked v3 `base.toml` plus experiment-overlay combinations at 50 steps.
- Confirmed all resolved v3 configs use group size 8, a concurrency ceiling of 2,048, trainer EP=8, and inference expert parallelism.
- Confirmed Mooncake and linear attention resolve to one inference node/replica, while P/D resolves to two inference nodes and one replica.
- Confirmed the Qwen3.5 overlay resolves trainer CP=1.
- Confirmed the configs disable full offload, router replay, and the length penalty.
- Syntax-checked each rendered v3 SLURM script with `bash -n`.
- Ran Ruff on the changed Python and experiment config paths.
- Ran a P/D launch on 24 H200 GPUs: both TP=8 vLLM engines became healthy, NIXL P/D transfer initialized, and the startup weight broadcast completed.


configs/experiments/adaptive-concurrency/base.toml
```diff
@@ -0,0 +1,67 @@
...
+[weight_broadcast]
+type = "nccl"
+timeout = 3600
+
+[model]
+name = "Qwen/Qwen3-30B-A3B-Thinking-2507"
+
+[trainer]
+dist_timeout_seconds = 3600
+enable_router_replay = true
```
mikasenghaas: lets not use r3

configs/experiments/adaptive-concurrency/base.toml
```diff
@@ -0,0 +1,67 @@
...
+
+[orchestrator]
+batch_size = 128
+group_size = 16
+max_off_policy_steps = 32
+
+[orchestrator.algo]
+type = "grpo"
+
+[orchestrator.algo.length_penalty]
```
mikasenghaas: remove this

PR #3545, opened 2026-09-14
fix(kernels): keep FP8 indexer quantization scalars in FP32

KavinKrishnan:
## Summary

Keeps the FP8 indexer's quantization scalars in FP32 inside the Triton kernel.

## Problem

`eps`, `fp8_min` and `fp8_max` arrive as Python floats. Inductor lowers those as
f64, so the quantization arithmetic is performed at double precision against
f32 tensor data. On GLM-5 this surfaces as a compile failure in the indexer
path rather than as a numerical difference, which makes it look unrelated to
precision.

## Fix

Cast the three scalars to `tl.float32` at the top of the kernel, so the
quantization math runs at the precision the tensors are already in.

```python
eps = eps.to(tl.float32)
fp8_min = fp8_min.to(tl.float32)
fp8_max = fp8_max.to(tl.float32)
```

## Notes

Two files, thirteen lines. No ModelExpress involvement — this is independent of
the `mx_refit` transport work in #3487 and is submitted separately so it does
not wait on that review.


KavinKrishnan:
Folded into #3487 and closing this in favour of a single pull request.

The change itself is unaltered — same two files, same thirteen lines, now
carried as one commit in #3487's series. GLM-5 is the motivating configuration
for the `mx_refit` transport and does not train without this fix, so reviewing
them together is less confusing than reviewing them apart, even though the
kernel fix has no ModelExpress involvement of its own.

Happy to reopen as a standalone PR if a maintainer would rather take the
thirteen-line precision fix on its own timeline.

PR #3546, opened 2026-09-14
feat(examples): glm-4.5-air scaleswe — muon 2 train + 4 infer, with 4-node and 2-node sign-SGD variants

samsja:
## Summary

Extends the GLM-4.5-Air `scaleswe` advanced example with budget variants, so the same recipe scales from six nodes down to two:

- `swe.toml` (updated): 2 trainer nodes with Muon instead of the single-node sign-SGD setup, and 4 one-node inference replicas — matching the research-prod air ablation base config (research-prod PR #198) and the sibling `search.toml`/`terminal.toml`.
- `swe-4-node.toml` (new): the previous default — 1 trainer + 3 inference replicas with the single-node sign-SGD trainer (full optimizer offload, skip-optimizer checkpointing), batch size 256.
- `swe-2-node.toml` (new): train a 100B+ MoE on only two nodes — 1 trainer + 1 inference replica, batch size 64 (4 groups of 16).

Router replay, `max_off_policy_steps`, and the length-penalty weights are unchanged everywhere.

## Changes

- **`swe.toml`**
  - **Deployment**: `1 train + 3 infer replicas` -> `2 train + 4 infer replicas` (one-node replicas)
  - **Optimizer**: `sign_sgd` -> `muon`; drop the sign-SGD-era `optim_cpu_offload = false` / `full_offload = true` and `[trainer.ckpt] skip_optimizer` (optimizer state is checkpointed again, so runs can resume)
  - **Trainer memory**: `fused_lm_head_token_chunk_size` (8192) and `optim_cpu_offload` (true) use their defaults (per review); keep cp=4 (ulysses) and activation checkpointing + offloading
  - **Orchestrator**: unchanged (`batch_size = 256`, `group_size = 16`, `max_off_policy_steps = 32`, linear length penalty with `num_output_tokens_weight = 0.0`, input/turns weights 0.1)
  - **Router replay**: unchanged, stays enabled (`trainer.enable_router_replay` + `inference.vllm.enable_return_routed_experts`)
  - **Inference**: added `enable_expert_parallel = true` and `gpu_memory_utilization = 0.85` (TP=8 + EP)
- **`swe-4-node.toml`** (new): the pre-PR `swe.toml` config as its own file — 1 train + 3 infer replicas, sign-SGD single-node trainer, batch size 256, per-variant SLURM job name / sandbox labels / W&B name.
- **`swe-2-node.toml`** (new): same as the 4-node variant but `num_infer_replicas = 1` and `batch_size = 64` — train a 100B+ MoE on only two nodes.
- **Docs**: example README table + topology paragraph updated for all three SWE configs; `docs/scaling.md` examples list and `docs/configuration.md` example index link the two-node variant.

## Validation

- `uv run rl @ examples/advanced/glm-4.5-air/swe.toml --dry-run`
- `uv run rl @ examples/advanced/glm-4.5-air/swe-4-node.toml --dry-run`
- `uv run rl @ examples/advanced/glm-4.5-air/swe-2-node.toml --dry-run`


examples/advanced/glm-4.5-air/swe.toml
```diff
@@ -63,9 +64,8 @@ impl = "custom"
 attn = "flash_attention_3"
 cp = 4
 cp_style = "ulysses"
-optim_cpu_offload = false
-full_offload = true
-fused_lm_head_token_chunk_size = 1024
+optim_cpu_offload = true
+fused_lm_head_token_chunk_size = 8192
```
mikasenghaas: fine to use default here? should be same

examples/advanced/glm-4.5-air/swe.toml
```diff
@@ -63,9 +64,8 @@ impl = "custom"
 attn = "flash_attention_3"
 cp = 4
 cp_style = "ulysses"
-optim_cpu_offload = false
-full_offload = true
-fused_lm_head_token_chunk_size = 1024
+optim_cpu_offload = true
```
mikasenghaas: this is also default i believe but whatever

mikasenghaas (dismissed):

PR #3552, opened 2026-09-15
perf(optim): reuse pinned CPU buffers in CPUOffloadOptimizer

garrett361:
Allocate `CPUOffloadOptimizer`'s CPU-side buffers once, pinned, and copy optimizer state into them non-blocking, instead of allocating a fresh pageable buffer plus a fresh pinned buffer on every step.

- The old path copied each state tensor into fresh pageable memory blocking, then `.pin_memory()` allocated a second buffer and copied again host to host. `Tensor.to` cannot allocate pinned memory, and `non_blocking` was set to `not self.pin_memory`, which disabled the async path in the one case where a pinned destination would have made it work.
- Buffers are keyed by `(parameter, state key)` and reallocated if a state tensor's shape or dtype changes, which covers `load_state_dict`.
- One `torch.cuda.synchronize()` now ends the offload, replacing the per-tensor stall. The `pin_memory=False` path and the host-to-device direction are unchanged, the latter deliberately: those tensors must be freshly allocated for the GPU memory to be released between steps.
- The first commit adds the tests with the buffer-reuse one marked `xfail`, since the three correctness tests pass on the old code and only that one exposes the bug. The second commit drops the `xfail`.
- Measured on DS V4-Flash SFT, one node of 8x H200 at 6 layers, `seq_len` 131072, `cp=8`, `ep=8`, micro-batch 1, 10 steps, steady-state medians, with the fix as the only change from production defaults: optimizer 26.9 s to 3.0 s and whole step 30.9 s to 6.5 s, at an unchanged 103.8 GiB peak. Trace of the old 26.7 s window: 18.2 s pageable D2H at 2.2 GB/s, 5.4 s `pin_memory` re-copy, 1.5 s pinned H2D, and 0.15 s of AdamW kernels. Per-rank optimizer state is what sizes this window, and the 6-layer proxy carries 4.93e9 parameters per rank against production's 4.44e9, so it is not a scaled-down number.

Verified on an H200: 3 passed / 1 xfailed at the test commit, 4 passed at the fix commit, with `tests/unit/train/test_cpu_adam.py` and `tests/unit/test_configs.py -k offload` still green.


tests/unit/train/test_state_offload.py
garrett361, line 1: Would cut this down prior to merging

src/prime_rl/trainer/optim/state_offload.py
```diff
@@ -28,24 +40,19 @@ def _move_states(self, device: str):
                 if isinstance(value, DTensor):
                     local_tensor = value._local_tensor
```
garrett361, line 41: Not sure why we access the private `_local_tensor` rather than `DTensor.to_local()` here. Is there a reason @samsja?

src/prime_rl/trainer/optim/state_offload.py
```diff
@@ -28,24 +40,19 @@ def _move_states(self, device: str):
...
                     if device == "cpu":
-                        non_blocking = not self.pin_memory
-                        new_local = local_tensor.to("cpu", non_blocking=non_blocking)
-                        if self.pin_memory and not new_local.is_pinned():
-                            new_local = new_local.pin_memory()
+                        new_local = self._offload_to_cpu(param, key, local_tensor)
                     else:
                         new_local = local_tensor.to(device, non_blocking=True)
                     new_dtensor = copy.copy(value)
                     new_dtensor._local_tensor = new_local
```
garrett361, line 47: Should also probably be `DTensor.from_local`

samsja (approved):

PR #3558, opened 2026-09-16
perf(deepseek-v4): fuse the mHC chain and route unweighted RMS norms through quack

garrett361:
DeepSeek V4's manifold-constrained hyper-connection (mHC) chain ran eager and cost far more than
its arithmetic warrants: a python Sinkhorn loop of roughly 119 tiny launches over stream mixing
matrices, an fp32 copy of the flattened streams materialized purely as an intermediate, a
multi-kernel write-back over the full (batch, sequence, stream, hidden) tensor, and a `q_b_norm`
whose input was non-contiguous. This fuses or reroutes all four.

- The unweighted RMS norms (the mHC input norms and attention's `q_b_norm`) dispatch to quack
  through the same gate the shared `RMSNorm` uses, with an optional `out_dtype`. `q_b_norm` moves
  ahead of the head transpose so quack's flatten is a view rather than a copy; RMS over the channel
  axis commutes with the transpose exactly.
- The input norms take the bf16 streams directly and emit fp32, so the fp32 copy is never
  materialized. The gate math downstream of the norm stays fp32.
- The Sinkhorn loop and the write-back become two Triton ops vendored from NVIDIA/Megatron-LM at
  `bb5dfd08f` (BSD-3, full license text reproduced in the file header). They require a power-of-two
  `hc_mult` and a CUDA input, both of which Triton itself enforces, and the eager implementations
  move to `eager_reference.py`, where the tests measure the kernels against them.
- The rank-4 mHC residual is named `mhc_states` throughout, so `hidden_states` always means the
  rank-3 collapsed sequence, and the write-back moves onto `DeepseekV4HyperConnection` as
  `update_states`.
- Both fused paths are covered on GPU, measured against the eager references rather than a
  runtime fallback branch.

Forward/backward ms per call on an H200 at batch 1, sequence 16384, `hc_mult` 4, hidden dimension
4096 (the Sinkhorn row is measured on the stream mixing tensor):

| Op | Before | After |
| --- | --- | --- |
| mHC front half | 2.89 / 8.10 | 1.05 / 2.54 |
| Sinkhorn | 1.05 / 3.14 | 0.10 / 0.81 |
| Stream write-back | 5.96 / 7.89 | 0.40 / 0.94 |
| `q_b_norm` | 0.99 / 2.85 | 0.30 / 0.96 |

A 4-node full-depth A/B goes from 15.97 to 13.11 s/step at 4.2 GiB lower peak. That was measured on
the SFT-optim stack, not this branch: on `main` step time is optimizer-offload dominated, so the
same A/B there would hide the change inside the offload's shadow.


src/prime_rl/trainer/models/kernels/deepseek_v4/dsv4_mhc.py
garrett361, line 1: Eventually plan to move all of these to prime-kernels? CC @MarioSieg

samsja: probably fine to have .py kernel for now, would move the heavy one that need compiled to be run to prime kernel so that we can ship wheel etcc. Would keep it here for now

src/prime_rl/trainer/models/deepseek_v4/hyperconnections.py
```diff
@@ -3,23 +3,36 @@
 from torch import nn
 
 from prime_rl.trainer.models.deepseek_v4.configuration_deepseek_v4 import DeepseekV4Config
+from prime_rl.trainer.models.kernels.deepseek_v4 import dsv4_mhc
+from prime_rl.trainer.models.layers import norms
+
+
+def can_use_fused_mhc(t: torch.Tensor, hc: int) -> bool:
+    """The vendored Triton kernels tile the stream axis with `tl.arange`, so `hc` must be a power of two."""
+    return t.is_cuda and hc & (hc - 1) == 0
```
samsja: a bit puzzle about this one, shouldn't we just fail if this is not working ? like deepseekv4 flash shape work with this right ? I guess the only usecase is smaller model but then for smaller model for debug  I would hope thats its using the same kernel (example when checking kl mismatch)

tldr: I don't think we should have silent routing to different code

garrett361: yeah thought about removing this as well, will do

garrett361: done

samsja (approved):

PR #3562, opened 2026-09-16
perf(deepseek-v4): optimize interleaved RoPE with torch.compile

garrett361:


garrett361:
Actually it looks like [vLLM forces the entire rope computation to fp32](https://github.com/vllm-project/vllm/blob/v0.28.0/csrc/libtorch_stable/pos_encoding_kernels.cu#L29-L30), which negates some of this work, so closing in favor of a future fp32-respecting PR.

PR #3570, opened 2026-09-17
chore(trainer): update the reduce_scatter API and fix a ring attention edge case

garrett361:
Drops a deprecated collective call in `ring_attn.py`. `torch.distributed.reduce_scatter_tensor` is
deprecated in the pinned torch and emitted a `FutureWarning` on every ring step:

```
FutureWarning: `torch.distributed.reduce_scatter_tensor` is deprecated. Please use `torch.distributed.reduce_scatter_single` instead.
```

It is now a thin wrapper delegating to `reduce_scatter_single` with an identical signature, so the
swap is a pure rename.

While in there, a small edge-case cleanup in `ring_attention_backward`. The gradient buffers were
allocated with `torch.empty_like`, which inherits the input's strides, and on the
`heads_k_stride == num_kv_heads` branch they are passed straight to the reduce-scatter. NCCL ignores
a destination's strides and fills its storage linearly, so a non-contiguous `dk`/`dv` would silently
come back scrambled. `new_empty` always allocates contiguous. Unreachable today, since
`heads_k_stride` is hardcoded to `1` with no config surface.

- `_ring_attention_backward_fake` has to move with it, or inductor bakes the old strides into an
  `assert_size_stride` guard.
- `dq` is unaffected: flash kernels honor output strides, NCCL does not.
- Not every deprecated call site is covered. `trainer/distributed/collectives.py:159,178` still use
  `all_gather_into_tensor` and `reduce_scatter_tensor`, left as a follow-up. The
  `all_gather_into_tensor` warning still emitted on this path comes from `AllGatherComm` in the
  `ring_flash_attn` package, not from our code.

## Verification

`tests/unit/train/models`: 181 passed, 5 skipped, unchanged. Checked red-green with a local 2-rank
NCCL script (not committed, needs 2 GPUs), comparing contiguous against head-major `k`/`v` with
bitwise-identical inputs.

Before, at `heads_k_stride = 4 == num_kv_heads`:

```
dk: bitwise_match=False max_abs_diff=1.27539 max_rel_diff=1.33811
dv: bitwise_match=False max_abs_diff=0.978516 max_rel_diff=0.97093
```

After:

```
dk: bitwise_match=True max_abs_diff=0 max_rel_diff=0
dv: bitwise_match=True max_abs_diff=0 max_rel_diff=0
```

`max_rel_diff` is normalized by the tensor's max magnitude. bf16 noise on this shape is around
`4e-3`, so the pre-fix difference is structural rather than numerical.


S1ro1 (approved):
lgtm

PR #3574, opened 2026-09-17
exp: sandbox runtime benchmark (prime, modal, e2b)

mikasenghaas:
## Summary

- `benchmarks/runtimes/{prime,modal,e2b}.toml`: one `uv run eval` config per sandbox runtime with everything else fixed. Terminal-Bench 2 at avg@4 (89 x 4), SWE-bench Verified at avg@1 (500), SWE-bench Pro at avg@1 (731), `bash` harness, `deepseek/deepseek-v4.1-flash` on Prime Inference, concurrency pinned at 1000, 3600 s rollout timeout. Retries cover provider errors only, so sandbox failures land as failed episodes and count against the runtime under test. Task images and resource requests come from each Harbor task; the runtime block only names the provider (a `runtime-bench` label on prime, a 2 h sandbox lifetime on e2b).
- `benchmarks/runtimes/analyze.py`: reads the trace streams of finished runs and prints, per run and source: rollouts / share ok / `SandboxError` count, p50 and p90 of boot (`timing.boot`), setup (`timing.setup`) and rollout wall clock (dispatch to arrival), the resources each sandbox requested, the summed sandbox lifetime (box up to end of scoring), and the cost at each provider's published per-resource rates (prime $0.05 core-h / $0.01 GB-h / $0.001 GB disk-h; modal $0.1419 core-h / $0.024 GiB-h; e2b $0.0504 vCPU-h / $0.0162 GiB-h; sources in the README). Prime rows split boot by cached vs built image.
- `benchmarks/runtimes/README.md`: run order (prime, modal, then e2b once its limits are raised), credentials, first-use image costs per provider, what the analysis reports and its assumptions.
- `deps/verifiers` moves to the e2b runtime branch (PrimeIntellect-ai/verifiers#2613, merged with verifiers main at `ceecc4d4`, the commit main pins), and the dependency becomes `verifiers[harbor,modal,e2b]` so both SDKs are in the environment. `uv.lock` re-resolved accordingly (adds e2b, bumps modal to 1.5.5).

## Verification

- `uv run eval @ benchmarks/runtimes/<runtime>.toml --dry-run` resolves all three configs: 3 sources each, `min_inflight = max_inflight = 1000`, the intended runtime type per source, `retries.include = ["ProviderError"]`.
- `analyze.py` run against an existing run directory on the dev box (subprocess runtime) produces the table; resources and cost read `n/a` there as expected.
- Not run: the benchmark itself.


benchmarks/runtimes/e2b.toml
```diff
@@ -0,0 +1,48 @@
...
+[[source]]
+name = "swebench-verified"
+env.taskset.id = "swebench-verified"
+env.agent.harness.id = "bash"
+env.agent.runtime.type = "e2b"
+env.agent.runtime.timeout = 7200
+env.agent.timeout.rollout = 3600
+env.agent.retries.max_retries = 2
+# Sandbox failures are a result of this benchmark, so only provider errors retry.
+env.agent.retries.include = ["ProviderError"]
```
mikasenghaas, line 36: remove the retries on all configs

PR #3596, opened 2026-09-18
fix(orchestrator): fail closed on non-404 /init_broadcaster errors

JannikSt:
## What
`AdminPlane.initialize_nccl` caught `httpx.HTTPStatusError` but only handled HTTP 404 (older engines without the route), silently swallowing every other status. This adds an `else: raise RuntimeError(...)` so non-404 failures fail closed.

## Why this matters
An engine-side 500 from `/init_broadcaster` — e.g. the TCPStore client retry window exhausting while the trainer is still in model setup — used to be swallowed, leaving the orchestrator to proceed with a dead/missing engine rendezvous. The result is a silent pre-step-1 wedge:

- the trainer binds its TCPStore and blocks in `TCPStore(wait_for_workers=True)` until its own ~1200 s timeout, for engines whose init RPC already failed and will never be retried;
- downstream `/update_weights` then surfaces only as offset/distant symptoms (the telus GLM-5.2 incident signature), far from the root cause.

Loud, immediate failure at orchestrator setup is strictly better: the error points at the actual broken engine and its HTTP status.

## Why no 5xx retry (deliberate)
- `/init_broadcaster` is non-idempotent: each engine-side call constructs a fresh NCCL broadcast receiver (`worker/nccl.py`), with ambiguous partial-completion semantics against the trainer's `TCPStore(wait_for_workers=True)`.
- The admin-plane RPC already blocks server-side through the engine's own TCPStore retry window (the httpx client uses no read timeout), so an HTTP 500 means that window is exhausted — an orchestrator-side retry cannot extend it.
- The 404 branch is kept and documented as an explicit compatibility decision for older engines.

## Out of scope
- Pre-request `httpx.ConnectError` retry (the only unambiguously safe retry class: the request never reached the worker).
- Hosted run template `weight_broadcast.timeout` tuning.
- W2 stale-marker hardening / W3 trainer-restart rejoin (orthogonal follow-ups).

## Testing
- Two new unit tests mirroring the existing `tests/unit/orchestrator/test_clients.py` patterns: 500 → `RuntimeError` with "HTTP 500" in the message; 404 → still skipped, no raise.
- `uv run pytest tests/unit/orchestrator/test_clients.py -q`: 12 passed.

---

> [!NOTE]
> **Medium Risk**
> Changes orchestrator NCCL setup failure behavior on distributed training paths; safer fail-closed semantics but runs that previously wedged silently may now fail fast at init.
> 
> **Overview**
> **`AdminPlane.initialize_nccl`** no longer ignores non-404 HTTP failures from **`/init_broadcaster`**. HTTP **404** still logs a warning and skips NCCL broadcast setup for older engines; any other status (e.g. **500**) now raises **`RuntimeError`** with the status and rank offset so setup fails immediately instead of continuing with a broken rendezvous.
> 
> Comments document why retries are avoided (non-idempotent engine-side receiver setup). Unit tests cover **500 → raise** and **404 → skip without raise**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9950ea7f0b55f71e6811763a1690534352c68348. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


JannikSt:
closing — reopening with corrected commit authorship

PR #2592, opened 2026-05-22
Add trainer token JSONL export

samsja:
## Summary
- Add an opt-in RL trainer token exporter under `trainer.experimental.token_export`.
- Export one JSONL record per sequence with aligned per-token arrays for ids, masks, rewards, advantages, entropy, KL/logprob/ratio data, probability deltas, and DPPO mask diagnostics.
- Keep the trainer integration to setup/export/close and leave token-export math in `src/prime_rl/trainer/rl/token_export.py`.
- Thread rollout rewards through trainer micro-batches so they can be included in exports.
- Document the JSONL-only workflow in the config skill.
- Use config attribute docstrings for the token export settings.

This PR intentionally does not include the HTML visualizer; PR #2551 is left unchanged for that follow-up work.

## Verification
- `uv run ruff check packages/prime-rl-configs/src/prime_rl/configs/trainer.py skills/configs/SKILL.md src/prime_rl/trainer/batch.py src/prime_rl/trainer/rl/data.py src/prime_rl/trainer/rl/train.py src/prime_rl/trainer/rl/token_export.py src/prime_rl/transport/types.py`
- `uv run ruff format --check packages/prime-rl-configs/src/prime_rl/configs/trainer.py src/prime_rl/trainer/batch.py src/prime_rl/trainer/rl/data.py src/prime_rl/trainer/rl/train.py src/prime_rl/trainer/rl/token_export.py src/prime_rl/transport/types.py`
- `uv run python -m py_compile src/prime_rl/trainer/rl/token_export.py packages/prime-rl-configs/src/prime_rl/configs/trainer.py src/prime_rl/trainer/batch.py src/prime_rl/trainer/rl/data.py src/prime_rl/trainer/rl/train.py src/prime_rl/transport/types.py`
- Inline `uv run python` JSONL serialization probe for `TokenExporter` with two response sequences.
- Latest Mika follow-up: `uv run ruff check packages/prime-rl-configs/src/prime_rl/configs/trainer.py`
- Latest Mika follow-up: `uv run ruff format --check packages/prime-rl-configs/src/prime_rl/configs/trainer.py`

---

> [!NOTE]
> **Medium Risk**
> Adds new per-step file I/O and threads `rewards` through micro-batch preparation/packing/padding and tensorization, which can affect training-loop performance and data-shape assumptions. Functionality is opt-in via config, limiting blast radius when disabled.
> 
> **Overview**
[...]

PR #3601, opened 2026-09-18
feat(dashboard): a flow tab for verifiers.v1.flow runs

eligotts:
## Summary

Add a Flow tab that follows a `verifiers.v1.flow` run through committed unit state, stage executions, calls, reports, and native traces.

- Project shared typed events into unit lanes and stage transitions. Execution and invocation IDs associate calls, retries, and cached attachments with their stages while preserving result provenance.
- Show affected-unit associations and unit steering directly in the inspector. Links do not invent a target execution, and steering remains visible before or during a stage.
- Keep stage completion separate from the resulting unit status. Operator holds are not execution failures; drain stops remain distinct from cancellation. Launch status follows lifecycle events and the launch lock.
- List Flow reports only when a recorded transition references their filename. Contents remain ordinary files under `reports/`, read and rendered by the existing report and citation paths. Missing requested reports show as unavailable instead of selecting a different report.
- Open completed agent traces in the existing viewer and render message images inline. Flow live snapshots remain available on disk for monitoring agents; they are not consumed by the dashboard or included in its graph refresh fingerprint.
- Serve `/api/runs/{run}/flow` with ETag refreshes and `/api/runs/{run}/flow/traces/{trace_id}` for trace lookup.
- Pin Verifiers to `a94159a` on `feat/flow-e2b`, including the core contracts from [Verifiers PR #2592](https://github.com/PrimeIntellect-ai/verifiers/pull/2592). Reuse the v1 credential reader to avoid the incompatible Prime CLI sandbox import during dashboard startup.

## Validation

Three focused dashboard tests pass. Fresh Taskforge and Repoforge stub runs and repeated initialization pass. HTTP and Chromium checks cover report publication filtering, transition report links, citation verification, completed trace viewing, missing/deleted reports, operator holds, concurrent unit associations, and live steering. Formatting, lint, and JavaScript syntax checks pass. No live model or sandbox runs were launched for validation.


src/prime_rl/entrypoints/flow.py
```diff
@@ -0,0 +1,109 @@
...
+        units=[Unit(path).inspect() for path in paths if (path / ".git").exists()],
+        events=root / TRANSITIONS,
+        traces=root / "traces.jsonl",
+        calls=root / "calls",
+        draining=(root / DRAIN_FILE).exists(),
+    )
+
+
+def main(argv: list[str] | None = None) -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
```
mikasenghaas: need pydantic config here

PR #3602, opened 2026-09-18
feat(eval): make task dispatch rate configurable

samsja:
> [!NOTE]
> **Low Risk**
> Optional config with default unchanged; reuses existing orchestrator dispatch rate limiting with no auth or data-path changes.
> 
> **Overview**
> Eval runs can now cap how fast episodes are dispatched via a new **`tasks_per_minute`** setting on **`ServedEvalConfig`** (shared by standalone eval and SFT online eval). **`None`** keeps the previous behavior with no rate limit; a positive value throttles dispatch through the existing **`Dispatcher`** **`AsyncLimiter`**, intended to pace sandbox provisioning during autoscaling.
> 
> The eval **`runner`** passes **`config.tasks_per_minute`** into the dispatcher instead of hard-coding **`None`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 31f8041b908234dc4c45f128540af8e4daa6435b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/unit/test_configs.py
```diff
@@ -176,6 +159,17 @@ def test_cli_overrides_toml(tmp_path):
...
 
+def test_eval_tasks_per_minute_accepts_toml_and_cli_overrides(tmp_path):
+    write_toml(
+        tmp_path / "eval.toml",
+        {"tasks_per_minute": 100, "source": [{"env": {"taskset": {"id": "gsm8k"}}}]},
+    )
+
+    config = cli(EvalConfig, args=["@", str(tmp_path / "eval.toml"), "--tasks-per-minute", "250"])
+
+    assert config.tasks_per_minute == 250
```
samsja: can ermove this please

samsja: Done — removed in eeecd673916e929252ea1bb46c8660d371de43ab.

packages/prime-rl-configs/src/prime_rl/configs/eval.py
```diff
@@ -21,6 +21,11 @@ class ServedEvalConfig(EvalSourcesConfig):
     """Adaptive in-flight episode concurrency, sized by the same controller as
     ``[orchestrator.concurrency]``. Set ``min_inflight = max_inflight`` to pin it."""
 
+    tasks_per_minute: int | None = Field(None, ge=1)
```
mikasenghaas, line 24: we should prob set a reasonable default here

mikasenghaas: high enough so it doesn't bother

mikasenghaas: also wondering if we should put this in a generic `DispatcherConfig.max_dispatch_per_minute` or smth

samsja: few thoughts on both:

- on the default: the right value is fleet-specific — the knob exists to pace sandbox provisioning while an autoscaled fleet catches up, so it only does anything when it binds. a default high enough to never get in the way would also be too high to prevent the sandbox-not-ready storms it's there for, and once the default is finite there's no way to express "unlimited" from TOML (no null) for big fast fleets. so kept it `None` to mirror the orchestrator knob — but happy to flip it to smth like 10k/min (that wouldn't bind for eval-shaped loads: 128 pinned @ ~5s/episode is ~1.5k/min) if you think the unconfigured-not-ready failure mode is worse
- on `DispatcherConfig.max_dispatch_per_minute`: agreed, the orchestrator's top-level `tasks_per_minute` is the same knob feeding the same `Dispatcher`. worth a follow-up that introduces the shared sub-config and migrates both in one go — nothing in-tree uses the orchestrator one yet, so now's the time. kept this aligned with the existing field so that stays a pure rename

samsja: ok, flipped it after all: `tasks_per_minute` now defaults to 10000 — high enough not to throttle realistic eval loads while still backstopping dispatch storms (accepted that you can't express unlimited from TOML anymore, but nothing realistic runs >166 eps/sec). keeping the current name for now, the `DispatcherConfig` unification can be a follow-up if we ever want it — 6a59204f8dcbd39ff340abca9065ed3b2d1578d9

samsja: settled on 1k/min as the default in the end (8634426705f18d42749773e217d6f43019b35ac7)

samsja: final call: went back to `None` in the end so the eval knob behaves exactly like the train one — same default (off), same opt-in semantics (31f8041b908234dc4c45f128540af8e4daa6435b). consistency between the two felt worth more than a finite default — and if we ever want a default for both, the `DispatcherConfig` unification is the right place to do it

mikasenghaas (approved):

PR #3603, opened 2026-09-18
feat: per-row reasoning effort for sft

mikasenghaas:
## Summary

- A `reasoning_effort` dataset column sets the renderer's `reasoning_effort` field per row, on top of the `[renderer]` config. Null rows use the configured renderer unchanged, so the global effort stays the default.
- Per-dataset effort is the same column with one constant value, so one run can mix efforts across rows or across HF subsets without any new config surface.
- `RendererResolver` in the SFT data module builds renderers lazily per resolved config and caches them, so rows that share a config share one instance. The trainer hands the renderer config and processor to `setup_dataset` instead of building a renderer itself.
- A column on a renderer without a `reasoning_effort` field, or on `renderer.name = "auto"`, raises a clear error.

Renderer audit for the effort field: `qwen3.8`, `hy3`, `inkling`, `gpt-oss`, and `deepseek-v4` expose `reasoning_effort`. `nemotron-3` / `nemotron-3-ultra` expose boolean `low_effort` / `medium_effort` instead, so the column does not apply there. No renderer names the field `effort`.

## Verification

Tiny local dataset (subsets `plain` and `mixed`, the latter with a `reasoning_effort` column of `low` / `medium` / `xhigh` / null) on `Qwen/Qwen3-0.6B` with `[renderer] name = "qwen3.8"`, `reasoning_effort = "medium"`, decoded through `setup_dataset`:

| Mode | Rendered effort per row |
|---|---|
| `plain` subset, no column | all 64 rows medium |
| `mixed` subset | low 16, medium 16, xhigh 16, null → medium 16 |
| `mixed` subset with `renderer.name = "qwen3"` | `ValueError: Renderer 'qwen3' has no reasoning_effort field, but a row sets 'low'` |

The `mixed` subset also ran `uv run sft` for 3 steps on one GPU and finished with decreasing loss.

`tests/unit/test_configs.py` and `tests/unit/train/sft` pass (236 tests).


docs/development.md
```diff
@@ -106,7 +106,7 @@ Warm up the random-weight mini model with SFT on reverse-text so KL divergence b
 ```bash
 uv run sft \
   --model.name ./mini-glm-moe \
-  --data.name PrimeIntellect/Reverse-Text-SFT \
+  --data.source '[{"dataset": "PrimeIntellect/Reverse-Text-SFT"}]' \
```
mikasenghaas: make this data.source.name

mikasenghaas: Dropped the source restructure entirely; `data.name` is back as on main. The row-level `reasoning_effort` column covers per-row and per-dataset effort on its own.

PR #3608, opened 2026-09-19
feat: default enable_ep_weight_filter to true

samsja:
## What

- Adds a typed `inference.vllm.enable_ep_weight_filter` field defaulting to `true`, so every vLLM server prime-rl launches now passes `--enable-ep-weight-filter` (previously this flag only appeared as an explicit pass-through in the GLM-5.3 advanced examples).
- Drops the now-redundant explicit `enable_ep_weight_filter = true` from the GLM-5.3 examples (`examples/advanced/glm-5.3/swe.toml`, `examples/advanced/glm-5.3/infer/pd-llmd.toml`).
- Asserts the default lands on the server namespace.

## Why

Under expert parallelism, vLLM's weight filter skips non-local per-expert tensors **before** reading them from disk: each rank reads only its own expert shard instead of the full checkpoint (experts are ~85-90% of MoE weight bytes). This cuts load I/O and startup time, especially on shared storage.

The default is safe elsewhere — vLLM only activates the filter when the model is MoE, `enable_expert_parallel` is on, and EPLB is off. It is a silent no-op for dense models, fused-expert checkpoints (GPT-OSS style), and EPLB runs. The repo pins `vllm>=0.28.0`, which has the flag.

Set `inference.vllm.enable_ep_weight_filter = false` to restore the old behavior.

Note: #3604 relocates these GLM-5.3 examples to `configs/advanced/` and its copies still carry the explicit line — drop it when rebasing on this PR.

## Test plan

- `uv run pytest tests/unit/test_parsers.py tests/unit/test_configs.py`
- Verified `to_namespace()` emits `enable_ep_weight_filter=True` for a real GLM-5.3 config, and that `--vllm.enable-ep-weight-filter false` overrides it.

---

> [!NOTE]
> **Low Risk**
> Behavior change is gated in vLLM to MoE + expert parallelism without EPLB; dense models and other modes are no-ops, with an explicit config escape hatch.
> 
> **Overview**
> **Expert-parallel MoE loads now opt into vLLM’s EP weight filter by default**, via a typed `inference.vllm.enable_ep_weight_filter` field (default `true`) that flows through `to_namespace()` to `--enable-ep-weight-filter`. That lets each rank read only its local expert shard at checkpoint load instead of full expert tensors, reducing I/O and startup time when EP is on and EPLB is off.
> 
> The GLM-5.3 advanced TOML recipes **drop redundant** `enable_ep_weight_filter = true` lines and rely on the default. A unit test **asserts** the resolved server namespace includes `enable_ep_weight_filter=True`. Set `inference.vllm.enable_ep_weight_filter = false` to disable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f26215dee423de2a19ec763106fd979183556cef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/inference.md
```diff
@@ -123,6 +123,8 @@ data_parallel_size = 8
 
 This configuration will run 2 vLLM processes, each with `data_parallel_size_local = 4` and `tensor_parallel_size = 2` and expert parallelism spanning 2 nodes. The requests are again routed to these processes via the `vllm-router`.
 
+Under expert parallelism, non-local expert weights are skipped at load time (`inference.vllm.enable_ep_weight_filter`, defaults to `true`), so each rank only reads its own expert shard from disk. Since experts make up most of the weight bytes of MoE models, this greatly reduces load I/O on shared storage. It is a no-op for non-MoE models and when EPLB is enabled.
```
samsja: remove this

samsja: Removed in cc73d0f.

mikasenghaas (approved):
nice

